PR #287, opened 2014-10-11
Intel10G: Offload checksums

javierguerragiraldez:
Second edition of checksum offloading.  This time verifies if the packet is asking to be checksummed before meddling with it.

It's also a little more robust, uses less `ffi.cast()` of pointers and removes `lib.bits{}`, which only works on 32-bit integers.  (it's easy to extend them to 64 bits (just replace `0` and `1` with `0ULL` and `1ULL`), but all the NIC registers are 32 bits, so it would recast values all the time)

### DIFF
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -28,6 +28,7 @@ local IP_UDP = 0x11
 local IP_TCP = 6
 local IP_ICMP = 1
 local IP_SCTP = 132
+local IP_L2TPV3 = 0x73
 local IPV6_ICMP = 0x3a
 
 local IPV4_IHL_OFFSET = 14
@@ -270,9 +271,19 @@ local function keep_offload_ctx(self, p)
    local b = p.data
    local ctx_desc_a = 0ULL
    local ctx_desc_b = 0x0000000fe0200000ULL -- DTYP=2, DEXT=1, BCNTLEN=0x3F
-   local iplen, maclen, ctx_l4t = 0ULL, 14ULL, 3ULL
+   local wraplen, maclen, iplen, ctx_l4t = 0ULL, 14ULL, 0ULL, 3ULL
    local ethtype, proto = ffi.cast('uint16_t *', b+ETHERTYPE_OFFSET)[0], 0ULL
    local offloadflags = 0ULL
+
+   if ethtype == ETHERTYPE_IPV6 then
+      iplen = 40ULL
+      proto = b[IPV6_NEXT_HEADER_OFFSET]
+      if proto == IP_L2TPV3 then
+         wraplen = 66     -- IP_L2TPV3 header size: 14 + 40 + 12
+         b = b + wraplen
+         ethtype = ffi.cast('uint16_t *', b+ETHERTYPE_OFFSET)[0]
+      end
+   end
 
    if ethtype == ETHERTYPE_IPV4 then
       ctx_desc_b = bor(ctx_desc_b, 0x0400ULL)     -- TUCMD.IPV4
@@ -304,7 +315,7 @@ local function keep_offload_ctx(self, p)
       b[maclen + iplen + SCTP_CSUM_OFFSET+1] = 0
    end
 
-   ctx_desc_a = bor(ctx_desc_a, lshift(band(maclen, 0x7F), 9), band(iplen, 0x1FF))
+   ctx_desc_a = bor(ctx_desc_a, lshift(band(maclen+wraplen, 0x7F), 9), band(iplen, 0x1FF))
    ctx_desc_b = bor(ctx_desc_b, lshift(ctx_l4t, 11))
    if ctx_desc_a ~= self.prev_ctx_desc_a or ctx_desc_b ~= self.prev_ctx_desc_b then
       self.txdesc[self.tdt].address = ctx_desc_a
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -323,7 +323,7 @@ local txdesc_flags = bits{ifcs=25, dext=29, dtyp0=20, dtyp1=21, eop=24}
 function M_sf:transmit (p)
    keep_offload_ctx(self, p)
    self.txdesc[self.tdt].address = memory.virtual_to_physical(p.data)
-   self.txdesc[self.tdt].options = bor(p.length, txdesc_flags, lshift(p.length+0ULL, 46))
+   self.txdesc[self.tdt].options = bor(p.length, txdesc_flags, self.offloadflags, lshift(p.length+0ULL, 46))
    self.txpackets[self.tdt] = p
    self.tdt = band(self.tdt + 1, num_descriptors - 1)
 end
@@ -335,7 +335,9 @@ function M_sf:sync_transmit ()
    -- Release processed buffers
    if old_tdh ~= self.tdh then
       while old_tdh ~= self.tdh do
-         packet.free(self.txpackets[old_tdh])
+         if self.txpackets[old_tdh] ~= nil then
+            packet.free(self.txpackets[old_tdh])
+         end
          self.txpackets[old_tdh] = nil
          old_tdh = band(old_tdh + 1, num_descriptors - 1)
       end
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -380,6 +380,8 @@ function M_sf:receive ()
       if band(wb.xstatus_xerror, checksdone) == 0 then
          p.flags = bor(p.flags, C.PACKET_CSUM_VALID)
       end
+   else
+      p.flags = bor(p.flags, C.PACKET_NEEDS_CSUM)
    end
    return p
 end
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -372,6 +372,15 @@ function M_sf:receive ()
    p.length = wb.pkt_len
    self.rxpackets[self.rxnext] = nil
    self.rxnext = band(self.rxnext + 1, num_descriptors - 1)
+
+   p.flags = 0
+   local checksdone = band(wb.xstatus_xerror, 0x60)  -- which checks have been done? (IPv4, L4)
+   if checksdone ~= 0 then
+      checksdone = lshift(checksdone, 24)            -- shift to errors area
+      if band(wb.xstatus_xerror, checksdone) == 0 then
+         p.flags = bor(p.flags, C.PACKET_CSUM_VALID)
+      end
+   end
    return p
 end
 
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -19,6 +19,26 @@ local macaddress = require("lib.macaddress")
 local bits, bitset = lib.bits, lib.bitset
 local band, bor, lshift = bit.band, bit.bor, bit.lshift
 
+local ETHERTYPE_OFFSET = 12
+-- network order
+local ETHERTYPE_IPV6 = 0xDD86
+local ETHERTYPE_IPV4 = 0x0008
+
+local IP_UDP = 0x11
+local IP_TCP = 6
+local IP_ICMP = 1
+local IP_SCTP = 132
+local IPV6_ICMP = 0x3a
+
+local IPV4_IHL_OFFSET = 14
+local IPV4_PROTOCOL_OFFSET = 23
+local IPV6_NEXT_HEADER_OFFSET = 20 -- protocol
+
+local IPV4_CSUM_OFFSET = 24
+local TCP_CSUM_OFFSET = 16 --50
+local UDP_CSUM_OFFSET = 6 --40
+local SCTP_CSUM_OFFSET = 8 --42
+
 num_descriptors = 512
 --num_descriptors = 32
 
@@ -44,7 +64,10 @@ function new_sf (pciaddress)
                  rxpackets = {},   -- Rx descriptor index -> packet mapping
                  rdh = 0,          -- Cache of receive head (RDH) register
                  rdt = 0,          -- Cache of receive tail (RDT) register
-                 rxnext = 0        -- Index of next buffer to receive
+                 rxnext = 0,        -- Index of next buffer to receive
+                 prev_ctx_desc_a = 0,
+                 prev_ctx_desc_b = 0,
+                 offloadflags = 0ULL,
               }
    return setmetatable(dev, M_sf)
 end
@@ -179,6 +202,7 @@ function M_sf:init_receive ()
    })
    self.r.MAXFRS(lshift(9000+18, 16))
    self:set_receive_descriptors()
+   self.r.RXCSUM(bits{IPPCSE=12})
    self.r.RXCTRL:set(bits{RXEN=0})
    return self
 end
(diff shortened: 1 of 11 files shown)


javierguerragiraldez:
added CPU-based checksum on `want_modify()`

eugeneia:
@javierguerragiraldez, @lukego 

> ERROR     testlog/lib.nfv.config

This failure is caused by a known regression, rebasing onto master fixes it.

> ERROR     testlog/apps.intel.intel_app

I can get the intel test to pass on `0000:82:00.0/1@chur` but it fails on `0000:84:00.0/1@chur` with "wrong proportion of packets passed/discarded".

Then again for some cards on davos I get the same error and for others I get "not enought packets somewhere".

> ERROR     testlog/lib.nfv.selftest.sh

On chur: The `vhost_user` app has trouble with memory, see this snabb log:

https://gist.github.com/eugeneia/fc06af4cc7b27fa84f4c

On davos: PING fails, I assume because the intel app is failing.

eugeneia:
I double checked the results on grindelwald and the `vhost_user` crash happens there too. So as far as I can tell we have at least one  regression that slipped into master and affects the `vhost_user` app.

lukego:
@eugeneia Great catch. I'm taking a look on chur.

lukego:
I rebooted `chur` with 1GB large pages in the grub config. Does this help? (I am not sure that is the root cause, but I had sneakily changed that to 2MB recently while testing the straight-line branch, and 1GB is the preferable setting for the NFV application.)

The error on Grindelwald is the same for you? (Failing to resolve a physical address?) 1GB pages are already enabled there (see /proc/meminfo).

eugeneia:
On Chur the master branch passes all tests now. On Grindelwald I get:

> Restarting Virtio_B (died at 1421076458.855534: lib/virtio/net_device.lua:382: mapping to host address failedcdata<void *>: 0xffff88003bd2cd00)

See the full log: https://gist.github.com/eugeneia/6489768c57e434036269

lukego:
Can you strace the snabb process?

eugeneia:
@lukego https://gist.github.com/eugeneia/f0434505c44c03313a1d

I used `grep -v nanosleep` to shrink the file, its >20MB otherwise.

eugeneia:
@javierguerragiraldez @lukego I now believe that we have a regression in the intel driver. To see for yourself run the intel_app selftest repeatedly on any server, on any card. The error varies (sometimes unbalanced packets, sometimes not enough packets, sometimes `sent == prevsent`, sometimes everything passes).

I have ran the intel selftest about a hundred times now, on branch master and keep-nfv, on davos and chur, on different cards. I can't see a pattern, the failures/successes seem random.

javierguerragiraldez:
On Mon, Jan 12, 2015 at 2:22 PM, Max Rottenkolber
<notifications@github.com> wrote:
>
> @javierguerragiraldez @lukego I now believe that we have a regression in the intel driver. To see for yourself run the intel_app selftest repeatedly on any server, on any card. The error varies (sometimes unbalanced packets, sometimes not enough packets, sometimes sent == prevsent, sometimes everything passes).


I'll take a long look tonight.

previously, i've seen that on Davos 0000:82:00.0/1 and 0000:87:00.0/1
do work consistently, while :02: and :84: fail most of the time.

on chur I use :82: by default, but there it usually worked quite good
on any card.


-- 
Javier

javierguerragiraldez:
I get the `selftest()` to consistently run with the :82: card in chur, but only after unbinding it (both /0 and /1) from the `ixgbe` driver.  The :84: card passes the 'in chip' tests, but fails tests that actually put bytes in the wire.
The NFV tests are better in that the script does connect with the VMs, but fails on the first ping, apparently the packets reach the intel_app but don't appear in the other end.

lukego:
I suppose that we should be automatically unbinding the NICs from the OS driver before we use them.

Javier, do the NFV tests work for you on the "good" NIC? That is the most interesting question from my perspective. (Should I replace some cables on davos in case it is a physical connectivity issue?)

eugeneia:
@javierguerragiraldez Doh! I didn't think of maybe unbinding the cards at all. Sorry for causing such a ruckus.;) I was honestly confused.

> The :84: card passes the 'in chip' tests, but fails tests that actually
put bytes in the wire.

This is semi-known, "some" cards have bad cables and we always avoid them instead of keeping track of them.

javierguerragiraldez:
> I suppose that we should be automatically unbinding the NICs from the OS
driver before we use them.

I think some very early versions did that. I don't remember why it was
removed, maybe around the issues with non-root users.

A similar preparation will be needed to migrate to SR/IOV, although that
would have to be on a selection process, not initialization.

> Javier, do the NFV tests work for you on the "good" NIC? That is the most
interesting question from my perspective.

Unfortunately no. I think there is a regression, but can be something
simple like a mismatch between the app and the scripts.

From the log it seems the packets reach the Intel app but are not
transmitted.

--
Javier

eugeneia:
After unbinding all cards on davos from the ixgbe driver I sill get this failure in about one of 20 runs:

> Send a bunch of packets from Am0
> ...
> wrong proportion of packets passed/discarded
> EXITCODE: 1

It really helps to do something like `for i in {1..20..1}; do RUNTEST; done`.

javierguerragiraldez:
> >Javier, do the NFV tests work for you on the "good" NIC? That is the most interesting question from my perspective.

> Unfortunately no. I think there is a regression, but can be something simple like a mismatch between the app and the scripts.

finally got the NFV test running, at least with "good" NICs (:82: in chur).  it was a silly thing, a MAC address mismatch between the config script and the VM images.  It was on my local config, so I guess that's why it didn't affect everybody.

All tests pass on the master branch, but on this (`offload`) the tunnel tests fail.  Just removing the `C.VIRTIO_NET_F_CSUM` flag from the advertised virtio features allows it to pass, without changing the code at all, so I guess it's because the Linux driver sends fragmented packets when this flag is active, and the tunnel hasn't been tested on this case.

I see that the `Virtio_A` app is dying at line https://github.com/SnabbCo/snabbswitch/blob/master/src/lib/virtio/net_device.lua#L247 because `b` was `nil`.  Adding lots of `print()` to code that I don't really understand, it seems that the `Virtio_B` app has 8192 packets on the `self.vring_transmit_buffers` freelist, but `Virtio_A` is empty.  Should it have been initialized by then?  Or is it because the app hasn't passed through a `:push()` cycle yet?

lukego:
Great detective work, Javier!

Question on my mind is whether the straightline branch will work well enough to merge within the next week or two. If so then this bug will probably disappear during the Great Simplification. If not then we will have to fix it. Sorry about this limbo situation - has to be short lived - really seems worth it if we can make the merge.

javierguerragiraldez:
> Maybe the intel selftest is really sensitive to minor hardware failures, e.g. packet loss on wire?

Not really.  For example, the switch-like 'filter by MAC' test builds two packets, one with the right destination address and the other to a bogus destiny.  Then sends both as many times as possible in a second.  Finally it tests that the receiving app gets roughly half as many packets as they were sent.  And when I mean "roughly half" I mean anywhere between 25% and 75%.  That way, if the filters are badly defined we get either 0 or 100% and the test fails.  But even lets say, 30% packet loss (which would be disastrous in a real network) would result in 35% received packets, well within the 25%-75% passing range.

javierguerragiraldez:
> this bug will probably disappear during the Great Simplification

Yes, it looks very probable.  Still, I tried with changing the `want_modify()` function to always return a coalesced clone, and didn't seem to make a difference (besides a huge slowdown!).

lukego:
Going to need a revision after the straightline merge! Sorry about that :-)

lukego:
(Oops should not have closed, only commented.)

lukego:
(Damned how did I accidentally close this twice. Fat fingers!)

javierguerragiraldez:
The `selftest()` is working again (after straightline merge), but I can't get the NFV `selftest.sh` to correctly boot the VM images on chur.  Is that a known issue?  I see some other PRs have similar snabbot reports...

lukego:
@eugeneia Can you see what the problem is with NFV selftest?

eugeneia:
@javierguerragiraldez The NFV selftest still runs fine for me (when I run it manually, the SnabbBot failures where PING fails are still a mystery to me). If the VMs don't boot its most likely an issue with the qemu settings in bench_conf.sh, you can see how qemu fails by removing the /dev/null pipes off the qemu calls in bench_env/common.sh.

javierguerragiraldez:
> you can see how qemu fails by removing the /dev/null pipes off the qemu calls in bench_env/common.sh

Great, that's what I was looking for.

lukego:
Looking forward to merging this! The NFV selftest we can perhaps troubleshoot on master. (This is not the only PR affected.)

src/apps/intel/intel10g.lua
```diff
@@ -224,11 +248,102 @@ end
...
+      self.txdesc[self.tdt].options = ctx_desc_b
+      self.txpackets[self.tdt] = nil
+      self.tdt = band(self.tdt + 1, num_descriptors - 1)
+      self.prev_ctx_desc_a = ctx_desc_a
+      self.prev_ctx_desc_b = ctx_desc_b
+      self.offloadflags = offloadflags
+   end
+end
+
+-- local txdesc_flags = 0x0000000022300000ULL                      -- DEXT=1, IFCS=1, DTYP=0011b
```
lukego: Large block of commented out code. Should be deleted? (I don't think you really want to push this change?)

javierguerragiraldez: right, i'm still cleaning it up.  during the after-straightline rebase there were some conflicts that confused me, so i commented out most to later compare with the previous history.

javierguerragiraldez:
I managed to get the NFV tests working on the master branch, it was mostly about configuration and not code.  This branch does pass the selftest but not NFV; I think there's something fishy between this and virtio code, but can't pinpoint it yet.

javierguerragiraldez:
... and `iperf` passes!

one fixed one small `-- TODO` in lib/vritio/net_device.lua: to copy packet flags from the link to virtio.

still lots to clean up, and not stellar performance (max so far is 8.80 Gbits/sec with jumbo frames), but enough for tonight!

lukego:
Great hacking! :beer:

javierguerragiraldez:
After several nights of very small performance improvements, average bandwidth when using jumboframes is around 9Gb/sec (as reported by `iperf` between 2 VMs over the NIC 'switch') on grindelwald.

Rebasing to `upstream/master` was almost uneventful, except that i'm no longer able to run the NFV tests (`selftest()` still works), so coudln't test bandwidth on interlaken.

This should be now ready to merge, but it would be nice to see it under NFV tests again...

eugeneia:
@javierguerragiraldez Can you tell me on which commit the selftest still runs? Could it be a vhost_user issue?

src/program/snabbnfv/traffic/traffic.lua
```diff
@@ -41,6 +41,11 @@ function traffic (pciaddr, confpath, sockpath)
          local mtime2 = C.stat_mtime(confpath)
          if mtime2 ~= mtime then
             print("Loading " .. confpath)
+            if mtime2 == 0 then
```
lukego: This change may have unintended consequences: The traffic process will be started as a daemon in the OpenStack context and initially its configuration file may not exist. This change would seem to cause the traffic process to terminate instead of polling for the configuration to be created.

Can we skip this change or make it in another way that does not change the high-level behavior of the traffic process?

javierguerragiraldez: > Can we skip this change or make it in another way that does not change the high-level behavior of the traffic process?

what I wanted to avoid was the forcefully killing of the process, so that the profiling output could be seen.  maybe it would be better to catch `SIGTERM` and do an ordered exit instead of just dying

javierguerragiraldez:
Full rebase and cleanup.
Performance is steady at little under 10Gb/sec on grindelwald.  Adding @lukego 's "don't allocate tx" patch [5b5df1928] (not included here) gets over 11.5Gb/sec.

javierguerragiraldez:
Finally passes all NFV tests!
It was mostly about setting checksum valid/needed flags when the NIC doesn't bother to check.

lukego:
Hold on! :-)

Let us make sure that the right thing is happening here. It would be easy for a problem with verifying the checksum to hide a problem with setting the checksum. I am thinking especially of the case with L2TPv3 tunnelling enabled.

Let's take an example packet that the VM has requested checksum offload for and then Snabb Switch has added encapsulation to. The headers marked with (*) are the ones that don't have valid checksums yet and need to have this fixed before transmission.

    Ethernet (outer)
    IPv6 (outer)
    Ethernet (inner)
    *IPv4 (inner)
    *TCP (inner)

I don't see from the code how this would happen today. The hardware needs to know to checksum the *inner* headers instead of the outer ones. Or, if hardware can't do that, then it has to be done in software by Snabb Switch.

Here is the code that I am looking for but don't find:

1. The Virtio code to set the `packet.csum_start` and `packet.csum_offset` fields in the packet sent by the VM so that we know where the not-yet-checksummed header is.
2. The L2TPv3 app to update these offsets when adding a prefix to the packet (e.g. in `packet.prepend()`)
3. The Intel10G driver to use these offsets to checksum the right header (the inner one) e.g. by telling the NIC that the ethernet header is really big in order to skip over the outer headers.

and then we also have to be very careful not to over-interpret the checksum validation done by hardware. Just because hardware says that the outer header's checksum is OK does not mean that the inner one is also OK. So perhaps the L2TPv3 app needs to clear the `PACKET_CSUM_VALID` flag when it removes the encapsulation too.

Does this make sense? I can add a few line-notes to help explain in more detail.

javierguerragiraldez:
The transmission and reception checksum issues are quite different.

About transmission:

> Let's take an example packet that the VM has requested checksum offload for and then Snabb Switch has added encapsulation to. The headers marked with (*) are the ones that don't have valid checksums yet and need to have this fixed before transmission.
>
> Ethernet (outer)
> IPv6 (outer)
> Ethernet (inner)
> *IPv4 (inner)
> *TCP (inner)
>
> I don't see from the code how this would happen today. The hardware needs to know to checksum the inner headers instead of the outer ones. Or, if hardware can't do that, then it has to be done in software by Snabb Switch.

I'm trying to make the TX checksum handle the inner headers by telling the NIC that there's a 80-byte ethernet header instead of the usual 14-byte one.  The outer IPv6 header doesn't have checksum field, so it can be ignored.  It seems to work already, but i want to test it more before committing.


About reception checksum verifying:

It seems that the Linux kernel admits that not every packet can be verified by every hardware checksum engine, so a packet might not have the PACKET_CSUM_VALID flag, but if it has PACKET_NEEDS_CSUM, then the kernel will check it in software.  If neither flag is set, the packet is blindly discarded.

There's no way to tell the RX checksum engine how it should check the packet, so I can't play with the offsets to make it verify the inner headers.  The docs say that it won't check any kind of encapsulation, and it seems to happen exactly so: the packets are signaled as "no checks performed"

So, we could either do our own check, or admit the hardware doesn't help and pass it to the kernel.  The pros of letting the kernel do it:

- it's already done and debugged!
- it can manage lots of different cases
- it's probably been optimized the hell out of it

the pros of doing it ourselves:

- it happens "outside" of the VM, on another CPU core.  might be better use of resources.
- we can focus on a few cases that matter most to us, maybe achieving better optimization than the all-powerful code in the Linux kernel.


Finally, i think there might be some opportunities to avoid some checks if we could pass some flags from the transmitting app(s) to the receiving end.  It seems there's no space for this in the L2TPv3 header... unless we use one or two of the 'reserved' bits mentioned in the RFC.

lukego:
This makes a lot of sense.

For the transmit path I am looking forward to the hack to checksum the inner packet :-)

For the receive path I like the idea of letting the guest do the checksum if it was not done in hardware. So for non-tunnelled traffic the hardware will checksum the TCP header and we will report that to the guest (PACKET_CSUM_VALID) but for tunnelled traffic the inner TCP header won't have been automatically checksummed and so the guest will need to check it (PACKET_NEEDS_CSUM).

The issue still in my head though is: do the packets actually get checksummed on TX and RX with the code on this pull request? Or is checksum skipped in both directions and that is what makes the tunnel test case pass? If the latter then it is too early to merge because it is only passing the test cases on a technicality (limitation on the test suite). I could be missing something obvious in the code though :)

lukego:
BTW: Is the "tell the NIC there is a large ethernet header so that it skips the inner header" trick based on the `packet.csum_start` field and initialized from Virtio? (That seems like the correct way, or?)

lukego:
Aside: Some friends kindly open sourced their CPU checksum implementation. In trivial testing at least it seems way faster than e.g. the DPDK one. It doesn't sound like we desperately need this right now but, anyway, the code is in #391 for reference.

javierguerragiraldez:
Finally includes the 'inner packet' checksum offload.

I've verified that it does update the correct field in the inner packet, and the Linux driver accepts it.  Also, modifying another byte of the packet makes it reject the packet, so I guess the checksum is correct.

This offload code doesn't use the `.csum_start` and `.csum_offset` fields of the packet struct, because the NIC uses `maclen` and `iplen` plus types (IPv4 vs. nothing (IPv6) and TCP/UDP/SCTP/nothing).  Still, I've added the update to these fields in the tunnel app, because the Linux kernel driver might want to use them at some point.

lukego:
Victory at last :-). This is a very impressive piece of programming! Celebration is in order :beers: :+1:

I have an second idea. (Uh-oh.) How about if we measure the benefit of hardware checksum offload before we commit to it?

Specifically I start to wonder about the "straightline" tradeoff here. Hardware offload turns out to be quite complex and the justification for this is to get a major performance boost. Now that we have an implementation that we can benchmark, why not check whether a more generic CPU-based checksum can be competitive?

Even more specifically I am imagining this experiment:

- Enable checksum offload at the Virtio level.
- VM to network: Calculate checksum using the SIMD checksum function from #391 instead of offloading it onto the NIC.
- network to VM: Keep it simple for proof of concept experiment - same as we have today, using hardware checksum if available and otherwise letting the guest do it.

Worth testing?

My thinking here is that the NIC offers a pretty crumby checksum offload interface, really. Ideally we would simply say "checksum from START to END and store the result at OFFSET" and the rest would be taken care of in software. Instead they have us parsing packets and explaining the data to the NIC in terms of a really limited set of protocols. The next Intel NIC is better in some respects (more protocols) and worse in others (complexity explosion while trying to simultaneously offload inner and outer checksums). God knows how nasty it will be making this nice across more vendor NICs.

So a reasonable person may ask, what is the payoff compared with implementing the checksum primitive that we actually want on the CPU? Especially if this can be offloaded onto the SIMD unit that has exponentially increasing throughput across contemporary CPU models?

I also have a slightly uneasy feeling about embracing special cases to make hardware happy. There are a lot of situations where we could either do everything well, slowly, or brokenly, and it's not obvious which. For example, all of these rhetorical questions would require some head scratching to answer in a satisfying way:

- Can we use VLANs on the outer headers? QinQ? MPLS? GTP?
- How about on the inner headers?
- Can the guest add its own encapsulations and ask us to offload its inner headers?
- How about if we and the guest both add encapsulations and wind up with three levels?
- How is the relative performance of different encapsulations (GRE, L2TPv3, Geneve, ...) on different hardware? Do users have to care about this when they are choosing protocols and/or buying hardware? (What do you do if the NIC only offloads VXLAN and the Cisco router only offloads L2TPv3? and whose fault is it if you find yourself in that situation?)

These problems would all seem to disappear if we could "do a straightline" and have a fast generic checksum that we can use guilt-free. (There are also cool userspace bonus points we could have a really fast SIMD checksum since the kernel does not allow itself to use SIMD in the network stack. Not that the kernel's life is not already hard enough these days.)

Having said all of that, I will have a beer to celebrate the fantastic driver hack you have got working :-)

javierguerragiraldez:
> Victory at last :-). This is a very impressive piece of programming! Celebration is in order

Glad you like it.  I'm always worried when it takes me so many days and the result is so few lines of code!


> I have an second idea. (Uh-oh.) How about if we measure the benefit of hardware checksum offload before we commit to it?

While testing it I haven't seen any obvious improvement in bandwidth.  I guess we could see less guest CPU usage.  If we can prove that, even a "same performance as the kernel" implementation could be beneficial.  Now, if we can use SSE and the kernel can't.... sounds tempting to test at the very least.

Now that you've found a fast-and-tested checksum function, it should be easy to write a simple app that turns PACKET_NEEDS_CSUM into PACKET_CSUM_VALID by adding checksum at one direction and by checking it at the other.  Other packets should just pass through.

But I still don't get how the `.csum_start`/`.csum_offset` fields work in the general case.  In our selftest, during iperf I saw the IPv6/TCP packets had 0x36/0x10 on those fields, pointing to the TCP checksum field.  But what about IPv4/TCP?  both headers need a checksum, so which one would those fields point to?

lukego:
I know what you mean about much effort and little code. But I like it best that way really. Even better when the final result is to delete code :-). Great to have the fun of writing the tricky code and then find a simpler solution that is easy to maintain :-).

I am not sure exactly how the checksum flags work but I can tell you how I *hope* they work:

- Guest calculates all checksums except one. For example, guest computes IPv4 checksum and TCP pseudo-header checksum, but not TCP data checksum.
- Guest requests offload of this last checksum. The offload is one dumb ones-complement checksum only. The checksum location is given in the `struct virtio_net_hdr` fields: from `csum_start` until the end of the packet, with the result stored at `csum_offset`.
- In the case of TCP, the guest will prepopulate the TCP checksum field with the pseudo-header checksum and that means only a dumb ones-complement checksum is needed.
- This would work for all protocols, including ones that are not invented yet, provided the guest tells us the right bytes to checksum. (The Snabb Switch code does not care what the protocol is -- none of this packet-parsing that the NIC makes us do.)

Could this be close to reality?

If so then the offload should be basically one function call to `update_checksum()` in either the virtio module or the intel10g module like e.g.

```
char *buffer = packet.data + net_hdr.csum_start;
int len = packet.length - net_hdr.csum_start;
int initial = 0; // assume pseudo-header is already in the checksum field
int csum_pos = net_hdr.csum_offset; // xxx is this relative to start of packet or to csum_start?
update_checksum(buffer, len, initial, csum_pos);
```

although there might be an incompatibility in `update_checksum()` because it is setting the checksum field value to 0 before doing the calculation, but if that is there the pseudo-header checksum is then it should not be cleared (and so you would need to comment out the line `*cksum_ptr = 0`).

Make any sense? Sound possibly like reality?? (I know that in the case of the NIC the actual programming interface turned out different than I hoped :-))

lukego:
Closing because the software checksum seems like a big success. (Great to also have this NIC offload code for future reference.)

PR #316, opened 2014-11-13
[lib.timer] Remove `core.timer' and supply a `lib.timer'

eugeneia:
[lib.timer] Remove `core.timer` and supply a `lib.timer` instead, which re-implements `core.timer` in an app-local fashion.

Adapted (and removed some, now impossible?) uses of `core.timer`.

After #302 lead to headaches this PR goes a different road. It changes the timer implementation quite a bit, so any auditing would be appreciated. Did I mention how I am especially scared of time? Most unpredictable concept in the universe.

**Notes:**

This uses `engine.now()` to get the "time". Which returns the number of seconds since 1970 *as a floating point number with up to 5 digitis of precision* **as far as I can tell**. So what I do is multiply by 1.000 and expect to get *milliseconds*.

I had to remove some uses (top level reporting timers in designs and selftests), since you kind of need an app with a pull method now to insert your `timers:run()`.

The interface of `lib.timer` goes like this (by example):

```
local timers = require("lib.timer").TimerTable.new()

timers:activate( 
  timers:timer(   -- This was `timer.new(...)' before.
    "name",
    <function>,
    <milliseconds>, -- This parameter changed to milliseconds (instead of nanoseconds)
    'repeating')
)

timers:run() -- Call this e.g. in `pull()'
```


lukego:
Late reply...

Agree that apps should have their own simple timer mechanism separate from the `timer` module.

I like the idea of a repeating lib.timer(). Is the mode argument actually a bit awkward? should we have separate `timer` and `repeating_timer` functions instead?

How should we represent time in Snabb Switch? integer nanoseconds or floating-point seconds?

eugeneia:
> should we have separate timer and repeating_timer functions instead?

Probably!

> How should we represent time in Snabb Switch? integer nanoseconds or floating-point seconds?

I dislike floating-point seconds because its hard to tell the precision of the timer. I find an API that says "returns time in nanoseconds" better than one that says "returns time in seconds (as a floating point number with nanosecond precision, except sometimes, because IEEE)".

plajjan:
I agree with @eugeneia, I prefer determinism of nanosecond integer.

lukego:
This is actually tricky.

Will our time value be represented as a native Lua number or an FFI number?

Native Lua numbers are convenient because you can do arithmetic on them without creating garbage and you can use them as table keys. On the other hand, native Lua numbers are double floats, and those can only count up to 2^53 before losing precision. That is limiting because 2^53 nanoseconds is only a few months. So if we want to use native Lua numbers as integers we might have to choose microseconds as the unit instead (2^53us ~= 285 years).

FFI number could be a uint64_t. That is precise enough to use for nanoseconds (2^64ns ~= 584 years). The downside is that we will have to be careful with time values: if you do arithmetic on time values inside an inner loop and LuaJIT's allocation sinking optimization doesn't apply for any reason then you will invoke the garbage collector. That kind of issue is a pain to debug.

The third possibility would be to use seconds as a double-float *if* it could be determined that the floating point behaviour is sane and easy to describe. Hypothetically, if time we measured time in double-float seconds since 1970 and we knew that precision would be within 10ns for the next 100 years then that might be something we could live with. (Hopefully the IEEE would develop quadruple floats before the year 2115?)

The Wikipedia page for double floats has some nice information but it would be nice to find a detailed discussion of exactly this problem.

Otherwise: integer microseconds represented as a double-float? and hope that quad-floats are invented before the year 2355 or so? :-)

lukego:
Closing this PR because I believe that a replacement version is coming.

eugeneia:
@lukego So I have been pondering about yet another timers rework, but satisfied I am not: https://github.com/SnabbCo/snabbswitch/compare/master...eugeneia:timers-v3

(This passes everything but the RateLimiter test :-1: )

The essence of this branch is to remove `core.timer` and replace it with a thin layer over a generalized `lib.timer`. Arguably the code does get a bit simpler at some points but I found the following caveats:

* There is a timer use case not covered by this: Defining and running timers from "outside" of the engine. (This is kind of known, but still begs the question if there is a feature mismatch.)
* The new `core.timer` has no notion of "stacking timers". E.g. if a repeating timer has an interval of 100ns but is only called every 200ns then its effective interval will be 200ns. This is the reason why the new timer can not really satisfy the old `core.timer` selftest.
* The `RateLimiter` selftest fails with the new timer (I assume because of the previous issue)

The more I mess around with new timer implementations the more I feel that the current one is simply better. Both in functionality and performance. Right now I feel it would make more sense to catch and log errors in timer callbacks separately from the app restart framework and just be happy with the existing implementation. Maybe someone else wants to take a shot at this with a fresh mind, but I personally don't feel like I am improving things. ;)

lukego:
Let's take a look at this again later.

There is another relevant development that is the idea of running a separate Lua VM for each app as @javierguerragiraldez is exploring. If we do this then it will change the constraints - it would be more awkward for timers to be global and shared.

PR #336, opened 2014-12-08
Avoid interpreter fallback in timer and app core modules (fixed and rebased)

alexandergall:
The core modules app.lua and time.lua are causing an excessive amount
of interpreter fallback, leading to a significant performance penalty.
This changeset removes one of the causes and offers a workaround for
the other one through a configuration option.

The definition of call_timers() in core/timer.lua is moved out of
run_to_time() to avoid generation of the non-compileable bytecode
UCLO (bytecode 51 in the current version of LuaJIT).

The code for detecting dead apps by wrapping them in pcall() leads to
frequent trace aborts due to "NYI: return to lower frame".  It is not
entirely clear what exactly this means and whether it can be avoided
somehow.  A new configuration option "dead_app_detection" is
introduced that disables this mechanism when set to false (defaults to
true).  The options for main() are passed on to breathe() to make this
work.

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -184,7 +184,18 @@ end
 
 -- Call this to "run snabb switch".
 function main (options)
-   options = options or {}
+   local default_options = {
+      dead_app_detection = true,
+   }
+   if options then
+      for k, v in pairs(default_options) do
+	 if options[k] == nil then
+	    options[k] = v
+	 end
+      end
+   else
+      options = default_options
+   end
    local done = options.done
    local no_timers = options.no_timers
    if options.duration then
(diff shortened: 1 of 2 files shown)


lukego:
I take your point and definitely agree with merging fixes that avoid JIT problems.

I would suggest not modifying the `options` table that is passed as an argument, just in case the caller wants to use it for something else too and is surprised when it changes.

alexandergall:
Agreed. A better choice to make this configurable would probably be to use a variable in the app module so one can do e.g.

`local app = require("core.app")`
`app.dead_app_detection = false`

PR #339, opened 2014-12-16
Minor fixes for #338

eugeneia:
Minor fixes alongside the documentation PR #338 :

Includes one possibly controversial change: Make "App modules" **not** return the app classes so you can do:

```
  App = require("path").App
```

I did this because *most* apps did it this way and I wanted all apps to behave the same. It's arguable if this is *right*.


eugeneia:
@lukego This should be merged alongside the documentation PR (already merged).

lukego:
Looks like a valuable activity to me. I'm really happy to see some editing to gradually bring our varied and experimental coding styles closer together.

@alexandergall what do you think? you wrote most of the affected code.

alexandergall:
It's certainly good to make the usage consistent across all apps. My only objection would be that a lot of the `require` code looks a bit awkward because most modules provide a single app that tends to have the same name as the module, e.g. 

`local baz = require("foo.bar.baz").baz`

A nice one is

`local RateLimiter = require("apps.rate_limiter.rate_limiter").RateLimiter`

But I can live with that :)

lukego:
Yeah. It would sure be nice to do away with those require lines entirely. Perhaps we could cook up a simple convention for giving the app name as a string and for `config.lua` to infer an app name.

One step at a time.. :-)

eugeneia:
Just so it shows up on the record: The precedent is set by `basic_apps`, `intel10g` and `pcap` which are modules each providing multiple apps. So this is not *only* consistency but also the most common denominator.

PR #340, opened 2014-12-26
Update authorized_keys

yyang13:
yyang13's ssh key


lukego:
Welcome! You should be able to access `chur` in the lab now with `ssh -p 2020 root@lab1.snabb.co` and then create your own user account.

PR #341, opened 2015-01-08
Add support for jit.dump and jit.profile to a file.

dpino:
src/designs/neutron/snabbnfv-bench has support for gettting a profile session via NFV_PROF env variable. This patch adds support for writing the result of the profile to a file (NFV_PROF_FILE) and implements support for dumps (NFV_DUMP, NFV_DUMP_FILE).

### DIFF
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -25,18 +25,19 @@ ffi.cdef[[
 local usage = [[
 Usage: snabb [options] <module> [args...]
 Available options are:
--P pathspec  Set library load path (Lua 'package.path').
--e chunk     Execute string 'chunk'.
--l name      Require library 'name'.
--t name      Test module 'name' with selftest().
--R           Start interactive Snabb REPL.
--dm          Enable developer mode. Enables debug prints and asserts.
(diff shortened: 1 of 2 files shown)


lukego:
SnabbBot has caught a real bug here: undeclared global variable.

dpino:
The undeclared var was actually not needed.

The main goal of the patch is to provide a way to do a -jdump using an environment variable like *NFV_PROF* does. It also adds support to write down the profiling and dumping to a file, which was missing in this case.

While taking a look at the undeclared var thing, I noticed the -jdump in /core/main.lua works different than -jp, although both commands have a similar syntax in LuaJIT.

```
sudo ./snabb -jp=Flvm,/tmp/output apps.fuzz.fuzz
```
```
sudo ./snabb -jdump /tmp/output apps.fuzz.fuzz
```
It's not possible to pass parameters to jdump (sometimes we run it with +rsT), so I refactored it to have the same syntax as -jp.

```
sudo ./snabb -jdump=+rsT,/tmp/output apps.fuzz.fuzz
```

PR #343, opened 2015-01-15
[neutron-sync] Use constant disk space for neutron-sync-master/neutron-sync-agent

eugeneia:
Uses `git commit --amend`, `git reflog expire` and `git prune` to operate sync master and agent with constant disk space usage.

Based on this test: https://gist.github.com/eugeneia/babaa842e6e34fcf5fc1

Note: I didn't try to prune periodically (instead of after each commit/pull) because I assume that pruning a single-commit-repo will always be a cheap operation.

### DIFF
--- a/src/designs/neutron/neutron-sync-agent
+++ b/src/designs/neutron/neutron-sync-agent
@@ -37,13 +37,15 @@ initial=true
 # Loop pulling/cloning the repo.
 while true; do
     if [ ! -d $NEUTRON_DIR ]; then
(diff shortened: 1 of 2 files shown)


lukego:
Awesome!

@n-nikolaev is the one field-testing the changes related to the Neutron code. We should probably find a nice workflow for him to test and review these changes and push them to me. (Currently I merge them onto master and then they land in his lap by surprise when he does a `git pull` :)).

eugeneia:
He could just add my Snabb fork (on GitHub) as a remote and then pull this branch into a branch on his repository? That's how I access third party branches and it works quite well as far as I can tell.

PR #345, opened 2015-01-18
Added code for Getting Started Guide

jeffbrl:
Added first configuration script, pcap file, and sprayer app for following the Getting Started Guide examples.


src/apps/sprayer/sprayer.lua
```diff
@@ -0,0 +1,36 @@
...
+
+   while not link.empty(i) and not link.full(o) do
+      self:process_packet(i, o)
+      self.packet_counter = self.packet_counter + 1
+   end
+end
+
+function Sprayer:process_packet(i, o)
+   local p = link.receive(i)
+   local iov = p.iovecs[0]
```
lukego: You could remove the `iovec` access because it is not used.

lukego:
Nice one!

The SnabbBot CI system says that tests failed but this seems to have been an error on the CI. `make` and `make test` work for me on `chur`.

I made one comment: would you like to change that or merge as-is?

jeffbrl:
I borrowed some logic in the ``process_packet`` method from @plajjan's DDoS app with his permission.  I should have spotted the ``iovec`` line. I'll fix it.

jeffbrl:
Should commits to address issues noted in PR comment be squashed by the PR author?

lukego:
Merged!

Historically I have asked for commits to be rebased before merge with fixups squashed. However, I spent this weekend deeply geeking out on Git workflows and I am changing my mind. On the one hand a rebase/squash will make the history shorter but on the other hand it won't actually be the history any more and we lose context such as this discussion. (I'll post to snabb-devel with latest Git workflow ideas for Snabb Switch.)

plajjan:
For the future.. I don't think the git history need to reflect this very exact discussion, so I would have voted for squash in this case. I think a detailed commit log can explain many peculiarities and similar that would otherwise be seen in a non-squashed history.

This is a simple mistake, why should we let the history reflect a mistake? Had it not been git we would be sending around .patch files and certainly would ask over mail (or whatever medium we use to exchange said .patch files) to fix the mistake and resubmit, or possibly the person merging would simply fix it before merging.

lukego:
I am flexible. If somebody prefers to close a pull request, rebase onto a new branch, and send a new cleaner pull request, that is absolutely fine with me.

On the other hand, rebasing does destroy context, and an authentic history will contain lots of mistakes.

How about if we try to model our history on the Linux kernel and try to follow Linus's tips:
http://www.mail-archive.com/dri-devel@lists.sourceforge.net/msg39091.html

PR #348, opened 2015-01-22
IPC method using file-backed shared memory

alexandergall:
I'd like to pick up on a suggestion for sharing counters with other processes by Luke in the thread 
https://groups.google.com/d/msg/snabb-devel/-Ka_gPfd88k/tL0wTqxcizcJ

I commented back then that I was thinking about something like that to implement an interface to a SNMP agent. I have done this by generalzing Luke's code from https://github.com/lukego/snabbswitch/blob/counter/src/core/counter.lua

I think that this (or something similar) might be worth adding to the Snabb core.

The "mib" subclass supports all basic SMIv2 types.  And, yes, I have working code for my VPN app and a SNMP sub-agent (written in Perl) that allows me to monitor my system with our standard monitoring system via SNMP (based on Nagios). I'm really pleased with that :)

I have also implemented Luke's "counter" module on top of it, which is a trivial extension of the shmem base class. The difference to his code, apart from using uint64 instead of double (open for discussion), is that he pre-allocates a memory region of a certain size and falls back to heap memory when it is exhausted, while my code starts with a zero-length mapping and extends it when new objects are registered. It will fail hard (assert)  when the underlying mmap() fails or there is a problem with the data or index file. Registering is expected to not be used in the fast path.

I suggest to put this under lib/ipc/shmem in anticipation of other IPC mechanisms that we may add in the future. But I don't have a strong opinion about where to put it.

### DIFF
--- a/src/lib/ipc/shmem/shmem.c
+++ b/src/lib/ipc/shmem/shmem.c
@@ -1,0 +1,38 @@
+#include <sys/mman.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <stdbool.h>
+
+bool shmem_unmap(void *mem, size_t size) {
+  if (munmap(mem, size) == -1) {
+    perror("munmap");
+    return(false);
+  }
+  return(true);
+}
+
+// Note: the io.* file handle passed to us is converted to a Unix
+// filehandle (FILE *) by LuaJIT
+char *shmem_grow(void *fh, void *old_mem, size_t old_size, size_t new_size) {
+  int fd;
+  void *mem;
+
+  if (old_mem != NULL) {
+    if (shmem_unmap(old_mem, old_size) == false) {
+      return(NULL);
+    }
+  }
+  fd = fileno(fh);
+  if (ftruncate(fd, new_size) == -1) {
+    perror("ftruncate");
+    return(NULL);
+  }
+  if ((mem = mmap(old_mem, new_size, PROT_READ|PROT_WRITE, MAP_SHARED,
+		  fd, 0)) == MAP_FAILED) {
+    perror("mmap");
+    return(NULL);
+  }
+  return((char *)mem);
+}
+
--- a/src/lib/ipc/shmem/shmem.lua
+++ b/src/lib/ipc/shmem/shmem.lua
@@ -1,0 +1,267 @@
+-- The shmem base class provides a simple IPC mechanism to exchange
+-- arbitrary cdata objects with other processes through a file-backed
+-- shared memeory region.
+--
+-- A new empty region is created by calling the constructor with a
+-- filename and optional directory, where the file should be created.
+--
+--  local shmem = require("lib.ipc.shmem.shmem")
+--  local foo = shmem:new({ filename = "foo", directory = "/tmp" })
+--
+-- If omitted, the directory defaults to "/tmp/snabb-shmem".  The
+-- constructor creates an empty file called the "data file" with the
+-- given name and maps it into the processe's virtual memory.  In
+-- addition, it creates an "index file" by appending the suffix
+-- ".index" to the data file name.  The first line of the index file
+-- contains a string that identifies the name space to which the
+-- memory region belongs, followed by a colon, followed by an integer
+-- version number.  The name space indicates how the rest of the index
+-- file needs to be interpreted.  A name space is tied to the subclass
+-- that implements it (but a subclass may inherit the name space of
+-- its ancestor class).  The version number allows for changes of the
+-- index format within the name space.
+--
+-- The base class provides the name space "default", i.e. the header
+-- line of the index file for version 1 contains the string "default:1".
+--
+-- The default name space contains a single line in the index file for
+-- every object stored in the memory region.  The order of the
+-- descriptions must be the same as that of the objects.  A
+-- description consists of an arbitrary name, followd by a colon,
+-- followed by the length of the corresponding object in bytes.  For
+-- example, the index file
+--
+--   default:1
+--   foo:4
+--   bar:7
+--
+-- Describes a memory region of length 11, which contains an object
+-- named 'foo' that consists of 4 bytes starting at offset 0 and an
+-- object named 'bar' consisting of 7 bytes starting at offset 4.  The
+-- type of the object is implied by its name and is not part of the
+-- description in the index. Each name must be unique.
+--
+-- An object is added to the region by calling the method register(),
+-- which takes a string, a ctype object and an optional value as
+-- arguments.  The ctype must refer to a complete type such that the
+-- size of the object is fixed when it is added to the index.  The
+-- following example adds an unsigned 32-bit number called "counter"
+-- and a struct named "bar":
+--
+--  local counter_t = ffi.typeof("uint32_t")
+--  local bar_t = ffi.typeof("struct { uint8_t x; char string[10]; }")
+--  foo:register("counter", counter_t, 42)
+--  foo:register("bar", bar_t)
+--
+-- The index file now contains
+--
+--  default:1
+--  counter:4
+--  bar:11
+--
+-- The contents of the objects can be changed by the set() method
+--
+--  foo:set("counter", 1)
+--  foo:set("bar", bar_t({ x = 1, string = 'bar' }))
+--
+-- The provided value must be of the correct type or must be
+-- convertible to it.  The assignment is performed by de-referencing
+-- the address of the object as a pointer, equivalent to (where
+-- "ctype" is the ctype object passed to the register() method)
+--
+--  ffi.cast(ffi.typeof("$*", ctype), address)[0] = value
+--
+-- The get() method returns just a reference to the object
+--
+--  ffi.cast(ffi.typeof("$*", ctype), address)[0]
+--
+-- For certain types, this will result in a Lua object, complex data
+-- types are represented as references
+--
+--  print(type(foo:get("counter")))  --> number
+--  print(type(foo:get("bar")))      --> cdata<struct 403 &>: 0x7f2ce4efb004
+--
+-- Be aware that assignments to the latter will change the underlying
+-- object, while simple objects obtained from get() are distinct from
+-- the underlying object.
+--
+-- To manipulate any type of object "in place", one first obtains a
+-- pointer to the object by calling the ptr() method and de-references
+-- that pointer
+--
+--  local c = test:ptr("counter")
+--  c[0] = 42
+--  print(test:get("counter")) --> 42
+--
+module(..., package.seeall)
+
+local ffi = require("ffi")
+local C = ffi.C
+
+require("lib.ipc.shmem.shmem_h")
+
+local shmem = subClass(nil)
+shmem._name = "shared memory base class"
+
+---- Class variables
+-- Should be overridden by derived classes
+shmem._namespace = "default:1"
+-- The character(s) used as field separator in the index file.  Can be
+-- overriden by derived classes.  Object names that contain the field
+-- separator are considered illegal by the register() method.  A value
+-- of '' suppresses the length field in the index file.
+shmem._fs = ':'
+
+local defaults =  {
+   directory = '/tmp/snabb-shmem',
+}
+
+---- Class methods
+
+-- Options:
+-- { filename = <filename>,
+--   [ directory = <directory>, ] Default: /tmp/snabb-shmem
+-- }
+--
+function shmem:new (options)
+   assert(options and options.filename) 
+   local o = shmem:superClass().new(self)
+   local dir = options.directory or defaults.directory
+   if dir ~= '' then
+      o._data_filename = table.concat({ dir, options.filename}, '/')
+   else
+      o._data_filename = options.filename
+   end
+   o._data_fh = assert(io.open(o._data_filename, 'w+'))
+   o._index_filename = o._data_filename..".index"
+   o._index_fh = assert(io.open(o._index_filename, 'w+'))
+   assert(o._index_fh:write(o._namespace, '\n'))
+   assert(o._index_fh:flush())
+   o._size = 0
+   o._base = nil
+   o._objs = {}
+   return o
+end
+
+---- Instance methods
+
+-- Append an object with the given name and ctype to the shared memory
+-- region and add its description to the index file.  If a value is
+-- supplied, the object is initialized with it via the set() method.
+--
+-- The method returns a pointer to the object via the ptr() method.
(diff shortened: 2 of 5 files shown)


alexandergall:
Darn, the procedural interface of counter.lua is broken.  Need to fix that before merging.

lukego:
This looks completely awesome :-)

src/lib/ipc/shmem/shmem.c
```diff
@@ -0,0 +1,38 @@
...
+  return(true);
+}
+
+// Note: the io.* file handle passed to us is converted to a Unix
+// filehandle (FILE *) by LuaJIT
+char *shmem_grow(void *fh, void *old_mem, size_t old_size, size_t new_size) {
+  int fd;
+  void *mem;
+
+  if (old_mem != NULL) {
```
lukego: Is this function vulnerable to creating dangling pointers due to remapping memory? If so, could we fix that?

Example:

1. Lua code calls `counter` module and gets the address of a counter for repeated use.
2. `munmap()` invalidates the old mapping and then `mmap()` returns a different address (for whatever reason).
3. Lua code tries to use the remembered counter value and causes a `SIGSEGV` because it is no longer mapped.

Is this a valid error scenario? (I do like the idea of counters having stable virtual addresses: then I expect they can be incremented very cheaply indeed and so we need not be shy of doing so in the fast path).

If indeed this would be a problem, could we solve it simply by skipping the `munmap()`? Then even if `mmap()` happened to choose a new address then all previous mappings would remain valid as separate mappings?

src/lib/ipc/shmem/shmem.lua
```diff
@@ -0,0 +1,267 @@
...
+--
+-- An object is added to the region by calling the method register(),
+-- which takes a string, a ctype object and an optional value as
+-- arguments.  The ctype must refer to a complete type such that the
+-- size of the object is fixed when it is added to the index.  The
+-- following example adds an unsigned 32-bit number called "counter"
+-- and a struct named "bar":
+--
+--  local counter_t = ffi.typeof("uint32_t")
+--  local bar_t = ffi.typeof("struct { uint8_t x; char string[10]; }")
```
lukego: I really like the generality of this design. Great that it is so easy to store structured data instead of single scalar values, and that the way this data is interpreted is flexible and open. (This data structure is starting to remind me of [Forth's Dictionary](http://en.wikipedia.org/wiki/Forth_(programming_language)#Structure_of_the_language).)

How do ensure that readers see a consistent value? The `counter` module depends on x86_64 loading and storing 64-bit values atomically so that reads and writes won't conflict. Do we need some kind of protocol to extend this to structured data, in case a write happens part-way through a read? (I wonder what would work if so.. a double buffer?)

src/lib/ipc/shmem/shmem.c
alexandergall: On Thu, Jan 22, 2015 at 5:51 PM, Luke Gorrie <notifications@github.com> wrote:
>
> In src/lib/ipc/shmem/shmem.c:
>
> > +
> > +bool shmem_unmap(void *mem, size_t size) {
> > +  if (munmap(mem, size) == -1) {
> > +    perror("munmap");
> > +    return(false);
> > +  }
> > +  return(true);
> > +}
> > +
> > +// Note: the io.* file handle passed to us is converted to a Unix
> > +// filehandle (FILE *) by LuaJIT
> > +char *shmem_grow(void *fh, void *old_mem, size_t old_size, size_t new_size) {
> > +  int fd;
> > +  void *mem;
> > +
> > +  if (old_mem != NULL) {
>
> Is this function vulnerable to creating dangling pointers due to remapping memory?

Yes. That's why the shmem class keeps the locations of objects as an
offset relative
to the base address. Also see the comment of the base() method.

Now, this makes the return value of the register() method rather
useless :(  I've added
that as an afterthought to implement the semantics of your original
counter.lua code.
Clearly, I did not think this through.

> If so, could we fix that?

>
> Example:
>
> Lua code calls counter module and gets the address of a counter for repeated use.
> munmap() invalidates the old mapping and then mmap() returns a different address (for whatever reason).
> Lua code tries to use the remembered counter value and causes a SIGSEGV because it is no longer mapped.
>
> Is this a valid error scenario? (I do like the idea of counters having stable virtual addresses: then I expect they can be incremented very cheaply indeed and so we need not be shy of doing so in the fast path).

For counters, we could simply register all counters first, then use
the base() method to
create a pointer to the array of counters and use indices to access
them. The indices could
even be used as names through an "enum"-style table. This would all be
resolved at compile
time and should be as efficient as using pointers with the current code.

That wouldn't work for segments with mixed types and the return value
of register() would still
be unsafe to use.

>
> If indeed this would be a problem, could we solve it simply by skipping the munmap()? Then even if mmap() happened to choose a new address then all previous mappings would remain valid as separate mappings?

Doesn't mmap() necessarily create a new mapping when the old one isn't
munmap()-ed? Not sure.

I wonder how likely it is for mmap() to chose a new address when
explicitely asking for
a mapping to the same address. We'd still have to deal with it, even
it is unikely.

src/lib/ipc/shmem/shmem.lua
alexandergall: On Thu, Jan 22, 2015 at 6:12 PM, Luke Gorrie <notifications@github.com> wrote:
>
> In src/lib/ipc/shmem/shmem.lua:
>
> > +--
> > +-- Describes a memory region of length 11, which contains an object
> > +-- named 'foo' that consists of 4 bytes starting at offset 0 and an
> > +-- object named 'bar' consisting of 7 bytes starting at offset 4.  The
> > +-- type of the object is implied by its name and is not part of the
> > +-- description in the index. Each name must be unique.
> > +--
> > +-- An object is added to the region by calling the method register(),
> > +-- which takes a string, a ctype object and an optional value as
> > +-- arguments.  The ctype must refer to a complete type such that the
> > +-- size of the object is fixed when it is added to the index.  The
> > +-- following example adds an unsigned 32-bit number called "counter"
> > +-- and a struct named "bar":
> > +--
> > +--  local counter_t = ffi.typeof("uint32_t")
> > +--  local bar_t = ffi.typeof("struct { uint8_t x; char string[10]; }")
>
> I really like the generality of this design. Great that it is so easy to store structured data instead of single scalar values, and that the way this data is interpreted is flexible and open. (This data structure is starting to remind me of Forth's Dictionary.)

Reminds me of a Forth extension module I once had for my C64 :)  I did
some programming in Forth, but forgot most of it in the 30+ years
since :(

>
> How do ensure that readers see a consistent value? The counter module depends on x86_64 loading and storing 64-bit values atomically so that reads and writes won't conflict. Do we need some kind of protocol to extend this to structured data, in case a write happens part-way through a read? (I wonder what would work if so.. a double buffer?)

I thought about that. My initial idea is to use semaphores. This could
be a named semaphore or an unnamed one that is placed directly in the
shared memory. I'm not sure about the performance implications, but
implementation would be simple, I guess.

javierguerragiraldez: I think the easiest would be some kind of double-buffering. Either with an atomic value to indicate which copy is the current one, or by having a monotonically increasing counter on the struct itself and just using the largest one.

It might be necessary some reordering barrier to make sure the critical atomic flag (the current index or the generation counter) is seen as updated last by every party.  Does anybody know what kind of non-reordering guarantees the x86 model provides?  I don't think `CAS` instructions apply too well on `mmap()`'ed blocks.

alexandergall:
futex(7) could just be what we need in this case. With a futex, there is no system call overhead in the uncontended case, which I assume will be, by far, the common case. I would suggest to use one futex per shared segment, but it would be easy to do that on a per-element basis as well.

lukego:
I am a little scared of locks and especially fine-grained locks. The model we have now keeps the traffic process independent of the monitoring tools. Locks would allow the monitors to screw up the traffic.

How about if:

* Traffic process is independent of monitoring tools (never knows if they are running or not).
* Monitoring tools can always access counters instantly.
* Monitoring tools see a consistent snapshot that is refresh only when they want.

That would be a great solution for our current use cases, right?

This could be done via the filesystem. The traffic process updates counters in its own private address space and then periodically writes snapshots to the filesystem (e.g. ramdisk). Each snapshot atomically replaces the old one (`mv counter.dat.new counter.dat`). Monitoring tools open a `counter.dat` snapshot at any time and can read it for as long as they want. Monitoring tools refresh counters by closing the file descriptor they have and opening the latest file (on the same filename).

What do you think?

alexandergall:
One could safeguard against a lockup caused by a faulty monitor by using a timeout for the FUTEX_WAIT operation and deactivate locking in the traffic process when this happens. But, yes, this could get a bit complicated.

What frequency for snapshots do you have in mind?  On my routers, I like to be able to see counter updates in near real-time when I'm logged in and looking at things. For the purpose of graphing with the standard 5 minute intervals, you could obviously tolerate a much bigger lag. For my SNMP data, I'd like to have near real-time data again, e.g. for the operational status of my VPN tunnels, even though most of the data is static. In any case, I see demand for fairly short snapshot intervals and this is probably a problem.

lukego:
I'd shoot for a 1ms update interval, and settle for 10ms or 100ms if that turned out to make a difference.

To write a snapshot should be quite fast: `open()`, `write()`, `close()` `link()`. This is comparable to the work that Apache does to serve a HTTP request.

lukego:
NB: I suppose I mean `rename()` rather than `link()`.

alexandergall:
Ok, that would certainly be sufficient :) I would have guessed that the performance impact would be more severe. Still, it looks a tad excessive to me...

Here's a crazy idea: let's declare the interface to not guarantee consistency of data for the consumer. The hypothesis here is that such events are very rare (are they?). The consumer can chose to accept this or implement checks that reduce the likelihood of accepting wrong data, e.g. reading the same object twice or comparing it with previous data to spot "outliers".  Too crazy?

lukego:
FYI: Quick benchmark suggests that the stat file on tmpfs can be rewritten around 100K times per second. That is 100us mean time and with uncertain variance. That seems a little too expensive to do on the thread that is processing packets, and to introduce a background thread seems excessively complex.

Here is the test for posterity anyway:

```
luke@interlaken:~$ cat filetest.c
#include <assert.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>

#ifndef RUNS
#  define RUNS 10000
#endif

#ifndef BYTES
#  define BYTES 8192
#endif

int main(int argc, char **argv) {
  int i;
  char *data;
  data = (char*)malloc(BYTES);
  assert(data != NULL);
  for (i = 0; i < RUNS; i++) {
    int fd;
    fd = open("/tmp/stat.new", O_CREAT|O_WRONLY);
    assert(fd >= 0);
    assert(write(fd, data, BYTES) == BYTES);
    close(fd);
    assert(rename("/tmp/stat.new", "/tmp/stat") == 0);
  }
}
luke@interlaken:~$ gcc -o filetest filetest.c && sudo time ./filetest
0.02user 0.49system 0:01.02elapsed 50%CPU (0avgtext+0avgdata 460maxresident)k
0inputs+160000outputs (0major+155minor)pagefaults 0swaps
```

lukego:
Here is another idea based on Javier's double-buffer-with-counter idea above:

* Counter file has a `uint64_t serialno` field initialised to 0.
* Traffic process uses a double-buffer: writes counters in private memory and periodically syncs to shared memory with `memcpy()`.
* Sync is in three steps: increment serial number; memcpy; increment serial number.

The consumer can then choose to do either a simple non-atomic read or a more complex atomic read.

atomic read algorithm:

* Wait until serial number is even. (No update in progress.)
* Read value(s).
* Check if serial number is the same: if yes then success, if no then -EAGAIN.

lukego:
If none of these schemes really seems ideal than I would agree that it is better to keep it simple and forget about atomic reads for now :).

lukego:
ARGH!

Who knew that `/tmp` is not mounted on `tmpfs` on ubuntu? The measurement of 100us update time above is for a normal file system on a spinning disk.

Testing on `/run/` instead (which *is* tmpfs) looks much nicer: I can do 150K updates in one second i.e. 6.7us per update. That sounds like something that one could realistically do e.g. once every 10ms.

Let me humbly repurpose that solution for your consideration :-)

alexandergall:
Correct me if I'm wrong, but the double-buffer method you describe can't guarantee atomicity without using a true atomic test-and-set operation, can it? Still not sure how far we should take this.

lukego:
I confess that I have gotten carried away :-).

How about if we keep the code as you have it without having atomic reads of compound datatypes if you don't need that.

If we need atomic reads in the future we have a plausible simple solution with tmpfs snapshot files.

javierguerragiraldez:
> Correct me if I'm wrong, but the double-buffer method you describe can't guarantee atomicity without using a true atomic test-and-set operation, can it? Still not sure how far we should take this.

AFAIK, test-and-set are critical to arbitrate between two or more writers.  If the writer/reader roles are well-defined, consistency checks (like Luke's proposal) can be enough.

In double-buffer techniques (like the one described in LMDB docs) you only need to guarantee that the 'guard' value (current index or version counter) is modified last, as seen by every party.  I've read somewhere that even if x86 CPUs have heavy instruction reordering, memory writes are guaranteed to be in-order.  But I don't really trust that kind of affirmations without some spec definitions, so it might be better to add some kind of memory barrier.  OTOH, most of the non-flush memory barrier primitives in GCC are NOPs in x86.

alexandergall:
Ack.  I see different use cases with respect to requirements (perfect vs "probabilistic" integrity) as well as access models (pure reader/writer vs mixed access).

I would classify my SNMP case as a pure reader/writer (producer only writes, consumer only reads) where probabilistic integrity is good enough (most data is static, writes are rare and reads infrequent, e.g. seconds or minutes apart). Couners for traffic statistics appear similar to me.

An example for a pure reader/write with a requirement for perfect integrity would be, for example, if we used a separate Snabb process to manage the ARP/neighbor cache and have traffic-forwarding processes performing lookups in these tables.

We might end up with different arbitration schemes and leave it up to the apps to chose on, including the option to use none.

lukego:
Merged. Even if there is an issue with the `counter` interface we can fix that when we add code that uses it.

alexandergall:
On Wed, Feb 4, 2015 at 7:13 AM, Luke Gorrie <notifications@github.com> wrote:
>
> Merged. Even if there is an issue with the counter interface we can fix that when we add code that uses it.

Ok. I'll probably suggest a change for the result of the register()
method, which is useless (and misleading) as it is now.

PR #351, opened 2015-01-27
Update Authorize keys

hcnimkar:


hcnimkar:
SSH key has been added to enable login to chur server

PR #356, opened 2015-01-28
NFV config fuzz

eugeneia:
This adds a fuzzing function to the NFV selftest. The fuzzer generates semi-random NFV configurations that allow a successful `iperf` run between two VMs to test the configuration. The fuzzer's output can be controlled using a "spec" file which might look like this:

```
return { ingress_filter = 2, -- generate two random ingress filter rules (in addition to rules for iperf)
 egress_filter = 2, -- generate two random egress filter rules (in addition to rules for iperf)
 tunnel = true, -- do generate a tunnel config
 tx_police_gbps = 10 } -- do generate random value 1...10 (rx_police_gbps is also recognized)
```

If a key is omitted or `nil` then the config part in question is omitted.

### DIFF
--- a/src/designs/neutron/snabbnfv-fuzz
+++ b/src/designs/neutron/snabbnfv-fuzz
@@ -1,0 +1,19 @@
+#!/usr/bin/env snabb
+
+local lib  = require("core.lib")
+local fuzz = require("lib.nfv.fuzz")
+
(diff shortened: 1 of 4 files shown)


andywingo:
Nice!  You might like http://luatime.org/archives/2015/01/27/generating-arbitrary-test-data for comparison.

PR #358, opened 2015-01-29
Rename snabbfv-traffic's rx_police_gbps to tx_police_gbps and add proper rx_police_gbps

eugeneia:
Rename snabbfv-traffic's rx_police_gbps to tx_police_gbps and add proper rx_police_gbps. I started a mailing list thread about my woes actually testing rate limiting but I believe solving these is not in the scope of this PR. Nevertheless this branch extends the NFV selftest with a function `test_rate_limited` that allows testing of effective rate limiting. This functions is used to test the changes for ballpark sanity.


lukego:
Howdy! Sorry about the SLOW response! Could you please rebase (or merge from master) so that I can merge it cleanly? Thanks !

PR #361, opened 2015-02-05
Improve efficiency of lib/ipc/shmem mechanism

alexandergall:
As mentioned in the discussion of PR#348, the original version was a bit half-baked concerning access to the shared memory segment between re-mappings induced by register(). My suggestion in this PR is to hand out a table with pre-computed pointers after all objects have been registered to avoid the pointer-arithemtic in the get() and set() methods.

This is implemented by the new dictionary() method, which also returns tables to translate object names into "handles" (integer index into the main object pointer table). The register() method now returns the object's handle instead of the rather useless pointer.

The "procedural interface" of the counter module needed some fixing as well (is it useful at all?).

### DIFF
--- a/src/lib/ipc/shmem/counter.lua
+++ b/src/lib/ipc/shmem/counter.lua
@@ -7,7 +7,8 @@
 --
 -- The procedural interface can be used to avoid method-call overhead.
 -- The set of counters may also be accessed by treating the entire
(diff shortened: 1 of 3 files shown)


lukego:
Merged to master. Let's see what feedback there is when putting this to work in more programs.

PR #367, opened 2015-02-10
Merge: Intel10G: better initialization / closing

lukego:
Merge Javier's fine work on more robust initialization and shutdown of the Intel10G app: #360.

### DIFF
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -70,10 +70,11 @@ function M_sf:close()
       self:discard_unsent_packets()
       C.usleep(1000)
    end
-   if self.fd then 
-      pci.close_pci_resource(self.fd) 
+   if self.fd then
+      pci.close_pci_resource(self.fd, self.base)
       self.fd = false
    end
+   self:free_dma_memory()
 end
 
 --- See data sheet section 4.6.3 "Initialization Sequence."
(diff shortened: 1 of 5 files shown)


lukego:
@eugeneia I think this test failure is due to SnabbBot running an older version of LuaJIT than the one reference in our deps/luajit submodule. Can you check? (Is there a special step needed for SnabbBot and developers in general to pick up the latest LuaJIT? If so I wonder if we can solve that in the Makefile?)

This is the tell-tale error that should mean out-of-date LuaJIT:

    Invalid DMA address: 0x500438aedf00
    core/memory.lua:95: DMA address tag check failed

eugeneia:
The reason this happens (i think)is because SnabbBot will:

* clone the repo
* checkout the base (old submodule)
* fetch submodules
* build / test
* checkout HEAD
* build / test again

Strictly speaking its an error SnabbBot I guess, it should update the submodules after every checkout.

lukego:
Guys, how is intel10g reliability for you now on master?

I am seeing diverse errors and warnings when I run in a loop: `never got link up`, `sq_sq: missing packets`, `mq_sq: wrong proprtion`, `transmit counter did not increase`. So to me it seems like we currently have a high-priority bug that NIC initialization is not dependable?

lukego:
Here is output of 100 runs on master for me, with the new Interlake server: https://gist.github.com/lukego/bda37923f504e0c4f71f

Could be that I broke something in merge too.. whatever the case, we need to put this problem behind us quickly.

PR #370, opened 2015-02-12
authorized_keys: Add myself, daurnimator.

daurnimator:


lukego:
done! you shuold be able to login to chur with `ssh -p 2020 root@lab1.snabb.co` and create yourself an account.

PR #373, opened 2015-02-15
Intel10G: extensive reconfig tests and better initalization

javierguerragiraldez:
1. The `manyreconf()` function does two different tests: "VF-only": a hundred reconfiguration cycles for the same apps, the hardware is mostly set up once.  and "full PF cycle": a hundred full tear-down and reinitialization cycles.  Now the `selftest` does both.

1. The `autonegotiate_sfi()` was shown to do more harm than good when not needed, so it's now an optional part of the initialization.

1. When everything is going well, the NIC is initialized in typically 160msec, so the timeout is now 250msec, as per the Intel docs (down from 500).

1. If the NIC isn't up after 250msec, we reinitialize it, and in the vast majority of cases (well over 98%) it's done in 150-180msec.

1. In a few cases it won't come up until we do an `autonegotiate_sfi()`.  After that, it behaves normally.

1. Some cards work most of the time on the first init, some always need exactly one reinit.

1. A few cards (notably :03: in grindelwald) do work consistently at one reinit most of the time, but sometimes it needs many more reinits (with `autonegotiate_sfi()` every five reinits).  Eventually it does come up and work normally.  Typically after 24-28 reinits, a few times it took ~60 reinits, once it was 87!  In every case it did work and then a few hundred times more with exactly one reinit.

In short, the initialization is a lot faster most of the time, and when needed, it does a few extra tricks automatically and always gets it to work.  I've run the full test (with over 100 reconfigs of each type) 500 times in a row with the most troublesome card and it didn't gave up even once.

### DIFF
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -80,25 +80,12 @@ end
 --- See data sheet section 4.6.3 "Initialization Sequence."
 
 function M_sf:init ()
-   return self
-      :init_dma_memory()
-      :disable_interrupts()
-      :global_reset()
-      :wait_eeprom_autoread()
-      :wait_dma()
-      :init_statistics()
-      :init_receive()
-      :init_transmit()
-      :wait_enable()
-end
+   self :init_dma_memory()
 
-
-function M_sf:recheck()
+   self.redos = 0
    local mask = bits{Link_up=30}
-   if band(self.r.LINKS(), mask) == mask then
-      return self
-   else
-      return self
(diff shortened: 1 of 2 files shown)


lukego:
Great work!

I ran this in a loop for over an hour on interlaken, cycling between NICs, and it worked every time.

One day when we have spare time I would love to review our init vs. the Intel ixgbe driver and also try to optimize the selftest down to ~ 10 seconds. However, reliability is what really matters. Great work!

PR #375, opened 2015-02-17
PacketFilter API refactor

eugeneia:
Split up configuration table to contain `rules` key plus keys for global options (`state_track`, `state_check`). In my opinion this makes the API more straightforward to use.


src/apps/packet_filter/README.md.src
```diff
@@ -54,15 +74,11 @@ is allowed.
...
+the rule are tracked in the specified connection table.
 
 — Key **state_check**
 
-*Optional*. Checks if a packet belongs to a connection in a named
-connection table.  If used inside a rule, a packet must belong to an
-existing connection in addition to any other condition in the rule to pass.
-If used 'outside' any rule, if a packet belongs to an existing connection
-it's enough to pass the app.
+*Optional*. A string denoting a connection table. If set, packets must
```
lukego: Inconsistent with the **state_check** description above?

I expect the logic to be that a packet must match the rules *OR* match a known connection, but not both. The connection table is essentially a set of exceptions to the rules that allow all packets belonging to connections that we allowed to be established. (In other words, if we allow the first packet for a connection then we will allow all the future packets too, even if they don't match the rules e.g. because they are flowing in the opposite direction.)

eugeneia: This describes the following portion:

> If used inside a rule, a packet must belong to an
> existing connection in addition to any other condition in the rule to pass.

The "out of rule" behavior is described in the global `state_check` option.

lukego: Thanks for explaining.

I don't immediately see the use case for `state_check`. Could it be that `state_track` is all we need in the API? cc @javierguerragiraldez

src/apps/packet_filter/packet_filter.lua
```diff
@@ -354,17 +354,18 @@ local function generateConformFunctionString(rules)
...
-   if rules.state_check then
-      conntrack.define (rules.state_check)
-      T('if state_pass("', rules.state_check, '", buffer) then return true end')
+
+   if options.state_check then
+      conntrack.define (options.state_check)
+      T('if state_pass("', options.state_check, '", buffer) then return true end')
    end
 
+   local rules = options.rules or {}
```
lukego: Documentation says that **rules** is required, but it seems to be optional in the code?

src/apps/packet_filter/README.md.src
javierguerragiraldez: More than once I've found the iptables connection tracking behaviour to be limiting.  In short, it tracks _all_ packets on a single table and then you can write rules that check a packet.  In this implementation, both the tracking and checking can be used on a rule or globally, and also to use separate, named tables.  This allows the user to only track packets that pass a certain rule, or to discriminate different connections in different tracking tables.

If we wanted to only emulate iptables, we could use only the 'inside rule' `state_check`, and an implicit global `state_track`, both storing in a single table.

javierguerragiraldez:
> if we have no rules ({}) then it seems all packets should be dropped?

Right, a no-rules filter would drop everything.

Currently, the behaviour of a global `state_check` is to check before any rule, and if it finds a match, the packet is passed.  Therefore, a filter with only a global `state_check` and no rules means "pass only known connections", which I think is a pretty common case.

src/apps/packet_filter/README.md.src
lukego: That sounds reasonable. Thanks for explaining. If the PacketFilter has an API that makes sense and is accurately documented then that is excellent.

I am also thinking about the effect of changes to one part of the code base on other parts. The "packet filtering ecosystem" that we have today looks like this:

- PacketFilter app.
- snabbnfv-traffic process (uses and configures PacketFilter app).
- neutron2snabb program (translates Neutron configuration to snabbnfv-traffic).
- [Neutron API Extensions](https://github.com/SnabbCo/snabb-nfv/wiki/Neutron-API-Extensions) where we need a section explaining our interpretation of Neutron Security Groups.

So Max and I need to describe our Neutron configuration interface and make sure that it is correctly mapped onto the PacketFilter app.

src/apps/packet_filter/packet_filter.lua
eugeneia: I guess the code could complain at that point instead of defaulting, a PF without rules is probably never what the user wants and mostly a mistake.

eugeneia: @lukego I could make this more strict, but since we can not reliably check for the correct type of the rules table anyways... We say it must be a *list*, which we might want to call "array" because that's the closest thing PIL describes. We can't AFAIK easily check that rules is indeed an array though.

I think the current code is OK: The manual says do it that way and the code handles user errors on a range between forgiving (default to empty rule array) to undefined (I don't know what happens when the table has e.g. string keys, or non-table values, it might ignore those keys or throw up. I don't think that in this specific case there is imminent danger for an API user to rely on undefined behavior).

PR #377, opened 2015-02-19
[lib/nfv/selftest.sh] Make TESTCONFPATH process specific ($$)

eugeneia:
This fixes a regression of the keep-nfv branch where a globally shared pathname (TESTCONFPATH) was erroneously introduced.

### DIFF
--- a/src/lib/nfv/selftest.sh
+++ b/src/lib/nfv/selftest.sh
@@ -28,7 +28,7 @@ fi
 
 export BENCH_ENV_PID
 
-TESTCONFPATH=/tmp/snabb_nfv_selftest_ports
+TESTCONFPATH="/tmp/snabb_nfv_selftest_ports.$$"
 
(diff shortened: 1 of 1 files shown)


src/lib/nfv/selftest.sh
```diff
@@ -70,6 +70,9 @@ function stop_bench_env {
 
     # Give VMs and snabbnfv-traffic time to shut down.
     sleep 5
+
+    # Clean up temporary config location.
+    rm -f "$TESTCONFPATH"
```
lukego: Can we do this on an [EXIT trap](http://redsymbol.net/articles/bash-exit-traps/) to make sure it happens even with `^C`? (Could be that this is how the whole `stop_bench_env` function should be invoked?)

eugeneia: Yes probably, I only learned about the existence of traps from `bench_env`, excuse my ignorance. I will open a PR to fix this.

PR #387, opened 2015-03-03
Remove garbage from core.apps.restart_dead_apps()

alexandergall:
The "action" table is allocated each time the function is called,
which creates a fair amount of garbage, even if the restart mechanism
isn't actually used, which is the default.

This commit includes a bypass of the function if use_restart is false.
If use_restart is true, the action table is allocated statically and
only the sub-tables that are used by the function are cleared.

The dead-app detection code probably contains more of this kind of
garbage, e.g. in the apply_config_action() function.

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -51,10 +51,13 @@ local function with_restart (app, method)
 end
 
 -- Restart dead apps.
+local actions = { start={}, restart={}, reconfig={}, keep={}, stop={} }
 function restart_dead_apps ()
+   if not use_restart then return end
    local restart_delay = 2 -- seconds
-   local actions = { start={}, restart={}, reconfig={}, keep={}, stop={} }
    local restart = false
+   actions.restart[1] = nil
+   actions.keep[1] = nil
 
    -- Collect 'restart' actions for dead apps and log their errors.
    for i = 1, #app_array do


src/core/app.lua
```diff
@@ -51,10 +51,13 @@ local function with_restart (app, method)
 end
 
 -- Restart dead apps.
+local actions = { start={}, restart={}, reconfig={}, keep={}, stop={} }
 function restart_dead_apps ()
+   if not use_restart then return end
    local restart_delay = 2 -- seconds
-   local actions = { start={}, restart={}, reconfig={}, keep={}, stop={} }
    local restart = false
+   actions.restart[1] = nil
```
lukego: What does the `actions.restart[1] = nil` do?

If this is a trick to empty an array then is there a risk that filling in the first element will revive all of the original following elements?

alexandergall: Oops, that was a little simple-minded by me. It will also not be reflected in #actions.restart. I guess the proper way to clear the array is to traverse it and set all elements to nil.

I'm sure this is a Lua FAQ and I fell right into the trap :)

javierguerragiraldez: > I guess the proper way to clear the array is to traverse it and set all elements to nil.

In plain Lua it's the only way, but LuaJIT (2.1) includes [`table.clear(t)`](http://repo.or.cz/w/luajit-2.0.git/blob_plain/v2.1:/doc/extensions.html) that does exactly that.  Of course, it has a lot of warnings that it's only for very specific cases, usually it's better to let the GC do its magic, etc.

alexandergall: I would favour the traversal method, but maybe we shouldn't bother at all as long as that logic is bypassed when `use_restart` is `false`. The dead-app detection code has other performance issues that need to be addressed at some point.

PR #390, opened 2015-03-05
More logging bench_env

lukego:
Log the output of QEMU and snabbnfv traffic to files in `/tmp/`.

This is intended partly to help debug SnabbBot test failures.


lukego:
Closing this one.

The patch to redirect snabbnfv traffic output is ill-conceived:

1. The documentation says that the output will be logged to `SNABB_LOG0` and the patch makes this false (logging to a hardcoded location instead).
2. Consequence of the above: NFV selftest fails to locate the traffic log file and that causes it to think that the tunnel test case fails (because it doesn't find the message indicating success).

PR #391, opened 2015-03-09
ipchecksum.c: Optimized IP checksum library in C [DRAFT]

lukego:
Import Tony Rogvall's SIMD (x86 SSE) IP checksum routines.

This branch contains code for future reference: optimized checksum routines that are linked into the build process but not actually called from anywhere. This code can be integrated into Snabb Switch when somebody needs a fast checksum implementation running on the CPU. (I don't plan to finalize this code right now: if you need it, take it.)

I have only benchmarked this quite trivially but I see around 5 bytes per cycle of checksum throughput. This seems to be significantly faster than traditional serial checksum routines.

This code uses 128-bit SIMD operations. I am very curious to know whether we could achieve large/exponential speedups with 256-bit (AVX2) and 512-bit (AVX512) variants. (Perhaps the alignment requirements could be relaxed too.)

For bonus points this also includes incremental checksum update routines that can be used when modifying single 16/32 bit values.

Thank you Tony Rogvall (the famous Erlang/Forth hacker) and Juho Snellman (the famous Lisp/C++ hacker) for kindly providing this code as open source.


javierguerragiraldez:
I've just drafted a simple checksum app using this.  Some observations:

- It only processes packets with flags `C.PACKET_NEEDS_CSUM` set and `C.PACKET_CSUM_VALID` clear.  Anything else is just passed.

- Processed packets get the flags reversed to signal that the checksum is valid and no need for further processing.

- Incoming packets are just verified, and the flags are set accordingly if they pass.  If not, they're dropped.

- At the point I'm almost sure this is not how these functions should be used.  It seems the packets we'll see have part of the checksum already done (the pseudo header?); but `update_checksum()` starts by clearing the checksum field.

- The current test show 2.5Mpps, but it seems to be a limitation of `basic_apps.Source` and `basic_apps.Sink`.  A big hint for this is that I get the same number if all packets are passed or all are calculated, and the same results either on grindelwald or interlaken.

lukego:
Cool :-)

Yes, it seems we don't really have a nice synthetic benchmarking setup for taking meaningful measurements of app performance in isolation. I would like to expand `programs/snabbmark` to serve this goal one day.

On the other hand, microbenchmarks are frequently misleading anyway. The best way to benchmark this could be to use iperf between virtual machines and check for an impact on end-to-end throughput. The numbers will be preliminary anyway because this code is based on SSE instructions and we may be able to get double speed by porting it to AVX2 for Haswell (interlaken) and then double again porting it to AVX512 for Intel's next CPU.

Regarding the API, yes I think these routines are not exactly matched to the way Virtio-net does checksum offload. The routine expects any pseudo-checksum to be passed as a parameter (`initial`) but Virtio-net does this implicitly by storing the value in the `checksum` field of the packet (instead of 0). The effect is the same. This should be resolved in "some nice way". (To start with it should be enough to change the routine to not clear the checksum field.)

Zooming out a little bit, now is the time to decide and document the exact semantics of checksum offload metadata in `struct packet`. We have two flags (`PACKET_NEEDS_CSUM` and `PACKET_CSUM_VALID`) and two numbers (`csum_start` and `csum_offset`). What do they mean and what do Snabb Switch users (people building app networks, people developing apps, etc) need to know about them?

Here is a suggestion for `PACKET_NEEDS_CSUM`:

* `PACKET_NEEDS_CSUM` means that one checksum operation is needed before transmission: the ones-complement checksum from `csum_start` until the end of the packet needs to be calculated and then stored at `csum_offset`.
* All apps that transmit packets externally need to take this flag into account. The simplest way to do this is to call a `finish_packet()` library function that finishes the checksum with SIMD if needed. The alternative is to use some offload that is available to that specific app e.g. based on a NIC or Virtio-net.

and for `PACKET_CSUM_VALID`:

* `PACKET_CSUM_VALID` is an optional hint to say that certain checksums are already known to be correct.
* The headers this refers to would initially be defined as exactly the same ones that Virtio-net uses. The exact details need to be checked, but approximately the outermost IP and TCP/UDP headers.
* All apps that modify packet payload need to be aware of this flag. The simplest way is to clear the `PACKET_CSUM_VALID` flag when you modify the packet. The alternative is to make sure that this flag is true when set e.g. by calling `checksum_update_incremental_32()` to incrementally update the checksum when modifying one field of packet payload.

In both cases the concept of checksum offload would be "under the hood". You need to think about it when writing an app to make sure you respect the flags. You don't need to think about it when writing an app network because you depend on all apps to handle this internally. So the code from the `Checksum` app would be refactored into a library routine that is called by all apps that need it.

lukego:
Here is yet another possible design for checksum offload in Snabb Switch:

Suppose that we removed all of the checksum offload metadata from `struct packet`. That is: packets moving around in Snabb Switch should already be checksummed and there is no native concept of offloading this.

Then apps don't have to think about checksum offload, but they do have to respect checksums when modifying packet data.

The Virtio-net app could still provide checksum offload from the VM perspective. Packets from the VM could be checksummed by the Virtio-net app before being sent onto the app network. Packets going to the VM could also be checksummed by the Virtio-net app and the result would be provided to the VM in the `virtio_net_hdr` struct.

This seems like an appealingly simple design that makes our `struct packet` even more streamlined. The question is whether this would require more checksum work and how expensive that would really be. Hard to say when it would be practical: Always? with AVX2 on Haswell? with AVX512 on future CPUs?

lukego:
For the sake of experiment I made an AVX2 version of the checksum routine and did a simple microbenchmark of cycles-per-packet for various "packet sizes" (really just an array). I compared the AVX2 code with the SSE code and with a reference implementation (from DPDK).

Here are the very preliminary results in cycles per packet:

PKTSIZE |   REF |  SSE |  AVX2
------- | ----- | ---- | -----
     64 |    26 |   37 |    39
    100 |    39 |   49 |    47
    256 |    83 |   65 |    53
    300 |   128 |   82 |    67
    512 |   182 |  123 |    76
    600 |   212 |  130 |   100
   1000 |   347 |  210 |   135
   1500 |   529 |  296 |   189

This was run on Interlaken (Haswell) and compiled with `gcc-4.9 -O2` with cycles measured by `perf stat`.

Thoughts:

* The cost of checksum seems to be significant. It's not trivial but it's not massive either.
* Checksum may be realistic to do on the CPU. Especially so if this code can be further optimized and if Intel SIMD performance continues to improve as expected.
* (Checksum offload is not free either because the hardware makes us parse the packet, create an extra descriptor, jump through hoops.)
* If bulk checksum is faster we could consider using segmentation offload to move larger chunks of data between VMs. This could potentially also be done in software and kept isolated in the Virtio-net code. (I don't suggest this now, but only observe that we will still have plenty of opportunities for further optimization within the simple design.)

I'll clean up the AVX2 port and submit it.

Standard disclaimer: This was a quick microbenchmark so the quality of the results is not very certain. However, I think that end-to-end benchmarks like iperf and packetblaster are more interesting for us anyway.

lukego:
You know @javierguerragiraldez, I am really warming up to the idea of deleting the checksum offload flags/fields `struct packet` and doing the whole thing locally in the Virtio-net module. If the guest sends us a packet that needs offload then we do that before putting it onto the app network. On the receive path we could either let the guest do the checksum or we could do that in the Virtio-net module and pass them the result.

This means one software checksum calculation on the VM->network path and potentially one on the network->VM path. So we have to measure the performance and potentially optimize the checksum code more.

The checksum code will be a bit tricky but it will make `struct packet` even simpler and that should help app developers.

What do you think, worth a try?

lukego:
(And of course if the performance is not competitive with the hardware checksum then we can stick with that. This would be an experiment that might simplify the code a lot or might be a dead-end.)

lukego:
I pushed an update that automatically selects between x86, SSE2, AVX2 based on the running CPU. For best results test on Interlaken which is the only lab server with the AVX2 CPU feature.

The API could possibly be improved e.g. to take a parameter saying where to store the checksum as the original did. also tip: if checksum doesn't work then check the endianness :).

javierguerragiraldez:
> I am really warming up to the idea of deleting the checksum offload flags/fields struct packet and doing the whole thing locally in the Virtio-net module.

it's working like that in my `ipchecksum-simd` branch!

performance on grindelwald seems to be roughly on par with the hardware checksum, around 9.8Gb/sec on jumboframes.

lukego:
Cool!

Let's focus on 1500-byte frames for now. That is the more important case. (Slow code can easily look fast with jumbo frames.)

javierguerragiraldez:
> Let's focus on 1500-byte frames for now. That is the more important case.

still way slower there: 3.6Gb/sec on grindelwald... it's profiling time!

lukego:
Code looks very good! I hope that approach pans out :-)

Is the current bottleneck Snabb or the guest? (Guest has to calculate its own checksums on received packets on this branch. Curious to know whether throughput is better if you tell the guest not to do that work by setting `PACKET_CSUM_VALID`.)

PR #395, opened 2015-03-10
Fix typo in mib shmem module

alexandergall:


eugeneia:
@alexandergall Can you rebase your failing PR's onto current master? I think they fail because of a previously fixed regression that's not in your current base.

PR #397, opened 2015-03-12
neutron_sync_master.sh: remove DB tables no longer used

thisisrmr:
The default value for DB_NEUTRON_TABLES variable contains: "qos_policies qoses portqosmappings" which is not used anymore by the neutron2snabb program.


lukego:
Nice catch! :-)

PR #403, opened 2015-03-13
Stateful NFV Packet Filtering

eugeneia:
Implement NFV packet filtering as described in [Neutron API Extensions/Stateless Packet Filtering](https://github.com/SnabbCo/snabb-nfv/wiki/Neutron-API-Extensions#stateless-packet-filtering).

E.g. if `vif_details` contains `packetfilter="stateless"` do stateless packet filtering. Otherwise do standard Neutron packet filtering.


src/program/snabbnfv/neutron2snabb/neutron2snabb.lua
```diff
@@ -59,8 +59,8 @@ function create_config (input_dir, output_dir, hostname)
                             { vlan = vif_details.zone_vlan,
                               mac_address = port.mac_address,
                               port_id = port.id,
-                              ingress_filter = filter(port, secbindings, secrules, 'ingress'),
-                              egress_filter = filter(port, secbindings, secrules, 'egress'),
+                              ingress_filter = filter(port, secbindings, secrules, 'ingress', vif_details.packetfilter),
```
lukego: The `packetfilter` setting should be in `profile` rather than `vif_details`.

`profile` contains configuration supplied by the user and `vif_details` contains information determined by the OpenStack code. (This is OpenStack Neutron semantics and not some funny invention of mine :-))

src/program/snabbnfv/neutron2snabb/neutron2snabb.lua
```diff
@@ -96,8 +96,13 @@ function filter (port, secbindings, secrules, direction)
          end
       end
    end
-   if #rules > 0 then return { rules = rules }
-                 else return nil end
+   if type == "stateless" then
```
lukego: Does this change mean that we now include a PacketFilter app in the app network, even if no rules are defined?

That would be nice in principle - consistency - but in practice I wonder if this is a potential performance regression when no filtering is defined, which is the most important case in practice (e.g. for hosting router-like devices).

(Is there also a potential functional regression e.g. in the difference between no filtering (nil) vs no rules ({})? I am not sure.)

lukego:
@n-nikolaev How do these changes look to you?

src/program/snabbnfv/neutron2snabb/neutron2snabb.lua
eugeneia: > Does this change mean that we now include a PacketFilter app in the app network, even if no rules are defined?

Yes. As per [Neutron/SecurityGroups](https://wiki.openstack.org/wiki/Neutron/SecurityGroups) specification:

> When there is no rule defined, all traffic are dropped.

src/program/snabbnfv/neutron2snabb/neutron2snabb.lua
```diff
@@ -59,8 +59,8 @@ function create_config (input_dir, output_dir, hostname)
                             { vlan = vif_details.zone_vlan,
                               mac_address = port.mac_address,
                               port_id = port.id,
-                              ingress_filter = filter(port, secbindings, secrules, 'ingress'),
-                              egress_filter = filter(port, secbindings, secrules, 'egress'),
+                              ingress_filter = filter(port, secbindings, secrules, 'ingress', vif_details.packetfilter),
+                              egress_filter = filter(port, secbindings, secrules, 'egress', vif_details.packetfilter),
                               gbps = vif_details.zone_gbps,
```
eugeneia: > The packetfilter setting should be in profile rather than vif_details.

This line looks a bit fishy too now that I think about it. We never use the `gbps` field in `nfvconfig`... (and some sort of `tx_police_gbps` is missing too).

src/program/snabbnfv/neutron2snabb/neutron2snabb.lua
lukego: OK.

We need a way to disable security group filtering entirely for a port. Do you know a suitable one?

n-nikolaev: @lukego Something like:

    neutron port-create ..... --no-security-groups

lukego: @n-nikolaev Does `--no-security-groups` work as expected today or do we need to implement that explicitly?

n-nikolaev: Works as expected - we get no rules in the generated port file.

eugeneia: @n-nikolaev No rules or no filter configs? E.g. the first case (an empty filter config) would drop all packets.

n-nikolaev: No filter config - all packets are passing effectively.

eugeneia: Ok I am pretty sure this branch will break this as of now. E.g. if we look at the code referenced above, even in case of `secbindings[port.id]` being false an empty filter config will be returned. I assume that `--no-security-groups` leads exactly to this case, so the config should only be returned when `secbindings[port.id]` is a true value, e.g. the return conditional should be moved up into `if secbindings[port.id] then`?

PR #409, opened 2015-03-17
Solarflare Rebased

eugeneia:
Rebased version of Hans' [solarflare](https://github.com/hanshuebner/snabbswitch/tree/solarflare) branch.

Removed were some obsolete commits. Some documentation fixes included in the branch were moved to my `documenttion-fixes` branch.

Includes a move of `program/solarflare` into `program/snabbmark` and adds a rudimentary `apps/solarflare/selftest.sh`.


lukego:
Looks like the Solarflare selftest needs to only run when hardware is available?

And we need to acquire another suitable Solarflare NIC for davos for use by SnabbBot CI.

eugeneia:
Uhm yeah I can explain that failure, I forgot about that case it should return a skipped code (43) I guess.

eugeneia:
@lukego On the other hand, that library is as far as I know unrelated to hardware. Should it be a submodule?

lukego:
Current thinking is to install the Solarflare library separately.

The Solarflare setup requires additional steps e.g. to have the correct kernel driver and firmware. Hans is working on a HOWTO for this. So to submodule the shared library would only solve part of the problem.

eugeneia:
@lukego I pushed a snabbmark test specific check that will skip the test if it can't load the solarflare module to solve this for now.

eugeneia:
Oh and I just realized I kind of broke `snabbmark solarflare`.... I added a 10 second timeout as a general error case, so that the benchmark makes sense as a selftest, but obviously that prevents you from testing with more packets.... just a second! :)

eugeneia:
Ok I added an optional `timeout` parameter to `snabbmark solarflare` to provide a safeguard for the selftest. Let's see what SnabbBot says but should be good to go.

src/lib/hardware/pci.lua
```diff
@@ -110,6 +113,24 @@ function set_bus_master (device, enable)
    C.close(fd)
 end
 
+--- ### Open a device
```
lukego: Do we really want to reintroduce `open_device()`? (I don't see it being used in this change set?)

src/program/snabbnfv/nfvconfig.lua
```diff
@@ -25,10 +31,11 @@ function load (file, pciaddr, sockpath)
       local name = port_name(t)
       local NIC = "NIC_"..name
       local Virtio = "Virtio_"..name
-      config.app(c, NIC, Intel82599, {pciaddr = pciaddr,
-                                      vmdq = true,
-                                      macaddr = mac_address,
-                                      vlan = vlan})
+      config.app(c, NIC, require(device_info.driver).driver,
```
lukego: (Just reminds me: Generally in Snabb Switch we need a better way to refer to apps and network interfaces. Global names for the driver like `apps.intel.intel10g.driver` are excessively verbose, and PCI addresses for network interfaces are excessively specific - won't work for e.g. raw socket on `eth0` even though some applications would work with one.)

src/lib/hardware/pci.lua
eugeneia: Good catch! This should be indeed removed.

src/apps/solarflare/solarflare.lua
```diff
@@ -0,0 +1,332 @@
...
+                          ffi.string(C.strerror(ffi.errno()))))
+   end
+   return rc
+end
+
+SolarFlareNic = {}
+SolarFlareNic.__index = SolarFlareNic
+SolarFlareNic.version = ef_vi_version
+
+driver = SolarFlareNic
```
lukego: I think the `driver = ...` lines in `solarflare.lua` and `intel10g.lua` could use a brief comment to say that these are to make it easier for driver modules to be used interchangably.

lukego:
SnabbBot didn't test?

eugeneia:
@lukego Yeah sorry was playing around with plotting our perf data and disabled it for a little while (and then forgot to reenable).

src/lib/hardware/pci.lua
```diff
@@ -54,9 +54,12 @@ function path(pcidev) return "/sys/bus/pci/devices/"..pcidev end
 -- Return the name of the Lua module that implements support for this device.
 function which_driver (vendor, device)
    if vendor == '0x8086' then
-      if device == '0x10fb' then return 'apps.intel.intel10g' end -- Intel 82599
-      if device == '0x10d3' then return 'apps.intel.intel' end    -- Intel 82574L
-      if device == '0x105e' then return 'apps.intel.intel' end    -- Intel 82571
+      if device == '0x10fb' then return 'apps.intel.intel_app' end -- Intel 82599
```
alexandergall: This change has broken the packetpblaster program, which checks for the driver `apps.intel.intel10g` explicitely.

eugeneia: @alexandergall Good catch! willfix asap :)

PR #411, opened 2015-03-18
[csum-offload-simd] Incoming checksum verification

javierguerragiraldez:
Checks incoming packets, works for both IPv4 and IPv6, TCP and UDP.

Performance seems terrible, I guess because of all the `ntohs()` calls.  I'm sure there's some way to factor them out, probably involving a little bit-shuffling.

### DIFF
--- a/src/lib/checksum.lua
+++ b/src/lib/checksum.lua
@@ -27,6 +27,12 @@ function finish_packet (buf, len, offset)
    ffi.cast('uint16_t *', buf+offset)[0] = lib.htons(ipsum(buf, len, 0))
 end
 
+function verify_packet (buf, len)
+   local phead = C.pseudo_header_initial(buf, len)
+   if phead.initial == 0 then return false end
+
+   return ipsum(buf+phead.headersize, len-phead.headersize, phead.initial) == 0
+end
 
 -- See checksum.h for more utility functions that can be added.
 
--- a/src/lib/checksum.lua
+++ b/src/lib/checksum.lua
@@ -55,14 +55,15 @@ function selftest ()
    for i = 0, n-1 do  array[i] = i  end
    local avx2ok, sse2ok = 0, 0
    for i = 1, tests do
-      local ref =   C.cksum_generic(array+i*2, i*10+i, 0)
-      if have_avx2 and C.cksum_avx2(array+i*2, i*10+i, 0) == ref then
-	 avx2ok = avx2ok + 1
+      local initial = math.random(0, 0xFFFF)
+      local ref =   C.cksum_generic(array+i*2, i*10+i, initial)
+      if have_avx2 and C.cksum_avx2(array+i*2, i*10+i, initial) == ref then
+         avx2ok = avx2ok + 1
       end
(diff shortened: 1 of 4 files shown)


lukego:
How is the state of this branch at the moment? (It would help to add a comment to the PR when pushing new commits so that I know when/what to test.)

I did some light testing locally. Observations:

1. iperf performance seems to be good.
2. Fishy experiment: commenting out the `finish_packet()` call does not break iperf, but I think it should cause all packets to be dropped on the receiving VM due to bad checksum.
3. Fishy experiment: adding counter/printout it seems like `verify_packet()` is returning false around 1/3 of the time, but I expect checksums to always be verifiably correct in practice with local VM TCP/IPv4 traffic.

Does this make any sense to you?

javierguerragiraldez:
Hi.  Yes, I had seen your point 3, it was really bugging me out, i had do the sum manually lots and lots of times, yet the small packets kept giving non-zero sums.  Even so, the kernel did accept those.

The thing resulted in a different interpretation of the `initial` value between the generic and the SIMD versions.  It works now but it would be much better to agree on either network order or host order.

lukego:
Interesting! That makes sense.

Can you on this branch pick either host or network byte order, fix the functions to be consistent, and make the `checksum` module unit test ensure this holds?

javierguerragiraldez:
Done!

The `if len >...` test is gone, now all checksum functions have to interpret the initial value as network order.  The `lib.checksum.selftest()` ensures that the functions are consistent with many random initial values.

lukego:
Awesome! Works well and runs fast for me. (Sorry about the endian bug, I should have covered initial value argument in the original unit tests!)

src/lib/virtio/net_device.lua
```diff
@@ -196,6 +196,21 @@ function VirtioNetDevice:transmit_packets_to_vm ()
    end
 end
 
+local function validflags(buf, len)
+   local valid = checksum.verify_packet(buf, len)
+
+   if valid == true then
+      return C.VIO_NET_HDR_F_DATA_VALID
+   elseif valid == false then
+      return 0
```
lukego: Curious: Why do we return 0 here? (What does that mean to Linux and Virtio-net, and is this better than returning `C.VIO_NET_HDR_F_NEEDS_CSUM` so that the guest will test the checksum and see the error?)

src/lib/checksum.lua
```diff
@@ -10,6 +10,7 @@ require("lib.checksum_h")
 local lib = require("core.lib")
 local ffi = require("ffi")
 local C = ffi.C
+local band = bit.band
 
 -- Select ipsum(pointer, len, initial) function based on hardware
```
lukego: Comment here seems like the right place to say that we expect network byte order?

src/lib/virtio/net_device.lua
```diff
@@ -205,6 +220,7 @@ function VirtioNetDevice:tx_packet_start(addr, len)
 
    -- TODO: copy the relevnat fields from the packet
```
lukego: This TODO can disappear now? (Maybe worth sweeping for more comments that we have obsoleted?)

PR #412, opened 2015-03-18
[snabbnfv selftest bench] Add default MTU=1500 iperf bench.

eugeneia:
Title says all really.


lukego:
How come SnabbBot is unhappy?

eugeneia:
@lukego Not sure wasn't able to reproduce. Although the "benchmarks" failure is a bug in the CI AWK script.

PR #422, opened 2015-03-24
[nfvconfig] Remove obsolete notion of "zerocopy pairs"

eugeneia:
[nfvconfig] Remove obsolete notion of "zerocopy pairs" and the related `nfvconfig.apply` function.

This also fixes a bug in `vhost_user` where it tried to `close(2)` a nil value. Additionally this makes `vhost_user:close` set `self.socket` to nil instead of -1 in order to avoid possible bugs when using `assert`.

### DIFF
--- a/src/apps/vhost/vhost_user.lua
+++ b/src/apps/vhost/vhost_user.lua
@@ -58,8 +58,10 @@ function VhostUser:stop()
    self.connected = false
    self.vhost_ready = false
    -- close the socket
-   C.close(self.socket)
-   self.socket = -1
+   if self.socket then
+      C.close(self.socket)
+      self.socket = nil
+   end
    -- clear the mmap-ed memory
    self:free_mem_table()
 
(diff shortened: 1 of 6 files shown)


lukego:
Rejected by SnabbBot, but I don't see an error in the log?

eugeneia:
@lukego Performance regression in the iperf benchmark. Seems reproducible but I can not explain why. At first I suspected my changes to `vhost_user` but they were not the issue. When run manually `cperf.sh check` reports:

```
iperf-1500 409a93f 3.082 0.0597997
iperf-1500 2babe20 3.136 0.134104

iperf-jumbo 409a93f 6.468 0.194874
iperf-jumbo 2babe20 6.692 0.271028
```

These results don't reproduce SnabbBot's results so I forced him/her/it to rerun a third time and hope this was just a fluke.

eugeneia:
eh... doh! Since cperf now persists results SnabbBot obviously repeatedly reported the fluke...

PR #423, opened 2015-03-24
neutron2snabb: get profile:bindings from vif_details

n-nikolaev:
profile:bindings is not stored in the ml2_port_bindigs table
after the actual binding happens (i.e. the vm boots).
mechanism_snabb will store the value in the vif_details for persistence.


lukego:
@eugeneia How does this look to you?

The idea is that Neutron has a strange behavior (bug?) where the binding:profile field is sometimes cleared and so the notion is for our code in Neutron to copy the value into vif_details and for neutron2snabb to access it there instead.

eugeneia:
@lukego If its necessary and it works I have no objections! :)

PR #426, opened 2015-03-26
Various Documentation Fixes

eugeneia:
The branch contains a collection of documentation fixes and updates of the last month.

@lukego If you want me to make the snabbnfv documentation Lua agnostic, I can amend that to this PR. Just give the word! :)


src/apps/solarflare/README.md
```diff
@@ -0,0 +1,29 @@
+# Solarflare Ethernet Controller Apps
+
+## Solarflare (apps.solarflare.solarflare)
+
+The `Solarflare` app drives one port of a Solarflare SFN7 Ethernet
```
lukego: Multiple instances of the Solarflare app can be instantiated on the same PCI device. Packets received from the network will be dispatched between apps based on destination MAC address and VLAN.

src/lib/protocol/README.md
```diff
@@ -741,14 +738,9 @@ pushed headers).
...
-The payload is incomplete if the underlying packet consists of more than
-one buffer. The caller has to access the remainder of the payload,
-e.g. any additional packet buffers, by his own.
+datagram payload and its byte size.
 
 If *pointer* and *length* are supplied then *length* bytes starting from
-*pointer* are appended to the datagram's payload. If there is not enough
-free space in the buffer to append *length* bytes or if the payload is
-incomplete an error will be raised.
+*pointer* are appended to the datagram's payload. The caller has to
```
lukego: I think the caller should assume there is always space available in the packet buffer to store new headers. The datagram module can internally check this with an assertion.

The intention of the current design (straightline) is that packet buffers are fixed-size and larger than any actual network packet. (Currently defined as 10KB in `packet.h`).

lukego:
@eugeneia I think it's worth making the `snabbnfv traffic` config file documentation stand-alone, like a manpage, so that users can write configurations without referring to the Lua API documentation for PacketFilter and so on.

src/lib/protocol/README.md
eugeneia: The overflow is checked for in `packet.append` (used by `datagram:payload`).

> The caller has to ensure that the underlying packet has enough free space to accommodate for *length* additional bytes.

I meant this to be read as "can raise an error on overflow". Should I reword to make that meaning more implicit?

lukego: I think that we can remove this sentence. The app developer should not have to think about how much space is available in the packet buffer, it is the job of the core developers to ensure that this is always sufficient for any realistic scenario.

(Like we don't ask programmers to check that they have enough stack space available before calling a subroutine and instead expect the runtime system to ensure that this is the case for any sane usage.)

eugeneia:
@lukego  Hrmm I just noticed this would mean to basically copy the whole PacketFilter configuration docs... and how formal should the grammar description be? E.g. if we don't mention that its a Lua expression we kind of need to define the syntax for strings and numbers....

lukego:
OK. Let's revisit this in the future. Requires some reflection. Perhaps we will want to target this as part of a larger management interface revamp (YANG, etc).

Overall I feel like we want to encourage power users to build tools with Snabb Switch using Lua, but the resulting tools should be accessible to everybody (just like `tcpdump`, `iperf`, `netcat`, etc). So refering to Lua in the documentation under `programs/` should perhaps be the exception (e.g. `snsh`) rather than the rule.

lukego:
Further thought to documentation refactoring:

Another solution to the "too many packet filtering APIs to document" is to replace the current PacketFilter app with a pflua based one. Then we shorten our Lua documentation by throwing away the current `PacketFilter` API and replacing it with a reference to [`pcap_filter(7)`](http://www.tcpdump.org/manpages/pcap-filter.7.html)). The original SecurityGroup-style `rules={...}` syntax can then be documented in the `snabbnfv traffic` manpage without being a duplicate of anything. (The `nfvconfig` module would be responsible for translating high-level rules into low-level pcap filter expressions.)

cc @andywingo @dpino

eugeneia:
eh this is merged already :(

PR #433, opened 2015-03-28
core: Count "frees per second" (fps) and do dynamic idling

lukego:
"Frees per second" (fps), literally the number of calls to
packet.free() per second, is introduced as a new basic metric for
measuring work done by a Snabb Switch process.

engine.Hz = false now uses a dynamic mechanism to decide how long to
sleep for depending on how busy/idle the process is, as observed from
fps. This is now the default setting and should be suitable for normal
usage.

Load on the process can be printed by calling
engine.report_load(). The load report looks like this:

    load: time: 1.00s  fps: 2,025,221   fpb: 304  bpp: 1280 sleep: 0   us

and the metrics mean:

    time - period of time that the metrics were collected over
    fps  - frees per second
    fpb  - frees per breath
    bpp  - bytes per packet (average packet size)

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -27,6 +27,7 @@ configuration = config.new()
 -- TODO: Move these over to the counters framework
 breaths = 0			-- Total breaths taken
 frees   = 0			-- Total packets freed
+freebits = 0			-- Total packet bits freed (for 10GbE)
 freebytes = 0			-- Total packet bytes freed
 
 -- Breathing regluation to reduce CPU usage when idle by calling usleep(3).
@@ -233,6 +234,7 @@ end
 
 local nextbreath
 local lastfrees = 0
+local lastfreebits = 0
 local lastfreebytes = 0
 -- Wait between breaths to keep frequency with Hz.
 function pace_breathing ()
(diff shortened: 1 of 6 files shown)


lukego:
Added some follow-on changes for improving the messages printed to stdout.

lukego:
Couldn't resist: I added `fpGbps` to estimate how much traffic is being processed in terms of on-the-wire physical resources. fpGbps = 10 would mean the total traffic corresponds to full utilization of one 10GbE link in one direction.

Example from an iperf run between two VMs connected via a local NIC. The VMs are showing 10.7 Gbps goodput in this example and one Snabb Switch process is serving both VMs:

    load: time: 1.00s  fps: 2,114,697 fpGbps: 22.570 fpb: 293 bpp: 1325 sleep: 0   us

Here fpGbps of 22.5 makes sense for a 10GbE interface that is being driven in both directions at more than 100% of line rate (possible here because the 82599 NIC can forward between local VMs at more than 10Gbps).

lukego:
I start to wonder whether `frees` is really a useful metric instead of simply counting ingress and egress.

lukego:
Merged because this feels like a big step in the right direction.

alexandergall:
Corner case: loadgen doesn't trigger dynamic idling because it doesn't perform any `frees`.

PR #434, opened 2015-03-30
[packet_filter] limit the number of streams tracked by stateful filters

javierguerragiraldez:
This patch adds a `counts` table to keep track of the number of entries in the `conntracks` tables.  When a track is over an arbitrary limit (2000 entries by default), the `track()` function won't do anything (not even analyse the packet).  This not only avoids extremely big (and unpredictably slow) tables, but also makes the 'tables full' state less susceptible to (intentional?) performance drops.

This limit doesn't affect the `check()` function, so a firewall won't drop existing streams, just won't keep track of new ones.

### DIFF
--- a/src/apps/packet_filter/conntrack.lua
+++ b/src/apps/packet_filter/conntrack.lua
@@ -221,8 +221,9 @@ return {
       counts[name] = counts[name] or 0
    end,
 
-   track = function (name, buffer)
-      if counts[name] > 1e6 then return end
+   track = function (name, buffer, limit)
+      limit = limit or 2000
+      if counts[name] > limit then return end
       local p = conntracks[name]
       local spec, flags = spec_from_header(buffer)
       if spec then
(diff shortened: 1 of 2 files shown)


lukego:
SnabbBot struck by the Github DDoS?

eugeneia:
Yup, will rerun!

lukego:
Oh it's luajit.org that is offline. I mailed Mike. Perhaps we should mirror luajit on Github.

lukego:
Turns out the Snabb Lab IP address has exceeded its quota for git requests to luajit.org. See #436 for an attempted fix.

lukego:
@javierguerragiraldez Could you merge or rebase onto master to pick up the change from #436? SnabbBot needs this change because the Snabb Lab is blocked currently from luajit.org.

javierguerragiraldez:
done!  it rebased cleanly, hope the 'bot likes it :-)

PR #438, opened 2015-04-04
[packet_filter] rewritten code generator

javierguerragiraldez:
Simpler code, both in the generator and generated.  Stays 100% compiled code, with simple rules manages 7-10 Mpps on grindelwald and interlaken.


lukego:
Superseded by #444 if all goes to plan.

Issue #330, opened 2014-12-01
Example on how to get going


plajjan:
When I first started up snabb I had no idea how to put together a "configuration file". I read the README file in src/ and while it does document how to link apps together and so forth it doesn't provide a complete example. I found src/designs/loadgen/loadgen which I think is about the closest to a working example in the source tree.

My next conundrum was over the naming of the input and output interface of each app. It was less than obvious that Intel10G used rx/tx while many other applications use input/output.

I'd like to write a very short introduction to help other folks out there. A start:
https://github.com/SnabbCo/snabbswitch/wiki/Writing-an-application-configuration

hit or shit?

lukego:
This is very good!

(Did I really miss that you opened this a month ago? Sorry!)

Issue #405, opened 2015-03-16
Getting Started Guide update needed for straightline, program


lukego:
The Getting Started Guide needs to be revised for the master branch.

The example program could now be built into a full stand-alone executable with its own command-line syntax, like the examples in `src/programs/`.

The straightline merge also means some small fixes are neded (see also #401).

lukego:
Fixed by #439.

PR #439, opened 2015-04-06
Move apps/sprayer to program/example and convert scripts to programs

eugeneia:
Move `apps/sprayer` to `program/example-spray` and `program/example-replay` respectively and convert scripts to programs. Move "Getting Started Guide" from Wiki to tree: https://github.com/eugeneia/snabbswitch/blob/fix-getting-started/src/doc/getting-started.md


src/program/example/example.lua
```diff
@@ -0,0 +1,17 @@
+module(..., package.seeall)
+
+local ffi = require("ffi")
+local C = ffi.C
+local lib = require("core.lib")
+local usage = require("program.example.README_inc")
+
+function run (args)
```
lukego: This code for dispatching on a subcommand (`sprayer <subcommand> ...`) is a bit tricky as an introduction to Snabb Switch. I think that we should simplify this somehow.

Some ideas:

* Stop using subcommands and simply flatten out the command-name space. (s/`snabbnfv traffic/snabbnfv-traffic/`)
* Standardize (move) subcommand logic somewhere else (e.g. `main.lua` or `lib.lua`) to make the example program simpler.
* Change the example to use a different command-line syntax (e.g. `-r <file>` for replay and `-s <file>` for spray).

eugeneia: Agree, I just blindly copied the `snabbnfv` hierarchy without thinking.

> Stop using subcommands and simply flatten out the command-name space. (s/snabbnfv traffic/snabbnfv-traffic/)

I think this is the smartest move, we're not exactly running out of namespace are we?

lukego:
This looks very well done.

However, I would really like the "Hello world" experience to be simpler than this, and I think that requires some revision of our APIs.

How about we spend some brain cycles in the next release on making it more trivial to write a first app and then revise and merge this then?

eugeneia:
@lukego I simplified this a bit and added detailed explanations of the code while adding the getting started guide to the source tree. We should be able to see if the explanation gets more or less complicated when revised for possible API changes. I have one problem though: I can not require `require("program.example-spray.sprayer")`. The file is there but I get this error:

```
program/example-spray/example-spray.lua:4: module 'program.example-spray.sprayer' not found:
	no field package.preload['program.example-spray.sprayer']
	no file './program/example-spray/sprayer.so'
	no file '/usr/local/lib/lua/5.1/program/example-spray/sprayer.so'
	no file '/home/max/snabbswitch/deps/luajit/usr/local/lib/lua/5.1/program/example-spray/sprayer.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
	no file './program.so'
	no file '/usr/local/lib/lua/5.1/program.so'
	no file '/home/max/snabbswitch/deps/luajit/usr/local/lib/lua/5.1/program.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
	core/main.lua:96: in function <core/main.lua:94>
	[C]: in function 'require'
	program/example-spray/example-spray.lua:4: in main chunk
	[C]: in function 'require'
	core/main.lua:46: in function <core/main.lua:32>
	[C]: in function 'xpcall'
	core/main.lua:101: in main chunk
	[C]: in function 'require'
	[string "require "core.main""]:1: in main chunk
```

Any hints in debugging this would be welcome. The file `src/obj/program/example-spray/sprayer_lua.o` *does* exist.

eugeneia:
Oh, could this be related to the dash (`-`) character in the directory name? Should I replace it with underscores?

lukego:
I think that it is the dash, yes. I have been substituting underscores, but maybe there is a way to make the dash work too.

eugeneia:
Yup it was the dashes. Fixed.

lukego:
I somehow missed the text on this PR and only saw the code. Awesome with friendly prose! Let me read...

eugeneia:
Oh I just added the prose like yesterday so you couldn't have known. :)

lukego:
I would have known if you had pushed it as a new commit instead of discarding the history with a rebase :-)

lukego:
@eugeneia Wow! I love the new getting started guide :-) Great job!

Logistic question. Will it be awkward to have the example programs upstream in the repository? In #470 I am making `snabb --help` print all the available programs, and it might be confusing if this included ones that are meant as sample programs rather than end-user tools.

Generally it may be a reasonable idea to support writing apps/applications in separate repositories rather than forks of the main repo. Have the Makefile pick them up somehow? But this is also something we can worry about later if it is not needed now.

eugeneia:
You know you kind of introduced me to `git rebase -i` so you reap what you sow I guess. ;P

I don't feel like it will be awkward to have the examples in the main tree. The "example_" prefix makes it quite clear that these are not "normal" programs. I personally am always glad if I clone a repository and have some simple examples included that I can digest.

If we had something like a "production LTS only operator use" release, then we could simply tag a HEAD with the examples removed.

One thing that I see is that the `example_replay` is actually a useful program. `example_spray` is not though (as far as I can tell). It would be really elegant to have the "examples" just be simple but useful production programs/apps that just happen to be written and documented with special care.

lukego:
If only I could rebase my Sent Mail folder so that I never recommended git rebase :-)

I take your point. There does not seem to be much harm in including a couple of examples, even positive if they would morph into real programs some time. I also think it makes a lot of sense to keep the Getting Started guide based on the current API because we do want to support that for a reasonable length of time and who knows when a new one will really land (it would be a shame to churn the API prematurely).

lukego:
So time to merge?

That would allow us to close #330 and #405 and welcome new scripters with welcome arms. @plajjan and @jeffbrl might have some feedback but we could take that on new Issues/PRs rather than holding up this one.

eugeneia:
Sure, go ahead! :)

PR #441, opened 2015-04-09
Enable MTU configuration and add SNMP objects to intel10g driver

alexandergall:
I need to be able to configure the MTU of individual interfaces. This requires functionality to pass configuration options down to the driver. I suggest a simple method, where the config supplied to Intel82599 is passed on to the intel10g module unchanged.

There is no consesus about what should be included in the MTU among vendors. For example, IOS does not include the Ethernet header, while IOS-XR and JunOS does.  Others include the CRC as well. I suggest the convention to include the Ethernet header but not the CRC.

Finally, I've added support for most of the objects in the ifMIB and interfaces MIB, which makes it possible to monitor an interface with standard tools.

### DIFF
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -25,10 +25,34 @@ num_descriptors = 512
 --num_descriptors = 32
 
 -- Defaults for configurable items
-local default = { snmp = {
-		     status_timer = 5, -- Interval for IF status check and MIB update
-		  }
-	       }
+local default = {
+   -- The MTU configured through the MAXFRS.MFS register includes the
+   -- Ethernet header and CRC. It is limited to 65535 bytes.  We use
+   -- the convention that the configurable MTU includes the Ethernet
+   -- header but not the CRC.  This is "natural" in the sense that the
+   -- unit of data handed to the driver contains a complete Ethernet
+   -- packet.
+   --
+   -- For untagged packets, the Ethernet overhead is 14 bytes.  If
+   -- 802.1q tagging is used, the Ethernet overhead is increased by 4
+   -- bytes per tag.  This overhead must be included in the MTU if
+   -- tags are handled by the software.  However, the NIC always
+   -- accepts packets of size MAXFRS.MFS+4 and MAXFRS.MFS+8 if single-
+   -- or double tagged packets are received (see section 8.2.3.22.13).
+   -- In this case, we will actually accept packets which exceed the
+   -- MTU.
+   --
+   -- XXX If hardware support for VLAN tag adding or stripping is
+   -- enabled, one should probably not include the tags in the MTU.
+   --
+   -- The default MTU allows for an IP packet of a total size of 9000
+   -- bytes without VLAN tagging.
+   mtu = 9014,
+
+   snmp = {
+      status_timer = 5, -- Interval for IF status check and MIB update
+   }
+}
 
 local function pass (...) return ... end
 
@@ -38,6 +62,7 @@ local M_sf = {}; M_sf.__index = M_sf
 
 function new_sf (conf)
    local dev = { pciaddress = conf.pciaddr, -- PCI device address
+                 mtu = (conf.mtu or default.mtu),
                  fd = false,       -- File descriptor for PCI memory
                  r = {},           -- Configuration registers
                  s = {},           -- Statistics registers
@@ -225,16 +250,23 @@ function M_sf:init_receive ()
    self.r.RXCTRL:clr(bits{RXEN=0})
    self:set_promiscuous_mode() -- NB: don't need to program MAC address filter
    self.r.HLREG0(bits{
-      TXCRCEN=0, RXCRCSTRP=1, JUMBOEN=2, rsv2=3, TXPADEN=10,
+      TXCRCEN=0, RXCRCSTRP=1, rsv2=3, TXPADEN=10,
       rsvd3=11, rsvd4=13, MDCSPD=16
    })
-   self.r.MAXFRS(lshift(9000+18, 16))
+   if self.mtu > 1514 then
+      self.r.HLREG0:set(bits{ JUMBOEN=2 })
+      -- MAXFRS is set to a hard-wired default of 1518 if JUMBOEN is
+      -- not set.  The MTU does *not* include the 4-byte CRC, but
+      -- MAXFRS does.
+      self.r.MAXFRS(lshift(self.mtu+4, 16))
+   end
    self:set_receive_descriptors()
    self.r.RXCTRL:set(bits{RXEN=0})
    return self
 end
 
 function M_sf:set_rx_buffersize(rx_buffersize)
+   assert(rx_buffersize >= self.mtu)
    rx_buffersize = math.min(16, math.floor((rx_buffersize or 16384) / 1024))  -- size in KB, max 16KB
    assert (rx_buffersize > 0, "rx_buffersize must be more than 1024")
    self.rx_buffersize = rx_buffersize * 1024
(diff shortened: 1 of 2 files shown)


alexandergall:
Sigh. I don't know how to run `program/snabbnfv/selftest.sh` myself on interlaken and I can't make sense of the testlog.

alexandergall:
Can someone please help me find the reason why the snabbnfv selftest fails?  I guess this has something to do with it:

qemu (/home/max/bench_env/vma.img) log:
QEMU waiting for connection on: unix:vhost_A.sock,server
qemu-system-x86_64: -netdev type=vhost-user,id=net0,chardev=char0: chardev "char0" went up
 
QEMU 2.1.0 monitor - type 'help' for more information
(qemu) 
(process:29278): GLib-CRITICAL **: g_io_channel_unref: assertion 'channel != NULL' failed
 
This is not enough to diagnose the actual problem.

There is also a failure of a ping with a large packet

ping6 -s 9000 -c 1 fe80::5054:ff:fe00:1%eth0
PING fe80::5054:ff:fe00:1%eth0(fe80::5054:ff:fe00:1) 9000 data bytes
Connection closed by foreign host.
JUMBOPING failed.

But that must have failed before, too.  The -s option only specifies the size of the ICMP payload, i.e. the entire packet will be 9000+8+40+18 = 9066 bytes, including Ethernet, CRC, IPv6 and ICMP header. The MAXFRS register used to be set statically to 9018 (and this default is unchanged by my code).

eugeneia:
>  (process:29278): GLib-CRITICAL **: g_io_channel_unref: assertion 'channel != NULL' failed

This is a phantom. I reproduced the jumboping failure without this message appearing.

When I increase `default.mtu` to 9062 jumboping succeeds but he subsequent iperf fails. When I set `default.mtu` to 16380 iperf succeeds too. So it appears to be indeed related to the `MAXFRS` register.

eugeneia:
Regarding running the snabbnfv selftest I can only refer to https://github.com/SnabbCo/snabbswitch/blob/master/src/program/snabbnfv/README-selftest.md

Do you have access to `chur`? If so you can copy over the references assets.

alexandergall:
Thanks. I'll try that. I can't access any of the lab servers since yesterday evening, though. Not sure whether Luke's IP address has changed and the DNS for lab1.snabb.co needs to be updated or it's just me.

eugeneia:
I can only access davos too.

alexandergall:
I suspect that the MTU problem is not due to MAXFRS but the code I've introduced that prevents the sending of frames which are larger than the MTU in `M_sf:transmit()`:

    if p.length > self.mtu then
      if self.snmp then
        local errors = self.snmp.ifTable:ptr('ifOutDiscards')
        errors[0] = errors[0] + 1
      end
      packet.free(p)
    else
      ...

I do think that this is correct and the app should not generate such frames. In fact, what is the MTU configuration of the VMs? They should use an MTU that is compatible with that of the snabbnfv instance. It probably doesn't, because otherwise it would not generate these large frames (i.e. TCP MSS and fragmentation of Pings). The NIC drops frames larger than MAXFRS, but maybe not if packets are only switched through it (not sent on the wire).

alexandergall:
Bingo. We should sort out the MTU business ASAP. This is fundamental stuff that we need to get right.

lukego:
Howdy! Sorry for slow response. I have had family visiting from Australia around Fred's first birthday :). Back to normal now.

Great to be getting strict about MTU and to relate MTU mismatches to standard counters like `ifOutDiscards`.

Question: how much do we want to do in software vs. hardware? I suppose the simplest and most flexible is to use software for enforcing the MTU and tracking all counters. Then we could simplify the driver by using fixed values of `MAXFRS` and `SRRCTL` based on the known size of `struct packet.data`. I suppose the downside is potential overhead of checking MTU, checking multicast/broadcast to bump the right counters, etc... not obvious whether that would be significant?

(Note to self: in both cases I think we can remove `set_rx_buffersize()` and the dynamic tuning of receive buffer size `SRRCTL` now that the straightline merge made our buffers all uniform sizes.)

I will check on lab machine connectivity when the baby sleeping on my lap awakens :)

lukego:
(Sorry if I am veering of topic after joining the thread late and still starting to think about applying the MIB-counter-goodness to other parts of the code base e.g. Virtio-net and Solarflare interfaces.)

lukego:
Crazy idea that should perhaps be discussed separately: Would it make sense to create an `Interface` app that enforces MTU, collects ifTable statistics, and otherwise supports externally visible logical interfaces?

This would allow app network designers to introduce logical interfaces at appropriate points and it would also free people writing I/O apps from worrying about tracking statistics and enforcing MTUs. Using a standard app would also give us consistent behaviour between different I/O interfaces (Intel10G, Intel1G, Virtio-net, RawSocket, etc).

Worth considering? Technical challenges?

alexandergall:
The hardware vs software question is a good one, indeed. Software has at least two advantages:

The code will work for any NIC, which implies that the same information will be available throughout, even if some hardware doesn't provide counters for everything.

It will cover cases which bypass the hardware stats registers, as appears to be the case for traffic switched by the NIC.

The only downside is possibly, as you say, performance. But the code will be extremely centralized (as part of the generic driver) and we can optmise the hell out of it. This should actually be fairly ease to check.

Personally, I have a strong tendency to get rid of as much hardware assistance as possible.

While my code currently uses the SNMP objects directly, I think it would be better to devise our own system of counters and statistics and generate all external representations (like the MIBs) out of that. The shmem mechanism seems like a good starting point because it solves two problems at the same time: communicating the data to other processes and provide stable storage for data across invocations of Snabb processes (e.g. interface counters).

alexandergall:
Now something specific to this PR. I still stand by my statement, that dropping giant frames in transmit() is correct and I suggest to add this check on receive() as well to make all use-cases behave the same way (i.e. when traffic bypasses the MAXFR logic on ingress). If I'm correct, this will demonstrate that the current setup for the snabbnfv selftest is flawed, probably because the VMs use an MTU that is too large.

This is a bit orthogonal to the question whether counters should be maintained in software. It's primarily code that makes snabbnfv behave correctly and consistently wrt to the MTU.

lukego:
> Personally, I have a strong tendency to get rid of as much hardware assistance as possible.

I agree. CPUs are amazing these days. NIC offloads feel messy and old-fashioned lately.

> I think it would be better to devise our own system of counters and statistics and generate all external representations

Sounds reasonable to me. I also like the shared memory files and the idea of persistence.

Dreaming of the future I also ponder having the app network `fork()` multiple processes, including both realtime traffic processes and non-realtime management processes that could conceivably run apps that implement SNMP, NETCONF, and so on. Just dreamin'.. in the short term it's very convenient to farm out all of the northbound interfaces to off-the-shelf software but I can imagine that we will prefer to have our own one day.

> If I'm correct, this will demonstrate that the current setup for the snabbnfv selftest is flawed, probably because the VMs use an MTU that is too large.

Do I understand correctly that you mean the NIC is configured with too small of an MTU compared with the VMs, but that the selftest does not detect this because the NIC is not enforcing MTU for locally switched packets?

alexandergall:
@lukego not crazy at all. I've been thinking about that for a long time. I believe that this is the way to go in the long run. Just like I would like to see control-plane stuff (including neighbor discovery/ARP) moved out of the packet-processing apps (could be done by the `Interface` app as well). The question is how to share an interface with other apps, but you also had some ideas about that, too.

alexandergall:
>> If I'm correct, this will demonstrate that the current setup for the snabbnfv selftest is flawed,   probably because the VMs use an MTU that is too large.

> Do I understand correctly that you mean the NIC is configured with too small of an MTU compared with the VMs, but that the selftest does not detect this because the NIC is not enforcing MTU for locally switched packets?

Exactly. My check in transmit() already broke this and I suspect that a similar check in receive() would, too.

lukego:
Let us integrate this code after the May release is tagged. Just because it makes substantial additions to an important module (intel10g) and so it would be nice to have it tested on master for a bit before shipping.

lukego:
@alexandergall Looks like we can merge this now and then follow-up with the MTU fix, which is an independent problem. Agree?

alexandergall:
Agreed. Thanks.

PR #443, opened 2015-04-10
Makefile: Pick gcc flags for C code and stick with them

lukego:
Compile C code with `-g -O2 -msse2 -mavx2`

The default build options should now be suitable for all C files in
the source tree: debug info included, optimization performed, SIMD
intrinsics available.

This does somewhat impact compile time (now ~3.5 sec) and object code
size (now 1.6 MB). Those are over budget (1 sec and 1 MB). However, I
propose we address that by removing unneeded C source files and
replacing them with calls to ljsyscall. These new Makefile options do
make sense for the C code that we actually want to keep in Snabb
Switch even if they may be overkill for system call glue.

This fixes an ugly 'make' warning where we had a duplicate rule for
the checksum library in order to apply these flags only to it.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -69,7 +69,7 @@ snabb: $(LUAOBJ) $(HOBJ) $(COBJ) $(ASMOBJ) $(INCOBJ) $(LUAJIT_A)
 	    exit 1; \
 	 fi
 	$(E) "LINK      $@"
-	$(Q) gcc $(DEBUG) -Wl,--no-as-needed -Wl,-E -Werror -Wall -o $@ $^ \
+	$(Q) gcc -g -Wl,--no-as-needed -Wl,-E -Werror -Wall -o $@ $^ \
 	    ../deps/luajit/src/libluajit.a \
 	    -lrt -lc -ldl -lm -lpthread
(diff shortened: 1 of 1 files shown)


lukego:
Great catch, @SnabbBot! This seemingly innocent Makefile change breaks runtime compatibility with GLIBC < 2.15 when compiled on Ubuntu.

lukego:
The problem here is toolchain voodoo: when compiled with `-O2` the `select()` call in `apps/socket/io.c` becomes dependent on GLIBC >= 2.15:

```
snabbswitch/src$ ./selftest.sh
selftest: ./snabb binary portability
Scanning for symbols requiring GLIBC > 2.7
GLIBC_2.15 __fdelt_chk
^^^ Error ^^^
```

So our choices are:

1. Leave things as they are (Makefile issues warnings due to overlapping rules).
2. Require GLIBC >= 2.15 instead of >= 2.7.
3. Rewrite `io.c` to use a voodoo-free alternative to `select()`.
4. Find a simpler solution.

I close this PR because I have not found a simpler solution yet and leaving things as they are for nwo seems the most reasonable alternative.

PR #446, opened 2015-04-21
Update authorized_keys

liuran2011:
add my host liuran.


liuran2011:
commit by liuran from china.

lukego:
Welcome! You should now be able to connect to chur with `ssh -p 2020 root@lab1.snabb.co` and create a private account for yourself.

PR #448, opened 2015-04-23
Apps: DNAT and Conntracking

dpino:
This is the initial development of a DNAT app and implementation of connection tracking.


dpino:
Code for the DNAT and connection tracking apps, are actually in one single app at this moment: apps/conntrack.

How to run the app:

```
sudo ./snabb snsh apps/conntrack/main.lua 
Results written at: /tmp/output.pcap
```

What the app does is to read packets from input.pcap and write down the results in output.pcap. The source and destination address will be changed according with the configuration of the ```proxy``` variable:

```lua
local proxy = {
   ip   = "192.168.1.1",
   mask = "255.255.255.0",
   port = "80"
}
```

To check that the addresses were successfully rewritten what I did was to open the output.pcap file with WireShark. BTW, you can run this app directly in your workstation, no need to log into Snabb Lab.

lukego:
Thanks for sharing this!

Let me put another interesting NAT-related problem on your radar: [lwAFTR](https://tools.ietf.org/html/draft-farrer-softwire-lw4o6-deterministic-arch-01) (lightweight IPv4/IPv6 Address Family Transition Router).

This is a stateless form of NAT for IPv6 networks to communicate with IPv4 networks. For example, Deutsche Telekom's TeraStream network is natively IPv6 and applies lwAFTR to all internet traffic. That is an important network element.

There is no production-quality open source lwAFTR implementaton today, and only one commercial one to my knowledge (from A10 Networks). If you guys are ever interested in prototyping an lwAFTR I would gladly put you in touch with [Ian Farrer](http://blogs.cisco.com/sp/dts-ian-farrer-on-the-all-ipv6-terastream-network) at Deutsche Telekom who lives and breathes this stuff (and is an author of the IETF draft linked above).

[Carrier Grade NAT](http://en.wikipedia.org/wiki/Carrier-grade_NAT) is another important application with additional requirements for ISPs e.g. being able to tell the police who was using an address/port at a given time.

kbara:
That was a lucky accident; we'll be working on it in https://github.com/igalia/snabbswitch , and hadn't meant to do a PR here. I'm quite curious about lwAFTR now, having read the draft. :)

PR #449, opened 2015-04-23
in get_time() make actual use of the argument when calling clock_gettime

noj:
Looks like the argument passed to get_time is not being used, this diff fixes it.


lukego:
Hoppsan! Nice catch :-)

noj:
Btw, turning on -Wextra would've caught it :-)

PR #450, opened 2015-04-28
Don't print the backtrace on error messages by default

lukego:
This is an idea for review: Do not print the Lua traceback (stack trace) for errors unless the environment variable SNABB_BACKTRACE is set.

The problem this addresses is that simple program errors often look pretty scary. For example, in `snabbnfv traffic` you will see a Lua traceback if you try to use a NIC that is already in use.

On the one hand I don't think we ever want end-users to see Lua tracebacks (unless when reporting bugs). On the other hand I do think that developers always want to see them.

Is this a good solution? If not then what might be better?

### DIFF
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -93,7 +93,9 @@ end
 
 function handler (reason)
    print(reason)
-   print(debug.traceback())
+   if debug_on_error or os.getenv("SNABB_BACKTRACE") then
+      print(debug.traceback())
+   end
    if debug_on_error then debug.debug() end
    os.exit(1)
 end


lukego:
@eugeneia SnabbBot reports a performance regression but I suspect this is spurious?

eugeneia:
@lukego Yup, reran it and passed.

lukego:
Sorry @darius. I fat fingered your name on the Assignee.

@eugeneia I am not honestly sure if this is a good idea or not. I suppose at the minimum it may require us to first improve our error messages so that they are useful even without the backtrace for context.

eugeneia:
@lukego It would be nice to have some way to differentiate between intentional errors (e.g. thrown using `error`) and cases we did not see coming. But since we can not "enumerate" all possible error conditions I guess we'll have to fix 'em when they pop up.

- [ ] Review all calls to `error` and their messages.

eugeneia:
The error messages we throw aren't that unclear (maybe driver errors should always print the PCI address of the device used?) but they don't advertise where they actually come from so a stack trace is the only way to figure that out. On the other hand end users won't care.

lukego:
Thanks for the review, Max!

Let me update this one and we can think about merging it onto `next` early in the next release cycle. That way we will have a few weeks to try it out and improve error messages and potentially revert the change before shipping it.

lukego:
I would like to give this a little more thought. Could be that it is better to introduce a separate function for exiting with a fatal error message. Then genuine uncaught `error()` messages would be bugs and it would make sense to include the backtrace. Closing this issue for now.

PR #451, opened 2015-04-28
apps/packet_filter: Rename pcap_filter.PcapFilter to bpf.BPF

lukego:
Give a new shorter name to the pflua based packet filter app.

The name BPF is a reference to the Berkeley Packet Filter where the pflang filter language originated (before being adopted by libpcap).

- http://en.wikipedia.org/wiki/Berkeley_Packet_Filter
- http://www.tcpdump.org/papers/bpf-usenix93.pdf

This name is somewhat ambiguous in that BPF can refer to a larger body of software including the `/dev/bpf` packet capture device on *BSD. However, the name Pcap has the same ambiguity of pcap-filter
vs libpcap, and networking people seem to prefer "BPF" to "PcapFilter".

This is motivated by comments [by Kristian](https://github.com/SnabbCo/snabbswitch/pull/444#issuecomment-91796187) and [by Simon](https://groups.google.com/d/msg/snabb-devel/eOMS8zZWi90/CgjGjmbcZIkJ).

Speak ye now or forever hold your peace on the name of the pflua filtering app! This PR can easily be replaced by a new one if we hit upon a better name.

### DIFF
--- a/src/apps/packet_filter/README.md
+++ b/src/apps/packet_filter/README.md
@@ -1,20 +1,20 @@
-# PcapFilter App (apps.packet_filter.pcap_filter)
+# BPF App (apps.packet_filter.bpf.BPF)
 
-The `PcapFilter` app receives packets on the `input` port and transmits
+The `BPF` app receives packets on the `input` port and transmits
 conforming packets to the `output` port. In order to conform, a packet
 must match the *[pcap-filter](http://www.tcpdump.org/manpages/pcap-filter.7.html)
-expression* of the `PcapFilter` instance and/or belong to a *sanctioned
+expression* of the `BPF` instance and/or belong to a *sanctioned
 connection*. For a connection to be sanctioned it must be tracked in a
-*state table* by a `PcapFilter` app using the same state table. All
-`PcapFilter` apps share a global namespace of *state table identifiers*.
-Multiple `PcapFilter` apps—e.g. for inbound and outbound traffic—can
+*state table* by a `BPF` app using the same state table. All
+`BPF` apps share a global namespace of *state table identifiers*.
+Multiple `BPF` apps—e.g. for inbound and outbound traffic—can
 refer to the same connection by sharing a state table identifer.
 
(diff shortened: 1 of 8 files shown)


andywingo:
I think BPF describes the bytecode / VM, more than the language itself.  Dunno.  Of course if there's a name that people use most, sure, makes sense.  But if BPF only describes the VM (and I don't know if that's the case) then it would be confusing to introduce the term, given that pflua in its default configuration doesn't produce or consume BPF bytecode.  Your call!  (Also FYI: am going on holiday for a few days.  See you on Tuesday!)

kbara:
I'm mildly against this change, as I think it's actively misleading. Most implementations that take "pflang"* input involve a transformation into BPF, which is then interpreted on a 6502-like virtual machine (or JIT compiled from Linux 3.0's implementation, and there are extensions to add registers, etc - but it's all still in this general 6502-like framework).

Pflua is fundamentally different. The main pipeline generates Lua code, which is then interpreted/JIT'd in the familiar way with LuaJIT. The main pipeline never involves BPF. (There is BPF-related code in pflua - a disassembler, and a bpf-lua pipeline that uses libpcap to compile to BPF after handling details like parsing - but this is an alternative pipeline, and even then, BPF is only an intermediate representation. The bpf-lua pipeline emits Lua, which in turn is also run through LuaJIT).

* pflang is a name invented by Andy. To the best of my team's knowledge, libpcap's input language was previously unnamed.

Check out https://www.kernel.org/doc/Documentation/networking/filter.txt and https://lwn.net/Articles/599755/ .
For added historical context about the 6502-like VM, see http://sharkfest.wireshark.org/sharkfest.11/presentations/McCanne-Sharkfest%2711_Keynote_Address.pdf

lukego:
Closed due to lack of overwhelming support. Soon we will make the May release and then it will be more disruptive to change the name.

The name PcapFilter at least has the virtue of being named after the pcap-filter(7) man page.

PR #460, opened 2015-05-06
Move SIMD code into arch/avx2.c and arch/sse2.c

lukego:
The Makefile treats these files specially: avx2.c compiles with -mavx2 and sse2.c with -msse2. This solves two problems.

First problem: The SSE2 code was actually compiling with AVX instructions because of '-mavx2'. This would cause a crash on hardware that does not support AVX or AVX2 such as an Atom CPU. (We missed this testing on the Sandy Bridge lab servers because they do support AVX even though they do not support AVX2.) Caught by Justin Cormack: https://groups.google.com/d/msg/snabb-devel/I_3dJvfl7gY/YHgHVzjMq0kJ

Second problem: The Makefile had overlapping rules for building the SIMD code and this causes warnings to be printed. That is because the default build rule applied to these files but then we also had specific rules. This is solved now because the arch/ directory is excluded from the default rules.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -18,9 +18,10 @@ OBJDIR = $(patsubst %,obj/%,$(SRCDIR))
 LUASRC = $(shell find . -regex '[^\#]*\.lua' -printf '%P ')
 PFLUASRC = $(shell cd ../deps/pflua/src && \
 	         find . -regex '[^\#]*\.lua' -printf '%P ')
-CSRC   = $(shell find . -regex '[^\#]*\.c'   -printf '%P ')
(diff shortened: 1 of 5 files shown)


justincormack:
Confirmed that is working correctly on Atom:

```
selftest: checksum
no avx2
sse2: 1000/1000
selftest: ok
```

lukego:
@eugeneia Do you think this is a real error or a spurious one?

justincormack:
It looks unrelated to this patch, but hard to tell what the issue is...

lukego:
I'll wait for a recheck before merge just to be safe.

PR #461, opened 2015-05-06
Fix a couple of bugs that can affect the packet filter

lukego:
This branch contains fixes to the packet filter:

- Fix bug in `engine.now()` that can return false when called before any breathe loop (happens when the initial app networks contains a stateful filter).
- Pulled in a fix for Igalia/pflua#178.


lukego:
CI failing because I had not pushed the latest pflua commits to our SnabbCo fork/mirror.

I will take the opportunity to close this and resubmit the two commits as separate PRs.

PR #468, opened 2015-05-11
[draft] Sandbox level 2: one LuaState VM for each app

javierguerragiraldez:
`lib.lua.statebox` implements the LuaState handling code using FFI calls to the Lua C API.  Handles passing simple values and non-recursive tables. Any userdata or FFI cdata value is cast into a lightuserdata.

`core.app_loader` uses `lib.lua.statebox` to create app-compatible objects.
`core.config` calls this loader when the app is specified with a dotted.name instead of the usual class table.

Translated `apps.basic.source`/`sink` to the new style.  It's very similar to the sandbox level 1 style, but now it's safe to call `require()` to reuse code within the app.  Usually there's no need to call `module()`, since we do want most of the code and variables to be 'global' within the sandbox.  Global functions and some variable names (currently zone and name) are copied to the app proxy object.

The start code within the sandbox sets the LuaJIT profiler, probably the `app:stop()` hook is not the right place to close it.

__Note__: the sample apps seem to be an order of magnitude slower than previously and dies if the `engine.main()` lasts more than 400msec.  Profiling within the sandbox shows almost all the time is spent allocating packets, and dies around 100k allocations.  I think the freelist object, being a Lua object, doesn't like the sandbox.  Should we somehow force all the freelist/allocation to happen in the main thread?  or simply turning the freelist into an FFI object (which is a trivial change) would help to make it VM-agnostic and shareable, just like links?

### DIFF
--- a/src/apps/basic/basic.lua
+++ b/src/apps/basic/basic.lua
@@ -1,0 +1,12 @@
+
+inputi, outputi = {}, {}
+
+function relink()
+   inputi, outputi = {}, {}
+   for _, l in pairs(output) do
+      table.insert(outputi, l)
+   end
+   for _, l in pairs(input) do
+      table.insert(inputi, l)
+   end
+end
--- a/src/apps/basic/sink.lua
+++ b/src/apps/basic/sink.lua
@@ -1,0 +1,11 @@
+require ('apps.basic.basic')
+
+
+function push()
+   for _, i in ipairs(inputi) do
+      for _ = 1, link.nreadable(i) do
+         local p = link.receive(i)
+         packet.free(p)
+      end
+   end
+end
--- a/src/apps/basic/source.lua
+++ b/src/apps/basic/source.lua
@@ -1,0 +1,17 @@
+require ('apps.basic.basic')
+
+local size = size or 60
+local pkt = packet.from_pointer (ffi.new("char[?]", size), size)
+
+function pull()
+   for _, o in ipairs(outputi) do
+      for i = 1, link.nwritable(o) do
+         link.transmit(o, packet.clone(pkt))
+      end
+   end
+end
+
+
+function stop()
+   packet.free(pkt)
+end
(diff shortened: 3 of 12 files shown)


lukego:
Awesome! Can't wait to read and understand :-)

Quick answer: Yes, it makes senes that private freelists would be a problem indeed. I suspect that a Source app would be allocating infinitely many packets and they would always end up on the private freelist of the Sink app. I like the idea of making the freelist into a shared FFI object in some simple way. (I'm not sure the simplest way to create shared FFI state: a C global variable perhaps?) This should work fine.

Or the quick workaround would be to test with a new app network that has the same app both allocating and freeing the packets. Like a SourceSink app. Then the packets should be correctly recycled. However, this wouldn't work for any other app networks, so maybe you want to skip this small step.

javierguerragiraldez:
> I like the idea of making the freelist into a shared FFI object in some simple way. (I'm not sure the simplest way to create shared FFI state: a C global variable perhaps?) This should work fine.

Yes, a global `static` C variable could work.  Or it could be injected by the loader, like the links are.

javierguerragiraldez:
Shared freelist, got 1.4MPkt/sec!  (roughly same as baseline)

lukego:
I am not sure what you are benchmarking exactly but 1.4 Mpps does not sound fast.

The `snabb snabbmark bench1 100e6` is supposed to be handy for exactly this scenario. That is, to feed packets through a simple app network and see what the throughput is. I would expect >= 20 Mpps as the baseline. For the multiple states it would be interesting to be able to make the app network longer e.g. to pass packets through N trivial app instances between the Source and the Sink. Then we would get a feel for whether the extra Lua states are slowing us down by chewing up cache space.

lukego:
also: Awesome that it is actually up and running so quickly!

Sorry to fixate immediately on the packet rate :-) reflex!

lukego:
Closed this PR now that we have successfully completed the proof-of-concept work.

PR #479, opened 2015-05-14
Move Wiki pages to Snabb Switch tree

eugeneia:
- [x] Documentation Guide (this still has a TODO, see comment below)
- [x] Snabb NFV Wiki
- [x] Review new documents
- [ ] Delete superseded Wiki pages


src/doc/documentation-guide.mk2
```diff
@@ -0,0 +1,189 @@
+# Contributing Documentation in SnabbSwitch
+
+*XXX - TODO: Document different documentation targets: Casual users,
+ power users, contributors.*
```
eugeneia: Might be a good time to figure this one out.

lukego: Meta-answer: I would say that each `program` can choose individually how it wants to present itself to users. For example, `snsh` will be very Lua-scripting-oriented but `packetblaster` might prefer to blend in with the usual Unix/C tools.

The decision is a little tricky in programs like `snabbnfv` where Lua is convenient but possibly distracting. I suppose the documentation guide does not have to tell people what to do in these situations beyond suggesting being thoughtful. If some kind of common practice arises over time then we could highlight that.

lukego:
I reckon the Snabb NFV wiki pages should move into `src/program/snabbnfv/doc/`.

src/doc/documentation-guide.mk2
eugeneia: I think last time we talked about our Lua centric configuration format etc. we (I at least) conceded that we should probably take the boring route of using JSON as a config format. Sure we can't have comments then and the format is sub-par for the writer but at least its very well defined and we can just point to the JSON spec.

Other than that the snabbnfv `--help` got much simpler with the `PcapFilter` app since we now just point to the `filter-expression` man page.

eugeneia:
The [Installation](https://github.com/eugeneia/snabbswitch/blob/77d33c7390a1cd234351744febb46f33c2f83070/src/program/snabbnfv/doc/installation.md) and [Configuration with Neutron (previously "Software Installation")](https://github.com/eugeneia/snabbswitch/blob/77d33c7390a1cd234351744febb46f33c2f83070/src/program/snabbnfv/doc/neutron-configuration.md) have intersecting content. We should either:

* Merge them into one document
* Split the scopes clearly (which?)

eugeneia:
Actually [Configuration with Neutron (previously "Software Installation")](https://github.com/eugeneia/snabbswitch/blob/77d33c7390a1cd234351744febb46f33c2f83070/src/program/snabbnfv/doc/neutron-configuration.md) seems outdated. I will `rm` it.

eugeneia:
@lukego Could you go through the Wikis and delete any pages that you can part with? Right now this PR imported:

* Documentation Guide
* (Getting Started Guide is already moved)
* NFV Architecture
* NFV Compute Node Requirements
* NFV Installation
* Neutron API Extensions

Or alternatively tell me what I missed. I can't find any more non-outdated or obsolete pages. One edge-case: "Snabb Lab" should probably just stay in the Wiki, right?

lukego:
This is awesome!

I have attempted to update the Snabb Switch and Snabb NFV wikis to remove unneeded pages and leave some forwarding pointers to the repo. (Currently they point to the `next` branch but could be updated when this documentation lands on master.)

There is some general information embedded in the `Scenario tests` page that we need to migrate into its own file i.e. how to configure the Neutron driver and tell it about the network setup. The challenge here is that currently the Neutron driver supports a very specific kind of network that is quite unlike what most people have. In the future it needs to be extended very slightly to support traditional `vlan` and `flat` network models too. Then the user will not have to fuss about this extra configuration.

Do you think more action is needed right now for this PR?

eugeneia:
Since you kept [Device Driver Programming](https://github.com/SnabbCo/snabbswitch/wiki/Device-Driver-Programming) I reckon I should import that too? Regarding [Writing an application configuration](https://github.com/SnabbCo/snabbswitch/wiki/Writing-an-application-configuration), should we really keep it around? Seems to duplicate info from the "Getting started guide".

Regarding "Scenario tests", I honestly don't understand much of its content.

lukego:
Device Driver Programming should die, but upon reading it I realized that I would like to start writing small introductory texts like that in the future, as simple entry points into the code. I was distracted by nostalgia and did not press Delete :).

Scenario Tests... That is alive because currently it is the only place where I can point to the description of the config file format for our Neutron ML2 plugin. That information needs to exist in some good place. The rest can go... it was a draft write-up while we were preparing the end-to-end tests for our OpenStack setup (which should also be documented Somewhere Else).

eugeneia:
Closing because this was merged into `next`.

PR #483, opened 2015-05-15
fix snabbswitch build on musl

comotion:
Compile-time errors fixed by adding includes and changing some types.
Avoiding the glibc memcpy demibug workaround builds luajit for non-glibc platforms like musl.

### DIFF
--- a/gcc-preinclude.h
+++ b/gcc-preinclude.h
@@ -2,7 +2,7 @@
 // See: http://www.win.tue.nl/~aeb/linux/misc/gcc-semibug.html
 //      https://rjpower9000.wordpress.com/2012/04/09/fun-with-shared-libraries-version-glibc_2-14-not-found/
 
-#ifndef __ASSEMBLER__
+#if !defined(__ASSEMBLER__) && defined(__GLIBC__)
 __asm__(".symver memcpy,memcpy@GLIBC_2.2.5");
 #endif
 
--- a/gcc-preinclude.h
+++ b/gcc-preinclude.h
@@ -2,7 +2,10 @@
 // See: http://www.win.tue.nl/~aeb/linux/misc/gcc-semibug.html
 //      https://rjpower9000.wordpress.com/2012/04/09/fun-with-shared-libraries-version-glibc_2-14-not-found/
 
-#if !defined(__ASSEMBLER__) && defined(__GLIBC__)
+#ifndef __ASSEMBLER__
+#include <string.h>
+#ifdef __GLIBC__
 __asm__(".symver memcpy,memcpy@GLIBC_2.2.5");
+#endif
 #endif
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -12,20 +12,20 @@ PFLUA_VSN     := "13b194e7a214faae5c3d2916a998418fc841dbc2"
 
 TEST_SKIPPED="43"
 
-SRCDIR = $(shell find . -type d -not -regex './obj.*' -printf '%P ')
+SRCDIR = $(shell find . -type d -not -regex './obj.*' -print | xargs)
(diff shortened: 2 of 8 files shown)


comotion:
Conditional back-versioning of memcpy  at c96f1e8 is incomplete and fixed by 30505c6.

lukego:
This is very interesting work! I am keen to make our Git workflow support simple ports like this. Like, working out what we need to do to land them on master, and making their development and maintenance smooth while they are living on branches.

More soon!

lukego:
I would like to bring compatibility with musl, 32-bit Linux, and nongnu tools onto master. There are some issues we will need to deal with along the way though and I think this effort should initially live on a long-lived branch (per #485) before landing on master. I would be glad if you want to maintain this branch! In that case please send a PR to add it to `src/doc/branches.md`.

Musl users would benefit from having a branch that is known to be musl-compatible. The easy short-term solution seems to be a dedicated branch whose maintainer regularly pushes from master with any required fixes for musl. Then people can run the code with musl and over time we can look at the branch to see what the compatibility issues are and how much interest it gathers.

The better long term solution would then be to make sure master is musl-compatible. This would require some new programming discipline and extended CI infrastructure. This might be awkward to adopt right this moment, but it does seem to fit in with future themes like replacing C code with Lua code and verifying support for more OS distributions and kernel versions.

How does that sound? (This is an early attempt at setting expectations for somebody who is submitting a PR for work that is valuable but cannot immediately land on master, please let me know how I am doing :-))

comotion:
Firstly I would like to figure out what is causing snabbswitch to fail despite compiling cleanly to musl. Making it compile cleanly was my initial stab at musl compatability,  However, I am also interested in finding out why my patches are causing snabbswitch to fail testing under glibc, as those patches were not meant to break anything. Do you have any idea why that might be? After I know more I can properly assess whether it is worth having a musl branch and if I can take responsibility for maintaining it.

lukego:
Interesting. I build this branch with GNU tools and the `./snabb` exits with status 1. I have seen this in the past when an error occurs very early during startup and we don't have any code to catch and report it.

I suspect the issue is this: we are using `luajit -bg` to compile Lua sources into `.o` files containing bytecode. Then when we `require()` these modules at runtime LuaJIT will expect to find them with `dlsym()`. Could it be that `dlsym()` does not work in this build for whatever reason?

@justincormack might know more.

justincormack:
I doubt if it is dlsym. I will have a look on one of my Musl systems and
see what's up.

lukego:
Right, scratch that. If I put a `print("foo") io.flush()` in `core/main.lua` then it prints. So Lua code is executing but perhaps hitting an error before we setup an error handler.

Could be a job for bisect? I can't dig into this right now.

lukego:
The root problem is `core.main` failing to load a Lua module that it depends on. This happens early before an error handler installed. If you merge #502 you should get a proper error message.

justincormack:
Yes it is failing to load 
local STP = require("lib.lua.StackTracePlus")

What is odd is there seems to be no change from master here, and the symbols in the binary look exactly the same, both under Musl and glibc... Other than the Musl embedded symbol names have three underscores not one eg luaJIT_BC___syscall_util

lukego:
These changes would be a very valuable starting point to a port away from GNU dependencies. However, somebody needs to follow them through all the way. Closing the PR for now.

justincormack:
A bunch of these are simple Posix compliance/portability/correctness issues, not in any way Musl specific, I might just submit these as individual issues so we can narrow down the Musl specific issues.

comotion:
Sorry I haven't been able to follow this up, am in my busy months, but did you figure out why STP was failing to load? With a working binary I would have less problems following up the other musl issues.

PR #486, opened 2015-05-17
neutron2snabb: Fix for security group -> pflua translation

lukego:
This change fixes two bugs in the translation of Security Groups into
pflang expressions for pflua:

1. Handle ICMP protocol.

2. Port restriction should be 'dst portrange ...' instead of 'portrange ...'.
   Otherwise source port can also match. (See issue Igalia/pflua#211.)


lukego:
This fix is tested in the OpenStack environment and makes our end-to-end test suite there pass 100%. That is currently 18 different test cases of booting VMs, adding/removing tunnels, etc. Woohoo :-)

I merged this into the `nfv` tree now and intend to synchronize that with master regularly (at least once for each release on master).

(The reason this issue closed automatically seems to be that I accidentally used the magic 'fix account/repo#issue' syntax when pushing this to another branch on github. I had not considered that things are so interlinked!)

lukego:
Merged onto `nfv` branch.

PR #487, opened 2015-05-18
Fix neutron2snabb parser db schema

dosaboy:
Fixes expected schema for ml2_port_bindings db table to
match that of 2014.1.4 Neutron schema.

Also adds debug print statements.


lukego:
Merged onto `nfv` branch.

PR #488, opened 2015-05-18
neutron2snabb: Added automatic database schema detection

lukego:
neutron2snabb now scans the *.sql files produced by mysqldump to determine the order of columns in the database dump.

The original selftest.sh regression test case is reintroduced to confirm that the schema is understood correctly and the same configuration files are generated.

This depends on the neutron_sync_master to store the *.sql files, which is the subject of a separate change. If these files are not present then the old hard-coded default is used.


lukego:
Merged onto `nfv` branch.

PR #489, opened 2015-05-18
snabbnfv traffic: Check PCI device exists and is supported

lukego:
This is now checked at startup so that bad command line arguments are found quickly. Previously the error would not be detected until a configuration file with a port was loaded and the error was cryptic.

Reported by @dosaboy.


lukego:
Merged onto `nfv` branch.

PR #490, opened 2015-05-18
NFV fixes since Jojoba

eugeneia:
This branch neutron-sync-master and neutron-sync-agent more robust and failure proof. Additionally their output is less verbose and now they print timestamped log messages instead.

### DIFF
--- a/src/program/snabbnfv/neutron2snabb/selftest.sh
+++ b/src/program/snabbnfv/neutron2snabb/selftest.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+export TESTDIR=/tmp/snabbtest
+
+mkdir -p $TESTDIR
+
+./snabb snabbnfv neutron2snabb \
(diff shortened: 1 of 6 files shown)


lukego:
I have merged this into the `nfv` branch for wider testing now. (Surprised Github doesn' show that? You can see it when examining the individual commits at least.)

eugeneia:
> Is SnabbBot process nfv btw? or do I need to ge my &*@! bench_env
> working again? :)

I assume you mean if SnabbBot will run for PRs on the snabb-nfv repo? No
it won't. I can make it so tomorrow. For now SnabbBot will run every
minute (e.g. as fast as it can) on SnabbCo/snabbswitch though. That is
with a lock so that it won't interfere with itself.

lukego:
Thanks. I would like to have SnabbBot runs on the `nfv` branch too please. I will PR this to master to trigger for now.

lukego:
Merged into `nfv` branch. Will be included on a future PR from there to master.

PR #493, opened 2015-05-19
Makefile: Fix dependency error that broke 'make -j'

### LINKED ISSUE #471: Makefile: parallel build (make -j) fails on fresh checkout
The build fails if you run:

```
git clone https://github.com/SnabbCo/snabbswitch
cd snabbswitch
make -j
```

Looks like a missing dependency where it will try to build LuaJIT before downloading it by initializing the submodule.

lukego:
There was a mistake in the LuaJIT dependency that would cause a parallel make to break. (It would expect to find the LuaJIT Makefile while asynchronously cloning the repo.)

Fixes #471.


lukego:
Closing because this has been merged onto the `fixes` branch.

PR #494, opened 2015-05-20
Modify checksum API

alexandergall:
I'm not happy with the current API for the IP checksum implementation.  I have discussed the issues in the thread https://groups.google.com/forum/#!topic/snabb-devel/nlifqWTDqzc

This PR contains the following proposed changes:

The type of the `initial` argument is changed to uint16_t throughout and its endianness is now expected to be the same as the output of ipsum(). The old code used endian conversions in an asymmetrical manner, which I find very awkward.

The code for cksum_generic() is cleaned up a bit.

The documentation in the code has been extended to describe the full semantics of ipsum(). In particular, the output of ipsum() was previously described as being in network byte order, when, in fact, it is in host byte order, i.e. an endian conversion is required to write the checksum to memory in network byte order.

I think that the convention that the result of ipsum() and the `initial` argument are in host byte order is the natural choice. Conversions have to be performed when a checksum is read from or written to a packet buffer.

### DIFF
--- a/src/arch/avx2.c
+++ b/src/arch/avx2.c
@@ -63,7 +63,7 @@ static inline uint32_t cksum_avx2_loop(unsigned char *p, size_t n)
 
 uint16_t cksum_avx2(unsigned char *p, size_t n, uint32_t initial)
 {
-    uint32_t sum = ntohs(initial);
+    uint32_t sum = initial;
 
     if (n < 128) { return cksum_generic(p, n, initial); }
     if (n >= 64) {
--- a/src/arch/sse2.c
+++ b/src/arch/sse2.c
@@ -66,7 +66,7 @@ static inline uint32_t cksum_sse2_loop(unsigned char *p, size_t n)
 
 uint16_t cksum_sse2(unsigned char *p, size_t n, uint32_t initial)
 {
-  uint32_t sum = ntohs(initial);
+  uint32_t sum = initial;
 
   if (n < 128) { return cksum_generic(p, n, initial); }
   int unaligned = (unsigned long) p & 0xf;
--- a/src/lib/checksum.c
+++ b/src/lib/checksum.c
@@ -14,11 +14,10 @@
 #include <stdint.h>
 #include <sys/time.h>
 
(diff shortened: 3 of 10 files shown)


lukego:
This API change makes sense to me.

The code looks nice too. I like the way you simplified the generic checksum routine. Is there any meaningful performance impact on that? If not then great. But we should remove the comment at the top of the file saying that the generic checksum is based on the DPDK one (http://dpdk.org/browse/dpdk/tree/lib/librte_net/rte_ip.h) because it will not be anymore.

alexandergall:
> The code looks nice too. I like the way you simplified the generic checksum routine. Is there any meaningful performance impact on that?

I didn't check, but I don't expect a difference, because of the inlining in the old code.

> But we should remove the comment at the top of the file saying that the generic checksum is based on the DPDK one (http://dpdk.org/browse/dpdk/tree/lib/librte_net/rte_ip.h) because it will not be anymore.

Ok.

alexandergall:
> On first reading it sounded to me like we are saying that the caller needs to pad the input with a zero-byte if it is odd-length. Could be worth emphasising that this is not required?

Maybe it shouldn't be mentioned at all, since it's part of the algorithm.

> Tony Rogvall who wrote the original SIMD checksum had an interesting idea for an API tweak for optimization btw. The idea is to actually insert zeros into the area following the payload so that no loop is needed to take care of the unaligned part at the end.

But how would one make sure that there is room for the padding bytes after the data?

lukego:
> But how would one make sure that there is room for the padding bytes after the data?

This seems doable in the special case that we are checksumming a `struct packet`: we could define a system-wide maximum packet data size and ensure there is some excess in the size of the `struct packet.data` array. Then every packet always has trailing padding. This would add complexity though so we wouldn't want to do it unless there was a meaningful payoff.

lukego:
This change is merged onto the `next` branch now so I close this PR.

I made a small change to the comment in the merge (cc0ebb9). If you would like to rephrase that you can just push a new commit on this branch and reopen the PR.

PR #495, opened 2015-05-20
Add a simple rate-limited logging facility

alexandergall:
The logger uses a simple token-bucket to limit the number of messages
writen to an arbitrary file handle. Usage:

    local logger = lib.logger_new( { rate = <rate>,
                                     fh = <fh>,
                                     flush = true|false,
                                     module = <module>
                                     date = true|false })
    logger:log(message)

    <rate> maximum rate of messages per second.  Additional
           messages are discarded. Default: 10
    <fh>   file handle to log to.  Default: io.stdout
    flush  flush <fh> after each message if true
    <module> name of the module to include in the message
    date   include date in messages if true

The output format is

    <date> <module>: message`

The date is generated via os.date() and has only second precision.

### DIFF
--- a/src/core/lib.lua
+++ b/src/core/lib.lua
@@ -432,6 +432,74 @@ function root_check (message)
    end
 end
 
+-- Simple rate-limited logging facility.  Usage:
+--
+--   local logger = lib.logger_new({ rate = <rate>,
+--                                   fh = <fh>,
+--                                   flush = true|false,
+--                                   module = <module>,
+--                                   date = true|false })
+--   logger:log(message)
+--
+-- <rate> maximum rate of messages per second.  Additional
+--        messages are discarded. Default: 10
+-- <fh>   file handle to log to.  Default: io.stdout
+-- flush  flush <fh> after each message if true
+-- <module> name of the module to include in the message
+-- date   include date in messages if true
+--
+-- The output format is
+-- <date> <module>: message
+--
+local logger = {}
+-- Default configuration
+logger.config = { rate = 10,
+		  fh = io.stdout,
+		  flush = true,
+		  module = '',
+		  date = true }
+
+function logger_new (config)
+   local config = config or {}
+   local l = { config = {} }
+   setmetatable(l.config, { __index = logger.config })
+   for k, v in pairs(config) do
+      assert(logger.config[k], "Unkown logger configuration "..k)
+      l.config[k] = v
+   end
+   l.tstamp = C.get_unix_time()
+   l.token_bucket = l.config.rate
+   l.discard = 0
+   return setmetatable(l, { __index = logger })
+end
+
+function logger:log (msg)
+   local rate = self.config.rate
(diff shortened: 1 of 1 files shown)


lukego:
Boy do we need something like this :-)

Couple of questions about intended usage:

1. How often can you reasonably call `logger:log()`? Is once per packet too often? Once per push? What is a practical limit to suggest?
2. How careful do you have to be about log messages that are expensive to compute? (Is it going to be a problem to use things like `..` string concatenation or `string.format()` even in cases where the message is suppressed?)

lukego:
Commit 979af8c3 is an example of a case where we had a log message during traffic processing and ended up hitting the pathological case where pretty much every packet causes a log message to be printed (and that message to be a computed string no less).

It would be awesome to have a generic logging capability that could be used in that situation without wrecking performance.

lukego:
Thinking aloud...

How about if we had separately a generic token bucket object and a function for printing well-formatted log messages?

So then we could write:

```lua
local logger = lib.tokenbucket(...)
...
if logger:take() then lib.log(...) end
```

with the nice benefits that:

1. The `take()` operation could be fast enough to run on every packet.
2. The rate limiter app could adopt this token bucket too (save on code).
3. The log message would only be constructed if there is a token so it would be okay to use expensive operations even on the traffic path (since we know it will seldom run).

alexandergall:
@lukego a generic token bucket is certainly a good idea. I would still prefer to integrate it in the logging function (and provide a take() method as a pass-through to the token bucket) to avoid a meltdown due to a badly written app.

lukego:
Good point. Logging the number of suppressed messages is important.

One compromise idea: How about if the logger took the same arguments as `string.format`? So you could write `logger:log("Packet too small: %d bytes", packet.length")` and the string would only be constructed if there is a token available? That might be nice if it meant you could expect good performance even when calling `log()` on every packet. (On the other hand maybe that varargs call is a LuaJIT NYI and would not be efficient anyway..)

alexandergall:
With this new version, you can do

    if logger:can_log() then logger:log(...) end

I tried this with one of my test cases and saw a drop of the packet rate from 7Mpps to 5.3Mpps (on interlaken) when logging a message for every single packet.  I believe that the performance drop is exclusively due to the call of clock_gettime().  Maybe this is the time to revisit the RTDSC issue :) With that, the logger could be essentially hitless, I think.

alexandergall:
Oh, and I've added an auto-throttling mechanism that will reduce the logging rate when too many messages are being discarded.  In the extreme case, like my test case with a message for every packet, only one message per 10 seconds will actually be logged to prevent clobbering of the log file.

lukego:
Cool :). This is merged onto the `next` branch for testing for the next release.

I am very keen to improve our logging and counter usage across the board so that we can troubleshoot production systems more easily.

lukego:
Merged onto `next` for inclusion in the next release. Closing PR.

Issue #481, opened 2015-05-14
snabbnfv traffic: Check PCI device on startup


lukego:
Make the `snabbnfv traffic` command check during startup that its network interface exists and is supported.

Currently the network device is not opened unless/until a port definition is supplied in the configuration file.

lukego:
Resolved by #489.

PR #496, opened 2015-05-21
[snabbnfv traffic] Fix issue #481: Check PCI device on startup

eugeneia:
This fixes issue #481 (snabbnfv traffic:Check PCI device on startup) .

### DIFF
--- a/src/program/snabbnfv/traffic/traffic.lua
+++ b/src/program/snabbnfv/traffic/traffic.lua
@@ -2,6 +2,7 @@ module(..., package.seeall)
 
 local lib = require("core.lib")
 local nfvconfig = require("program.snabbnfv.nfvconfig")
+local pci = require("lib.hardware.pci")
 local usage = require("program.snabbnfv.traffic.README_inc")
 local ffi = require("ffi")
 local C = ffi.C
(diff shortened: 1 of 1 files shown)


lukego:
Hah! I also addressed this in #481. 

Looks like I did two things wrong to obscure that change:

1. Failed to reference the bug Issue in the Pull Request, so that Github would cross reference.
2. Closed the PR after having merged it onto the `nfv` branch (but not yet onto master).

Is there one that we should keep or should we merge them together?

eugeneia:
Whoops I did see that PR at some point, well there goes short term memory. Your PR (#489) solves this in a generally neater way (no useless loop) but it doesn't check for `is_usable`. I am not sure if it should since the driver(s) unbind(?) the device anyways?

lukego:
Good point.

In a perfect world it could be nice to refuse to operate on ports that the kernel is currently using and insist that the user do `ifconfig ethxx down`. That would avoid really annoying problems like wiping out networking on the machine by accidentally using the wrong PCI address.

However, this would not be backwards compatible behavior for `snabbnfv traffic` and so I suspect we would do more harm than good by introducing this now. I close this PR for now.

PR #497, opened 2015-05-23
Document git workflow with new 'next' and 'fixes' branches

lukego:
Define a new Git workflow that is supposed to make life easier for users, contributors, application developers, and maintainers. Specifically aim to keep `master` stable for users and application developers and create a new `next` branch 

I know everybody surely has "git workflow fatigue" by now. Sorry. This is an attempt to solve an important problem: relieve the tension between providing a stable base and quickly bringing in new code for wider testing. I am not satisfied with the old workflow of landing individual new changes directly on master: it creates an awkward compromise of somewhat stable code landing at a somewhat reasonable pace.

@eugeneia and I will have to work out how to properly fit this into the release workflow.

### DIFF
--- a/src/doc/branches.md
+++ b/src/doc/branches.md
@@ -11,13 +11,35 @@ The current state of each branch with respect to master is visible here:
 #### master
 
     BRANCH: master git://github.com/lukego/snabbswitch
-    Synchronization point with changes recommended for all users.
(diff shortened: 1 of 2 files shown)


lukego:
I am pushing an update that expands the content and revises the language a bit.

PR #500, opened 2015-05-24
fixes: bug fixes collected for next v2015.05 maintenance release

lukego:


eugeneia:
The title confuses me. Shouldn't it be updated to say 2015.06(.1)? If so should I do some sort of bug fix release tomorrow? If I remember correctly you requested weekly bug fix releases (which is quite frequent, no?).

lukego:
Hoi :).

Yeah, somewhere in the Git workflow discussion I got it into my head that we should make weekly maintenance releases and just started acting as if this was something we had actually discussed. Excuse me :).

The idea is just to reduce the maximum latency of small but important fixes landing on master. If we only update master once per month then dumb bugs may end up bothering people for weeks longer than necessary. For example, all of our tutorial material says to `make -j` but that often doesn't work because of a race condition in `Makefile`. I don't want to regress into the habit of pushing individual fixes like this directly onto master and so "merge `fixes` onto `master` on mondays if it happens to contain new commits" seems like a reasonable compromise between "merge right now" or "merge next month".

What do you think?

lukego:
I am not sure what is the optimal process for bugfix releases.

One idea would be:

1. If `fixes` contains no new commits, then skip.
2. Merge `fixes` onto master.
3. Tag as `vYYYY.MM.N` with the first available `N`.
4. Create Github release page with the shortlog.

Could be that we could automate the release process over the longer term so that we humans focus only on having the branches in the right state at the right time. Having humans pushing to `master` is a bit scary. I actually very briefly screwed up `master` last week by accidentally pushing a couple of commits and then quickly rebased them away again. I would love to have a process that doesn't permit such goofs.

eugeneia:
Seems sane to me. So as of now `next` does not contain new commits.

lukego:
To me it seems that `fixes` does contain new commits: the ones shown above on this PR.

Admittedly this is only one small fix to Makefile (ab563cc4098792e6f7bc699cdb0c3a0405fc5467 ) plus some excessively noisy merge commits. However, that Makefile fix does seem to be needed to make the Getting Started instructions work reliably i.e. for `make -j` to not fail.

Or do I have a mistake somewhere here?

(Too bad that the lab is offline and so SnabbBot can't run tests.)

eugeneia:
Oh right I mistakenly did mention that commit in the Kiwi release even though it wasn't in. My bad, fixed.

lukego:
I should have merged `fixes` onto `next` before the release but goofed. have not mastered the multibranch workflow yet.

lukego:
These landed on `next` and will be included in the July release.

PR #472, opened 2015-05-13
Integrate snabb top

eugeneia:
This adds a `snabb top` program. `snabb top` is powered by `core.shm` and `core.counter`. It is not a curses program but uses VT100 term codes to emulate a single updating screen. This works Okayish, suggestions welcome. Consider this a work in progress, its not done yet:

- [x] Implement working `snabb top`
- [ ] Agree on shared memory mapping layout

PR #504, opened 2015-05-26
Vm sandbox

javierguerragiraldez:
A cleaner and less disrupting version of what we've seen on previous 'style proposal' branches.

- There's no 'general purpose' wrapping of the Lua C API, just enough methods to make `sandbox_loader.lua` source readable.

- Doesn't depend on `app.lua` calling `app:relink()`. Instead, adds an explicit `:inject_links()` method.

- Cleaner method to share the packet freelist.

- The `engine.frees`, `.freebytes` and `.freebits` are moved to an `ipc.shmem.counters` file, allowing any sandbox to update it.  It could also be an FFI struct shared in a similar manner to `packet_fl`. In a multithreaded future it could be a per-thread struct, updating the central statistics after each `breath()` cycle.

### DIFF
--- a/src/apps/basic/sink.lua
+++ b/src/apps/basic/sink.lua
@@ -1,0 +1,9 @@
+
+function push()
+   for _, i in ipairs(input) do
+      for _ = 1, link.nreadable(i) do
+         local p = link.receive(i)
+         packet.free(p)
+      end
+   end
+end
--- a/src/apps/basic/source.lua
+++ b/src/apps/basic/source.lua
@@ -1,0 +1,16 @@
+
+local size = size or 60
+local pkt = packet.from_pointer (ffi.new("char[?]", size), size)
+
+function pull()
+   for _, o in ipairs(output) do
+      for i = 1, link.nwritable(o) do
+         link.transmit(o, packet.clone(pkt))
+      end
+   end
+end
+
+
+function stop()
+   packet.free(pkt)
+end
--- a/src/apps/basic/test.lua
+++ b/src/apps/basic/test.lua
@@ -1,0 +1,11 @@
+module(..., package.seeall)
+
+function selftest()
+   local c = config.new()
+   config.app(c, 'source', 'apps.basic.source', {size=120})
+   config.app(c, 'sink', 'apps.basic.sink')
+   config.link(c, 'source.output -> sink.input')
+   engine.configure(c)
+
+   engine.main{duration=1, report={showlinks=true}}
+end
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -211,6 +211,11 @@ function apply_config_actions (actions, conf)
    -- commit changes
    app_table, link_table = new_app_table, new_link_table
    app_array, link_array = new_app_array, new_link_array
+   for _, app in pairs(app_table) do
+      if app.inject_links then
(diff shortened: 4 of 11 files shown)


lukego:
This is impressively simple on a few different levels.

I really like how simple basic apps like [sink](https://github.com/javierguerragiraldez/snabbswitch/blob/vm-sandbox/src/apps/basic/sink.lua) have become.

I also really like the way the Lua API is designed for VMs to be managed by C but then the C FFI is so powerful that you can even more easily control them from Lua :-).

Looking forward to taking it for a spin. Looks like the file `core/stats.lua` is missing at the moment.

javierguerragiraldez:
> Looks like the file core/stats.lua is missing at the moment.

[D'oh!](http://www.oed.com/view/Entry/249869)

i'll push it at lunchtime...

lukego:
Thanks Javier! This runs now:

```
$ sudo ./snabb snsh -t apps.basic.test
link report:
          34,917,405 sent on source.output -> sink.input (loss rate: 0%)
```

I would like to spend some time with this and aim to merge it for next month's release.

I suppose there is nothing stopping you from attacking N:M threading now :-)

lukego:
This is very interesting code to read. I am quite fascinated that `engine.configure()` does not need to know that some apps are now in their own sandboxes, because behind the scenes `config.app()` is creating backwards-compatible app "classes" that automatically create a sandboxed object when instantiated with `new()` and forward method calls to `pull()`, `push()`, etc. I had expected the meat of this change to be in `engine.configure()` but that code did not actually have to be touched at all. Clever.

I will come back and read through `sandbox_loader.lua` carefully. That code is quite intimidating at first glance :-)

I am also reflecting on the increasingly many use cases that we have for shared memory these days: sharing memory between Lua states, sharing memory between threads (potentially), sharing memory with other Snabb processes (SnabbTop), sharing memory with external application (Alex's NetSNMP plugin). I wonder if there is a simple way that we could consolidate the code for these.

lukego:
@javierguerragiraldez This branch seems to be in quite good shape now. How about if you move your new development onto a new branch based on this one and shoot for a 1

lukego:
@javierguerragiraldez I would like to merge this code onto a branch of my own and work on integrating it in the smoothest way with `core.shm` (#513) and SnabbTop (#472). I would close this PR and open a new PR from my branch that includes these commits. I am hoping there will be opportunities for cross-module simplifications. Would that be okay with you? (You should be able to keep doing the parallel hackery based on this same version and we can do a merge later when all parts are ready.)

javierguerragiraldez:
Great!  if you want to do it yourself, the first things that could be simplified are the `core.freelist`, which wouldn't need the `receive()` function, and `core.packet` would just call `freelist.new('core/packet/freelist', 'struct packet *', max_packets)` instead of checking if it has the global lightuserdata pointer already injected.

Similarly with the `core.link` objects, the `app:inject_links()` method could simply pass the registered names of the links and then do `for i,name in ipairs(inputlinknames) do input[i]=shm.map(name,'struct link*') end` from the 'inside'.

lukego:
Closing this PR because we have had a productive review and I think if/when/how we merge sandboxing will depend on the parallelism model we choose. (Can reopen if my understanding here is dodgy.)

PR #474, opened 2015-05-13
Improve handling of program list (for snabb --help)

lukego:
Several small fixes:

- Add src/program/programs.inc to gitignore
- Fix `FORCE` rule trick in Makefile to ensure it is updated
- Fix usage printout to program names with punctuation (e.g. example_foo)

PR #512, opened 2015-05-28
.gitignore: Ignore src/program/programs.inc

lukego:


lukego:
(Nevermind: This is already bundled in #474.)

PR #513, opened 2015-05-30
core.shm: shared memory alternative to ffi.new()

lukego:
This is a request for comments idea for a new primitive to unify our shared memory code.

The idea is to add a really simple mechanism that can handle simple cases directly and otherwise provide a good base for adding to.

These days we have multiple use cases for shared memory:

1. Share counters with external software (NetSNMP for Alex's VPN).
2. Share counters with other Snabb Switch process (SnabbTop #472).
3. Share objects between Lua VMs (counters, freelists, links, etc, in the sandbox branch #504).

The idea here is that each logical object would be a ramdisk file with a well-known name that can be `mmap()`'ed into any program. The object could be a single `uint64_t[1]` value for a counter, a `struct link` being accessed by an app, or a larger object like a `char buffer[102400]`.

The main idea this would offer for `lib.ipc` shm is that we could split objects into individual files and let the filesystem deal with naming and associating metadata. The main idea it would offer to the sandbox code is that we could use shared FFI structs for passing state instead of passing individual values via the Lua C API.

I will look more closely at how the use cases really do match or not match this mechanism but I wanted to post this early draft for feedback from @alexandergall, @javierguerragiraldez, and @eugeneia on whether this would potentially handle all our use cases or not.


lukego:
My motivation here is that I would like to shrink the size of the core code.

This is growing a bit now in that SnabbTop is using `lib.ipc` from the `core.app` module and the sandbox branch is also doing significant work to pass objects between VMs. I would be really happy if we could have a second look at the whole topic of shared memory IPC and find a simpler solution that works for everybody.

javierguerragiraldez:
This is very nice!   In the sandbox branch I certainly wanted a namespace for shared objects, but didn't made the leap from that concept to the SysV IPC shmem.  I was only thinking on a FFI name->blob+type dictionary.  (And just now noted it's called POSIX shm... am I that old?)

But I think this API is a little too "filesystem-like".  I'd rather have something like just a "named `ffi.new()`".  Maybe something like this (very untested!!):

    function shm_get(name, type)
       local ctype = ffi.typeof(type)
       local size = ffi.sizeof(ctype)
   
       local fd, err = S.shm_open(name, 'rdwr')
       if not fd then error("shm_open " .. name .. ": " .. tostring(err)) end
   
       local mem, err = S.mmap(nil, size, "read, write", "shared", fd, 0)
       if mem == nil then error("mmap failed: " .. tostring(err)) end
   
       S.shm_unlink(name)
   
       local obj = ffi.cast(ffi.typeof("$&", ctype), mem)
       return ffi.gc(obj, function(o)
          if o.__gc then o.__gc(o) end
          S.munmap(o, size)
       end)
    end

In short, this does an `shm_open()` (maybe creating), `mmap()` and `shm_unlink()`.  The unlink won't destroy the object as long as there's an active `mmap`.  The `ffi.gc()` call sets the `munmap()` to happen on GC.  I don't know if there's a way to check if the `ctype` already has a `__gc` metamethod, which the `ffi.gc()` would override.  The `o.__gc` check would only work in the common case where the metatable is it's own `__index` (so, not in the `ffi.metatype(ctype, {__index=methods})` style, which I personally don't like).

With this, when two different parts of the code do an `shm_get()` with the same name, they would get pointers to the same blob, without having to care which one creates and which 'receives' the object.  When the last instance is GC'ed, the shm object would be destroyed (since all `open()` are immediately `close()`d)

The most obvious con of this vs. the current proposal is that two different pieces of code could use the same object with different types and hilarity ensues.  Although, I don't see this as a too serious danger; at least not worse than currently with C or FFI structs.

About the current shm lib... could it be reimplemented on top of either of these ideas?  it could be as simple as documenting: "the frobnulator system keeps status data on a `struct { int fribles, unsigned long tribbles}` shared object named "frobble".  get it with `local frobble = shm_get("frobble", "struct{....}")`."  Mmmmh... maybe it's worthwhile to add a readonly option?

alexandergall:
This differs from lib.ipc.shmem in two ways: the ctype is part of the metadata and each object exists as a single entity in the file system.  The ctype can only be used by receivers that use LuaJIT, so the type still needs to be inferred from the name in the general case (e.g. my SNMP agent written in Perl).

Apart from the ctype, which could easily be added to the index file of lib.ipc.shmem, this is a degenerate case of that framework where every index contains just one object, right? The goal is to make the code simpler and this version certainly is. But will it remain so?

I think it's the common case that an application is handling a whole bunch of objects, which would be collected in a directory, I suppose.  So, the application will have to perform all the stuff that comes with that (creating it, walking it to find all objects etc.). I would immediately add code to the module that would do that for me, because I don't want to repeat it in every app. My guess is that we would end up with something not unlike lib.ipc.shmem...

alexandergall:
Actually, I have never used this mechanism. How exactly would one structure the objects? The name space is flat, isn't it? Maybe I should read up on that before posting lengthy comments :)

lukego:
I have pushed a redesign based on the early feedback so far. How does it look?

@javierguerragiraldez I like your interface! I have rewritten the code to be similar. I kept the explicit `unmap()` and `unlink()` operations because I think that doing these things on a garbage collection hook is scary and will create problems that are hard to debug. (I think the `unlink()` is done too early in your example btw and that each call would allocate and return a different object.)

@alexandergall Thanks for the initial assessment, I am interested to hear your follow on thoughts :-). If this code would simplify the SNMP plugin then that would be the icing on the cake. The main motivation though is to support new use cases for which `lib.ipc.shmem` would be overkill e.g. for multiple Lua VMs to have a shared freelist for packets, shared link structs, and so on. This would allow us to remove code from the SnabbTop and SandboxVM branches. (I would also be happy to avoid having `core.*` calling `lib.*` as on the SnabbTop branch because that feels like a layering violation.)

lukego:
The reason I dropped storing the type with the object is that this was an awkward feature. You don't need to store the type in simple cases, and you probably don't want to store it in that exact way in complex cases. So now it is out of the way and the caller can deal with types however it wants e.g. hard-code them or store them in suitable metadata file(s).

alexandergall:
I think that it would be easy to re-write `lib.ipc.shmem` to use this as basis and just add the index and handling of custom types (e.g. `OctetStr` in lib.ipc.shmem.mib').  The API could probably remain unchanged. The index file would simply include the full path of the object (the size would no longer be needed).  I think I would keep the name space versioning, because I feel that this will be something that we need later on.

lukego:
I have extended the syntax for object names. Does the idea come across clearly? Is this better or worse?

The intention is to make it easier to name and organize objects. Particularly, the engine could set `shm.path = app.name` when switching between apps and then it would be easy to create app-local objects e.g. `shm.map("counter/drops")` creating `/var/run/snabb/1234/app/myapp/counter/drops`. Then we would have no excuse not to define multiple counters for each and every app & find a way to conveniently monitor them.

This also hopefully brings the code and directory layout more into line with SnabbTop on #472.

lukego:
@alexandergall Cool. Great that the low-level shared memory objects and the higher-level abstractions like well-defined SNMP MIBs can be independent layers. I suspect we will hit some new challenges on the higher levels soon e.g. to aggregate counters split across multiple processes/VMs. One step at a time though :).

javierguerragiraldez:
> Particularly, the engine could set shm.path = app.name when switching between apps

on sandboxed apps, each one would have its own copy of the `shm` namespace, so `shm.path` could be set at instantiation time and left there.

lukego:
I added a `core.counter` module intended to make counters so incredibly easy to work with that literally every module in Snabb Switch can use them. This is a replacement for `lib.ipc.shmem.counter` which is roughly my previous attempt at facilitating ubiquitous counters that never really took off.

So now we have three layers in terms of counters:

1. Generic shared memory objects: `core.shm`.
2. Counters as `uint64_t` shared memory objects for ad-hoc use: `core.counter`.
3. Curated collections of objects intended for external applications to consume: `lib.ipc.shmem.mib` providing data for a NetSNMP agent.

Three layers is a lot but I feel like this is moving in a reasonable direction. The `counter` module is not really an abstraction but rather a few convenience functions.

How does this look?

If this direction still looks promising then the future steps to gradually take would be:

- Use `core.shm` for internal objects e.g. freelists, links, etc so that they are available.
- Take advantage of this in the VM sandbox branch.
- Use `core.counter` to start tracking all manner of interesting events inside Snabb Switch.
- Take advantage of this in SnabbTop to give more visibility into Snabb Switch processes.
- Update the MIB code to use `core.shm` and `core.counter` in whatever way makes the most sense.

javierguerragiraldez:
this looks much better than the original `lib.ipc.shmem.counter`, which I had to fix a bit to make it work on the sandbox branch.  In that case there were 9 closely related counters, so if done now, I'd use a struct to bind them all and share with `core.shm`, maybe even a two level struct like this:

    struct {
        struct {
            uint64_t packets, bytes, bits;
        } current, previous, reported;
    }

lukego:
This PR was based on v2015.05. I updated it to v2015.06 with `git merge master`. Seems to have work out neatly! Guess we don't need to rebase for this case.

lukego:
Ouch, no, I goofed that. I should have used `git merge origin/master` because my local master branch was not clean. So I have a couple of unwaned submodule updates bundled onto this branch now. I will live with that for now because a rebase would probably be too destructive. Have to resolve somehow before merging this branch.

lukego:
Yuck... I made a mess, fixed it with rebase, and *hopefully* this does not have unintended consequences. Seems like in this case rebase did not completely screw up the pull request when I rebased back onto an earlier commit of this branch. This was risky and probably was not a good idea.

The net result is that I have merged `v2015.06` into this branch cleanly now whereas before I had merged a non-pristine version of `master` that introduced some regressions with submodules.

eugeneia:
@lukego I started working on porting Snabb Top to `core.counter`, which will only work together well if I update the existing core statistics counters to use the new `core.shm` too (`frees`, ..., `breaths`). Just wanted to let you know to avoid us both potentially doing the same thing.

lukego:
@eugeneia I am happy if you want to charge ahead on that and make use of this code. This might be a good time to merge this PR onto `next` to make it easier to build on for you and @javierguerragiraldez?

I have an app that I want to write to explore some of the new ideas we have been talking about lately so I would happily context-switch to that.

src/core/counter.lua
```diff
@@ -0,0 +1,51 @@
...
+
+module(..., package.seeall)
+
+local shm = require("core.shm")
+local ffi = require("ffi")
+local counter_t = ffi.typeof("struct { uint64_t c; }")
+
+function open (name)          return shm.map(name, counter_t) end
+function set (counter, value) counter.c = value               end
+function add (counter, value) counter.c = counter.c + value   end
```
eugeneia: Could `value` default to 1 if not supplied to `add`? E.g. like Common Lisps `INCF`.

src/core/counter.lua
```diff
@@ -0,0 +1,51 @@
...
+
+
+module(..., package.seeall)
+
+local shm = require("core.shm")
+local ffi = require("ffi")
+local counter_t = ffi.typeof("struct { uint64_t c; }")
+
+function open (name)          return shm.map(name, counter_t) end
+function set (counter, value) counter.c = value               end
```
eugeneia: Are counters initialized to 0 or do I have to initialize them using `set`?

lukego:
Shm objects are all initialized to 0 by ftruncate (). Should be documented

lukego:
Value defaulting to 1 sounds eminently sane

lukego:
I merged this branch onto `next`.

@eugeneia and @javierguerragiraldez if you want to PR code that depends on core.shm you could make `next` the PR target instead of `master` and then it should be less noisy i.e. not show these core.shm changes that you merge when they are already on the target branch.

lukego:
Merged onto `next`. Closing PR.

PR #506, opened 2015-05-27
luajit: upgrade feb 2015 (04dc64b) to may 2015 (5de95ed)

lukego:
Changes:

    * 5de95ed x64/LJ_GC64: Fix setmetatable().
    * 60fb3fe Fix table allocation bump optimization.
    * 4a146ca LJ_GC64: Fix frame unwinding.
    * bca5da8 Fix for last commit.
    * b82fc3d Bump table allocations retroactively if they grow later on.
    *   d8cfc37 Merge branch 'master' into v2.1
    |\
    | * 69e5342 RELEASE LuaJIT-2.0.4
    | * cd4c59f Update changelog.
    * | a836b7e Cleanup.
    * |   4cec6a5 Merge branch 'master' into v2.1
    |\ \
    | |/
    | * 9622d6b x86/x64: Fix argument check for bit shifts.
    * | a92e730 x64: Add LJ_GC64 mode interpreter.
    * | 5caf53d Fix broken Windows build.
    * | c2924c3 Disable trace stitching (for now) due to a design mistake.
    * |   74caac9 Merge branch 'master' into v2.1
    |\ \
    | |/
    | * 11106aa Fix stack check in narrowing optimization.
    * | 1e67a4c ARM64: Fix math.floor/math.ceil for string args.
    * | 1c968d5 DynASM/PPC: Add sub/shift/rotate/clear instruction aliases.
    * | b2a5cc8 DynASM/PPC: Add support for parameterized shifts/masks.
    * | dec4acc PPC: Fix cross-endian builds.
    * | 8002871 PPC: Fix write barrier in BC_TSETR.
    * |   4e52d1f Merge branch 'master' into v2.1
    |\ \
    | |/
    | * a9fd686 Doc cleanup.
    * | 361827c PPC64: Add build infrastructure.
    * |   c821e0a Merge branch 'master' into v2.1
    |\ \
    | |/
    | * 8b6af89 Fix Lua/C API typecheck error for special indexes.
    * |   2c0748f Merge branch 'master' into v2.1
    |\ \
    | |/
    | * 1f578bd FFI: Fix FOLD rule for TOBIT + CONV num.u32.
    * |   e1a8f3b Merge branch 'master' into v2.1
    |\ \
    | |/
    | * 5c0e33c ARM: Handle more arch defines.
    | * 7ff4768 Properly fail unsupported cross-compile to MIPS64.
    * |   906008d Merge branch 'master' into v2.1
    |\ \
    | |/
[...]

PR #514, opened 2015-06-01
next: collected changes for v2015.07 release

lukego:
Commits are added to this Pull Request as they are merged for integration into the v2015.07 release.


lukego:
Interesting test failure. Looks like luajit and pflua upgrades each worked individually (#506 and #507 passed CI) but seem to have trouble when combined together and with the latest Snabb release? I will take a look at this tomorrow.

lukego:
Interesting. SnabbBot is failing the current build with `program.snabbnfv.nfvconfig` triggering an error in pflua ([log](https://gist.github.com/SnabbBot/8fbed2912cfa49228602#file-log-L490-L514)). However, this is not reproducible on my development machine.

I wonder if recent LuaJIT versions have introduce changes that break pflua? That seemed to happen with issue Igalia/pflua#225 which reproducibly happens only with the LuaJIT update that we have on this branch.

kbara:
This is still a work in progress, but the analysis on 225 (especially by mpeterv) shows pretty conclusively that this is a LuaJIT issue. I'd be tempted to stick with LuaJIT 2.0.x if I were you; when I benchmarked them last year, it was faster, and it's also more stable.
Diego and I have both been able to intermittently reproduce the problem on our development machines by running ```sudo ./snabb snsh -t apps.packet_filter.pcap_filter``` (with version 8ac912d21c817f86618b6588ad13e23260deeea6 from the 'next' branch).
It can also be reproduced within pflua head without Snabb by running ```./env pflua-compile "(icmp6 and src net 3ffe:501:0:1001::2/128 and dst net 3ffe:507:0:1:200:86ff:fe05:8000/116) or (ip6 and udp and src net 3ffe:500::/28 and dst net 3ffe:0501:4819::/64 and src portrange 2397-2399 and dst port 53)"```; see mpeterv's analysis for how the details change between specific LuaJIT 2.1 commits.

lukego:
Interesting idea to use the stable LuaJIT v2.0.x. I am not sure if that would be realistic or not. Snabb Switch has used the v2.1 branch from the beginning and there are at least some new features of interest, like the profiler and 64-bit bitops.

So var v2.1 has been quite good to use and I do think it's supposed to be fairly stable. I think that Cloudflare run v2.1 in production on billions of HTTP requests per day. Could be an idea to ask them how they handle updates.

kbara:
Good points, and having 64-bit bitops is pretty nice; I'd be curious to know how Cloudflare's handling upgrades too.

lukego:
Seems like the LuaJIT+pflua upgrade problem is caused by bug(s) in the latest LuaJIT. So with hindsight it would have been better to revert LuaJIT upgrade instead of pflua.

Hopefully LuaJIT will have a fix shortly and we can merge that. Otherwise we could revert the LuaJIT upgrade and un-revert the pflua upgrade for this month. Let's give it a few days and see.

lukego:
@eugeneia any insight into this test failure? Seems like the tests passed by performance regressed to 0? :-)

eugeneia:
@lukego I had SnabbBot rerun it and it seems fine now.

kbara:
@lukego The problem is fixed with LuaJIT fe565222a1cbf28fbae266da35c8a703fdcfa0dd , which disables the optimization you correctly bisected the problem to. I've checked the fix with both pflua-compile and ```sudo ./snabb snsh -t apps.packet_filter.pcap_filter```, and everything appears to be fine, though I haven't run extensive tests beyond that.

lukego:
Awesome work @kbara and the pflua team! I prepared PR #520 to reintroduce the pflua upgrade together with the compatible LuaJIT and will merge that here once it passes CI.

Your hard work to create minimal test cases is a real contribution to LuaJIT and Mike's to collection of reproducible test cases for the future.

eugeneia:
If you wonder while this failed: don't. Rerunning it now.

eugeneia:
@lukego How are we doing on the release? I could do it today. Couldn't hurt to include #530 and #531 as well.

lukego:
@eugeneia Merged as requested. I am ready for release if you/SnabbBot don't see problems.

eugeneia:
https://github.com/SnabbCo/snabbswitch/releases/tag/v2015.07

PR #515, opened 2015-06-11
remove mixed tabs/space

javierguerragiraldez:
Is it here the right place for this?  Just the simplest of whitespace fixes (remove mixed tabs/spaces), but have been driving me nuts!

### DIFF
--- a/src/apps/basic/basic_apps.lua
+++ b/src/apps/basic/basic_apps.lua
@@ -41,7 +41,7 @@ function Join:new()
    return setmetatable({}, {__index=Join})
 end
 
-function Join:push () 
+function Join:push ()
    for _, inport in ipairs(self.input) do
       for n = 1,math.min(link.nreadable(inport), link.nwritable(self.output.out)) do
          transmit(self.output.out, receive(inport))
@@ -105,11 +105,11 @@ function Tee:push ()
          for _ = 1, math.min(link.nreadable(i), maxoutput) do
             local p = receive(i)
             maxoutput = maxoutput - 1
-	    do local output = self.output
-	       for k = 1, #output do
-		  transmit(output[k], k == #output and p or packet.clone(p))
-	       end
-	    end
+            do local output = self.output
+               for k = 1, #output do
+                  transmit(output[k], k == #output and p or packet.clone(p))
+               end
+            end
          end
       end
    end
--- a/src/apps/bridge/base.lua
+++ b/src/apps/bridge/base.lua
@@ -12,7 +12,7 @@
 --
 -- config = { ports = { <free-port1>, <free-port2>, ... },
 --            split_horizon_groups = {
---              <sh_group1> = { <shg1-port1>, <shg1-port2>, ...}, 
+--              <sh_group1> = { <shg1-port1>, <shg1-port2>, ...},
 --              ...},
 --            config = { <bridge-specific-config> } }
 --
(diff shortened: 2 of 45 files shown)


lukego:
This looks good to me.

Incidentally I recently read a way to globally disable tabs in Emacs:

    (setq-default indent-tabs-mode nil)

... whereas I have been doing this on mode-specific hooks for untold years.

@alexandergall what editor do you use? can you easily switch from tabs to spaces?

lukego:
It is very interesting to consider what is the best way to merge cleanups like this.

There is a bunch of development happening in parallel, mostly based on `master`, and changes that touch many files have potential to cause merge conflicts. So how should we deal with that?

If there are conflicts then it seems best to deal with them all at once and for this to be done by somebody who understands the context of what is going on. It would be bad if lots of people get surprised by conflicts when running `git pull master` and have to work out what is going on.

How about if we merge this onto `next` towards the end of the monthly release cycle? That way we have a chance to merge all of the currently open PRs ahead of this change without conflicts. Then we can merge this (resolving any conflicts from the merged PRs) and quickly release it onto master. The only conflicts for contributors will be for PRs that don't merge on this release: they might see a conflict when they `git pull master` next month to sync with the latest release.

In this case we would leave this PR as-is for now and in a week or two I will ask you to please `git merge next`onto this PR branch and resolve any conflicts and then resubmit the PR with branch target `next`.

Thanks for this initiative Javier. It's great for a few reasons: cleans up the code, helps us work out a cleanup workflow, and perhaps opens the door for more nice cleanups. There are a couple of other cleanups that I have been considering but did not do because I was too lazy to think through the implications with respect to conflicts.

lukego:
btw: another interesting topic is how to choose the right target branch for a PR.

The safe choice is `master` and that should always be fine when in doubt. PRs sent to master can easily be merged onto any other branch (`fixes`, `next`, etc) because it is based on a common ancestor.

The exception is when you are building on other changes that are not on master yet. Then it is better to send the PR to a branch that includes those changes (or to wait until they land on master). So for example code that depends on `core.shm` or the new logging facility that Alex wrote can be sent to `next` since those changes won't land on master until next month.

Cleanups like this are a grey area. On the one hand it does not explicitly depend on any other changes but on the other hand it does implicitly depend on every other change that touches the same code. So I think the above strategy of sending it to `next` at a strategic point in time makes sense.

I am still enjoying the Git workflow geekout... :-)

lukego:
also: once more thanks Javier for taking the initiative on this. This kind of "gardening" to keep the project neat and tidy is very valuable.

I like the idea of making cleanups and consistency improvements into a pleasantly meditative background process that we can all do when we feel like it. That seems much more pleasant than trying to enforce strict rules at all times e.g. having a CI that automatically rejects changes containing a tab character or a line over 80 columns or making code reviewers into style police.

Further thought: one day somebody could maintain a `cleanups` branch. Then when people send in cleanups to `master` at random times they could be collected on the `cleanups` branch and then merged cleanly with `next` towards the end of the release cycle. The way to volunteer for all such work as this would be to send a PR advertising a branch to `branches.md`.

eugeneia:
@javierguerragiraldez Thanks for doing this. I was slightly annoyed by tabs every now and then too.

@lukego Merging this right before release sounds good to me.

alexandergall:
@lukego I use Emacs. Will change to spaces. BTW, I'm slightly annoyed by the way the Lua mode handles indention when splitting long commands on multiple lines.  This sometimes looks awful and makes it hard to obey the 80-columns limit. Is anybody else bothered by that?

lukego:
@alexandergall I am also bothered by that. I am occasionally running over 80 columns as a consequence. I wonder whether we should adapt our programming style or look into customizing/extending lua-mode. Do you have some particular examples to point to? I don't have any in mind right now.

lukego:
Merged onto next.

Looks like there is even a command `git blame -w` that will skip over whitespace changes so that should make the impact on our Git history more manageable.

PR #517, opened 2015-06-12
test_env: Automagical replacement for bench_env

eugeneia:
`bench_env` is not trivial to set up correctly, and thus keeps people from using the NFV selftest etc. `test_env` is a sloppy rewrite of `bench_env` that is supposed to *just work*. See its [README.md](https://github.com/eugeneia/snabbswitch/blob/test_env/src/scripts/test_env/README.md).

As of now it implements just enough of `bench_env` to host `src/program/snabbnfv/selftest.sh`. `test_env` itself doesn't require configuration: It computes things like numa node bindings by itself, sets reasonable defaults (and tells you which ones you can change) and will fetch and compile all test assets automatically (qemu, kernel image, root filesystem image).

I would like to know which test environments of `bench_env` do you use? Which use cases should `test_env` support?

@lukego Can you enable a port forward from `lab1.snabb.co:2008` to `davos:80` for the new modem, so I can use davos as an asset host?


lukego:
Hallelujah!

This is really important. Seems like every time I hack my bench_env config I end up burning half a day to make it work again. It would be wonderful to make this *much* easier to run.

(Aside: I also pondered our e.g. Intel NIC selftests. Could be that they could automatically find a safe NIC to test and use VMDq or hardware loopback mode instead of depending on two ports cabled together. Would be neat if those tests would also run with zero setup on any machine that has suitable hardware not already in use.)

lukego:
I have setup the port forward now.

lukego:
Can you please put a tag like `[draft]` into the PR title until you are happy to see it merged into other branches? That will be more obvious than the note in the PR text because it is visible also when browsing the PR list.

eugeneia:
I consider `test_env` to be ready: I cold-tested it on grindelwald and it worked without requiring user intervention (except having to upgrade g++ to 4.9, since qemu wouldn't build otherwise). The work on `test_env` won't be done until it supports all use cases currently supported by `bench_env` (what are those?).

In this branch the snabbnfv selftest automatically uses `test_env` (when run via SnabbBot) davos. `bench_env` is not removed and will work unchanged.

lukego:
Exciting!

I would like to host the assets somewhere other than the lab. The lab is not always online and its uplink is much slower than its downlink.

I wonder what is the easiest way to host them on a CDN? (Dropbox? S3? Google Drive?)

eugeneia:
> I would like to host the assets somewhere other than the lab. The lab is not always online and its uplink is much slower than its downlink.

I consider the assets being hosted on davos OK for now because the uplink (2 MB/s) is enough to fetch the (compressed) assets in two minutes and there will not be much demand for the assets (you get them once and they pretty much never change).

> I wonder what is the easiest way to host them on a CDN? (Dropbox? S3? Google Drive?)

If we move them we should make sure they are delivered over HTTPS.

lukego:
I disagree. The lab is a pain compared with hosted services. It's much more prone to connection outages, hardware failures, accidents at the root shell, getting hacked, etc. It has no redundancy, no RAID, no backups, it gets powered off during storms, and it doesn't always have a pair of hands available to fix it.

I would like to work on diversifying. I think our core infrastructure should all be on professional hosted platforms (Github, etc) and that our supplementary infrastructure is more spread out (e.g. servers with interesting hardware to run tests on).

(Just as I write this there are two lab servers offline, `chur` and `jura`, and the server hosting `snabb.co` also went offline today. Solving these problems with unix sysadmin'ery feels very old fashioned to me.)

lukego:
One very important test case that is often hard to setup is the benchmark pushing packets through a DPDK virtual machine. The benchmark should run N parallel instances and measure how many packets/second can be forwarded through them. This has been the key benchmark in our NFV optimization efforts and it is really a pity that we don't have it hooked into the CI yet.

Can test_env help with that?

lukego:
This is a great step forward, Max :). The NFV selftest is much easier to run now. Thoughts follow!

Rhetorical question: Who should use test_env and why? The README defines it as a replacement for bench_env but most people will not know what that is. Is test_env a component of the snabbnfv test suite? When would you use it directly instead of via `program/snabbnfv/selftest.sh`?

Could `tmux` simplify test_env? I mean to take care of spawning processes, logging their output, terminating them as a group, and being able to attach to and manually restart individual processes when troubleshooting. That works quite well with the OpenStack `devstack` test script (using screen). Managing processes manually is less intuitive for me... e.g. to terminate the processes am I supposed to kill the shell that sourced test_env so that the shutdown hook runs?

Could this be the basis for a "Getting Started with snabbnfv" guide? That is sorely lacking in snabbswitch now: a really trivial runnable reference example of how to boot up a VM inside snabbnfv. (Have to think harder about the mailing list threads where people have struggled with qemu version, qemu args, iommu settings, etc.)

Should we move this under `program/snabbnfv`? Is it usable for other programs? (@alexandergall, did you end up using bench_env for your VPN? How?)

Can we delete bench_env once we merge test_env? If not then why not?

lukego:
Incidentally this is how I ended up running test_env:

    sudo TESTPCI=0000:01:00.0 TELNET_PORT0=5000 TELNET_PORT1=5001 program/snabbnfv/selftest.sh

and it looks like the test suite ran OK. Slightly ambiguous because I do see errors scrolling by and have been assuming those are expected e.g. this timeout:

```
Trying ::1...
Connected to localhost.
Escape character is '^]'.
nc -w 1 -q 1 -v  fe80::5054:ff:fe00:0000%eth0 12340
nc: connect to fe80::5054:ff:fe00:0000%eth0 port 12340 (tcp) timed out: Operation now in progress
root@vma:~# 
root@vma:~# Connection closed by foreign host.
FILTER succeded.
```

alexandergall:
@lukego  No, I didn't use bench_env. I don't need VMs for for my performance tests, just a `loadgen` on one of the looped NICs and an instance of my app. I simply do that manually.

eugeneia:
> I would like to work on diversifying. I think our core infrastructure should all be on professional hosted platforms (Github, etc) and that our supplementary infrastructure is more spread out (e.g. servers with interesting hardware to run tests on).

I can look into setting up a static hosting solution based on Amazon S3 for snabb.co (which could then also host other files) but the easiest way to get a HTTPS capable download for the assets would be to use an existing DropBox account I guess.

> One very important test case that is often hard to setup is the benchmark pushing packets through a DPDK virtual machine. The benchmark should run N parallel instances and measure how many packets/second can be forwarded through them. This has been the key benchmark in our NFV optimization efforts and it is really a pity that we don't have it hooked into the CI yet.

This should be the next test case supported by `test_env` then. On it!

> Rhetorical question: Who should use test_env and why?

It should be used by the people writing integration tests. It's purpose is to launch reproducible test environments without requiring setup or configuration by the person running the test.

> The README defines it as a replacement for bench_env but most people will not know what that is. Is test_env a component of the snabbnfv test suite?

`test_env` is indeed very snabbnfv centric as of now. I am not sure if it will ever be useful for testing anything else. I would categorize it as "Snabb Development Glue".

> When would you use it directly instead of via program/snabbnfv/selftest.sh?

When you want to run VM(s) connected via snabbnfv to telnet into. E.g. poke around, manual testing.

> Could tmux simplify test_env?

I am not sure, not a big user of tmux or screen. Can you propose a pseudo session using a tmux based `test_env`? 

> to terminate the processes am I supposed to kill the shell that sourced test_env so that the shutdown hook runs?

Yes, `test_env` works just like `bench_env` in that regard. E.g. when your test script using `test_env` exists all VMs and snabb instances will be killed and temporary files cleaned up.

> Can we delete bench_env once we merge test_env? If not then why not?

We can delete `bench_env` once all its use cases are supported by `test_env` (e.g. DPDK).

> and it looks like the test suite ran OK. Slightly ambiguous because I do see errors scrolling by and have been assuming those are expected e.g. this timeout:

That's all fine. The timeout is `PcapFilter` successfully dropping unwanted packets. :)

eugeneia:
@lukego Do we have a working setup of the DPDK based test somewhere?

lukego:
@eugeneia Recently I have been running the DPDK-based test manually. Here is what I do: I start a `loadgen` process on one port and a `qemu` process running the DPDK guest+kernel and I leave those running. Then when I want a test result I run `snabbnfv` and it connects everything together until I kill it. I find it really convenient to leave the loadgen and guest running because it means I can take a performance measurement in under a minute.

It is a long time since I ran the whole shebang under bench_env. That is very nice when it works but I always get tangled when commenting/uncommenting bits of my `bench_env.conf` to switch between test scenarios.

So I don't have a fully working reference setup at the moment.

lukego:
Can we host the test assets on Github? They could have their own repo and we could even link that as a submodule e.g. `program/snabbnfv/test-assets`. I have the impression that Git has some extensions for dealing with large files and that Github supports this use case up to a point, but I am not sure what point that is :-).

Sounds like `test_env` can be considered a part of snabbnfv. Let's move it under `program/snabbnfv/`. If other programs adopt it in the future then we could move it somewhere more common again.

`tmux` is basically a scripted multi-window terminal app. You can use commands/scripts to create a new window (terminal/shell) and you can send input to the window and log its output and so on. The killer feature from my perspective is that you can drive it both manually and with scripts. So even if `test_env` has everything highly automated you could attach your terminal to any of the processes to see their output directly, hit them with an ^C, etc. It is a bit like if everything were in Emacs and you could switch to the e.g. the `*snabbnfv*` buffer and poke the process directly when you want to.

Could even be that tmux would simplify sending commands to VMs via their stdio but I don't know if that would really be an improvement.

Check it out and see what you think?

lukego:
also a thought: is it important for `test_env` to be based on shell functions that are sourced, or could it also be based on stand-alone scripts e.g. to start QEMU with the right arguments? Separate scripts could be handy for ad-hoc reuse in other contexts. Generally I am a lot more comfortable with adding a directory to my `$PATH` to access some handy scripts like `snabbnfv-qemu` than with sourcing a library of shell functions.

kbara:
Nice work. The use case I'd most like to see is this level of automation with setting up an environment with two QEMU instances and an arbitrary Snabb application between them, so getting started would be a matter of running one script and perhaps changing IP addresses and a few similar details in the QEMU instances.

One step beyond that, it would be lovely to be able to have one more level of abstraction, so that it would be equally easy to get started without Intel NICs. http://blog.ipspace.net/2014/12/l2vpn-over-ipv6-with-snabb-switch-on.html mentions Alexander Gall's initial setup without an Intel NIC, using raw IO on a standard Linux socket with an app that opens a raw socket instead of Snabbswith's driver.  (at time 32:57-34:16). He says it gets a couple hundred megabits per second. It's a good option both for getting started, and for developing on a laptop without extra hardware.

I've found getting started with Snabb applications that are being fed a pcap file is quite easy, while getting a working setup with QEMU on both ends and Snabb in the middle on chur is surprisingly time-consuming.

eugeneia:
### Hosting assets

> Can we host the test assets on Github? They could have their own repo and we could even link that as a submodule e.g. program/snabbnfv/test-assets. I have the impression that Git has some extensions for dealing with large files and that Github supports this use case up to a point, but I am not sure what point that is :-).

https://help.github.com/articles/conditions-for-large-files/ says:

> "If you push a file larger than 100 MB to GitHub, Git will reject the push and tell you which file is too large"

### tmux

> Could even be that tmux would simplify sending commands to VMs via their stdio but I don't know if that would really be an improvement.

As far as I can tell `tmux` only supports `send-keys` as far as scripted input goes, which is no improvement over `telnet` (on the contrary it seems complicated to get the command output).

> Check it [tmux] out and see what you think?

I have a working branch that uses `tmux` instead of Bash's `&` (https://github.com/eugeneia/snabbswitch/tree/test_env-tmux-refactor), it doesn't change the original behavior except that you can do a `tmux attach-session` and find windows for each process started by `test_env`. I think this might come in handy for debugging and should not be intrusive otherwise.

### Shell functions vs Shell scripts

> also a thought: is it important for test_env to be based on shell functions that are sourced, or could it also be based on stand-alone scripts e.g. to start QEMU with the right arguments? Separate scripts could be handy for ad-hoc reuse in other contexts. Generally I am a lot more comfortable with adding a directory to my $PATH to access some handy scripts like snabbnfv-qemu than with sourcing a library of shell functions.

The main difference between sourcing a script and running a script is that sourcing a script can modify the *current* environment. So while shell functions can set up exit hooks and keep track of PIDs, if `test_env` was based on shell scripts then the user would have to manage the processes themselves. (E.g.: `snabb_pid=$(test_env/start_sabb.sh ...)`)

lukego:
@kbara Good point. It would be neat to have more flexibility about which app instances are used for an application. For example, the NFV application could just as well connect a VM with a PCAP file, or a raw socket, or whatever, as to a VMDq queue of a NIC. Likely we could even write a simple load generation app that has equivalent behaviour to a hardware NIC (in terms of cache behavior etc) and do a bunch of testing on laptops with that.

Could be that we could build generic `Input` and `Output` apps that take a URL-like string to say what they should attach to (`raw:eth0`, `pci:01:00.0`, `pcap:foo.cap`, etc?) This ties in a little bit with topic #522 and whether it makes most sense to have discrete `Input` and `Output` apps vs combined `IO` apps. That feels like a pretty fundamental design crossroads.

lukego:
@eugeneia I love the new test_env :). I am testing the tmux branch and I really like being able to flip between the processes (snabb, qemu0, qemu1) and see them running. Running nested tmux instances even seems to be quite straightforward (hit escape key once for outer tmux, twice for inner tmux).

If we are using tmux then we don't need to track `$snabb_pid` anymore, right? We can just kill the tmux session and it should take down all the child processes with it? (Looks like that is how you are doing it already?) So maybe shell scripts instead of shell functions is easier to adopt in this model too?

It is easy to imagine generalizing this to other use cases like @kbara's too, e.g. if you want to automatically start two QEMU guests with well-known vhost-user socket names, but then be able to start/restart the snabbswitch instance manually. But if the qemu startup is a stand-alone script then perhaps you just call that manually instead of using the framework to start the tmux.

I am keen to merge this :).

eugeneia:
> If we are using tmux then we don't need to track $snabb_pid anymore, right? We can just kill the tmux session and it should take down all the child processes with it? (Looks like that is how you are doing it already?) So maybe shell scripts instead of shell functions is easier to adopt in this model too?

Generally yes, its doable. Practically its still super annoying though. `test_env` has some implicit counters `$snabb_n` and `$qemu_n` that are incremented and spliced into log and asset paths (so the user doesn't have to say, log this first snabb switch instance to snabb0.log. etc). Also it still cleans up Vritio socket files on exit. Plus handy functions such as `ip` and `mac` used by the NFV selftest to get network addresses of VMs by index.

All these things can be solved using temporary files (as of now only one `test_env` instance per user is supported anyways) but after starting to refactor to shell scripts I want to stop and reflect a bit. A shell script based `test_env` would end up being more awkward for the NFV selftest to use. Maybe the concerns using `source` can be alleviated by running `bash -rcfile <test_env>`? Are we trying to do too many things at once here? If we want a "swiss army knife" for automatically and manually testing multiple NFV scenarios is Bash the right language? (remember there is a DPDK case yet to come, and on the ML we have been talkling about testing with multiple QEMU versions, this is getting more complex fast). I am comfortable with a 200 line shell script but once `test_env` gets as large as `bench_env` it will be a pain to develop.

> It is easy to imagine generalizing this to other use cases like @kbara's too, e.g. if you want to automatically start two QEMU guests with well-known vhost-user socket names, but then be able to start/restart the snabbswitch instance manually. But if the qemu startup is a stand-alone script then perhaps you just call that manually instead of using the framework to start the tmux.

The only thing missing to support @kbara  is that `test_env`'s `snabb` and `qemu` wrappers also accept NUMA nodes directly instead of only PCI addresses. I have a added a commit that does this to the tmux branch, now you can:

```
bash -rcfile program/snabbnfv/test_env/test_env.sh

qemu <numanode> <socket> <telnet_port>
qemu <numanode> <socket> <telnet_port>

snabb <numanode> <your-program-args>
```

lukego:
OK. Maybe I am pushing to do too much at the same time. Could be better to keep it simple and merge test_env as a basic replacement for the old NFV selftest. That is an important thing to have and it does already do this now.

Could be that manual testing should be done in a different way instead of trying to reuse parts of test_env e.g. for starting QEMU processes. This is something we could look at separately.

I definitely think it's more important to keep these scripts simple than to try and cover lots of new use cases from the start.

Which of the branches are you most happy with?

eugeneia:
I fast-forwarded the tmux branch into this one. I think is strikes a good balance between simplicity and interactivity. My next goal would be to add a DPDK based environment.

eugeneia:
@lukego I need to know though: What QEMU parameters should a DPDK powered VM be started with? What kind of snabb instance fits in between packetblaster and a DPDK VM specifically?

lukego:
Basically yes. The `same_vlan.ports` should work fine for example. The only caveat is that if loadgen is blasting garbage packets then you need to be careful about things like packet filters that could drop all the traffic.

lukego:
Merged onto `next`. Closing PR. Please create a new one for any follow-on changes.

PR #521, opened 2015-06-18
Remove scripts/bench_env

lukego:
This PR would remove `bench_env` from the repository.

This will happen if the test_env scripts in PR #517 are deemed to be a suitable replacement. If you are using bench_env and would miss it then please comment on #517 to make sure that we know about your use case.


lukego:
Merged onto `next`.

PR #524, opened 2015-06-23
intel10g: Fix RX DMA corruption in SF mode

lukego:
Fix a problem where packet payload could be corrupted on receive on certain hardware. There was an exotic register flag that needs to be cleared and this was taken care of in VF mode (when using VMDq) but not in SF mode (when using a raw NIC). Now both cases work the same.

Reported here:
https://groups.google.com/d/msg/snabb-devel/sP3wJ-8fEEA/O1WMAG96SlQJ

May also resolve a previously reported problem:
https://groups.google.com/forum/#!topic/snabb-devel/MrzImre1gbM

Obsoletes #523.


lukego:
Merged onto `next`. Closing PR.

PR #526, opened 2015-06-25
Packetblaster test and fix

lukego:
Fix a crash-on-startup bug in `packetblaster` and add a test case to avoid regressions.

Thanks Basil Ahmed for the bug report.


eugeneia:
Heh, I just ran into that myself.

lukego:
Merged onto `next`.

PR #527, opened 2015-06-26
include unistd.h correctly

justincormack:
Cherrypicking obviously correct commit from #483 - sys/unistd is not a correct header, in order to reduce the Musl specific diffs.


lukego:
Sorry to be boring but closing this one because I think we need a different portability strategy than POSIX compliance, for example rewriting all the C code with ljsyscall.

PR #528, opened 2015-06-26
include string.h for memcpy and friends

justincormack:
Cherrypicking obviously correct commit from #483 - this file uses memcpy so needs to include string.h

### DIFF
--- a/src/apps/vhost/vhost_user.c
+++ b/src/apps/vhost/vhost_user.c
@@ -7,6 +7,7 @@
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <unistd.h>
+#include <string.h>
 
 #include "lib/virtio/virtio_vring.h"
 #include "vhost.h"


lukego:
Thanks Justin! Would you please patiently help me to understand some portability issues from first principles? :-)

I'm thinking that we need two things:

1. Definition of which platforms we support (hypothesis).
2. CI test coverage of multiple platforms (attempt to disprove hypothesis).

So for example suppose we say that today the master branch supports "Compile on Linux/x86-64 with GNU toolchain, deploy on Linux with GLIBC >= 2.7". On the one hand that is quite narrow compared with all the interesting platforms in the world (POSIX, musl, NetBSD, Windows, etc). On the other hand it is quite broad compared with what our CI is testing (only Ubuntu 14.04 LTS w/ check for linkage to newer GLIBC symbols).

So we have multiple related problems to work on here:

1. Maintain a correct definition of the platforms that we support.
2. Shrink that definition when we are able to demonstrate non-working platforms.
3. Expand that definition when we are able to demonstrate more working platforms.

Does that make sense?

If we run with this mental framework then it seems like the portability-related changes that we want to land on the master branch are those that directly contribute to one of these areas. That is, we want to merge changes that improve test coverage, fix problems with already supported platforms, or add support for new platforms.

Coming slowly back on topic :-)

I think it makes most sense for portability code to be developed on a branch and then merged onto master when it reaches the point where we can say "This branch adds support for platform FOO, we know it works because CI TESTS, and the branch history shows that the maintenance workload is modest and people are willing to do it."

The risk I see of merging piecemeal changes onto master is that on the one hand a port only benefits users when it is complete, and on the other hand it immediately makes general maintenance harder because there will be a bunch of stuff that works fine and passes tests but is vaguely taboo e.g. using features of GNU make, bash, GNU awk, and so on.

Coming even closer to being on topic :-)
I really don't want to chase away people who are interested in other platforms. I just want to make sure that portability work is done in such a way that it is a net benefit to the community.
How about we do one of two things with patches like this:

1. Find a way to describe and test how it improves our platform support. For example, could these small fixes be bundled together as a way to switch to using stricter CFLAGS?
2. Maintain the patches on a branch (e.g. `posix`) if it will take some time to collect enough changes to really improve support for a platform, have tests to show that it works, and be confident that the community will maintain it.

Is this making any sense?

I hope that I am not chasing away people who are interested in portability. It is a very important topic. I just want to make sure we do it in a good way.

justincormack:
Arguably this change is not a portability change, it is a correctness change. The documentation for memcpy tells you to use `#include <string.h>`  but you decided to wing it without. It is unclear if it happens to work in glibc or you have a missing compiler warning, are you sure? Just fix it.

In terms of easy wins on portability, the easiest of all is to stop hard coding gcc in the Makefile and build with clang too.This gives you a different sett of warnings, always useful.

Next up is to build and test on more Linuxes. I have started using Docker for this, eg https://github.com/justincormack/docker-frankenlibc which builds at https://registry.hub.docker.com/u/justincormack/docker-frankenlibc/builds_history/234523/ - it is not quite a perfect test, as it does not run with the correct kernel for each system, but I have a reasonably comprehensive collection of virtual machines too.

Then porting to Musl will fix general Posix compatibility and glibc-specific oddness, plus giving you a smaller static binary.

Then there are two more complicated things, other operating systems eg FreeBSD, which is going to have different interfaces for some things, and other architectures, which may have different synchronisation issues.

But generally, you want to merge correctness fixes irrespective of a particular target, and general changes that make it easier to progress towards portability and testability, eg I am inclined to send a patch that replaces `gcc` in the Makefile with `$(CC)` in order to be able to test with gcc5 or clang, because now the Makefile is hostile to testing different compilers, I dont want to have to do that in the context of eg making Snabb work with gcc5, as it probably already does but its annoying to even find out now.

lukego:
To me it seems like we need somebody to take charge of portability by maintaining a branch that is more portable. You seem very well qualified to be that person, having a lot of relevant experience and strong opinions. Are you interested?

I am happy to merge portability code onto my branches when it adds support for some identifiable and relevant new platform (whole ports, not partial ones), when it is maintainable (has tests that will catch regressions), and when it is actively maintained (people seem to care about it enough to keep it working).

To me a new "platform" could be as simple as a different CFLAGS that would cause `gcc` to reject code without this fixes, provided there is a reasonable justification for why those CFLAGS are better.

(You know, what I *really* want to do is rewrite all of this C code in Lua with ljsyscall instead of spending our time on porting it. To me it is not yet clear whether substantial ports of the C code will be possible without making dealbreaker changes like introducing autoconf. I see a risk that the traditional ways of making Unix software portable are not going to be palatable for Snabb Switch.)

justincormack:
Yes I have to agree that the C code is a problem... and Luarizing it gets rid of most of the issues. The main necessary C code is gcc intrinsics for now (and they are simple with no deps).

(and ljsyscall just got ppc64 and arm64 support in the last week, which are the two interesting not amd64 architectures right now).

PR #530, opened 2015-06-29
packetblaster 64 byte packets benchmark

eugeneia:
Run like so:

```
cd snabbswitch
PCIADDR=0000:84:00.1 src/scripts/cperf/benchmarks/packetblaster-64
```

Output is in Mpps.


lukego:
@eugeneia Looks like SnabbBot is not running this packetblaster benchmark. I don't see the result in the log. Something missing? (Can fix in next release / bugfix release.)

eugeneia:
@lukego Its simply a matter of enabling it in SnabbBot, which I will do upon release. Otherwise SnabbBot would fail this bench on PRs targeting 2015.06.

PR #539, opened 2015-07-02
[vhost_user] Do not crash when report() is called while unconnected.

### LINKED ISSUE #535: VhostUser:report() error when not connected
The VhostUser app causes an error when its `report()` method is called before the socket has connected. This should not trigger an error.

See backtrace in this [blog entry](https://marcelwiget.wordpress.com/2015/07/01/simple-snabbnfv-traffic-test-between-2-vms-over-a-10g-loopback-cable/) by @mwiget.

eugeneia:
Fixes #535 by printing "Not connected." on report while unconnected.

### DIFF
--- a/src/apps/vhost/vhost_user.lua
+++ b/src/apps/vhost/vhost_user.lua
@@ -323,7 +323,8 @@ function VhostUser:reply (req)
 end
 
 function VhostUser:report()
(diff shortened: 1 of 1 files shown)


eugeneia:
This should go to next, ignore!

Issue #538, opened 2015-07-02
snabbnfv traffic default settings should not print so much debug info


lukego:
The `snabbnfv traffic` program should default to not printing the link report or the app report (only the load report). Printing these reports has nontrivial overhead and should only be used for debugging.

The real solution to get detailed visibility without overhead will be SnabbTop (#472).

lukego:
@eugeneia Could I tempt you to take a look at this one and the related #537 and #535?

eugeneia:
@lukego Yup, on it!

PR #541, opened 2015-07-02
App report fixes

### LINKED ISSUE #537: engine:report_apps() not handling errors
The `engine:report_apps()` function should use `pcall` to catch errors in apps that have faulty report methods. This lack of error handling means that bugs like #535 are treated as fatal. It would be better to catch the error and print a warning.

eugeneia:
Fixes #537 and #538.


eugeneia:
@lukego There actually was an error in this fix. Added a fix.

PR #543, opened 2015-07-03
fixes: Collected bug fixes for the next maintenance release

lukego:


lukego:
@eugeneia I'd like to release this as v2015.07.1. There are good fixes here that will reduce user annoyance. Should I just do that or would you like to?

eugeneia:
@lukego On it!

lukego:
@eugeneia Actually, I would be quite happy to do this one myself. OK with you? with both of us doing releases (e.g. you feature releases and me fix releases) then we can build on each others' tricks and workflows.

eugeneia:
Sure! :)

lukego:
(I imagine that this time next year the actual releases will be done by cronjobs and we will only be responsible for having the RC branch in shape when the time comes around. However, I don't know how we would come to that point, or if that is really how it will work out :-))

eugeneia:
Have a script that interacts with the GitHub API to publish releases? I imagine it would:

1. Get the status of the next branch
2. If OK, extract merges to build change log and push release using GitHub API
3. Maybe check a custom "release" status that indicates manual blocking of a release and/or points to a hand written change log.

E.g. I imagine the release bot could check for a `snabbswitch/VERSION` file that contains the release name. If it doesn't exist it sets a "no release" status.

lukego:
This time I generated the release notes automatically with Git commands. The hand-written version could still be better for larger feature releases though. I don't think we have an urgent need for more automation, it is just something in the back of my head.

lukego:
I missed merging #548 before the release. Oh well - next week.

PR #545, opened 2015-07-03
Ensure /var/run/snabb exists

n-nikolaev:
Before creating the PID subfolder, ensure the root folder exists.

### DIFF
--- a/src/lib/ipc/fs.lua
+++ b/src/lib/ipc/fs.lua
@@ -21,6 +21,7 @@ function fs:new (pid, root)
    local root = root or default_root
    local pid = pid or syscall.getpid()
    local o = { directory = fs_directory(root, pid) }
+   syscall.mkdir(root, "RWXU")
    syscall.mkdir(o.directory, "RWXU")
    return setmetatable(o, {__index = fs})
 end


lukego:
Thanks!

Is this fixing a bug that will annoy users? (What happens?) I can merge it onto the `fixes` branch for quicker release in that case.

n-nikolaev:
For me /var/run/snabb does not exsitby default so it has to be added explicitly, you can verify by deleting the folder and try run snabb.

lukego:
OK. Generally it would help me a lot for the commit message to give the "why" of bug fixes i.e. how the problem affects the user. That helps me work out how urgent it is to release the fix.

Currently I'm not sure whether this is a "snabb doesn't start due to file not found error" problem or a "directory is mysteriously missing when you look for it in `ls`" problem. I can test that myself this time :).

lukego:
@n-nikolaev Can you explain more context please? I am testing on the master branch and if I remove `/var/run/snabb` and then run `make test` it all passes and I do have a `/var/run/snabb` directory created afterwards. How are you testing?

n-nikolaev:
Ok, so this was probably a false alarm related to write permissions.
Sorry for that - closing it.

lukego:
No problem. Thanks for the follow up.

There could conceivably also be a code path somewhere that is not creating the directory properly since we have not ported everything onto `core.shm` yet, but if there is I don't immediately know what it would be.

lukego:
I have seen this problem now when running `packetblaster` after a reboot:

```
$ sudo ./snabb packetblaster replay ./program/packetblaster/64.pcap 0000:01:00.1
lib/ipc/shmem/shmem.lua:196: /var/run/snabb/3644/core-stats: No such file or directory
stack traceback:
        core/main.lua:118: in function <core/main.lua:116>
        [C]: in function 'assert'
        lib/ipc/shmem/shmem.lua:196: in function 'init'
        lib/ipc/shmem/shmem.lua:218: in function 'new'
        lib/ipc/shmem/top.lua:23: in function 'new'
```

Thanks for the fix. I will merge. I really took in the SnabbTop branch too early. This release we need to bring consistency to our counters and shared memory objects and so on.

Issue #482, opened 2015-05-14
Github workflow for "merge early merge often"


lukego:
I would like to experiment with adding more merge steps into our workflow. The intention of this idea is to speed up getting good changes onto master by using smaller faster steps.

The process today is:

- Send PR to master.
- Spend a random amount of time waiting for testing and feedback from the crowd.
- Luke merges onto master.

The idea is to add this simple way for more people to do merges:

- People volunteer to be testers and reviewers of certain loosely-defined kinds of changes.
- Pull Requests submitted to master are *assigned* (Github feature) to a tester.
- That tester takes the lead on giving/gathering feedback and then merges onto their own branch.
- The tester sends Pull Requests containing changes that they have merged and are satisfied with.

The basic idea is that this would be a fairly fluid and flexible way to have more than one person approve the changes being submitted. This is based on the Linux model where changes are accepted by a chain of subsystem gurus before landing on master and every step in the chain represents somebody's review and approval.

I would be especially pleased that this process would let me send my own changes to other people for review and acceptance instead of dumping them straight onto master.

Once more: The purpose of this idea is to make the changes land on master faster and in better shape. I would have an easier time pressing Merge if I know that I am not the only one who has reviewed a change and thinks it is right.

lukego:
Anybody want to volunteer for this? If so then what kind of changes are you interested in testing? (Check out the Pull Requests lists for inspiration on what kind of changes come down the pipeline.)

plajjan:
I think I'm confused by step 3 in your proposed process. Why would a tester merge a PR onto their own branch? If I review a PR and I think it looks good, you are saying I should merge it on my branch and then I should open a PR that to 100% contains code that someone else has written, ie my PR is identical to the original coders PR. Why? Seems easier to just comment on the original PR then :)

You can send your own changes as PR to the project, wait for someone to review and write "+1" on the PR then you push merge, or you let someone else have RW on the repo in addition to yourself, then you could assign your own PRs to be reviewed of them (this is how I typically work in small projects).

lukego:
> Why would a tester merge a PR onto their own branch?

This is to adopt the distributed model from the Linux kernel workflow.

The idea is that we treat our branches like IP routers and our commits like packets. Each branch (router) makes sure that the commit (packet) looks okay and then forwards it (pull request) to the next hop. The rules are all applied locally and independently of the size of the network. The chain of merges that a commit passes through on its way to master will read like a traceroute of the path that it took and who tested it.

[...]

PR #546, opened 2015-07-05
Add snabbnfv getting-started.md page

mwiget:
Here's my proposed page, explaining how to install, build and run two virtual machines in qemu over snabb switch using the snabbnfv traffic application. Please send me any suggestions for changes, particularly for useful links to add at the bottom of the page.


lukego:
Great! I will read through and give detailed feedback.

Let me braindump a little on how a documentation update like this will make it into a release, based on the new [Git workflow](https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/git-workflow.md) that we have recently started using. (I write this for the benefit of everybody involved in the process so that we will all be on the same page - important since we are still bootstrapping this model.)

From your perspective as the author of the change it should look pretty straightforward:

- Submit update in PR.
- Get feedback and make revisions.
- Change is merged upstream.

The wrinkle then is that "upstream" doesn't mean landing directly on the master branch but rather starting on the most appropriate [well-known branch](https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/branches.md) and then migrating onto master via a short series of hops. For a documentation change the path to master is likely to be `topic->documentation_fixes->next->master`.

In more detail:

- @eugeneia (Max) merges the change onto his `documentation_fixes` branch where he is integrating all changes to the documentation. He might make some edits himself on this branch e.g. for the global consistency of our documentation.
- @eugeneia sends a Pull Request to merge all changes on `documentation_fixes` upstream.
- @lukego (me) merges these documentation fixes onto the `next` branch to queue it for the next release.
- @eugeneia merges the `next` branch onto `master` at the start of a calendar month to make a release.

This sounds complicated, and it is to some extent, but the value is that it will make sure a short chain of people (@mwiget, @eugeneia, @lukego) are all happy with the change and that it hits the master branch as part of a coordinated release. Meanwhile everybody in the community is able to give feedback on the changes at every step in the process. Github makes it pretty easy to ask specific people for their feedback by mentioning their name (e.g. @lukego) too.

mwiget:
This sounds indeed complicated. Have you looked at https://code.google.com/p/gerrit/ ? We use this for opencontrail and allows people to reject or accept, after CI gave the green light. Once the predefined gatekeepers all accepted, the changes merge automatically into master.

lukego:
We are consciously adopting the Linux kernel workflow instead of the Gerrit one.

Fortunately the short version "Submit PR to master, respond to feedback, relax once it is merge onto any upstream branch" is all that's needed for contributing improvements.

I spelled out the long version mostly for the sake of getting me and Max onto the same page since this will be the first time a documentation addition is handled under this workflow. Maybe this was gratuitously confusing to do on the PR containing the change :-). This tends to happen since the project is young and a bunch of stuff needs to be worked out as we go along.

The backstory on the workflow can be seen in e.g. #482. This was definitely a situation where many people have different strong preferences and it's not possible to make everybody happy :-).

mwiget:
sounds good to me ;-).  Thanks for the detailed feedback. Looking forward to any suggestions to improve and correct my write-up.

src/program/snabbnfv/doc/getting-started.md
```diff
@@ -0,0 +1,610 @@
...
+load: time: 1.00s  fps: 3         fpGbps: 0.000 fpb: 0   bpp: 98   sleep: 100 us
+load: time: 1.00s  fps: 4         fpGbps: 0.000 fpb: 0   bpp: 98   sleep: 100 us
+link report:
+                 120 sent on id2_NIC.tx -> id2_Virtio.rx (loss rate: 0%)
+                 120 sent on id2_Virtio.tx -> id2_NIC.rx (loss rate: 0%)
+load: time: 1.00s  fps: 3         fpGbps: 0.000 fpb: 0   bpp: 98   sleep: 100 us
+load: time: 1.00s  fps: 3         fpGbps: 0.000 fpb: 0   bpp: 98   sleep: 100 us
+```
+
+The difference in packet counters is a result of me stopping and starting one of the snabbnfv processes mid-flight. According to the documentation thats ok and it does indeed work just fine.
```
lukego: Little background: Yes, you can restart the Snabb Switch process and it will reconnect to the guests and continue to provide their networking. This is currently the one-and-only change on the `v2.1.0-vhostuser` branch that is not upstream in QEMU. (The other changes on that branch have landed upstream in QEMU after the 2.1 release.)

This guide should also work with the latest upstream QEMU with the exception that restarting the Snabb Switch and continuing to do networking will only work if you apply commit [f393aea](https://github.com/SnabbCo/qemu/commit/f393aea2301734647fdf470724433f44702e3fb9). That change requires some thinking about how to do "right" for QEMU, the trouble is that there is no analogous situation where you reboot the kernel underneath a running QEMU process. The other two changes on our branch are already upstream in QEMU.

lukego:
Your blog entry also prompted me to consider another issue: latency. Your ping times in the screenshot are often >1ms. Snabb Switch generally has consistent low latency (~100us) except that under low traffic we make `usleep()` calls to reduce CPU utilization, but the kernel does not seem to honor the requested sleeping time particularly well. You might ask for a 1us sleep and actually get 2,000us. This becomes visible in `ping`.

So the out-of-the-box setup of snabbnfv is not really ideal for latency testing. Could be that we should add a common-line argument to busy-wait instead of sleeping, which does provide consistent low latency in my tests. Even better would be a way to make the kernel do the right thing with usleep if that is possible e.g. with `nice` process priorities but I am not sure how much we dare to depend on that.

lukego:
This guide looks really excellent to me, Marcel. Thank you for writing it.

The main change I would like to make is to not have to tell people to disable the IOMMU or tune their NUMA. However, this is not a problem with the guide, but rather a problem with the software that we don't yet handle these things automatically. It is great to have an explicit guide that can inspire us to make the software simpler to deploy. (Otherwise it's easy to forget all these little steps that will be a pain in the backside to new users.)

src/program/snabbnfv/doc/getting-started.md
```diff
@@ -0,0 +1,610 @@
...
+  },
+}
+$ cat port2.cfg
+return {
+  { vlan = 431,
+    mac_address = "52:54:00:00:00:02",
+    port_id = "id2",
+  },
+}
+```
```
eugeneia: Might be worthwhile to point out that every Snabb NFV configuration can define multiple ports so that `n` VMs can connect to one 10G port.

jeffbrl:
@mwiget Well done!

Can Snabb NFV work with raw sockets to obviate the hardware dependency?

eugeneia:
@jeffbrl No it can't as of now. It would be certainly possible to make it do so though. Keep in mind that Snabb NFV uses the built-in switching of its underlying NIC, so you would need to plug a `bridge` app between `RawSocket` and the ports. You might want to checkout our ongoing brainstorm on `labswitch`, [here](https://groups.google.com/d/msg/snabb-devel/gdMF5Vmd-YI/L0MoOdx0ol8J) is a very generic program that lets  you define switched ports in an arbitrary ways.

jeffbrl:
@eugeneia Thanks for the pointer to your labswitch hack. I'll have to examine it further.

lukego:
Cool thought. The labswitch really does have potential here as a "sponge" to soak up reusable functionality (i.e. apps) built as part of more special-purpose applications (e.g. alex and nfv).

lukego:
@eugeneia as editor-in-chief please merge this one onto one of your branches, make any edits you need, and then PR again to master for me to merge. (I don't want to merge documentation update PRs myself because then you would not have an opportunity to edit for global consistency etc.)

authorized_keys
```diff
@@ -85,3 +85,5 @@ ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDAzg8e0cW4io7PVp+ID/f8wHldye9eMXv3wwp1TRpo
...
 
 ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDmUqhU3Eg6zPJ8MEKpCcBkYRHXd/FVBxmy0DOj3GD5Hdxd/+cBDk/mfIJOy+r4HG2qQJBls9EvvZzcSnF3mqqURfWUDfKJ+fNZ+0vdeTodmgHiOERO5pUyf4h1YxqbxX+cVfELxpPwkGjtWIsvdt+CKuQ1b1rauKgv33YVkm1VJHCBJk8mh/JN5jsplnyu0nR5sPOJQ8rnDOr3vRGa0nGc+G7S03bAUBTmgjbEsRUOG2TJ1TPmkXoCVlR1I4yWUmgSMiKAO8nOFEwQaOQbLhLsvcYF7T4KzF86BUIG6hkm3qfHLPCmdbOfXfFnp8oNZiBq/S0nrHMLT9BGXjfXdbGv sm101
+
+ssh-dss AAAAB3NzaC1kc3MAAACBAMdlCk1NNT2O+np4uzFWFDHP+zTS8uAC6c0mv2miSgAgJxFyfZpJH+HbOuLALCoyPrQbAPb+yPeXvl7xQwAUd94QW3dsX8B70skaxGQMXJdvEu3iDSnpxdeNMW+Ctl4JDHwNoZ93dCxqUqiF5tIE9ock8r1vEZ4d4Xy/LWe+mneVAAAAFQCZ3YEG7uDAfKRxcIK7v4XJyCknCwAAAIA4l8xAexLrEiheg8w8YYGvTtTV20xDaFObLI0fWFpYM0n6g80xkGoM409/1ne6PPqOydCp6dfNcbqf2vCq2WxffjEetMSE5BNk02JctdafO8wiGVFnQd39I+n70SCU/48s/NX+RqWcRgTlwDzp034ZiclDrmrBGVmz5TAJWXT8BgAAAIBbYv/+kxyNdM0HLiQn6/ShTCqK6gkhumDn3a/SS0nHx3LpdlACX9x49a7VTf4tYqctW6LUkE9ei0cvsHWq2ec6Q00UAypCaTtwUjt7vr7HmwuTKV6XOsLkupnEED5jtRgeEz5fuWPIMH6Xg/GENJ5z7N/6AlaOz3Emu6TQtkdwPw== mwiget@mwiget-mba13
```
eugeneia: I am not going to include this in my merge for reasons of hygiene, please add your public key in a different PR.

lukego: How are you going to exclude it? I worry that if we `rebase` PRs then we risk creating more problems than we solve...

lukego: (I wonder if this is a matter of policy that we need to agree on or whether every subsystem branch can have its own policy about rebases... have to ponder that.)

lukego: Oh I suppose you don't need a rebase: in this case you can simply `merge` up to the commits that you want, since the ones that you don't want come later.

(Sorry for rapid-fire comments... :-))

eugeneia: Yeah basically I checked out the PR as a branch, chopped off the last commit and am now appending my edits.

mwiget: No worries. I wanted to get a separate fork for the public key, but git automatically used the fork I had already open ;-)

eugeneia:
Included in #577

PR #558, opened 2015-07-10
Add new programs 'top' and 'gc'

eugeneia:
This is a clean-slate rebase of the #472 PR. It moves links and statistics counters into the `core.shm` framework and introduces the programs `top` to monitor live statistics and `gc` to clean up Snabb runtime files.

Summary of changes:

b2cc0fa

* Extract `struct counter` to its own header file so it can be easily embedded in other structs.

a3ffff3 9977f24

* Some minor extensions to `core.counter` and `core.shm`.

490eae6 874e48b 1c9c26a b351682 6d66741 099d0cf

* `core.app` implements its statistics counters using `core.counter` and maps them to `/engine/...`. It also has a new `configs` counter that is incremented for every call to `configure`.

* The counters in`struct link` are now pointers to shm counters. `link.new` now accepts a mandatory `name` argument. When a link is created, it is mapped to `/llinks/<name>` and its counters are created and mapped to `/counters/<name>/<counter>`.

* The new function `link.free` unmaps and unlinks the link and its counters.

* The function `link.stats` now returns a fresh table containing the names and values of the link's counters. The counter values are converted to Lua numbers using `tonumber`. All code (mostly tests) that was accessing the `stats` field of links directly now uses `link.stat`. This has the advantage of not having to worry about type conversions, or how to access the counter. A common pitfall of mine was wrong math due to unconverted ulongs, e.g. `100ULL * 0.45 = 0`.

b38d609

* Added `gc` program.

8d0c256

* Added `top` program.


src/core/counter.h
```diff
@@ -0,0 +1 @@
+struct counter { uint64_t c; };
```
javierguerragiraldez: Is this for the benefit of C files?  In Lua code, if you `require('core.counter')` you have this, and can use `struct counter` in any `ffi.typeof()` or `ffi.cdef[[]]`, no matter if this struct is defined on its own file or inline inside `core/counter.lua`.

src/core/link.lua
```diff
@@ -2,11 +2,16 @@ module(...,package.seeall)
...
 
+local shm = require("core.shm")
 local ffi = require("ffi")
 local C = ffi.C
 
 local packet = require("core.packet")
 require("core.packet_h")
+
+local counter = require("core.counter")
+require("core.counter_h")
```
javierguerragiraldez: this is redundant, since `core.counter` already defines the `struct counter`.  In fact, the `require()` function won't do anything because `package.loaded.counter_h` is already non-nil.

src/core/link.lua
```diff
@@ -14,8 +19,23 @@ local band = require("bit").band
 local size = C.LINK_RING_SIZE         -- NB: Huge slow-down if this is not local
 max        = C.LINK_MAX_PACKETS
 
-function new (receiving_app)
-   return ffi.new("struct link", {receiving_app = receiving_app})
+function new (name)
+   local r = shm.map("links/"..name, "struct link")
+   for _, c
+   in ipairs({"rxpackets", "txpackets", "rxbytes", "txbytes", "txdrop"}) do
```
javierguerragiraldez: any reason for this two-line `for` style?  i haven't seen it anywhere before.

src/core/counter.h
eugeneia: `link.h` uses `struct counter`. Is that what you mean by "for the benefit of C files"?

src/core/link.lua
eugeneia: I copied the pattern from the `core.packet_h` require just above. Indeed the requires of both `packet_h` and `counter_h` are redundant. I suggest we fix this in a separate "style branch", since we have a **lot** of unnecessary `require` calls (especially to core packages) in most modules.

src/core/counter.h
javierguerragiraldez: don't think so... in `link.lua` the `require('core.link_h')` is after the `require('core.counter')`, so it should work without pulling this specifically.
In fact, unless a struct or const is used in a C function, I see no value in using `.h` files.  Maybe somebody dislike the `ffi.cdef[[...]]` embedded definitions?
The only C code I've written since we pulled ljsyscall was for the pthreads experiment, but that was very short-lived.  The current `fork()`based doesn't need a single `.c` file.

src/core/link.lua
javierguerragiraldez: Agreed, this is fit for a style branch.

src/core/link.lua
eugeneia: That's just so the lines aren't too long. Too distracting?

src/core/counter.h
eugeneia: I have no personal preference, I wasn't sure about the scope of `cdef`'s so this was my cautious approach to just mimic what I knew worked. I could imagine a C program wanting to read snabb counters so that would be an argument for having the header file. Since this also applies to the other header files in core as well I say lets move this into a seperate general discussion on whether we want C header files and in wjich cases.

eugeneia:
So we have a heavy performance regression on the `basic1` benchmark here that warrants some investigation. I am not surprised that there some penalty to using `counter.add` instead of `x = x +1`.

See this graph (last commit is this PR's HEAD):

benchmarks

eugeneia:
I am trying to get certain about what causes the slowdown here, I used `require("jit.p").start('5vlm1')` to get some measurements:

On the `snabbtop2` branch the profiler reports 100% or 99% compiled and in the latter case 1% GC or 1% JIT compiler (differs from run to run). Notable is that the top quarter is held by calls to `core.counter`.

On the `master` branch the profiler always reports 100% compiled. The top quarter varies fro run to run but consists mostly of transmitting/freeing packets.

Its a bit confusing that the reports differ quite a bit between runs, and I can not really tell were the C is coming from, but I think its safe to say that `core.counter` is the bottleneck in the `snabbtop2` branch.

**snabbtop2**

```
Processed 100.0 million packets in 10.20 seconds (rate: 9.8 Mpps).
99%  Compiled
  --  8%  counter.lua:34 < link.lua:64 < basic_apps.lua:110 < app.lua:72 < app.lua:294
  --  7%  packet.lua:47 < basic_apps.lua:27 < app.lua:72 < app.lua:278 < app.lua:233
  --  6%  counter.lua:34 < link.lua:63 < basic_apps.lua:110 < app.lua:72 < app.lua:294
  --  5%  packet.lua:47 < basic_apps.lua:110 < app.lua:72 < app.lua:294 < app.lua:233
  --  5%  counter.lua:34 < link.lua:47 < basic_apps.lua:83 < app.lua:72 < app.lua:294
  --  5%  counter.lua:34 < packet.lua:87 < basic_apps.lua:84 < app.lua:72 < app.lua:294
  --  5%  counter.lua:34 < link.lua:64 < basic_apps.lua:27 < app.lua:72 < app.lua:278
  --  4%  counter.lua:34 < packet.lua:90 < basic_apps.lua:84 < app.lua:72 < app.lua:294
  --  4%  counter.lua:34 < link.lua:46 < basic_apps.lua:83 < app.lua:72 < app.lua:294
  --  4%  packet.lua:90 < basic_apps.lua:84 < app.lua:72 < app.lua:294 < app.lua:233
  --  3%  counter.lua:34 < link.lua:46 < basic_apps.lua:106 < app.lua:72 < app.lua:294
  --  3%  link.lua:46 < basic_apps.lua:83 < app.lua:72 < app.lua:294 < app.lua:233
  --  3%  counter.lua:34 < link.lua:47 < basic_apps.lua:106 < app.lua:72 < app.lua:294
  --  2%  link.lua:76 < link.lua:57 < basic_apps.lua:110 < app.lua:72 < app.lua:294
  --  2%  link.lua:63 < basic_apps.lua:110 < app.lua:72 < app.lua:294 < app.lua:233
  --  2%  counter.lua:34 < packet.lua:86 < basic_apps.lua:84 < app.lua:72 < app.lua:294
  [...]
 1%  Garbage Collector
  -- 100%  app.lua:289 < app.lua:233 < snabbmark.lua:48 < snabbmark.lua:17 < main.lua:56
```

**master**

```
Processed 100.1 million packets in 8.58 seconds (rate: 11.7 Mpps).
100%  Compiled
  -- 14%  packet.lua:90 < basic_apps.lua:84 < app.lua:76 < app.lua:297 < app.lua:234
  -- 11%  link.lua:28 < basic_apps.lua:106 < app.lua:76 < app.lua:297 < app.lua:234
  --  9%  link.lua:28 < basic_apps.lua:83 < app.lua:76 < app.lua:297 < app.lua:234
  --  8%  packet.lua:46 < basic_apps.lua:27 < app.lua:76 < app.lua:281 < app.lua:234
  --  5%  packet.lua:46 < basic_apps.lua:110 < app.lua:76 < app.lua:297 < app.lua:234
  --  4%  link.lua:56 < link.lua:37 < basic_apps.lua:110 < app.lua:76 < app.lua:297
  --  3%  link.lua:45 < basic_apps.lua:110 < app.lua:76 < app.lua:297 < app.lua:234
  --  3%  freelist.lua:16 < packet.lua:81 < packet.lua:90 < basic_apps.lua:84 < app.lua:76
  --  2%  link.lua:44 < basic_apps.lua:27 < app.lua:76 < app.lua:281 < app.lua:234
  --  2%  packet.lua:86 < basic_apps.lua:84 < app.lua:76 < app.lua:297 < app.lua:234
  --  2%  packet.lua:47 < basic_apps.lua:27 < app.lua:76 < app.lua:281 < app.lua:234
  --  2%  freelist.lua:24 < packet.lua:44 < basic_apps.lua:27 < app.lua:76 < app.lua:281
  --  2%  link.lua:44 < basic_apps.lua:110 < app.lua:76 < app.lua:297 < app.lua:234
  [...]
```

lukego:
I had a quick look at this today. I haven't found the problem yet. I can reproduce though and I am suitably impressed that our performance regression tests are doing their job :).

javierguerragiraldez:
well, calling `counter.add()` is known to be significantly slower than either doing the 'local function cache' dance: `local c_add=counter.add` at the top, or adding FFI metatype methods and just call `engine.frees:add()`, `engine.freebytes:add(p.length)`.

lukego:
@javierguerragiraldez That was my first thought too i.e. that the slow down comes from function call overhead in an inner loop. However, I am not so sure. Caching the counter functions locally doesn't seem to improve performance. Also, digging into the LuaJIT `-jdump` trace it looks like the machine code is quite streamlined in the code that is increasing link counters:

```
0bcae77a  mov r13, [r12+0x810]
0bcae782  mov r14, [r13+0x0]
0bcae786  add r14, +0x01
0bcae78a  mov [r13+0x0], r14   ; bump link packets counter
0bcae78e  mov r14, [r12+0x800]
0bcae796  add r15, [r14]
0bcae799  mov [r14], r15       ; bump link bytes counter
```

... but I need to come back and check this more thoroughly and it can be that I am misinterpreting or not looking at the most relevant trace.

Currently I am using the [LuaJIT profiler patch](http://www.freelists.org/post/luajit/Profile-at-trace-granularity,1) from Mike show which traces use the most CPU and then generating the JIT dump with:

    ./snabb snsh -jdump=+rs,dump.txt -jp=v \
                 -e 'require("program.snabbmark.snabbmark").run({"basic1", "10e6"})'

Probably we should improve the `snsh` command-line to make it easier to run arbitrary Snabb programs with customized environments e.g. profiler enabled.

lukego:
@eugeneia Looks like we also need to create `/var/run/snabb` with more liberal permissions:

```
$ ./snabb snsh --help
core/shm.lua:86: shm open error (17174/engine/breaths):Permission denied
$ ls -ld /var/run/snabb
drwx------ 169 root root 3380 Jul 17 05:29 /var/run/snabb
```

lukego:
Slight digression, forgive me. It's kind of fascinating to look at microbenchmarks with modern technology like LuaJIT and x86. (Leaving aside for this moment the practical relevance of such benchmarks.)

Here is a tiny program that bumps a counter in a billion-iteration loop:

```
local c = counter.open("test")
for i = 1, 1e9 do
   counter.add(c, 1)
end
```

and here are the 5 instructions that LuaJIT generates for the inner loop that I slightly annotated:

```
->LOOP:
0bcaf930  add rbp, +0x01  ; bump counter value
0bcaf934  mov [rcx], rbp  ; store counter to memory
0bcaf937  add ebx, +0x01  ; bump loop counter
0bcaf93a  cmp ebx, eax    ; check for loop exit condition
0bcaf93c  jle 0x0bcaf930        ->LOOP
```

and amazingly here is a Haswell CPU being able to execute all five instructions in one cycle:

```
 Performance counter stats for './snabb snsh -jdump=+rs script.lua':
     1,018,356,818      cycles                    #    2.211 GHz                    
     5,021,774,051      instructions              #    4.93  insns per cycle        
```

This really makes me want to spend more time with my new copy of [CSAPP3](http://www.amazon.com/Computer-Systems-Programmers-Perspective-Edition/dp/013409266X) :-)

eugeneia:
> @eugeneia Looks like we also need to create `/var/run/snabb` with more liberal permissions:

Yeah. Make it sticky? Like `/tmp` so everyone can create PID sub directories but only the owner and superuser can delete?

lukego:
That sounds reasonable to me (without wearing any kind of unix guru hat).

lukego:
This is really interesting. I have been digging a bit more and for the moment it looks like CPU voodoo rather than LuaJIT voodoo.

The performance drop seems to be directly linked to making `struct link` contain pointers to counters instead of counter values directly. If you remove the `*` characters from this line of `link.h` then the performance is high again:

    struct counter *txbytes, *rxbytes, *txpackets, *rxpackets, *txdrop;

So for some reason it seems that the performance penalty comes from the indirection of `struct link` having pointers to counters (shm objects) instead of having them inline.

I had imagined this would be annoying LuaJIT somehow but that does not immediately seem to be the case. I ran `snabbmark basic1` and compared the machine code for the trace that the profiler says is most used. The immediate difference I see besides a few different register names is just an extra pointer indirection. I have the [full pretty diff](https://www.diffchecker.com/lnklofur) (including some noisy register name differences) and here is an example of what I am interpreting as bumping a counter (left is with indirection removed, right is  code on this branch):

      $ diff -W 80 -y fast slow
      ...
      mov [rsi+0x828], r14d               |   mov [rdx+0x828], r14d
      mov r14, [rsi+0x818]                |   mov r13, [rdx+0x818]
                                          >   mov r14, [r13+0x0]
      add r14, +0x01                          add r14, +0x01
      mov [rsi+0x818], r14                |   mov [r13+0x0], r14
      ...

So now I am tempted to formulate a hypothesis for why this could annoy the CPU and then look for a way to disprove that e.g. relevant `perf` performance counters. I should also consider the possibility that I am looking in the wrong trace but I don't think so -- there were only three showing up in the profile and this one at the top.

Fun. :-)

src/core/link.h
```diff
@@ -5,9 +5,9 @@ enum { LINK_RING_SIZE    = 256,
 struct link {
   // this is a circular ring buffer, as described at:
   //   http://en.wikipedia.org/wiki/Circular_buffer
-  struct packet* packets[LINK_RING_SIZE];
+  struct packet *packets[LINK_RING_SIZE];
   struct {
-    double txbytes, rxbytes, txpackets, rxpackets, txdrop;
+    struct counter *txbytes, *rxbytes, *txpackets, *rxpackets, *txdrop;
```
lukego: This change highlights a subtle issue with `shm` shared memory objects: you have to be very careful about storing pointers in them.

Generally speaking pointers are only valid in one process because each process has a private virtual address space. This looks like a bug on the `struct counter *` pointers: they will only be valid for the process that initializes the link and if another process uses those pointers it will lead to memory corruption or a SEGFAULT.

The exception is pointers that are valid in all processes. The `struct packet *` pointers are uniquely okay because we do actually ensure that the same packet has the same address in every process.

So a new solution has to be found for the counters: we can have them inline in the struct, or we can remove them entirely, but we can't have them as pointers (even to other shm objects because `shm.map()` does not return consistent addresses for a given object).

Aside: I am also having a really hard time tracking down a bug that is caused by setting these `struct counter *` pointers to counters allocated with `ffi.new()` instead of `shm.map()` but I think that this is unrelated.

lukego:
I have been playing around with the performance regression. The basic approach for now is to use [ocperf](https://github.com/andikleen/pmu-tools) to look for interestingly different CPU performance counters on the slow code vs the fast one. The next step is to try in-between changes, but for the moment I am struggling with that. So here is a preliminary braindump.

I didn't find anything interesting in the standard `perf stat` counters but maybe some ideas in the extended `ocperf` ones.

Here is the profile for the slowed down code on this branch:

```
luke@interlaken:~/git/snabbswitch/src$ make && sudo ~/git/pmu-tools/ocperf.py stat -e cycles,instructions,cycle_activity.stalls_l1d_pending,cycle_activity.stalls_l2_pending,cycle_activity.stalls_ldm_pending,resource_stalls.any,resource_stalls.rob,resource_stalls.rs,resource_stalls.sb,icache.ifetch_stall ./snabb snabbmark basic1 1e9
make: 'snabb' is up to date.
perf stat -e cycles,instructions,cpu/event=0xa3,umask=0xc,cmask=12,name=cycle_activity_stalls_l1d_pending/,cpu/event=0xa3,umask=0x5,cmask=5,name=cycle_activity_stalls_l2_pending/,cpu/event=0xa3,umask=0x6,cmask=6,name=cycle_activity_stalls_ldm_pending/,cpu/event=0xa2,umask=0x1,name=resource_stalls_any/,cpu/event=0xa2,umask=0x10,name=resource_stalls_rob/,cpu/event=0xa2,umask=0x4,name=resource_stalls_rs/,cpu/event=0xa2,umask=0x8,name=resource_stalls_sb/,cpu/event=0x80,umask=0x4,name=icache_ifetch_stall/ ./snabb snabbmark basic1 1e9

Processed 1000.2 million packets in 57.44 seconds (rate: 17.4 Mpps).

 Performance counter stats for './snabb snabbmark basic1 1e9':

   137,610,684,367      cycles                    [40.00%]
   265,654,638,390      instructions              #    1.93  insns per cycle         [50.01%]
                 0      cycle_activity_stalls_l1d_pending                            [50.01%]
       577,030,750      cycle_activity_stalls_l2_pending                             [50.01%]
    15,187,350,549      cycle_activity_stalls_ldm_pending                            [50.01%]
    65,936,797,294      resource_stalls_any                                          [50.01%]
        22,569,426      resource_stalls_rob                                          [40.00%]
    39,972,536,873      resource_stalls_rs                                           [39.99%]
    26,848,268,009      resource_stalls_sb                                           [39.99%]
       152,017,985      icache_ifetch_stall                                          [40.00%]

      57.457470008 seconds time elapsed
```

and here is when I remove the `*` characters in the counter declarations of `struct link`:

```
Processed 1000.1 million packets in 49.22 seconds (rate: 20.3 Mpps).

 Performance counter stats for './snabb snabbmark basic1 1e9':

   117,936,355,494      cycles                    [40.00%]
   252,311,482,143      instructions              #    2.14  insns per cycle         [50.00%]
                 0      cycle_activity_stalls_l1d_pending                            [50.00%]
       902,218,405      cycle_activity_stalls_l2_pending                             [50.00%]
    11,441,697,088      cycle_activity_stalls_ldm_pending                            [50.00%]
    50,650,953,527      resource_stalls_any                                          [50.00%]
        29,498,053      resource_stalls_rob                                          [40.01%]
    21,189,725,460      resource_stalls_rs                                           [40.01%]
    30,318,684,145      resource_stalls_sb                                           [40.00%]
       177,455,398      icache_ifetch_stall                                          [40.00%]

      49.241633824 seconds time elapsed
```

The test is with one billion packets so it is easy to convert these big numbers into per-packet values. The fast one takes 117 cycles per packet (+17% for slow), 252 instructions (+5% for slow), loses 50 cycles to CPU resource stalls (+30% for slow) and 12 cycles stalled on pending L2 or memory accesses (+27% for slow).

I am not really sure how to interpret these counters. The Haswell CPU has many execution units: are the stalls when one execution unit can't run or when none can run? etc. Some background digging is needed there, and for now I don't claim my interpretation is any more meaningful than reading tea leaves.

The main theory I have is that the performance penalty comes from the address patterns of the counter objects in either one of two ways:

1. Counters allocated as `shm` objects will be on different cache lines but inline counters will be on the same. This potentially allows the CPU to coalesce memory operations better for the inline counters.
2. Counters allocated as `shm` objects may have page-aligned addresses i.e. starting on a 4096-byte boundary. (This is due to the way we allocate each counter with a separate mmap()ed file.) This can cause cache collisions because the cache is looking only at the low bits of the address and assigning them to the same bucket ([example](http://www.lshift.net/blog/2013/10/08/cpu-cache-collisions-in-the-context-of-performance/)).

and the next experiment that I want to compare is with counters still on separate cache lines but with more variation in the low bits of the addresses.

lukego:
Looks like it is problem 2: Counter are colliding in L1/L2 cache because they are allocated at 4096-byte aligned addresses.

Tested by adding some random padding at the start of the counters to unalign their addresses:

```diff
diff --git a/src/core/shm.lua b/src/core/shm.lua
index 3daeabc..650a455 100644
--- a/src/core/shm.lua
+++ b/src/core/shm.lua
@@ -85,11 +85,12 @@ function map (name, type,  readonly)
    local fd, err = S.open(root..'/'..path, "creat, rdwr", "rwxu")
    if not fd then error("shm open error ("..path.."):"..tostring(err)) end
    assert(fd:ftruncate(size), "shm: ftruncate failed")
-   local mem, err = S.mmap(nil, size, mapmode, "shared", fd, 0)
+   local r = math.random(100)*16
+   local mem, err = S.mmap(nil, r+size, mapmode, "shared", fd, 0)
    fd:close()
    if mem == nil then error("mmap failed: " .. tostring(err)) end
    mappings[pointer_to_number(mem)] = size
-   return ffi.cast(ffi.typeof("$&", ctype), mem)
+   return ffi.cast(ffi.typeof("$&", ctype), ffi.cast("char*",mem)+r)
 end

 function resolve (name)
```

The result is that performance comes back to normal:

```
Processed 1000.1 million packets in 48.88 seconds (rate: 20.5 Mpps).

 Performance counter stats for './snabb snabbmark basic1 1e9':

   117,094,002,854      cycles                    [40.00%]
   265,802,027,480      instructions              #    2.27  insns per cycle         [50.01%]
                 0      cycle_activity_stalls_l1d_pending                            [50.01%]
       544,000,698      cycle_activity_stalls_l2_pending                             [50.02%]
    10,019,546,691      cycle_activity_stalls_ldm_pending                            [50.02%]
    45,995,283,277      resource_stalls_any                                          [50.01%]
        19,196,410      resource_stalls_rob                                          [40.00%]
    17,526,932,629      resource_stalls_rs                                           [39.99%]
    28,985,851,873      resource_stalls_sb                                           [39.98%]
       130,095,179      icache_ifetch_stall                                          [40.00%]

      48.900446887 seconds time elapsed
```

This is an interesting class of problem since you could read the machine code for the inner loop forever and never see the problem: it's not in the code but rather in the data. The only sensible way to find such problems that I can think of is via performance counters, which I have used quite inexpertly above. (The reason I found this relatively quickly now is that @n-nikolaev and I were already bitten by the same issue last year when packet buffers were being allocated on 4096-byte boundaries and touching payload became very expensive.)

How to fix this for real? That is tricky. My hack above has the side-effect of adding padding bytes to the counter files on disk, which mucks up their nice representation as one counter per file. I don't immediately see a better solution that is fast and preserves the same file format but let us think about that for a bit.

javierguerragiraldez:
What about using a single shm object with all the counters in it?

lukego:
@javierguerragiraldez I like having a separate file for each counter and shm object. This way we outsource work to the file system instead of having to invent our own directory lookup mechanism. (The initial implementation of counters in Snabb Switch did keep them in a common shared memory blob with a separate index.) I don't want to give up on that right away if we don't have to.

I am tempted to try a simple double-buffer. Mirror the counters into ordinary FFI objects and then commit them to the shared memory locations once per breath cycle. This could potentially also potentially let us support atomic reads of shm objects if we need to in the future.

I also think we could probably drop most of the counters from the link objects and instead derive them once per breath from the deltas of the ring indexes. However, the issue will remain for all the other counters.

Question is also whether this cache-associativity performance problem will apply to other shm objects such as links or whether that will be okay because we don't operate on so many of them at the same time.

eugeneia:
Superseded by #568

PR #559, opened 2015-07-13
Update authorized_keys

mwiget:
Add my public key


lukego:
Got it! Your key is installed on `chur` now and you should be able to login and create an account with `ssh -p 2020 root@lab1.snabb.co`.

Issue #560, opened 2015-07-14
lib/virtio/net_device.lua:404: attempt to index a nil value


mwiget:
While trying to run Juniper's virtual router vMX on top of snabb snabbnfv traffic. The lab setup is described in detail in my blog post at [https://marcelwiget.wordpress.com/2015/07/12/juniper-networks-vmx-on-snabb-nfv/](https://marcelwiget.wordpress.com/2015/07/12/juniper-networks-vmx-on-snabb-nfv/).

I added a print statement to show the value of idx, and it shows an illegal value 407011329.

So somehow things get corrupt after successfully passing some traffic for a minute or so.

```
load: time: 1.00s  fps: 41,726    fpGbps: 0.046 fpb: 4   bpp: 127  sleep: 0   us
load: time: 1.00s  fps: 52,998    fpGbps: 0.047 fpb: 5   bpp: 100  sleep: 0   us
load: time: 1.00s  fps: 44,970    fpGbps: 0.048 fpb: 5   bpp: 124  sleep: 44  us
load: time: 1.00s  fps: 26,576    fpGbps: 0.034 fpb: 3   bpp: 150  sleep: 3   us
load: time: 1.00s  fps: 2,851     fpGbps: 0.002 fpb: 0   bpp: 89   sleep: 4   us
load: time: 1.00s  fps: 30,449    fpGbps: 0.028 fpb: 3   bpp: 105  sleep: 12  us
link report:
             246,389 sent on int1_NIC.tx -> int1_Virtio.rx (loss rate: 0%)
           1,867,917 sent on int1_Virtio.tx -> int1_NIC.rx (loss rate: 0%)
           1,867,915 sent on int2_NIC.tx -> int2_Virtio.rx (loss rate: 0%)
             246,389 sent on int2_Virtio.tx -> int2_NIC.rx (loss rate: 0%)
                   0 sent on mgmt1_NIC.tx -> mgmt1_Virtio.rx (loss rate: 0%)
                   0 sent on mgmt1_Virtio.tx -> mgmt1_NIC.rx (loss rate: 0%)
                   0 sent on mgmt2_NIC.tx -> mgmt2_Virtio.rx (loss rate: 0%)
                   0 sent on mgmt2_Virtio.tx -> mgmt2_NIC.rx (loss rate: 0%)
                   0 sent on xe0_NIC.tx -> xe0_Virtio.rx (loss rate: 0%)
[...]

PR #562, opened 2015-07-14
vhost.h, vhost_user.h: Use more explicit types

lukego:
Switch to stricter C declarations to ensure that the memory layout of vhost-user requests matches the vhost-user specification. For example, use 'uint32_t' instead of 'unsigned int'. Add __attribute__((packed)) to struct declarations.

This is to ensure that the on-the-wire memory layout follows the vhost-user spec rather than the C compiler. Inspired by #560 and *may conceivably* be helpful there.

vhost-user spec:
https://github.com/qemu/qemu/blob/master/docs/specs/vhost-user.txt

### DIFF
--- a/src/apps/vhost/vhost.h
+++ b/src/apps/vhost/vhost.h
@@ -45,12 +45,13 @@ struct vhost {
 // Below are structures imported from Linux headers.
 // This is purely to avoid a compile-time dependency on those headers,
 // which has been an problem on certain development machines.
-struct vhostu_vring_state { unsigned int index, num; };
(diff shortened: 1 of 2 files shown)


lukego:
@virtualopensystems-nnikolaev Do you think this change makes sense? Could otherwise drop it.

virtualopensystems-nnikolaev:
For me it is OK. Have tested it here and was working.

lukego:
@virtualopensystems-nnikolaev Thanks! Since you approve I have merged it onto `next`.

PR #565, opened 2015-07-20
Multiprocessing capabilities

javierguerragiraldez:
A minimal test of multiple-process SnabbSwitch architecture, based in `fork()`ing the initial process and LuaVM.

There's no handling of extra LuaVMs, the initial one is forked with the process, so they all inherit the initialization code.  No sandbox style, nor "single app VM", but any of these schemes could work with this architecture, if necessary.

I've added the OOP syntax discussed some time ago to both link and packet structures, in part because it helps with the polymorphism of having two kinds of link objects: intra-process and inter-process.  That meant rewriting apps to the OOP syntax to make them work with inter-process links.  This patch changes only the `Source` and `Sink` basic_apps.

Some modules (so far `core.memory`, `core.app` and `core.packet`) need some initialization code just after the fork, or during the prefork time.  To help with that, the `core.fork` module calls any `prefork()`, `postfork()`, `endfork()` and `reapfork()` function in any module at well-defined points in time: just before the first fork, just after each fork, just before exiting a child process and just after the child process has died, respectively.  Of these, the most valuable has been `postfork()`, the others could be handled for each specific case separately.

Each process keeps its own packet freelist.  To keep it from starving if it sends other processes more packets than it receives, the inter-process links swap a free packet for each one that gets transferred.  It must be noted that it seems the original link structure works with very little modifications if there's any better way to avoid freelist starvation.  This means that the number of packets processed globally is half the sum of the packets allocated or freed form each freelist.  p.ej, on grindelwald this test reports approx 6million frees in a second-long run, so it means we're transferring around 3million packets per second.

freelists now are fully FFI objects, allocated with `shm.map()` on a name derived from the stored structure.  The only case so far is the `struct packet *` on each process, mapped into files with name of the form: `/<pid>/freelists/struct_packet_#`

The per-process statistics simplify some `engine.main()` details, but it these could be done by local variables, especially if the link counters issues are solved in a better way.

The `memory.lua` DMA pool is global to all processes, and pointers of memory allocated there are equally valid on all processes.  To keep it simple, it now can't grow anymore after the first fork.  Allocations are valid from any process, and protected by a file lock to avoid race conditions.

A global `engine_state` flag is modified by the master `engine.main()` function, and observed from all child processes to know when to start and when to quit.

Github reports that this PR "can't automatically merge", I guess because of the `if not busywait then pace_breathing() end` patched line in `core/app.lua`


lukego:
Exciting progress :-)

lukego:
So it looks like this PR includes solutions to most of the big problems in parallelizing:

1. Sharing link objects between processes: make them into `shm` objects.
2. Sharing packet mappings between processes: allocate DMA memory before `fork` and allocate from a shared pool. The pool is a `shm` and access is synchronized by an `flock` mutex on the `shm` file inside `dma_alloc()`.
3. Returning packet buffers to the freelists of the processes that need to allocate them: the link struct is extended so that free packets can be chained backwards to where they came from. This part I don't understand yet and have to read more closely.
4. Spawning multiple processes: the app network defines which processes will exist and the engine will `fork` the children that are needed.

One part that seems to be missing is incremental updates of the app network. Currently it looks like the engine restarts all child processes every time the app network changes. That would be too drastic. Instead each child should independently inspect the new app network and update its apps as needed (restarting them, reconfiguring them, or leaving them alone).

Restarting after a configuration change is a valid design choice for an individual app but we can't give that luxury to the engine. Imagine if you tweak a small obscure config item and suddenly 20 cores worth of traffic processes all restart, re-discover their gateway address with ARP, reset their connection tracking state, etc... that would be no fun. It's very valuable for apps to be able to expect to stay up and running and be allowed to keep some simple state in local variables.

Cool to see so many pieces of the puzzle together in one place :-). Are there some other parts of the puzzle that still need to be filled in that I haven't thought of?

lukego:
btw really impressive work on streamlining the design and eliminating tricky cases. I really like the fact that this code makes sense on its own without needing to refer to really heavy-weight concepts e.g. computer science papers about lockless data structures, fine details of the x86 memory model, and so on. It is hard to work making software simple :).

lukego:
... and getting the diff down to 500 lines! That is also an impressive and important feat :).

javierguerragiraldez:
> Returning packet buffers to the freelists of the processes that need to allocate them: the link struct is extended so that free packets can be chained backwards to where they came from. This part I don't understand yet and have to read more closely.

I think this is a crucial point and there's quite some room for improvement, in all three different forms:

1. cleaning/optimizing the current implementation (removing some "just in case" code)

1. tightening the theory (i guess the head/mid/tail rings can be replaced with 'normal' head/tail kind; maybe some invariant analysis might even allow folding the two rings into one, with full packets from head to tail and empty "payback" packets from tail to head)

1. using normal rings and replenish freelists some other way.

I would really like to explore option 2, in the best case it might even converge with current links, just don't pay back packets when used intra-process.

javierguerragiraldez:
> One part that seems to be missing is incremental updates of the app network.

Yes it's totally missing.  In the current form, child forks simply call `engine.main()` until the master finishes.  But since each child has the complete code available, it shouldn't be a problem to update their sub-networks too... if we get the config to them.  Three obvious options:

1. kill everything and re-fork.  drastic and maybe slow.  (hundreds of miliseconds? maybe a couple secs? i guess it depends on app initialization code)

1. serialize the config and send (or share) to the childs, with a signal so everybody calls `engine.configure()`.  should each child resume working as soon as reconfigured? or should everybody pause and restart only after everything is consistent again?

1. formalize the "config reading" method and let every child discover the change in the config source and reupdate iself.  Any real world setup must get the config from somewhere and then feed to the `core.config` object (config source -> Lua code -> core.config -> core.engine).  we could say: "write your configs to a JSON file (or yaml, or Lua, or csv) and switch this flag, the engine will reread and reconfigure", each child would do the same and change its own small part.  (and the master would spawn/kill processes if needed)

A related point is that the current `engine_state` "metaglobal" flag is checked on a big, wide loop, where `engine.main()` runs for a second between checks.  If we keep this flag, it could easily be checked on every breath cycle.

Of course I'd like to kill this flag and use some kind of signals, but couldn't really understand how does ljsyscall manages them.  `sigsuspend()` and `SIGSTOP`/`SIGCONT` are really tempting, but I didn't manage to setup signal handlers, so only got killed forks.  :-(

lukego:
So the radical idea that keeps coming back to me is that the code might become a lot simpler if we allowed each process to be started independently instead of assuming they are forked.

The forking is convenient at first but then it gets things tangled up. For example the app network being in copy-on-write memory and awkward to update across all processes; the parent needing to wait/synchronize on terminating child processes; every module having to think about whether it is "fork-safe" with prefork/postfork methods.

How about if we switch the model to allowing processes to be started separately? We could still `fork()` the processes when creating an app network but this would be a convenience and not a requirement: you could also run genuinely separate app networks and still connect them with cross-process links.

I have a few ideas that would help that I need to flesh out a bit more:

1. A way to privately allocated DMA memory and then map it into new processes on-demand as they receive packets. (Hint: catch `SIGSEGV` and `mmap()` the mapping.)
2. A way to pass packets between different processes over ordinary links and still recover them to the right freelist -- even when processes crash and restart. (Hint: Garbage collect packets.)

I don't have any idea yet for sending updates app network descriptions between processes. I suppose we could put each new configuration into a well-known location (e.g. file or shm object called `config.47`) and expect that everybody will find it sooner or later?

I will write these up ASAP and you guys can shoot them down :).

javierguerragiraldez:
> How about if we switch the model to allowing processes to be started separately? We could still fork() the processes when creating an app network but this would be a convenience and not a requirement: you could also run genuinely separate app networks and still connect them with cross-process links.

hum.... i think we still need some way to associate all the intervening process into a group.  Maybe the easiest would be to specify a 'home' directory in the command line; then all the shm objects are relative to that dir.  Another nice thing is getting rid of the `S.getpgid()` i'm using for several global objects; simply use constant, "home-relative" filenames.


> 1. A way to privately allocated DMA memory and then map it into new processes on-demand as they receive packets. (Hint: catch SIGSEGV and mmap() the mapping.)

I like this.  how about a subdir with all the DMA mappings, and from time to time (on SISEGV?, SIGHUP?, inotify on the dir?) `mmap()` everything there.  would this need a "map directory"? or just good, descriptive names?

> 2. A way to pass packets between different processes over ordinary links and still recover them to the right freelist -- even when processes crash and restart. (Hint: Garbage collect packets.)

hum... if each process "owns" a subset of those DMA chunks, it could be easy to determine where each packet should return.  what i don't like is to allow any process to update the free list, it should be private... maybe just marking the packet as available, so the originating process can GC it.  i think it would need some careful tuning to avoid it becoming a big drain, but reverting to the basic link structs is tempting.

lukego:
> i think we still need some way to associate all the intervening process into a group

I would be interested to understand how fundamental this is. Could we make things simpler and more orthogonal by having a "flat" design that potentially connects every process running on the machine?

Just wondering whether group-shared objects are potentially a "code smell" indicating bits of state that have complex ownership rules and may end up being difficult to synchronize access to. For example having a common DMA pool seems more complex than having separate ones, all other things being equal.

(I suppose we may find entirely other reasons to separate objects into groups e.g. for running multiple versions of Snabb Switch that each expect shared memory files to be in a different format. But that is perhaps a separate problem.)

javierguerragiraldez:
Maybe superfluous if the packet GC idea pays out, but I really wanted to simplify the packet-swapping inter-process link.  Now it's 130 lines less!  :-)

lukego:
Wow! Great effort. That is really crystal clear now. I see now that the idea is very simple and neat.

Is the difference between `link` and `interlink` now so small that we could merge them together? e.g. simply have a flag on `struct link` saying whether you need to put back a free packet when you receive a full one? That would also potentially shrink the code-and-concepts footprint of the diff.

One nagging feeling is that LuaJIT may make us regret introducing unpredictable control flow on `link.transmit()`. I am thinking that the interlink will become `full` much more frequently (since both sides have to free up resources) and that this will execute a `while` loop that will fail to unroll. This seems likely to create a whole bunch of side-traces in every app that might make them harder to profile. I don't have an immediate solution.

lukego:
I do have one idea to simplify the control flow actually.

Suppose that `transmit()` on an interlink always frees exactly one packet. Then `full()` returns a flag saying whether there is a packet to free or not. The loops in app `pull` and `push` methods would always be successfully sending one packet and freeing one packet, until the link fills up, which is a termination condition for the loop anyway.

javierguerragiraldez:
> Suppose that transmit() on an interlink always frees exactly one packet.

I had a related idea: keep the whole ring full, from `head` to `tail` with data packets, from `tail` to `head` with payback packets.  To `transmit()`, first free a packet to make space; on `receive()` fill back the spot with an empty packet.

What I don't like about it is that it keeps a number of packets tied up.

Also, this last change doesn't improve (or degrade) performance a bit.  I guess that means the loop is being throttled by something else...

javierguerragiraldez:
.... and done.  no `free` index, no harvesting on `:full()`.

javierguerragiraldez:
now the interlink struct is usable to link apps in the same process, replacing the old link (it's almost identical now).  A simple source->sink in a child process runs at 35-38Mpps, so it seems the overhead is low enough.

Inter process links still run at 3-7Mpps, and when there's both kinds of link, the local link gets only 200-400Kpps, around than 1% of the local only case.

in this sample one second run, `source` and `'localsink` reside in process 1, and `sink` in process 2.  The inter process link gets 6.7Mpps, while the intraprocess is only 390Kpps

           6,774,414 sent on source.output -> sink.input (loss rate: 0%)
             390,915 sent on source2.output -> localsink.input (loss rate: 0%)

The profiler always reports 100% compiled, and the numbers are only slightly affected by commenting out the `pace_breathing()` call to avoid any possibility of sleeping in the loop.  Changing initialization order, so the local link appears first in the list of output links of source, or using two separate Source apps (in the same process) doesn't change situation either.

Increasing the ringsize made it even worse.  This is with 1024 packet rings:

           6,615,785 sent on source.output -> sink.input (loss rate: 0%)
              72,633 sent on source.output2 -> localsink.input (loss rate: 0%)

The low inter-process performance, and the wide variation range (3.5-7 Mpps) are starting to worry me, as I don't see much opportunity to optimize.  And the way intra-process links get penalized really confuses me.  It looks like an algorithm failure, not just code overhead.

javierguerragiraldez:
Another confusing test: adding a third link.

two inter-process and one intra-process link: the sum of the two inter-process remains slightly more than 3Mpps.

Sometimes it's roughly balanced:

           1,765,518 sent on source.output -> sink.input (loss rate: 0%)
             129,285 sent on source.output2 -> localsink.input (loss rate: 0%)
           2,199,103 sent on source2.output -> sink.input (loss rate: 0%)

Sometimes it's not:

           2,434,132 sent on source.output -> sink.input (loss rate: 0%)
             148,410 sent on source.output2 -> localsink.input (loss rate: 0%)
             954,612 sent on source2.output -> sink.input (loss rate: 0%)

One inter-process and two intra-process links: the sum of the two intra-process is limited to a around 150Kpps: (and both local links get _exactly_ the same number of transfers)

           3,473,584 sent on source.output -> sink.input (loss rate: 0%)
              87,465 sent on source.output2 -> localsink.input (loss rate: 0%)
              87,465 sent on source2.output -> localsink.input (loss rate: 0%)

It really looks like there's some funny bug blocking local links, but only if there's at least one inter-process; with two intra-process, the total transfer is over 50Mpps, there's _very_ little variation, and both links seem to be in lockstep:

          27,872,775 sent on source.output2 -> localsink.input (loss rate: 0%)
          27,872,775 sent on source2.output -> localsink.input (loss rate: 0%)

lukego:
How do I run these benchmarks?

lukego:
(Sounds like what you need is a second pair of eyes to try it out and ask dumb questions. Happy to volunteer.)

javierguerragiraldez:
It's just a simple configuration in `src/apps/basic/test_fork.lua` `selftest()`, so I just do `snabbswitch snsh -t apps.basic.test_fork`.

javierguerragiraldez:
I've got a vague feeling that a test so tight (just Source and Sink, no processing at either side), the whole system becomes just a test of trashing one or two cache lines between two cores.  The profiler reports a significant part of the time in just setting the `.has_new_data` flag; that sounds like most writes become inter-process sensitive.

I'll try to double-buffer the inter-link struct, to allow the transmitting app to quickly push into a non-shared buffer, and then commit into the main ring as blindly as possible.

The first step, of course is to use `:nwritable()` instead of constantly testing `not l:full()`.  It's encouraging that this simple change is the first one to really affect performance.  Although in some counter-intuitive way... it doesn't make the inter-process link any faster, just reduces the impact on the local link.

lukego:
Should move read and write indexes into different cache lines. Don't want two cores writing to the same cache line.

javierguerragiraldez:
hum... separating the `read` and `write` indexes by putting them at either side of the `stats` struct, or even of the packet ring didn't make any noticeable difference.  Neither did a simple private front buffer.

What did make a small difference was only storing the link in the `core.app` table/array at the process that holds the receiving app.  It made the breathe loop slightly simpler, avoiding a few checks in the link struct.

javierguerragiraldez:
This last patch includes some different small fixes.  Those that seem to help but might not be worth it:

- Each link has two pointer arrays, one for full packets and one for the empty "payback" packets.  It seems to slightly reduce cache thrashing.

- Allow the `config.cpu()` to specify the `engine.busywait` flag for each process.  Useful because in tests it's common that a fork doesn't include any "receiving" app, so the main loop doesn't see any work and starts to sleep.

- Don't check for full link on each `:transmit()`, assume that the sending app has already called `:full()` or `:nwritable()`.  Sending a packet on a full link becomes an error.  Just like calling `:receive()` on an empty link corrupts it.

But what seems to be today's key improvement is the front buffer: a private packet ring that holds packets until it gets full, then pushes to the real link as fast as possible.  In fact it doesn't seem to be much faster than calling `:transmit()`; the advantage is that it avoids small cycles, getting more packets across for each inter-process cache flushing.

Other ideas to get the same effect include:

- do it implicitly in the sending app. something like:
```Lua
local nw = l:nwritable()
if nw >= xx then for i=1, nw do
    l:transmit(....)
end end
```
con: this would have to be done for any app that might be used as "the last app before an inter-process link".  Simply preferring this style for _all_ apps doesn't sound right, much less coder-friendly.

- write a "buffer" app that does the same thing and have the "network implementer" add to the specs.

If this approach proves to be the way to go, then I think it could be transparently added to inter-process links, either by the engine configuration code or by the link itself.

Alternatively, thinking in the "don't fork, let separate executions connect themselves" scheme, I guess that we could replace the interprocess links by transmit/receive apps, paired by a common name.  something like this:

in config 1:
```Lua
    config.app(c, 'trans1', transmit, {linkname='link_1'})
    config.link(c, 'someapp.output -> trans1.input')
```
in config 2:
```Lua
    config.app(c, 'recv1', receive, {linkname='link_1'})
    config.link(c, 'recv1.output -> getthis.input')
```

Where `transmit` and `receive` both `shm.map()` the packet ring in the `'link_1'` file and implement the write and read of the ring, buffering packets if necessary.

lukego:
Transmit and Receive apps is an interesting idea. I suppose really that this is the simpler design and that making all links interprocess friendly is mostly a (premature) optimization. If you think the implementation can be simplifed with separate apps then I would say go for it and let's see what we can learn from that.

lukego:
I have finally taken this for a spin. However, now I need to get `perftools` installed on NixOS so that I can get visibility into the performance counters :). On it... See also NixOS/nixpkgs#9095.

javierguerragiraldez:
some simple tests with the simple `inter-proc` app and returning to 'normal' links seem to be stable around 8Mpps on grindelwald.  Once I clean it up, it would make the minimum PR much smaller, especially if I remove the whole fork idea in favor of the multiple executions.  I still haven't been able to make that work cleanly because the biggest modifications are in `memory.lua`, not my favorite part of SnS...

lukego:
@javierguerragiraldez would the memory situation be simplified if I implement #571? (Or what would be useful assistance with `memory.lua` hacking?)

javierguerragiraldez:
This last change (the ability to pin a process to a set of CPU cores) finally makes performance stable!

testing on grindelwald, which is 2x12cores @2.7GHz

Source -> Transmit / Receive -> Sink | packet count on 1 sec
-------------------- | -------------
No CPU pinning | 4-8Mpps, sometimes 10...
both to the same core | 10-11Kpps (ouch!)
different cores, same chip | 11-12Mpps (best!)
different chips | 5.7Mpps

would be interesting to check on two sides of a hiperthreading core, I guess it would be slower than two separate cores, but better than same core.

javierguerragiraldez:
> would the memory situation be simplified if I implement #571?

Yes, i've tried to do it myself, but no luck so far.  I think it must be straightforward, but the exact enchantments on hugetables are evading me (and that's before trying to catch SIGSEGV).

javierguerragiraldez:
Lot's of cleanup to do (mostly removing `fork.lua`, all the `pre/postfork()` hooks and the "per-fork" configurations), but I got 15Mpps between two different SnS instances :smile:.  Happy to confirm that the packet size doesn't affect this figure, it doesn't matter if it's 60 or 1500 bytes per packet.

To reproduce, start a packet-producer (just a `basic_apps.Source -> inter_proc.Transmit('spoon_link')` setup).  It will sit there until killed:

    snabbswitch snsh program/spoon/produce.lua

Then on a different terminal, start the packet consumer (a simple `inter_proc.Receive('spoon_link') -> basic_apps.Sink`).  It will pull packets for a second and then report:

    snabbswitch snsh program/spoon/consume.lua

javierguerragiraldez:
I think this looks a little less hacky: use a simple shared struct (via `core.shm`) to store the DMA hugepage ids.  In the C code it looks like a simple `int` array; no need to arbitrate, because the addresses (and therefore the array indexes) are unique.

The included `program/spoon/director.sh` is an idea of how to run several Snabb processes: start everything in background and wait for any of them to die.  I guess it could check if it should be restarted; but if not, then kill everything else.

The `cleanup` function removes 'global' objects: any inter-process link and the DMA ids, previously unmapping them.

javierguerragiraldez:
With 6 inter-process links crossing 12 processes, each on a different core, all on the same chip... the per-link throughput falls slightly, from 12-1.3Mpps to 11.5-11.7Mpps.  In all, this test clocks around 70Mpps of cumulative throughput.  (and over 100Mpps when doing 9 links on 18 cores, enough to saturate a 100GbE link if the packet averages 125 bytes or more)

lukego:
Howdy! Sorry about the slow feedback.

This is now a *very* impressively short change! The idea of using dedicates apps to transport packets between processes seems to be extremely good. I really like the way it will not impact the runtime behavior of existing applications that don't use it, and that it takes so extremely little code.

I think we have a winning design for the initial multiprocessing support :). It is also nice for the future that multiple such solutions could co-exist which makes experimentation and refinement easy in case we have new ideas that we want to try later.

Super duper cool! and to think we came *this* close to entering pthreads hell. Great save :-)

I have a few mundane housekeeping feedback before merging:

Can we have another pass at shrinking the diff? Even though it is small it seems like there are some changes that are not required to be bundled with the multiprocessing: particularly the new FFI metatypes in `packet` and `link` modules seem like an unrelated change (?). I do think that FFI metatypes are the right facade for our API, after your previous benchmarking experiments, but I suspect we can take that separately.

Can we have a long comment or README.md explaining the `interproc` app design? (So that people will know what the app is for and what `ret_packets` is and why it works to pass pointers between processes.)

Can we also have a comment explaining how the extended type syntax in `freelist` works? (Or remove that change if it is not needed...?)

Can we have a README.md and some in-file comments explaining the `spoon` program too?

Just amazingly cool. This is exactly how I hoped the multiprocessing support topic would play out. i.e. that we consider and prototype all of the most extreme options and then eventually settle on something extremely simple. Great hacking :).

lukego:
oh one more thought: perhaps rename `app.lua` in interprocess/ to something else? I really like to have globally unique filenames so that it is easy to find the right Emacs buffer :) and `app.lua` is already taken (though that is overdue to be renamed `engine.lua`...)

src/core/memory.h
```diff
@@ -2,3 +2,10 @@ int      lock_memory();
 void    *allocate_huge_page(int size);
 uint64_t phys_page(uint64_t virt_page);
 
+void setup_signal();
+
+
+struct map_ids_t {
+   int huge_page_bits;
+   int ids[16384];
```
nnikolaev-virtualopensystems: I had to increase this number on my setup, otherwise I was getting:
```
snabb: core/memory.c:104: allocate_huge_page: Assertion `offst < (sizeof(map_ids->ids) / sizeof(int))' failed.
```
when inspected the `offst` value was int the range `17001` to `170013`.

javierguerragiraldez: I'd like to know what does it depend on... for a while I dabbled on giving it an extra offset, since it seems to pick a semi-random point and then grow downwards.  But then I realized that any unused page-size segment of this array wouldn't get paged in, so by giving it an outrageous size we only waste hard disk space, not RAM.  So, it wouldn't be crazy to just set it at 8*1024*1024 to cover the whole segment.

javierguerragiraldez:
superceded by #601

PR #566, opened 2015-07-20
core.counter: Add double-buffering

lukego:
This is primarily intended to avoid a performance problem discovered on PR #558 where frequently accessing many public counter objects is expensive because their addresses are page-aligned and this clashes with the set-associative L1/L2 cache on the Xeon.

Double buffering is a handy technique though and it could become useful in the future for publishing structured shared memory objects and supporting atomic reads.


lukego:
Further work on this feature is being done on #568.

PR #567, opened 2015-07-20
SnabbTop + performance fix

lukego:
Use double-buffered counters (#566) to resolve the performance regression in SnabbTop (#558).


lukego:
Superceded by #568.

This integration branch is based on `next` and the new PR is sent directly there instead of master. (The list of changes looks very long on this PR because it includes unrelated stuff from `next`.)

PR #568, opened 2015-07-20
SnabbTop + performance fix

lukego:
Use double-buffered counters (#566) to resolve the performance regression in SnabbTop (#558).

### DIFF
--- a/src/core/counter.lua
+++ b/src/core/counter.lua
@@ -55,15 +55,15 @@ function open (name, readonly)
 end
 
 function delete (name)
-   local ptr = public[numbers[name]]
-   if not ptr then error("counter not found for deletion: " .. name) end
+   local number = numbers[name]
+   if not number then error("counter not found for deletion: " .. name) end
    -- Free shm object
-   shm.unmap(ptr)
+   shm.unmap(public[number])
    shm.unlink(name)
    -- Free local state
    numbers[name] = false
-   public[ptr] = false
-   private[ptr] = false
+   public[number] = false
+   private[number] = false
 end
 
 -- Copy counter private counter values to public shared memory.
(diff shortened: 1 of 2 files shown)


lukego:
The double buffering change causes some problems (as we see in SnabbBot tests):

1. Counters are not ordinary `shm` objects anymore. You can't perform `shm` operations directly on objects you get from `counter.open()`. This is breaking `link.free()` that attempts to delete counters directly via `shm` operations. Could be solved by adding a `counter.delete()` function and using that instead of `shm` calls.
2. Counters will lose updates if you map them read/write more than once (whether in the same process or different ones). Have to avoid this situation. One idea would be to use `flock` to raise an error if a counter is opened read/write more than once. Another idea would be for `counter.publish()` to only update the public shm counter when the private value has been changed.

eugeneia:
I can't build this branch as of now?

```
make[1]: *** No rule to make target `jit/vmdef.lua', needed by `obj/jit/vmdef_lua.o'.  Stop.
make[1]: Leaving directory `/home/max/snabbswitch/src'
make: *** [all] Error 2
```

lukego:
You probably need #576

eugeneia:
The [segfault](https://gist.github.com/SnabbBot/fa9592de7b3fdaa094f5#file-log-L498) is weird. Did we miss something? I can't think of a counter (e.q. `configs`) that is e.g. open for writing twice. The NFV selftest failing is probably related.

Can't explain the (reproducable) intel failure either...

eugeneia:
@lukego I found the bug and fixed it in a PR on your fork: https://github.com/lukego/snabbswitch/pull/4

If SnabbBot will be happy I think this is fit for merge and we can tackle the remaining issues when required (e.g. when multi-process snabb lands)?

lukego:
Great fix, Max! I was wondering what the problem could be.

Advanced Github-fu to send a PR to a PR too :-)

eugeneia:
@lukego Ugh... I just manually tested `next` for release and realized `snabb top` doesn't actually work. First there is a bug in `snabb top` that makes it crash when deallocating counters. I actually have a fix for that and its not a big issue but now the link counters always read 0, still troubleshooting that one.

eugeneia:
Ok got it, see #582.

PR #570, opened 2015-07-21
core.link: Streamline counters [DRAFT]

lukego:
Reduce the amount of counter-related code and data in `struct link`.

The ring indexes (read, write) and now upgraded to 64-bit integers that do not wrap around [*]. This means that we don't need to use separate packet tx/rx counters because the ring indexes serve that purpose directly. (We do have to be careful to handle wrap-around when actually looking up elements in the ring based on the index.)

The byte counters are simply removed. Do we really need them? (How about if we tracked only packets-per-second plus perhaps a system-wide average packet size?)

Dropped packets are still counted directly because this is important information and I don't immediately see a way to derive it indirectly.

[*] 64-bit packet counters really should not wrap around. Even at one billion packets per second it would take more than 500 years. However, if this is not a safe assumption then we could detect wrap around and account for it on the counters.

(Google: "2^64/(1 billion per second)" => "584.554531 years".)

@eugeneia @javierguerragiraldez @alexandergall what do you guys thing about this idea? I mean in terms of information loss, impact on the code, and conflicts with other open PRs?

### DIFF
--- a/src/core/link.h
+++ b/src/core/link.h
@@ -12,7 +12,9 @@ struct link {
   // Two cursors:
   //   read:  the next element to be read
   //   write: the next element to be written
-  int read, write;
+  int64_t read, write;
+  // How many packets have been dropped due to ring overflow?
+  int64_t dropped;
   // Index (into the Lua app.active_apps array) of the app that
   // receives from this link.
   int receiving_app;
--- a/src/core/link.lua
+++ b/src/core/link.lua
@@ -20,11 +20,8 @@ end
 
 function receive (r)
 --   if debug then assert(not empty(r), "receive on empty link") end
-   local p = r.packets[r.read]
-   r.read = band(r.read + 1, size - 1)
-
-   r.stats.rxpackets = r.stats.rxpackets + 1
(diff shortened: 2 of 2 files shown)


javierguerragiraldez:
I think `full()` is wrong when `read >= size-1`.  this should work:

    function full (r)
        return r.write == r.read + size - 1
    end

But in all, is there a measurable difference?  I remember doing some benchmarks and finding no difference between

    while not empty(in) and not full(out) do ... end

and

    for _ = 1, min(nreadable(in),nwritable(out)) do ... end

even if the first form does a check on every iteration, while the second one does only at the start and it's an integer count from there on.  I guess the index pointers become essentially free once they get cached in.

eugeneia:
I am always for simplicity but I wonder what is the benefit here? Did rx/txbytes actually cost us much? It seems like an interesting metric to me so I would consider keeping it.

lukego:
@eugeneia That is a fair point. I see no urgent need to make this change now. We could revisit this idea in the future if/when we are trying to squeeze more performance out of the basic app network and see if it really helps.

I close this PR for now.

PR #575, opened 2015-07-27
dynasm: Switch from C-based assembler to Lua-based assembler

lukego:
This is an upgrade for a little-known feature in the Snabb Switch code base: convenient support for generating machine code at runtime. That is, ad-hoc just-in-time compilers, which in simple cases are much like ordinary inline assembler.

The dynamic assembler capability is intended to make it easier for Snabb Switch hackers to connect with the machine when we want to. If you happen to be studying material such as Intel's Assembly Programming Optimization Guidelines, or Agner Fog's software optimization manuals, or a book like "Computer Systems: A Programmer's Perspective" then this should be a convenient vehicle to connect theory with practice. (If you want to use it in production then that is fine too.)

This is based on a dynamic assembler called `dynasm` which now has multiple incarnations.

The old code that this commit replaces uses Mike Pall's C-based dynasm. I say "C-based" because with this version of dynasm you write C code that generates machine code. This is used internally by LuaJIT.

The new code that this commit adds switches us over to Cosmin Apreutesei's Lua-based adaptation of dynasm. In this version you write Lua code that generates machine code directly (no need for C).

The integration into Snabb Switch is such that any file named *.dasl will automatically be preprocessed with dynasm. In other respects these files are treated as Lua source. There is a simple example and test case in src/apps/example/asm/asm.dasl.

The weakness of this integration is that I have copied and slightly tweaked the dynasm code directly into snabbswitch/src/. This is messy and should be seen only as a first step.

Here is the documentation for Cosmin's Lua-based dynasm:
https://github.com/luapower/dynasm/blob/master/dynasm.md

Here is the documentation for Mike's original C-based dynasm:
http://luajit.org/dynasm.html


pavel-odintsov:
Wow! We could build so much interesting things with it now :) Looking forward for dynamically generated packet parsers deeply optimized for specific architecture.

lukego:
I merged this onto `next` even though the integration is a little messy and we will need to find a workflow for keeping it in sync with upstream. For now it seems okay because it is a small amount of code that is not changing much.

PR #580, opened 2015-07-30
Fix checksum-related iperf performance regression

lukego:
This is a work in progress PR about a bug that I have been tracking down for the past couple of days.

The bug was first reported as a performance regression on `iperf` that cuts throughput to half and was introduced some months back. This was not visible in the `iperf` performance regression test in our CI or in the testing that I have been doing myself. But I have now been able to reproduce the problem on `chur` (albeit not as severely) and I have tracked it down to a checksum bug that is not caught by our checksum unit tests.

Here is what I know:

1. The problem is essentially caused by insufficient test coverage in our checksum routines.
2. The bug was introduced in PR #494 (Modify checksum API) and commit 2deca0b097dbe1795ef1f3da14a24f6f56c45cef.
3. The specific problem is that `lib.checksum.verify_packet()` is returning `false` (bad checksum) for packets with good checksums.
4. The reason this can appear as a performance regression is that it prevents the Virtio-net code from telling the VM that the packet checksum is already verified, and that causes the VM to checksum the packets itself with the CPU, and this will impact performance in setups where the guest becomes a bottleneck. (Have confirmed that forcing the "checksum valid" flag in Virtio-net solves the performance problem.)
5. The checksum seems to work OK on transmit and but not on receive. This is interesting because it is essentially the same code. One detail that jumps out is that the `initial` argument is always 0 on transmit and so the bug may be triggered by non-zero `initial` (e.g. if we are missing an endian swap on the value). However, nothing I have tried has fixed the problem yet.

I write this partly to collect my thoughts and partly to invite a second pair of eyeballs. Maybe @alexandergall or @javierguerragiraldez still have this code swapped in? If so ideas welcome :).

The bug can be reproduced by checking out this branch and running `./snabb snsh -t lib.checksum`.

I am sure that I have the right commit because first `git bisect` showed me that this is where performance reduced and then I found that this test case actually does work if you run it against the previous commit.


javierguerragiraldez:
I think it is a byte-order issue; commit 2deca0b made all checksum functions to explicitly work only in host order, but `C.pseudo_header_initial()` returns the 'initial checksum' in network order.

It seems the missing `ntohs()` is in checksum.lua line 70, like this:

    return ipsum(buf+headersize, len-headersize, ntohs(initial)) == 0

lukego:
Does that make the test case pass? I have been trying this kind of thing without success so far.

lukego:
I also think that fix should go inside the pseudo checksum function in order to be consistent with the comments in checksum.lua

lukego:
@eugeneia SnabbBot not testing this change?

lukego:
@javierguerragiraldez Oh! You are right! Looks like I outsmarted myself: I broke the test case when I moved it from an ad-hoc script file into `selftest()`. I suspect the code on this branch now is correct since 84b497a and that I need to verify this on the real test environment and then correct the selftest function.

lukego:
This does seem to resolve the issue. Great. @javierguerragiraldez thanks!

@eugeneia I reckon I understand why this was not detected by the SnabbBot CI iperf performance regression test. This problem causes performance to suffer only if the guest VM becomes the limiting factor in performance. SnabbBot tests with two guests attached to one Snabb Switch and in that case the limit is the Snabb Switch (not surprising especially since CI is running without AVX2 CPU). The setup where the problem has been seen is when each guest is attached to a separate Snabb Switch instance and in that case this checksum issue causes the bottleneck to move from the Snabb Switch into the guest that is receiving packets (iperf server).

eugeneia:
> @eugeneia SnabbBot not testing this change?

@lukego I can't ssh into davos (or grindelwald for that matter), so maybe a network issue?

lukego:
(davos up again now. The Ubuntu install there is a bit sad.. default GRUB entry does not boot. Bear with me.)

PR #589, opened 2015-08-07
core.memory: Catch SIGSEGV to map DMA memory [PoC]

lukego:
This is a quick and dirty proof-of-concept hack done to support SnabbCo/snabbswitch#565 (multicore tests):

Extend core.memory so that the virtual addresses of all packets are
valid in all Snabb Switch processes. The memory allocated by
memory.dma_alloc() is treated specially: if another process tries to
access it then this triggers a SIGSEGV handler that will detect which
memory is needed and automatically map it into the process.

Two simple test scripts are provided: s1 that allocates memory and s2
that modifies it without any explicit mapping.

Run like this:

    $ sudo ./snabb snsh s1
    Installing SIGSEGV handler
    address cdata<char *>: 0x5009f7c00000
    current value = 0
    current value = 0
    current value = 0
    ...

Then use the same address to change the value with s2 that simply
converts it to a pointer and pokes it:

    $ make && sudo strace  ./snabb snsh s2 0x5009f7c00000 42

and now you see the new value in s1:

    current value = 42

This is really quick and dirty! I will clean it up if it works and we decide to keep it.

cc @javierguerragiraldez

### DIFF
--- a/src/core/memory.c
+++ b/src/core/memory.c
@@ -13,6 +13,7 @@
 
 #include <assert.h>
 #include <fcntl.h>
+#include <signal.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
@@ -95,7 +96,18 @@ void *allocate_huge_page(int size)
   if (mlock(realptr, size) != 0) { goto fail; }
   memset(realptr, 0, size); // zero memory to avoid potential surprises
   shmdt(tmpptr);
-  shmctl(shmid, IPC_RMID, 0);
+  //shmctl(shmid, IPC_RMID, 0);
+  
+  // proof-of-concept hack: write shm id to a file.
+  {
+    char fname[256];
+    FILE *f;
+    snprintf(fname, sizeof(fname), "/tmp/%p", (void*)physical_address);
+    f = fopen(fname, "w");
+    fprintf(f, "%d\n", shmid);
+    fclose(f);
+  }
+  
   return realptr;
  fail:
   if (tmpptr  != MAP_FAILED) { shmdt(tmpptr); }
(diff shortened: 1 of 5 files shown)


lukego:
This was the least disgusting implementation I could come up with when browsing kernel sources etc. The shm syscalls are really annoying and non-interoperable with other kernel features, but we seem to be stuck with them because `mremap()` doesn't work with hugetlb from mmap(). C'est la vie.

Weekend!

javierguerragiraldez:
great!  it's much easier to hack around if there's a working PoC.

I'll go hacking with this tonight, next stop: remove fork()s!

lukego:
btw: this version is going to leak huge pages because it doesn't unlink the shm IDs anymore. If you start running low on huge pages then you can use the command line tools `ipcs` to list the allocated ones and `ipcrm` to delete them.

eugeneia:
@lukego That is another reason why SnabbBot broke down: It failed to allocate huge pages....

lukego:
@eugeneia Good point. Have to think about potential impact of running these rough draft patches in the CI environment. (I don't think it will have been a problem here: much more likely it was the reboot. QEMU depends on a bunch of hugepages to be preallocated whereas Snabb Switch allocates them on demand.)

I wonder if Docker-fication will help with this kind of problem. That is, if rough draft code makes a mess in `/var/run` and `/dev/shm` and so on, that this would be discarded with the one-shot container that runs the tests. Or indeed that the test suite could detect such messes since it knows that the tests had started running in a pristine environment.

eugeneia:
@lukego Nope its the leaking (try `sudo ipcs | wc -l` on davos). Can I safely reboot davos nowadays? `ipcrm` doesn't seem fit for the task.

eugeneia:
@lukego Do you have any pointers of what Docker sandboxing actually provides? I have a hard time finding out what it sandboxes and what not.

lukego:
@eugeneia Interesting. I suppose the problem is that I "innocently" pushed a change with a resource leak on memory allocation, but then the CI is running zillions of tests (booting VMs, etc) with that leak enabled. So whereas I imagined that one or two pages would leak in fact this is happening a lot.

Here is a command-line to clean it up. Perhaps we want to run this in a loop for the time being:

    for m in $(sudo ipcs -m | awk 'NF == 6 {print $2}'); do sudo ipcrm -m $m; done

It is very unlikely to cause a problem for the CI. The master Snabb Switch is only allocating the shared memory object for a millisecond or so in order to map it into memory.

Docker is using Linux Containers (cgroups). The kernel hackers have been busy and added support for multiple namespaces for filesystem / processes / sockets / routing table / etc.

eugeneia:
But we also use shared memory objects for the counters? Won't that lead to segfaults?

lukego:
Counters are a different kind of shared memory. The ipc commands are only operating on "System V" shared memory objects created by `shmget()`. That is how we allocate DMA memory (due to obscure Linux kernel limitations) but `core.shm` and counters are done differently (we `mmap()` a file instead).

However my command is pretty brutal in that it is deleting all of the root user's shared memory objects system-wide. So maybe it will crash some other process installed on the server. I suppose we will find out.

I do actually think that Docker/containers is a good solution to this problem. That should also protect against other interesting kinds of problems e.g. I tried to write the `shm` module test suite carefully to avoid potentially doing an `rm -rf /` if there would be a bug in the path resolving code. That would not be a problem if the test runs inside a container but it would be a major headache if it nuked davos' root file system.

javierguerragiraldez:
i've been cleaning those in my tests "manually" with a `grep` line to only `ipcsrm -m` those with zero active references (the last column in `ipcs`).  I've also done a slightly longer bash script to use the ids written to `/tmp/0x*0000` files.

in the end, i think any multi-process gathering of SnabbSwitch instances would be managed by  a 'director' that includes some cleanup options.

lukego:

PR #590, opened 2015-08-11
Virtualopensystems contributions

nnikolaev-virtualopensystems:
A set of 3 patches:
 - introduce virtualopensystems-contributions branch
 - automatic unbinding of the PCI device in the Intel 82599 driver
 - counter functions caching in app, link and packet

### DIFF
--- a/src/apps/intel/intel_app.lua
+++ b/src/apps/intel/intel_app.lua
@@ -32,6 +32,16 @@ end
 function Intel82599:new (arg)
    local conf = config.parse_app_arg(arg)
 
+--[[
+   Unbind the device from the Linux kernel. If the device is already unbound, io.open fails
+   and we silently skip the write operation.
+]]--
+   local unbind_fd = io.open("/sys/bus/pci/devices/"..conf.pciaddr.."/driver/unbind", "w")
+   if unbind_fd then
+      unbind_fd:write(conf.pciaddr)
+      unbind_fd:close()
+   end
+
    if conf.vmdq then
       if devices[conf.pciaddr] == nil then
          devices[conf.pciaddr] = {pf=intel10g.new_pf(conf):open(), vflist={}}
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -31,6 +31,8 @@ freebits  = counter.open("engine/freebits")  -- Total packet bits freed (for 10G
 freebytes = counter.open("engine/freebytes") -- Total packet bytes freed
 configs   = counter.open("engine/configs")   -- Total configurations loaded
 
+local counter_add, counter_read, counter_commit = counter.add, counter.read, counter.commit
+
 -- Breathing regluation to reduce CPU usage when idle by calling usleep(3).
 --
 -- There are two modes available:
@@ -109,7 +111,7 @@ function configure (new_config)
    local actions = compute_config_actions(configuration, new_config)
    apply_config_actions(actions, new_config)
    configuration = new_config
-   counter.add(configs)
+   counter_add(configs)
 end
 
 -- Return the configuration actions needed to migrate from old config to new.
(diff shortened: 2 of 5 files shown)


lukego:
Great that you will maintain a branch! This seems like the key to scaling development to me. You and I can gradually thrash out a working/testing/reviewing/etc style that works for us and then you can do most of the review of the changes on the branch before PR'ing them to master. Very good :). Let's get started!

Generally I need commit messages to explain *why* changes are being made. If the change is a bug fix then how is the bug triggered and how does it affect users? If the change is a performance optimization then what is the impact and how can it be measured? This will help every interested person to evaluate and test the change and potentially improve on it in the future. When this is not explained in the commit messages then I need to ask a bunch of questions on the PR and then other people will have to find that discussion in the future.

lukego:
The level of corporate branding on this PR seems excessive to me. I would like to find a better way to make sure that both individuals and companies are recognized for their contributions.

Maybe I started that with all of the "Snabb" names and logos. The intention of that is really to be a brand for the project - like "Linux" and its penguin - and not a corporate thing of mine. However I am using the same brand for my professional services, mostly for historical reasons, and that may not be appropriate anymore. I am open to feedback about what would make people more comfortable.

To be clear the reason I say it seems excessive is that the virtualopensystems brand is in the PR title, the branch name, the Github usernames of all the contributions, and the company logo is being used for all of the user avatars. If everybody would brand their contributions like this then it could dehumanize the upstream project and make everything feel corporate.

Can you explain a bit about what is important to you guys and perhaps we can find a better solution for giving credit where credit is due?

virtualopensystems-nnikolaev:
@lukego we can rename the branch to `upstream-contributions` will this fit better with the project.
We would like to keep the names and avatars like they are now.

eugeneia:
@lukego Seems like davos was rebooted and didn't `sysctl -w vm.nr_hugepages` so qemu wouldn't run. I've added `sysctl -w vm.nr_hugepages=10240` to `rc.local`.

lukego:
@eugeneia ah! we will fix this for all lab servers forever once we have a unified nixos grub configuration stored in Git :-).

lukego:
> We would like to keep the names and avatars like they are now.

@virtualopensystems-nnikolaev Could you say a few words to help make me more comfortable with this? For example, are there other projects on Github that taking branded contributions like this and where that is working out well? I have never seen this practice before so I am curious.

lukego:
re: branding a few more practical concerns about how it interacts with the way Github works:

1. The names are really long. "@virtualopensystems-nnikolaev" is 28 characters. Completion is only a partial help because Github doesn't always have it available and the prefix is not unique anyway.
2. This PR includes commits from multiple people but this is completely invisible on the main PR page (just see thumbnails of the vosys logo).
3. Even if you look at the detailed [commit list](https://github.com/SnabbCo/snabbswitch/pull/590/commits) it is hard to notice that there are multiple authors. The user avatar images are exactly the same (company logo) and so are the first 19 characters of the usernames.

These could probably be solved with some tweaks e.g. naming `nnikolaev-vosys` and using unique images (e.g. photo with company logo added, etc).

The ideal though from my perspective would be to have a productive discussion about how to give people and companies proper credit for the work that they do. This is an important problem and worth thinking about. I don't really know what people are concerned about and what I can do to help.

virtualopensystems-nnikolaev:
@lukego 

OK so we rename the branch to `vosys-nfv`, and our account names to `nnikolaev-virtualopensystems`. The logo we can't change for the time being.

Giving credit to people/companies should be enough to be able to see their names/logos on the cotributors page.

lukego:
@virtualopensystems-nnikolaev They are your accounts and you can do as you please. My feedback is that I find this confusing.

Issue #588, opened 2015-08-07
Dockerization of CI


lukego:
I propose that we Dockerize our test and CI infrastructure. That is, we would use Docker to make the test setup completely bootstrappable (everything done with Dockerfiles) and also instantly installable (as a versioned prebuilt binary from http://hub.docker.com/). Then that we actually execute CI tests using Dockerfiles and expand the set of hardware and configurations that we test.

This is following on from related threads:

1. #574 troubleshooting of an NFV test scenario.
2. [Docker for the lab?](https://groups.google.com/forum/#!topic/snabb-devel/zSWRH3Zq5Mo) on snabb-devel.
3. [Nix, NixOS, NixOps](https://groups.google.com/forum/#!topic/snabb-devel/8BTftZL0oMo) on snabb-devel.

Everything that we have done by hand would be automated. It would become taboo to rely on instructions in README files and manually created binaries: our new religion would be automation and reproducibility.

This requires work and especially for the NFV application:

1. Create a `Dockerfile` that builds a complete NFV guest operating system. For example with a specific Linux kernel version, kernel configuration, and Ubuntu version. (This would be generalized to support more kernels, kernel configurations, distros, and operating systems.) The build results (e.g. `bzImage` and `disk.img`) would be stored read-only inside the container and could be accessed externally from other containers via Docker "volumes". The container with completed build would be uploaded to http://hub.docker.com/ as free storage of the image files with proper naming and versioning.
2. Change `test_env` to access image files via a Docker "volume" (that the user can override e.g. for Ubuntu vs FreeBSD guest) and to use the system-installed `qemu` (etc) instead of manually building its own.
3. Create a `Dockerfile` to run the CI test suite (or a well-defined subset thereof). The "build" phase of this container would install all necessary software including building a specific version of QEMU. The "run" phase would checkout Snabb Switch (version provided as env/arg) and execute the tests and post the results to Github. See [draft Dockerfile](https://gist.github.com/lukego/850a037c7d9ecdf594af).

@eugeneia you might be interested in taking the lead on this? (or on shooting the idea down :-))

kbara:
```Everything that we have done by hand would be automated. It would become taboo to rely on instructions in README files and manually created binaries: our new religion would be automation and reproducibility.```
I would love to see this.

eugeneia:
@lukego Yup already started working on this, see: https://github.com/eugeneia/snabbswitch-docker

eugeneia:
[...]

PR #593, opened 2015-08-12
next: Collected changes for the v2015.09 release

lukego:
This Pull Request contains the changes that are being tested for inclusion in the v2015.09 branch.


lukego:
Looks like the test failure here is an operational issue (hugepage availability on the CI host). That is no biggie: the next merge will trigger a rerun anyway.

Does raise a couple of interesting thoughts though:

1. How come the hugepage is failing to allocate? In the log we see that Snabb Switch makes three attempts to provision a new huge page in the kernel and then allocate that. I wonder why those attempts are all failing and whether we could improve that algorithm to make it more robust.
2. I see that it would be valuable to run the CI on multiple hosts so that CI operational issues are easier to distinguish. If the CI result were that tests passed on 4/5 CI servers then it would be easier to estimate confidence in the code before digging into the root cause. Once more I see docker (Issue #588) as the great hope to make it easy to operate a CI fleet.

eugeneia:
@lukego I take it we're ready for release here? If so, I will go ahead and do the routine.

lukego:
Roger!

lukego:
Humble request: Could you put the release name in the first line of the commit message when you merge onto master? Just so that it is immediately visible in one-line summaries e.g. `git log --oneline`.

For example

    Merged PR #593 (v2015.09 release) onto master

PR #599, opened 2015-08-20
Add TCP and IPv4 checksum calculation tests

dpino:
It took me a while to figure out how to use the checksum methods in ```lib.protocol.ipv4``` and ```lib.protocol.tcp``` to calculate the checksum of a packet. I decided to add two selftest functions to verify the checksum calculation is correct. The main purpose is to show how to use the ```checksum``` functions.

### DIFF
--- a/src/lib/protocol/ipv4.lua
+++ b/src/lib/protocol/ipv4.lua
@@ -196,10 +196,28 @@ function ipv4:pseudo_header (ulplen, proto)
    return(ph)
 end
 
+local function test_ipv4_checksum ()
+   local IP_BASE      = 14
+   local IP_HDR_SIZE  = 20
+
+   local p = packet.from_string(lib.hexundump([[
+      52:54:00:02:02:02 52:54:00:01:01:01 08 00 45 00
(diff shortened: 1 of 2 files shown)


lukego:
@alexandergall whaddayareckon?

aside: I reflect that one day it would be neat to have a notation for generating synthetic traffic. (Or maybe just a really massively mixed pcap file that we could filter to extract the kind of packets we care about.)

alexandergall:
Well, tests are good :)  Note that I didn't write any of those modules and I never really looked at that code (though the ipv4 and tcp checksum and pseudo-header code looks like it has been taken over from the ipv6 and icmp modules, which I did write). Maybe the `checksum` methods should be documented better?

The `datagram` module can be used to create synthetic packets without having to use hexdumps.  The API is a bit awkward if one wants to relocate a pre-allocated header into a packet buffer (e.g. to update the checksum in a packet), but that can be done as well.

nnikolaev-virtualopensystems:
I did wrote those and I can confirm that they were based on what was already available in`ipv6` and `icmp` (why reinvent the wheel anyway).
I can also confirm the tests are OK to me.

PR #601, opened 2015-08-22
Inter proc: multiprocessing capabilities, take 2

javierguerragiraldez:
Replaces #565 with just the minimal changes:
- [x] `core.memory`: make packet pointers valid system-wide
- [x] `program/gc`: unmap hugepages.
- [x] `apps.inter_proc`: the transmit/receive apps
- [x] `program/spoon`: sample test program
- [x] document everything!

Everything else (OOP-syntax for links and packets, freelists as FFI structs (with clever type wrapping!)) is optional, so will come in their own PR.


src/core/memory.c
```diff
@@ -104,3 +123,47 @@ void *allocate_huge_page(int size)
...
+void allocate_on_sigsegv(int sig, siginfo_t *si, void *unused)
+{
+  uint64_t address = (uint64_t)si->si_addr;
+  if ((address & 0x500000000000ULL) != 0x500000000000ULL) {
+    // This is not DMA memory: die.
+    //exit(139);
+  } else {
+    uint64_t physaddr = address  & ~0x500000000000ULL;
+    uint64_t physpage = physaddr & ~(2*1024*1024-1);
+    uint64_t virtpage = address  & ~(2*1024*1024-1);
```
lukego: There is a bug here inherited from my hacky demand mapping:

Huge page size is hard-coded to 2MB but this is not safe and we should use the value detected in `memory.lua`.

src/program/spoon/README.md
```diff
@@ -0,0 +1,41 @@
+## Inter-process sample program
+
+The `program/spoon` directory contains a sample (and test, and benchmark)
+for the inter_proc app.
```
lukego: Include explanation of the name "spoon"? (I suspect there is a pun that I am missing completely?)

src/program/gc/gc.lua
```diff
@@ -2,8 +2,9 @@ module(..., package.seeall)
 
 local lib = require("core.lib")
 local shm = require("core.shm")
+local memory = require("core.memory")
 local syscall = require("syscall")
-local usage = require("program.gc.README_inc")
+local usage = 'help' --require("program.gc.README_inc")
```
lukego: Accidental change?

src/program/gc/gc.lua
```diff
@@ -18,10 +19,14 @@ function run (args)
...
    -- Unlink stale snabb resources.
    for _, pid in ipairs(shm.children("//")) do
-      if not syscall.kill(tonumber(pid), 0) then
+      if tonumber(pid) and not syscall.kill(tonumber(pid), 0) then
          shm.unlink("//"..pid)
       end
    end
+
+   -- unmap hugepages
+   memory.cleanup()
```
lukego: Is this safe?

I expect the `gc` program to only remove "stale" resources that are not in use by any running process. On casual reading though it seems like this could be disruptive? (`spoon` test script would break if there was a GC during startup?)

eugeneia:
This is super cool! I love how simple the interface is and can't wait to find a good reason to experiment with this. Its also a very small patch considering how much it provides in terms of functionality. Awesome!

Regarding the documentation:

* I would prefer if the implementation details are not mentioned in the user manual. Either move it into source code comments or a separate file? The hints regarding busyloop can be a hint, e.g. "Note: you should enable busyloop to gain optimal performance and latency. See. *imp notes*".
* It might be worth to explicitly state that inter-proc link names are global (within the system). Might seem obvious but these are the kind of statements which sync with the reader. E.g. Reading the docs a user might think "These names are probably system wide?" and then get explicit confirmation and be like "OK. That's what I thought".
* Run a spell checker. Always helps me catch typos that I'd miss otherwise.

src/program/gc/gc.lua
javierguerragiraldez: The `memory.cleanup()` marks for deletion all the hugepages registered in the `map_ids` structure, but that only means they will be deleted when the last process to use a given hugepage unmaps it (or finishes).  While a hugepage is used by a process, it can be used by another process.  I tried to make it fail by marking at several times, but it always worked.

I'm even tempted to restore the `shmctl()` call just after allocation, as it was before, making this `cleanup()` function unnecessary.  I guess it would only fail if:

- process A allocates some packets from its pool
- those packets are transferred to process B
- process A finishes
- since no other process has mapped A pool, it's destroyed
- process B receives the packet pointers and tries to map it.

It not only seems very unlikely, but also useless.... can anybody come up with a different scenario that could fail because of marking pages as soon as allocated?

src/program/spoon/README.md
javierguerragiraldez: yes, it's a bad joke (not a fork and piles nicely, matching convex with concave).  i plead the [second hard thing](https://twitter.com/codinghorror/status/506010907021828096).

lukego:
Sorry about the slow response on this. Newborn baby. I am really looking forward to thinking this lovely little diff through more carefully and merging the code.

This change also stimulates a lot of ideas for me e.g. related to memory allocation and whether we could make that code simpler (must be too hairy if even you are not comfortable with it :)). However, this is stuff to experiment with in the future and not to hold up getting this code merged.

src/core/memory.c
```diff
@@ -104,3 +123,48 @@ void *allocate_huge_page(int size)
   return NULL;
 }
 
+// SIGSEGV handler: attempts map packet memory on demand
+void allocate_on_sigsegv(int sig, siginfo_t *si, void *unused)
+{
+  uint64_t address = (uint64_t)si->si_addr;
+  if ((address & 0x500000000000ULL) != 0x500000000000ULL) {
+    // This is not DMA memory: die.
+    //exit(139);
```
lukego: This line should probably be uncommented right? I mean: if a genuine segfault occurs then we want the process to exit with status 139.

Currently it looks like the segfault behavior has turned into an infinite loop:

```
$ sudo ./snabb snsh -e 'require("ffi").cast("int*", 0)[0] = 0'
^C
```

Sorry about the slow feedback. I think this change is in good shape. Just want to let my subconscious ponder a little bit if there are any subtle issues that we may have missed particularly regarding the memory sharing. (Can also debug these after a merge to next of course.)

lukego: (I can also just make any fixes like this during merge. This is in code that I wrote anyway.)

lukego:
Sorry to leave this unmerged for so long. I meant to merge it at the start of this month but that didn't happen.

I have a couple of questions niggling me a bit:

1. Is catching SIGSEGV this way going to cause problems? I do think we have at least one bug where a genuine segfault will turn into an infinite loop instead of terminating the process.
2. The `snabb --help` output is becoming less helpful for end-users every time we add a new program that is really just an internal test case or example program. I would like to avoid this problem and make `snabb --help` only offer to run stuff that is actually useful to users. However, this should probably be taken as a separate issue rather than blocking this PR.

I will try to merge this at the start of the next release cycle. It would be very helpful if somebody wants to review the segfault handling code and make sure it is not introducing overly exciting new failure modes.

lukego:
The program name `spoon` also bothers me because it sounds intriguing and people will probably be tempted to dig in and see what the program does, only to be disappointed that it is an internal test case :).

lukego:
... also I am wondering more broadly about the whole `core.shm` mechanism. This has been a useful abstraction for exploring different multiprocessing models but all of the remaining use cases we have for it actually seem fairly awkward to me. I have been meaning to look into whether we could delete that whole concept and cover the few remaining use cases in a simpler way.

That is another whole can of worms though and should be taken separately..

eugeneia:
@lukego thinks this patch should be merged into a project branch that uses and battle tests this functionality before we merge it into mainline. So to anybody reading this: Go ahead and use this code and let us know how it went. We think its good, but we can't guarantee it. :+1:

kbara:
@eugeneia I agree with @lukego . Battle testing is planned. :-)

lukego:
I am also planning to use this "in anger" for 100G support in the immediate future. More on that to follow. (This was the reason that Javier and I decided to pursue the multiprocessing support last summer - to prepare a clear path for applications running 100G+ of traffic.)

src/core/memory.c
xrme: I believe that the correct way to exit here would be to re-signal the SIGSEGV.  Something like
```
signal(SIGSEGV, SIG_DFL);
return kill(getpid(), SIGSEGV);
```

See http://www.cons.org/cracauer/sigint.html for a detailed explanation.

PR #602, opened 2015-08-22
Fix ipv4 header initialization.

tlaberge:
The version/header length/TOS field is 16 not 32 bits wide.
Also add a self-test.


lukego:
Looks good to me. Thanks!

PR #603, opened 2015-08-23
snabbmark: Add first step towards porting basic1 to asm [sketch]

lukego:
Super early sketch of the code! shared just for fun.

The idea is simple: clone the 'basic1' benchmark in assembler to see what performance is possible. Then we could use this as a benchmark for deciding how optimized our Lua code is. Could be that in writing the assembler code we get some bright ideas for improving the Lua code too.

This is very rough code that implements:

    source -> sink

and recycles packets via a freelist.

The full implementation would have to implement the real basic1 topology i.e. include a "Tee" app that duplicates packets.

This is not optimized, etc, so don't draw any sweeping conclusions yet :-).

Note also that I did most of my assembler programming 20+ years ago on the MC68000 so it will take a little practice to find a decent style :).

### DIFF
--- a/src/program/snabbmark/snabbmark_asm.dasl
+++ b/src/program/snabbmark/snabbmark_asm.dasl
@@ -24,7 +24,9 @@ local function ptr (x) return ffi.cast("void *", x) end
 -- source -> sink
 
 local fl   = ffi.new("struct ring")
-local source_to_sink = ffi.new("struct ring")
+local source_to_tee = ffi.new("struct ring")
+local tee1_to_sink  = ffi.new("struct ring")
+local tee2_to_sink  = ffi.new("struct ring")
 
 -- Populate the freelist
 for i = 0, 255 do
@@ -33,15 +35,21 @@ for i = 0, 255 do
    fl.write = i
 end
 
--- Setup fixed-purpose registers
+  -- Setup fixed-purpose registers
 |  push r15
 |  push r14
 |  push r13
 |  push r12
+|  push r11
+|  push r10
+|  push r9
 |.type freelist, struct ring,   r15
 |.type input,    struct ring,   r14
+|.type input2,   struct ring,   r9
 |.type output,   struct ring,   r13
+|.type output2,  struct ring,   r11
 |.type packet,   struct packet, r12
+|.type packet2,  struct packet, r10
 
 |  mov64 freelist, ptr(fl)
 
@@ -51,17 +59,8 @@ end
 -- Source: allocate from freelist and write to output
 -- ----------------------------------------
 
-|  mov64 output, ptr(source_to_sink)
+|  mov64 output, ptr(source_to_tee)
 |->source:
--- Load number of freelist elements into rax.
--- (Handle wrap around.)
-|  mov rdx, 255
-|  xor rcx, rcx
-|  mov rax, freelist->write
-|  sub rax, freelist->read
-|  cmovs rcx, rdx
-|  add rax, rcx
-|
 -- Load number of input packets into rdx (like above).
 |  mov rdx, 255
 |  xor rcx, rcx
@@ -69,11 +68,12 @@ end
 |  sub rdx, input->read
 |  cmovs rcx, rdx
 |  add rdx, rcx
-|
--- Load max packets to process into r8 (min of the values above)
-|  mov r8, rax
-|  cmp r8, rdx
-|  cmovs r8, rdx
+
+-- (Assume the freelist has enough packets available).
+
+-- Max burst: 100.
+-- (XXX This is actually important to keep the freelist from running empty.)
+|  mov r8, 100
 |
 -- Transfer packets
 |  xor rcx, rcx
@@ -83,7 +83,7 @@ end
 |  mov rax, freelist->read
 |  add rax, rcx
 |  and rax, 255
-|  mov packet, [freelist + rax * 8] -- XXX better syntax to index freelist->packets?
+|  mov packet, [freelist + rax * 8]
 
 -- Initialize packet
 |  mov word packet->length, 64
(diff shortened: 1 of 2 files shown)


lukego:
Currently the performance I see with `perf` is 11 cycles per packet to run a `Source` and `Sink` app and recycle packets between them via freelist.

I have been reading [CSAPP3](http://www.amazon.com/Computer-Systems-Programmers-Perspective-Edition/dp/013409266X) and so I can see some obviously bad patterns in my coding style like this series of instructions where each depends on the result of the previous one and probably sucks for [ILP](https://en.wikipedia.org/wiki/Instruction-level_parallelism):

```
-- Write onto output link                                                                                                             
|  mov rax, output->write
|  add rax, rcx
|  and rax, 255
|  mov [output + rax * 8], packet
```

CPUs have changed a lot since the last time I wrote assembler code in anger (Amiga 500...) and so for now I am giving myself permission to suck :-).

lukego:
... I will try to implement the Tee app to make the program equivalent to `basic1` before getting carried away with tweaking the instructions.

lukego:
See also [The Unofficial Dynasm Documentation](http://corsix.github.io/dynasm-doc/) that I only discovered recently. Really great! Note that this is based on the [standard C-based dynasm](http://luajit.org/dynasm.html) while Snabb Switch is now using the [extended Lua-based dynasm](https://github.com/luapower/dynasm/blob/master/dynasm.md).

lukego:
The code is still extremely rough but I added a Tee app so that it implements the same topology as `snabbmark basic1`.

Each packet is allocated by a Source from a freelist and lightly initialized, transmitted to a Tee which makes a full copy and transmits on two output links, on to a Sink that puts the packets back onto the freelist. This is roughly equivalent to the work done by the `snabbmark basic1` benchmark. I have cut some corners but hopefully not important ones.

Current performance is around 23 cycles per packet. On Interlaken this is 107 Mpps and that is about 5x faster than `snabbmark basic1`. I suspect that both the asm code and the Lua code could be optimized substantially.

Note: The reason we have not dived into optimizing these basic operations in the past is that in real applications the performance bottlenecks have been elsewhere i.e. inside the main loops of the apps and not in the machinery that connects the apps together.

So why is this interesting at all? Well: I do think that sooner or later we will be in the situation of wanting to squeeze maximum performance out of the app network. There are already some apps that bypass the app network for performance (packetblaster and firehose) and this may actually turn out to be unnecessary. Likely we will also want to write an app that dispatches a 100G traffic stream across different links in an app network and it would be valuable to understand what the maximum performance we could expect from that would be (and maybe this is an app that we really would want to write in JIT'd assembler).

I also have a basic feeling that our machinery link `link.transmit` is not sufficiently minimalist but I have not been able to quantify this either by measuring performance or by showing all of the instructions. However, a 5x potential performance gain would make it easier to motivate work like #570 and speeding up the basic1 benchmark.

Here is a cute feature of the code: copying a 64-byte packet using SIMD. This version uses 128-bit registers:

```
|  movdqu xmm0, [packet];    movdqu [packet2], xmm0
|  movdqu xmm0, [packet+16]; movdqu [packet2+16], xmm0
|  movdqu xmm0, [packet+32]; movdqu [packet2+32], xmm0
|  movdqu xmm0, [packet+48]; movdqu [packet2+48], xmm0
```

and on AVX2 (Haswell) you would only need half as many instructions (pardon my fudged opcodes):

```
|  mov.qu ymm0, [packet];    movdqu [packet2], ymm0
|  mov.qu ymm0, [packet+32]; movdqu [packet2+32], ymm0
```

and on AVX512 (Skylake) you can actually fit a whole 64-byte packet in a register:

```
|  mov.qu zmm0, [packet];    mov.qu [packet2], zmm0
```

Funny to think about "packet copies" that will take only 2 instructions and execute in 1 cycle! (Of course not all packets are 64-bytes but small packets are usually where performance hurts.)

lukego:
Oh and if you want to actually run it then you can do this:

    sudo perf stat ./snabb snsh -l program.snabbnfv.snabbmark_asm

and it will print the link statistics and CPU performance counters after running 1 billion packets from the source through the app network:

```
tee1.read       1000000000ULL   tee1.write      1000000000ULL
tee2.read       1000000000ULL   tee2.write      1000000000ULL
fl  .read       2000000000ULL   fl  .write      2000000255ULL
s2t .read       0ULL    s2t .write      1000000000ULL
done

 Performance counter stats for './snabb snsh -l program.snabbmark.snabbmark_asm':

    64,488,027,290 instructions              #    2.90  insns per cycle
    22,243,300,039 cycles

       9.276217798 seconds time elapsed
```

lukego:
Just having a little more fun here: I ported the assembler code back to Lua with the same simplifications (i.e. no real breathe loop, etc). The notion is that the exercise of writing the code in assembler can help to write an efficient Lua version due to fresh [mechanical sympathy](https://dzone.com/articles/mechanical-sympathy).

Here is how it runs:

| version | cycles/packet | cycles/step |
|---|---|---|
|asm|23|4.6|
|asmlua|51|10.2|
|basic1|116|23|

So more-or-less the Lua port of the asm code is twice as fast as the original basic1 benchmark and then the actual asm code is twice as fast again.

I added "per step" values that take the average number of cycles for each of the five "steps" in processing a packet. The app network looks like this:

```
                  .------.                                                                                                          
    Source ---> Tee      Sink                                                                                                       
                  `------'                                                                                                          
```

and so the five processing "steps" are:

1. Source app allocates and transmits a packet.
2. Tee app receives the packet and transmits a duplicate.
3. Tee app transmits the original.
4. Sink app receives the duplicate and frees.
5. Sink app receives the original and frees.

4.6 cycles per step must be really approaching the limit of what is possible, particularly since one of the processing steps is making a copy of the packet (albeit with only 64 bytes of payload).

Overall the impression I am forming is that:

1. The app network abstraction is fundamentally pretty cheap.
2. The implementation we have can possibly be tuned to double or quadruple speed.
3. We should remember this if/when we have an app that is limited by basic app network operations e.g. an app that splits up a 100Gbps traffic stream and has to perform `link.transmit()` at a ferocious rate.

javierguerragiraldez:
Nice way to get big, impressive numbers :smile:.

One thing that does stand out is that there's not a single `if..then` in the whole Lua file!  Of course, as the comment says, this code assumes that links never overflow nor underflow.

As comparison, the current SnS can get over 50Mpps when doing a simple Source/Sink network.  Of course, that's much less steps than this demo, but it includes a fully working link, a sizeable freelist and copying of a 120-byte packet (yes, the same one over and over, so it's probably in cache all the time).

I don't see the 'original' code that's being compared to the `asm` and `lua-asm`, but I guess it does two packet clones (one at the source and one at the tee), three link transmissions (source->tee and 2*tee->sink) and two allocations/frees (the original and the copy) for each source->tee packet.  That is, slightly more than twice the steps, which correlates neatly with the 27Mpps you mention.

lukego:
> Nice way to get big, impressive numbers

This is a problem and it could be that we need to be more disciplined in our use of metrics.

Generally the term "Mpps" can refer to any code that is operating on packets but we might want to consider reserving this for benchmarks that are actually performing I/O. Then we would need to find a different metric for tests like this one and also the multiprocessor tests that are moving packets between rings in memory rather than physical ethernet links.

I also think we should move towards metrics based on *cycles* rather than *seconds* for synthetic benchmarks. Most benchmark scores are directly proportional to CPU frequency and so if you tested with (say) both a 2GHz and 4GHz then you should expect the same cycles/packet on both but a factor of 2 difference in Mpps. To me this says that cycles/packet is the more useful metric for talking about the performance of a given bit of code.

This is the thinking also that lead me to `cycles/step` in the description above. That is supposed to tell the average cost of processing a packet *per app* and be somewhat independent of how complex the app network is.

I actually have a fantasy of making a benchmarking environment that will take any app and generate a "data sheet" that says how it performs under various configurations and workloads and also how dependent it is on cache behavior e.g. test separately with packets that fit in L2 cache, in L3 cache, or in DRAM. Just a notion at this stage though.

lukego:
> I don't see the 'original' code that's being compared

The reference code is running `snabb snabbmark 1e9`:
```
```

This is a standard benchmark of the app network with simple Source/Sink/Tee apps. SnabbBot CI is running this benchmark on every PR and will report a test failure if the score regresses. This means that when we optimize the app network we should see an improvement in the snabbmark score and the CI will defend this improvement in the future.

Ideally I would like to have performance regression tests in place for all applications and only merge optimizations that can be shown to improve the benchmark scores. This is the optimization equivalent to requiring bug fixes to be submitted with a test case to show that it really works and to ensure that it will keep working in the future.

lukego:

PR #606, opened 2015-08-30
SnabbSwitch build as dynamic library

pavel-odintsov:
Hello, folks!

I have implemented ability to build Snabb Switch as dynamic library, please review this pull request and offer some feedback. 

Why I need this? I want to use SnabbSwitch as traffic capture backend for my C/C++ application because it's really fast. So with current implementation as stand alone binary I could not integrate it. Actually I want to use SnabbSwitch in similar to Netmap/PF_RING way.

So I decided to build SnabbSwitch as dynamic library and link it with my app.

For tests I have used Firehose app.

I built .so for callback code:

``` C
gcc -O3 -shared capturecallback.c -o capturecallback.so -fPIC
-o/root/capturecallback.so
```

And build test app which linked against SnabbSwitch:

``` C
gcc test_app_linked_to_libsnabb.c -Lsrc -lsnabb
```

So finally I could run it:`LD_LIBRARY_PATH=src ./a.out`

And get output like this:

``` bash
Loading shared object: /root/capturecallback.so
Initializing NIC: 0000:02:00.0
Initializing NIC: 0000:02:00.1
Initializing callback library
Processing traffic...
We process: 153059 pps
We process: 572443 pps
We process: 564516 pps
```

I will be very glad if you could help me with integrating this code to
Snabb Switch master. Because I haven't idea how we could build Snabb binary and libsnabb.so with same Makefile without breaking anything.


lukego:
Howdy!

Sorry for the slow response.

So do I understand correctly that the two main changes that need to be integrated here are:

1. Makefile builds `libsnabb.so` instead of `snabb`. (Should have option to build either/both?)
2. Firehose expects to find the callback functions directly without loading a separate shared library (need to support both modes or just this one?)

I also had another thought: would it be interesting to use snabbswitch as a separate program instead of a .so? I mean so that you could run `firehose` to initialize a NIC and put the DMA buffers in a shared memory file, then fastnetmon could separately use a small library to map that file and access the packets. Could simplify matters potentially, or?

pavel-odintsov:
Hello, Luke!

1. There are some additional fixes about -fPIC option. We could not build both binaries together because .so need -fPIC but app do not need it. 
2. Will be fine to support both. If user specified path to .so we use it. Instead we try to find it in global namespace.

Option with mapping file is nic but my case is more flexible in case when I want multiple snabb instances for multiple NIC's.

lukego:
Howdy!

So wondering what the most expedient way is to support this on the master branch.

Could use e.g. `make EXTRACFLAGS=-fPIC` and a separate target to build the shared library?

I am also tempted to make it `libfirehose.so` instead of `libsnabb.so`. The reason being that this is still quite a rough first cut at integration. If we wanted to be a really neat shared library then I suspect we should change a bunch of things (more conservative use of the C symbol namespace, document/rethink any potential conflicts e.g. calls to library functions that may be non-reentrant, etc). Right?

pavel-odintsov:
Hello!

Will be fine to get libfirehose.so too. It will be PoC for this idea. So if it success we could move forward.

lukego:
Can you explain more definitely what support you need here Pavel? (Do you actually need this upstream for your PoC? what is your plan?)

pavel-odintsov:
I want to use SnabbSwitch driver from C. It's my main idea. So I need some way to build this .so file from upstream SnabbSwitch repository.

lukego:
So! This is actually a good test case for the Snabb Switch [git workflow](https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/git-workflow.md). Bear with me here please :-)

Just now you seem to be stuck waiting to synchronize with upstream but actually that is not supposed to happen. Instead application developers should be able to hack up exactly what they need, ship their application, and then sync with upstream as a background process over time. This should be good for everybody because it means that application developers don't get blocked and also code landing upstream has already had some battle-testing and is not so experimental.

WIth me so far? :-)

So my suggestion would be to create a branch called e.g. `firehose_fastnetmon` and add a patch that builds `libfirehose_fastnetmon.so` instead of `snabb`. Then you can use this to build and ship your application from. Then you are self-sufficient and have complete control over what you ship to your users.

If you want to stay in sync with upstream then you only have to `git merge master` once per month after a new release is made. If you want to give visibility into your changes you can also publish your branch in [branches.md](https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/branches.md) where the rest of us can easily see them (everybody would get your branch when they clone the upstream repo).

Over time we can work out what is the best integration with upstream. The answer to this could change in the coming weeks/months e.g. we have #589 landing soon that makes it possible to transparently share all DMA memory between any processes on the machine without having to link them together in any way. That could make it possible to have the same tight integration between fastnetmon and firehose but without any object code linkage. That would potentially be a lot easier to maintain.

I feel like the optimal solution to linking processes together is emerging now but we don't have the complete picture yet.

What do you think? Crazy talk? :-)

PR #608, opened 2015-09-04
abcJIT: a short-lived fork of LuaJIT for Snabb Switch

lukego:
Here is an idea for some good clean fun :-)

LuaJIT is transitioning from being maintained by Mike Pall to being
maintained by the community. This is fun and exciting. It is also a
process that will take a while to find a good workflow. A month? A
year? I don't know: but I see an opportunity for us to have fun and
learn new things during the process.

This PR introduces abcJIT: a short-lived Snabb-specific fork of
LuaJIT. I propose that we maintain abcJIT as part of the Snabb Switch
repository, using the normal Snabb Switch Git/Github workflow, while
we wait for the new LuaJIT upstream to establish itself.

I also propose that we develop abcJIT with a mix of conservatism and
radicalism. Conservatism in the sense of avoiding changes that could
lead to instabilities in Snabb Switch e.g. complex new
optimizations. Radicalism in the sense of stripping the code base down
to the bare essentials that we are actually using at the moment.

For example I would like to strip out all `#ifdef` statements from the
code. We only need to support x84-64 (not 32-bit, not ARM, not PPC,
etc) and Linux (not Windows, not Xbox, not Android, etc). I would like
to pick the most suitable NUMMODE and delete the rest. I would like to
keep the features that we want to have enabled and delete the
rest. Potentially we could even remove optimizations that are not
important to Snabb Switch e.g. anything specific to the Lua C API.

Why? I expect that we will learn a lot about LuaJIT and maybe come up
with some nifty ideas during this process. Then we can take that
experience and apply it to the new "upstream" LuaJIT when that is ready.

The key is to understand that this is a short-lived fork. That means
that we can be absolutely ruthless about removing code because really
we are going to get it all back in the future anyway.

What do we think?

Here is the README that I wrote:

    abcJIT: a JIT for Snabb Switch (descended from LuaJIT)

    abcJIT is a descendent of LuaJIT that is targeted at Snabb Switch. The
    development philosophy of "what would Mike do?" is replaced with "what
    would be best for Snabb Switch?".

    abcJIT is a response of sorts to Mike Pall's call for the community to
    take charge of the future development of LuaJIT:
    http://www.freelists.org/post/luajit/Looking-for-new-LuaJIT-maintainers

    abcJIT moves in a new direction:
    - Ruthlessly eliminate code that Snabb Switch does not use.
    - Helpfully write and review comments to explain the code.
    - Experimentally add features that make Snabb Switch easier to develop.

    abcJIT benefits Snabb Switch hackers:
    - Learn about JIT internals by making simple changes.
    - Optimize more effectively thanks to better community expertise.
    - Defend the self-sufficiency that is one of the core Snabb values.

    abcJIT benefits LuaJIT hackers:
    - Grow the pool of people with maintenance experience.
    - Explore design ideas in the post-"what would Mike do?"-era.
    - Cooperate with maintainers of other trees to share code and ideas.

    abcJIT exists only while it is beneficial to Snabb Switch. In all
    likelihood it will be merged into another more mainstream LuaJIT tree
    in the future. However, we would keep the valuable experience of
    having taken charge of the code we depend on, even if only for a while.

    The Snabb Switch community forked abcJIT from LuaJIT v2.1 in September 2015.

P.S. The name "abcJIT" is from the Jackson 5 song "A-B-C as easy as
1-2-3". The whole idea is to make the software simpler by deleting as
much code and as many concepts as possible.

P.P.S. The mechanism I used to import LuaJIT into our source tree is
[git subtree](http://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/)
with `--squash` to compress the history.


kbara:
I think gaining experience with LuaJIT internals is vital at this point. I'm happy with stripping out as much as we can as an experiment and to learn, but I do hope that whichever fork of LuaJIT predominates in a few years is still multi-platform, etc.

lukego:
@kbara Agree. Really the problem we want to solve is to increase our mastery of and contribution to LuaJIT and other dependencies. Stripping out features is only one idea to that end and potentially a really dumb one, particularly since anybody could want to port snabbswitch to ARM or make a Windows port of SnabbNFV for HyperV instead of KVM.

#609 is a more low-key alternative formulation of this idea. (I close this PR now in favor of that one.)

PR #609, opened 2015-09-05
deps: Use git subtree instead of submodule

lukego:
Switch from `git submodule` to [`git subtree`](https://github.com/git/git/blob/master/contrib/subtree/git-subtree.txt) (with squashed history) for tracking dependencies.

Git subtrees are very similar to submodules. The main difference is that the subtree stores the complete sources for the dependencies (luajit, ljsyscall, pflua) in our repository instead of linking to a separate repository. With the code imported into our repository we can work on it exactly like the rest of our code i.e. you could send a PR that includes changes to Snabb Switch, LuaJIT, pflua, and ljsyscall. Git subtree then provides commands for pulling updates from the upstream projects and also extracting our own changes for submission.

The main practical implication is that now if you have an idea for a change to LuaJIT (or another dependency) then you can PR it to `snabbswitch` to get feedback and testing from other people in our community. This did not really work with submodules. I hope this will make it easier to experiment with changes to the other projects that we depend on and start to think of them as read/write rather than read-only.

This change does not imply any particular policy for if/when/how we merge changes to LuaJIT/pflua/ljsyscall into the Snabb Switch repo. Changes submitted on PRs do not have to be merged directly into our repository and could instead be tested and reviewed prior to submission to upstream.

This is a less extreme alternative to #608.


justincormack:
Subtrees are much more useful in this sort of project and have all sorts of advantages (like all the code being there on checkout).

lukego:
@justincormack Sounds good. Thanks for the words of experience.

@eugeneia @SnabbBot not testing?

lukego:
Superceded by #617.

PR #617, opened 2015-09-14
deps: Use git subtree instead of submodule (v2)

lukego:
New version of #609. Move subtree dependencies (LuaJIT, pflua, ljsyscall) into `lib/` instead of `deps/`. Using the same names as the submodules seemed to cause confusing conflicts when switching between branches and I would like to spare people the fight with git over that.


lukego:
The version of LuaJIT that I imported is the vanilla upstream v2.1-beta1. This is missing some features that we have added on the Snabb version of LuaJIT. I reintroduce these features in #619, #620, and #621.

PR #623, opened 2015-09-23
traceprof: a new trace-oriented profiler for LuaJIT

lukego:
This is a new profiler designed to help understand LuaJIT trace behavior and performance.

You can test it with `snabb snsh -jtprof ...` combined with `-p` (run a program) or `-t` (run a selftest) or `-e` (run a Lua expression) or directly on a script file.

This is similar to our enhanced `-jp=v` support (#621) but seems to give more accurate results. Specifically when "back testing" on lukego/blog#8 the information from traceprof is consistent with the analysis:

```
traceprof report (recorded 10583/10583 samples):
 62% TRACE   3               counter2.lua:5
 26% TRACE   4      (3/5)    counter2.lua:9
 10% TRACE   3:LOOP          counter2.lua:5
```

while the `jit.p` result is completely misleading:

```
100%  TRACE   3          ->loop     counter2.lua:5
```

which seems to be due to limitations with the method that `jit.p` uses to collect samples. I don't fully understand this method and after studying it for a while I decided to try writing a simpler one instead.

(With traceprof I did have to increase the number of iterations on that test case to 10e9 to see the last trace reported. That is surprising but could be consistent with CPU out-of-order execution. Even with this many iterations I only see one trace reported from `jit.p`.)

More from the comments in `traceprof.lua`:

### traceprof.lua: Low-level trace profiler

Traceprof analyzes the time spent in JIT-compiled traces.  It is an
alternative to 'jit.p' with fewer features and based on a simpler
and (hopefully) more accurate sampling method.

API:

- `start()`: Start profiling.
- `stop()`:  Stop profiling and print a report.

... and `start()` has some undocumented optional parameters too.

Here is an example report:

    traceprof report (recorded 659/659 samples):
     50% TRACE  20      (14/4)   rate_limiter.lua:82
     13% TRACE  12:LOOP          basic_apps.lua:26
     10% TRACE  14               rate_limiter.lua:73
      3% TRACE  14:LOOP          rate_limiter.lua:73
      2% TRACE  18:LOOP          basic_apps.lua:82
      1% TRACE  22      (12/13)  basic_apps.lua:25
      1% TRACE  25      (20/5)   link.lua:70

The report includes some useful information:

* Which traces are hotspots? (Cross reference with -jdump)
* Where does each trace begin? (Source code line)
* How to traces connect? Side traces show (PARENT/EXIT).
* How much time is spent in the "LOOP" part of a trace vs outside?

Traceprof uses an interval timer to periodically write the CPU
Instruction Pointer value to a log (array). The timer fires every 1
millisecond (default) and invokes a tiny C signal handler to write
the next value.

This log is analyzed after measurement in a separate reporting
step. The logged Instruction Pointer values are compared with the
machine code addresses of all compiled traces (and their loop offsets).

Traceprof was originally written due to confusion about
interpreting the results of 'jit.p' and not understanding exactly
how its sampling method works.

Future work:

* Handle JIT "flush" event when existing traces are dropped.
* Dump annotated IR/mcode for hot traces (like -jdump).
* Fix warning introduced in LuaJIT change.

cc @alexandergall @kbara

### DIFF
--- a/lib/luajit/src/lib_jit.c
+++ b/lib/luajit/src/lib_jit.c
@@ -299,6 +299,9 @@ LJLIB_CF(jit_util_traceinfo)
     setintfield(L, t, "nk", REF_BIAS - (int32_t)T->nk);
     setintfield(L, t, "link", T->link);
     setintfield(L, t, "nexit", T->nsnap);
+    setintfield(L, t, "szmcode", T->szmcode);
+    setintfield(L, t, "mcode", T->mcode);
+    setintfield(L, t, "mcloop", T->mcloop);
     setstrV(L, L->top++, lj_str_newz(L, jit_trlinkname[T->linktype]));
     lua_setfield(L, -2, "linktype");
     /* There are many more fields. Add them only when needed. */
--- a/src/lib/traceprof/traceprof.c
+++ b/src/lib/traceprof/traceprof.c
@@ -1,0 +1,61 @@
+// Location where the next instruction pointer value will be stored.
+
+#include <assert.h>
+#include <stdio.h>
+#define __USE_GNU
+#include <stdint.h>
+#include <stdlib.h>
+#include <sys/time.h>
+#include <signal.h>
+#include <ucontext.h>
+
+#include "traceprof.h"
+
+static int samples;
+static int logsize;
+static uint64_t *log;
+
+// Callback function to handle sigprof.
+void traceprof_cb(int sig, siginfo_t *info, void *data)
+{
+  (void)sig;
+  (void)data;
+  if (samples < logsize) {
+    uint64_t ip = (uint64_t)((ucontext_t*)data)->uc_mcontext.gregs[REG_RIP];
+    log[samples] = ip;
+  }
+  samples++;
+}
+
+void traceprof_start(uint64_t *logptr, int maxsamples, int usecs)
+{
+  // Initialize state
+  samples = 0;
+  logsize = maxsamples;
+  log = logptr;
+
+  // Setup signal handler
+  struct sigaction sa = {
+    .sa_flags = SA_RESTART|SA_SIGINFO,
+    .sa_sigaction = traceprof_cb
+  };
+  sigemptyset(&sa.sa_mask);
+  sigaction(SIGPROF, &sa, NULL);
+
+  // Start interal timer (signal source)
+  struct itimerval tm;
+  tm.it_value.tv_sec  = tm.it_interval.tv_sec  = usecs/1000000;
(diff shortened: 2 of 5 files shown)


lukego:
I pushed a fix that includes the linkage between traces now too. Report looks like this on lukego/blog#8 example:

```
traceprof report (recorded 10572/10572 samples):
 63% TRACE   3               ->loop    counter2.lua:5
 25% TRACE   4      (3/5)    ->3       counter2.lua:9
 11% TRACE   3:LOOP          ->loop    counter2.lua:5
```

I am really happy with how closely that matches the analysis that I did by hand. Could be that the next time I have this situation I can diagnose the problem by looking at `traceprof` output and without opening the JIT dump file.

lukego:
I had a somewhat [epic Twitter discussion](https://twitter.com/lukego/status/646651023587692544) with a couple of LuaJIT internals gurus about this.

Overall this has reinforced my feeling that it makes sense to maintain `traceprof`. The problem I want to solve is making the trace-level compilation behavior made by LuaJIT transparent to the application programmer. The JIT behavior to observe is very subtle and complex and in my mind this calls for a very predictable and simple tool. Traceprof is such a tool: it simply logs the instruction pointer at regular intervals and then summarizes those values in terms of traces.

The standard `jit.p` uses a more complex method of sampling: it dynamically instruments the bytecode/IR to cause trace exits that call into the profiler. The exact points of these exits are a bit subtle and it can happen that many profiler events are coalesced together because execution did not reach an exit to the profiler in a reasonable timeframe. (That seems to be the reason that even with 10 billion loop iterations `jit.p` is only seeing one of the traces in the example above.)

This may also be the reason that upstream LuaJIT `jit.p` module does not offer per-trace reports. The code we are using for this was written by Mike but I don't believe he ever actually adopted it himself. This could be because he understood that it was not necessarily well matched for the sampling method that is being used.

I do still see `jit.p` as the right solution for basic source-level profiling and especially when you want to see call trees. However, I prefer the traceprof approach for more low-level trace-oriented profiling like in #612 and lukego/blog#8.

Incidentally in the tweet stream there is also a reference to [ll_prof.py](https://github.com/v8/v8-git-mirror/blob/master/tools/ll_prof.py) which is a tool that the V8 (Google/Javascript) developers use. They are using `perftools` as the backend to record data and then writing tools to analyze the raw dumps afterwards. This is an interesting idea. For now I prefer the method of doing everything from first principles (e.g. `lib.pmu`) and within our own process (with access to all internal data structres e.g. bytecode/IR/mcode). However, maybe the trade-offs will look different in the future e.g. if we want to support many CPU architectures with different PMU hardware.

PR #624, opened 2015-09-24
virtio/net_device: don't initialize 'avail' from 'used'

lukego:
*This is a subtle change that requires more thought and testing.*

The issue is: at what position do we start processing elements from
the virtio 'avail' ring? This change starts from the initial value
declared by the virtual machine (avail->idx). That essentially means
that we assume that the ring is initially empty.

The previous behavior is to start reading the 'avail' ring from the
position of 'used->idx'. The intention here is to allow for some valid
elements to already exist on the avail ring. We assume that the
'avail' and 'used' rings have their indexes synchronized (start from
the same index) and that the avail ring elements between
used->idx..avail->idx are the elements that we have not processed yet.

Why this question at all?

Well - our Virtio-net/vhost-user implementation attempts to allow the
snabb process to restart and reattach itself to a running virtual
machine. This is not a standard feature of Virtio-net but rather extra
that we want to support. The issue is that when we reattach to a
virtual machine we want to decide which of its buffers we have already
processed and which we have not. This is important so that the VM's
receive buffers don't leak/disappear when the snabb process restarts.

The old behavior before this patch has been working will in this way.

However, now we are interoperability testing with a new VM and this
seems to be causing an error. Specifically the issue seems to be that
the guest's 'avail' and 'used' rings do _not_ start with the same
index. This means that it is too simplistic to use the used->idx field
to decide where to start processing the avail ring. And so this patch
is intended to potentially improve compatability at the expense of
breaking the restart-and-reconnect functionality.

Any thoughts @nnikolaev-virtualopensystems @javierguerragiraldez?

### DIFF
--- a/src/lib/virtio/net_device.lua
+++ b/src/lib/virtio/net_device.lua
@@ -417,9 +417,9 @@ end
 function VirtioNetDevice:set_vring_addr(idx, ring)
 
    self.virtq[idx].virtq = ring
-   self.virtq[idx].avail = tonumber(ring.used.idx)
+   self.virtq[idx].avail = tonumber(ring.avail.idx)
    self.virtq[idx].used = tonumber(ring.used.idx)
-   print(string.format("rxavail = %d rxused = %d", self.virtq[idx].avail, self.virtq[idx].used))
+   debug(string.format("rxavail = %d rxused = %d", self.virtq[idx].avail, self.virtq[idx].used))
    ring.used.flags = C.VRING_F_NO_NOTIFY
 end
 


javierguerragiraldez:
what are the cons of not applying the patch?  i guess depending on the VM guest, it could transmit garbage packets, or resend old packets.
with the patch, the con seems to be just losing packets in the event of Snabb restarting.

garbage packets are ugly, and things could crash if an uninitialized ring element has bad pointers.  resending old data is also bad news on the networks, maybe it could cause connections to be dropped.

losing packets when some critical piece of the network restarts seem a lot more manageable.

nnikolaev-virtualopensystems:
If we are not "re-connecting" we do not need those assignments at all. AFAIK these values should always be initialized with 0. Currently used.idx is 0 at start. Of course `ring.avai.idx` can be some other value (the driver will push some empty buffers in the avail ring on its Rx virtq). I am not sure this will work properly no matter if you are reconnecting or not.
Maybe if we can detect reconnection it could be something like this:

```lua
function VirtioNetDevice:set_vring_addr(idx, ring)

   self.virtq[idx].virtq = ring
  if <RECONNECT> then
   self.virtq[idx].avail = tonumber(ring.used.idx)
   self.virtq[idx].used = tonumber(ring.used.idx)
  else
   self.virtq[idx].avail = 0
   self.virtq[idx].used = 0
  end
   debug(string.format("rxavail = %d rxused = %d", self.virtq[idx].avail, self.virtq[idx].used))
   ring.used.flags = C.VRING_F_NO_NOTIFY
end
```

What is your use case? Which kind of virtio driver your are testing with?

lukego:
Thought of the morning: Could it be a problem with mapping the vring memory in the first place?

That is, rather than this guest starting from nonzero indexes on its used/avail rings, could this simply be garbage memory and not a real vring? I am testing with our QEMU v2.1 at the moment but I believe some fixes have landed in newer QEMU versions for vhost-user memory mappings. I will rebase the reconnect feature onto a newer release and see if that makes a difference.

@nnikolaev-virtualopensystems Thanks for the ideas! I am testing with an A10 vThunder virtual appliance.

@javierguerragiraldez Good thought! I think it is not acceptable to do garbage DMA. That would potentially mean transmit/receive of arbitrary bits of guest memory (same as if we would put garbage DMA descriptors with random addresses into a physical NIC). I would be less concerned about retransmitting valid packets than this personally.

Incidentally here is the error message that triggered this whole investigation:

```
Set features 0x10020000
VIRTIO_NET_F_CTRL_VQ VIRTIO_RING_F_INDIRECT_DESC
rxavail = 0 rxused = 0
rxavail = 8738 rxused = 8738
vhost_user: Connected and initialized: /root/vhost-sock/B.socket
load: time: 1.00s  fps: 56,797    fpGbps: 0.025 fpb: 0   bpp: 0    sleep: 100 us
lib/virtio/net_device.lua:357: mapping to host address failedcdata<void *>: 0x7fe091030cb8
stack traceback:
        core/main.lua:116: in function <core/main.lua:114>
        [C]: in function 'error'
        lib/virtio/net_device.lua:357: in function 'map_from_guest'
        lib/virtio/net_device.lua:129: in function 'packet_start'
        lib/virtio/virtq.lua:67: in function 'get_buffers'
        lib/virtio/net_device.lua:122: in function 'receive_packets_from_vm'
        lib/virtio/net_device.lua:108: in function 'poll_vring_receive'
        apps/vhost/vhost_user.lua:76: in function 'method'
        core/app.lua:75: in function 'with_restart'
        core/app.lua:276: in function 'breathe'
        core/app.lua:231: in function 'main'
        program/snabbnfv/traffic/traffic.lua:85: in function 'traffic'
        program/snabbnfv/traffic/traffic.lua:61: in function 'run'
        program/snabbnfv/snabbnfv.lua:15: in function 'run'
        core/main.lua:56: in function <core/main.lua:32>
        [C]: in function 'xpcall'
        core/main.lua:121: in main chunk
        [C]: at 0x0044dd60
        [C]: in function 'pcall'
        [C]: in function 'require'
```

That is an assertion failure when we can't translate the address of a vring descriptor (consistent with the address value being garbage).

lukego:
This seems to be a QEMU issue. Everything looks much more normal when testing with the latest QEMU v2.4.0. I suspect the issue is a problem of memory mapping i.e. we are not seeing the real vrings in the cases that fail. More testing is needed to confirm.

I have rebased the one patch we have that is not upstream and pushed a `v2.4.0-snabb` branch to the `SnabbCo/qemu` repo. Perhaps we should start recommending this version to users instead? (Except for people have specific requirements e.g. want to backport features/fixes onto a specific version to match a distro.)

Link: https://github.com/SnabbCo/qemu/tree/v2.4.0-snabb

lukego:
Closing: This is assumed to be a QEMU issue and that the existing Snabb behavior is good.

PR #626, opened 2015-09-25
Streamlining the Snabb Switch CI (SnabbBot)

eugeneia:
This branch includes changes to the CI as discussed in #588 .
The counterpart to this branch is: https://github.com/eugeneia/snabbswitch-docker

Notable changes:

* Normalize environment variables used by Snabb Switch tests and benchmarks
* Drop `scripts/cperf` in favor of `make benchmarks` (nobody except me used cperf, hopefully `make benchmarks` will be convenient)
* Strip asset fetching functionality from `test_env` and adjust it to “new” assets (for instance there is only one kernel image for use in both DPDK based and regular NFV tests)
* Added `scripts/dock.sh` to do run tests and benchmarks in Docker container (should this automatically do a `docker pull`?)
* Rework of SnabbBot to use `scripts/dock.sh` and `make benchmark`


lukego:
What a lot of work! This is going to be awesome :-)

I am taking this for a spin on NixOS (chur). Apologies in advance :). Having a different distro inside and outside the containers is probably going to uncover some obscure bugs but I hope that is a valuable exercise.

The test scripts (e.g. `make benchmarks`) don't run on NixOS. bash is not called `/bin/bash` and instead you need `#!/usr/bin/env bash`. No biggie: running tests inside docker should avoid this.

The `snabb` binary compiled on NixOS does not run inside the container. I think that it is looking for glibc in the wrong place (interesting data point re: Linux binary portability). No biggie: building the binary inside docker should avoid this.

Building and running inside the container with `sudo scripts/dock.sh bash -i` does seem to basically work! I did not get success on all of the benchmarks though and I am not sure how to troubleshoot:

```
# SNABB_PCI0=0000:03:00.0 SNABB_PCI1=0000:03:00.1 make benchmarks    
basic1-100e6 20.4
packetblaster-64 9.47
snabbnfv-iperf-1500 0
snabbnfv-iperf-jumbo 0
Terminated
snabbnfv-loadgen-dpdk 0
```

I don't really understand how to build the assets e.g. a custom kernel? Looks like these live in a separate repository called `eugeneia/snabbswitch-docker` but cloning that and running the Makefile didn't work:

```
$ make
MKDIR	assets
COPY	QEMU SnabbCo
MKDIR	context
COMPILE	Linux ubuntu-trusty
make[1]: Entering directory `/home/luke/git/snabbswitch-docker/kernel/ubuntu-trusty'
cat: debian.master/config/config.common.ubuntu: No such file or directory
cat: debian.master/config/amd64/config.common.amd64: No such file or directory
cat: debian.master/config/amd64/config.flavour.generic: No such file or directory
make[2]: Entering directory `/home/luke/git/snabbswitch-docker/kernel/ubuntu-trusty/ubuntu-trusty'
make[2]: *** No rule to make target `bzImage'.  Stop.
make[2]: Leaving directory `/home/luke/git/snabbswitch-docker/kernel/ubuntu-trusty/ubuntu-trusty'
make[1]: *** [ubuntu-trusty/arch/x86/boot/bzImage] Error 2
make[1]: Leaving directory `/home/luke/git/snabbswitch-docker/kernel/ubuntu-trusty'
make: *** [assets/bzImage] Error 2
```

So some clarification of the workflow for that would be very helpful. Maybe another README that explains the "big picture" like how to bootstrap everything and how to know what versions of stuff you are really running?

eugeneia:
> So some clarification of the workflow for that would be very helpful. Maybe another README that explains the "big picture" like how to bootstrap everything and how to know what versions of stuff you are really running?

You need to do a `git submodule update --init`, the kernel/qemu/dpdk sources are included as submodules. A README is definitely needed here.

lukego:
One more drive-by testing thought:

The original intention of `make test` was to be a quick and broad test to run before committing code. That only works when it takes around 30 seconds to run. This broke down when we added longer running tests e.g. rate limiter benchmark, extended NIC selftest, NFV, etc. I don't think anybody routinely runs the test suite locally anymore because it takes too long.

It would be very good to strip `make test` back down to 30 seconds or less. For example, skip the NFV tests, run 5 iterations of the NIC tests instead of 100, and so on. Then we could have a separate way to run the full test suite e.g. in CI context.

eugeneia:
@lukego On the other hand I think its crucial to be able to reproduce CI results manually. One of the core goals I have in mind for the updated CI is to reduce its divergence from manual testing. E.g. I think the value of a CI degrades when its substantially different from running `make test`. I was often asked how to reproduce CI results locally and I think I want to avoid that in the future.

Also note that the intel selftest is skipped by defaul too, as a counter point, here I run `make test` in 20 seconds:

```
$ time sudo make test
DIR       obj/testlog
TEST      core.lib
TEST      core.shm
TEST      core.link
TEST      core.app
SKIPPED   testlog/core.app
TEST      core.timer
TEST      core.memory
TEST      core.counter
TEST      apps.intel.intel_app
SKIPPED   testlog/apps.intel.intel_app
TEST      apps.keyed_ipv6_tunnel.tunnel
TEST      apps.rate_limiter.rate_limiter
TEST      apps.packet_filter.pcap_filter
TEST      apps.socket.raw
TEST      apps.vpn.vpws
TEST      apps.vhost.vhost_user
SKIPPED   testlog/apps.vhost.vhost_user
TEST      lib.protocol.tcp
TEST      lib.protocol.ipv4
TEST      lib.protocol.ipv6
TEST      lib.hardware.pci
TEST      lib.hash.murmur
TEST      lib.checksum
TEST      lib.pmu
SKIPPED   testlog/lib.pmu
TEST      lib.bloom_filter
TEST      lib.ipc.shmem.shmem
TEST      program.snabbnfv.nfvconfig
SKIPPED   testlog/program.snabbnfv.nfvconfig
TEST      program.snabbnfv.neutron2snabb.neutron2snabb_schema
TEST      program.snabbnfv.neutron2snabb.neutron2snabb
TEST      apps/solarflare/selftest.sh
SKIPPED   testlog/apps.solarflare.selftest.sh
TEST      selftest.sh
TEST      lib/watchdog/selftest.sh
TEST      program/snabbnfv/neutron2snabb/selftest.sh
TEST      program/snabbnfv/selftest.sh
SKIPPED   testlog/program.snabbnfv.selftest.sh
TEST      program/packetblaster/selftest.sh
SKIPPED   testlog/program.packetblaster.selftest.sh

real	0m19.544s
user	0m7.180s
sys	0m0.798s
```

lukego:
@eugeneia Good point and well argued with data :).

eugeneia:
@lukego  regarding the failed benchmarks, your best bet would be to run them individually. To see the actual command run e.g. by `snabbnfv-iperf-1500`:

```
$ cat src/bench/snabbnfv-iperf-1500 
#!/bin/bash
set -e
out=$(program/snabbnfv/selftest.sh bench)
# Extract floating point Gbits number from output.
echo "$out" | grep IPERF-1500 | cut -d " " -f 2
% 
```

E.g. for this case, see what running `program/snabbnfv/selftest.sh bench` gives.

eugeneia:
I wonder if we should change all instances of `#!/bin/bash` to `#!/usr/bin/env bash` to improve portability? I noticed that the linux kernel sources as well as the qemu sources contain numerous “`#!/bin/bash` scripts” as well so this seems to be a widespread problem.

lukego:
In the short term I reckon it makes sense to focus on running the tests inside containers where we can safely assume that all the right software is installed where we expect it. Then we can consider to what extent it is worth porting the test infrastructure to run "natively" in other environments e.g. NixOS.

eugeneia:
https://github.com/eugeneia/snabbswitch-docker now has a README. Its not very friendly yet but should give a good overview of the current state.

lukego:
Still getting the hang of this... :)

Any tips for debugging?

```
[luke@chur:~/git/snabbswitch/src]$ sudo SNABB_PCI0=0000:01:00.0 SNABB_PCI1=0000:01:00.1 scripts/dock.sh program/snabbnfv/packetblaster_bench.sh | tee log.txt
Defaulting to SNABB_TELNET0=5000
Defaulting to PACKETS=100e6
Defaulting to CAPFILE=64
Defaulting to MAC=52:54:00:00:00:
Defaulting to IP=fe80::5054:ff:fe00:
Defaulting to GUEST_MEM=512
Defaulting to HUGETLBFS=/hugetlbfs
Defaulting to QUEUES=1
snabbnfv traffic starting (benchmark mode)
Loading program/snabbnfv/test_fixtures/nfvconfig/test_functions/snabbnfv-bench1.port
engine: start app B_NIC
engine: start app B_Virtio
Get features 0x18428001
 VIRTIO_F_ANY_LAYOUT VIRTIO_NET_F_MQ VIRTIO_NET_F_CTRL_VQ VIRTIO_NET_F_MRG_RXBUF VIRTIO_RING_F_INDIRECT_DESC VIRTIO_NET_F_CSUM
...
```

looks like no traffic is processed.

eugeneia:
I would check out what DPDK VM is up to:

```
SNABB_PCI0=0000:01:00.0 SNABB_PCI1=0000:01:00.1 scripts/dock.sh /bin/bash
program/snabbnfv/packetblaster_bench.sh &
telnet localhost 5000
screen -r
```

I assume you gave it a bit of time to start up the DPDK VM.

lukego:
I'm a bit lost here: I can't connect to a VM on port 5000 or 5001 and I can't find the tmux session to attach to either. Can you take a look and see if you can run the tests on chur please?

eugeneia:
Sure, I can't in to to chur right now though (lab1.snabb.co:2020, right?). I think you need to add my public key. Are you sure that you are *in* the docker container when trying telnet/tmux?

lukego:
Yep. I modified `dock.sh` to run `docker -t` so that I can run the container interactively.

The whole configuration of chur is on Github and you should have an account defined here:
https://github.com/SnabbCo/snabblab-nixos/blob/master/chur-configuration.nix#L170

eugeneia:
SnabbBot did not gain consciousness, I was accidentally logged in as him.

Regarding the missing `-t` option, that's fairly annoying... I had to remove it because it would break stuff by converting `\n` to `\r\n`.

lukego:
How did you find that error message? (Any refresher on troubleshooting problems during test?)

I believe the problem is that DPDK was compiled on a newer CPU than what chur has. DPDK is tricky to deploy: they choose a specific CPU microarchitecture at compile-time and then if you the binary on a different CPU you will get an error, a crash, or bad performance.

Could be that we can update the scripts for bootstrapping the VM to compile with more conservative hardware expectations. I'll take the new README on bootstrapping the assets for a spin.

lukego:
The kernel takes a very long time to clone. Maybe instead of a submodule we should use a `git clone --depth=1` command?

eugeneia:
@lukego Try `git submodule update --depth 1`, I think that is what I did.

Regading finding that error message (I had to add the `-t` flag as you did, I also had to do `scripts/dock.sh "cd ..; make"` to build everything):

```
SNABB_PCI0=0000:88:00.0 SNABB_PCI1=0000:88:00.1 scripts/dock.sh /bin/bash
scripts/dock.sh program/snabbnfv/packetblaster_bench.sh > bench.log &
telnet localhost 5000
screen -r # (there was no screen to resume)
cat /etc/rc.local # Will tell you how DPDK is started
[...]
/root/dpdk/examples/l2fwd/x86_64-native-linuxapp-gcc/l2fwd -c 0x1 -n1 -- -p 0x1
```

lukego:
are you running docker inside docker? I did not expect the `scripts/dock.sh` on the second line..?

lukego:
More recursive dockery...

`make image` doesn't run on NixOS. should I be running that inside docker? (any particular requirements?) The problem I hit was `bc` not being available in PATH.

eugeneia:
Oh that's a mistake, no docker inside docker. Sorry!

eugeneia:
> make image doesn't run on NixOS. should I be running that inside docker? (any particular requirements?) The problem I hit was bc not being available in PATH.

Regarding `make image`, it should *not* be run in docker, and it should not call `bc` either, the rule looks like this:

```
image: assets/qemu assets/bzImage assets/qemu.img assets/qemu-dpdk.img
	$(E) "DOCKER BUILD	$(IMAGE)"
	$(Q) (cp image/$(IMAGE)/Dockerfile assets/ && docker build -t $(IMAGE) assets/)
```

Now the `eugeneia/snabb-nfv-test` image *does* install `bc` inside the container but I don't see right now how that could be related.

lukego:
I think we need to automate the asset building with a `Dockerfile` instead of a `Makefile`. I'm testing on NixOS and the first few dependencies were minor (`bc` and `openssl`) but now it is trying to run `dpkg-gencontrol` that looks Debian/Ubuntu specific and at this point it seems like I am bumping into the same problem that docker is supposed to solve.

eugeneia:
Oh that sounds like its failing on the kernel's `make deb-pkg`? Could that be it? I see your point, let me think about building the kernel inside docker.

eugeneia:
The solution to this would be to use an intermediary container for building the kernel (can do, not that much of an issue).

> it seems like I am bumping into the same problem that docker is supposed to solve.

That statement is a bit unfair: We solved the problem of running snabb on machines that only have docker. Building the image that is used to run snabb on these machines, on machines that only have docker is another thing. Double trouble.

> I think we need to automate the asset building with a Dockerfile instead of a Makefile.

Inside the Docker builder we can not, “burn” VM images so we would end up with a Makefile that does things like: docker build, docker run, docker commit, ... This was indeed my first approach and it was much messier than what we have now.

lukego:
I think I need to look more closely at the scripts and draw some pictures of the whole container workflow. I am still in the naive docker-newbie world where every problem can be solved by one more application of containers (and where any other solution e.g. manual package installation is taboo). I don't really appreciate the limitations of `docker build` etc yet.

Really what I want to accomplish now are:

1. Work out how to run the benchmarks/tests on chur and attach to the tmux to see all of the processes running.
2. Try rebuilding the guest with different DPDK build options so that it will run on chur's CPU.

lukego:
The kernel's build dependencies are actually more than I expected. bc, openssl, dpkg, and now some missing perl modules. I start to feel some sympathy for the people who have struggled to compile Snabb Switch on different environments like 32-bit or musl-libc.

lukego:
(I will keep thinking aloud as I work this out. Please bear with for a while :). I am excited to be able to run the CI suite on the rest of the lab servers starting with chur.)

lukego:
The scripts on this PR all look really nice and simple and neat :).

I am struggling with running things a bit though. I am still not sure how you were able to see the error message from DPDK but I didn't find it. It might make life easier if there is an easy way to attach to the tmux and see the processes?

I tried building the assets inside an Ubuntu container with a Dockerfile like this:

```
FROM ubuntu:15.04
RUN apt-get update
RUN apt-get install -y build-essential gcc pkg-config glib-2.0 libglib2.0-dev libsdl1.2-dev libaio-dev libcap-dev libattr1-dev libpixman-1-dev libncurses5 libncurses5-dev git telnet tmux numactl wget
RUN git clone https://github.com/eugeneia/snabbswitch-docker
```

but for some reason I can't check stuff out:

```
# git submodule update --init --depth=1
...
fatal: reference is not a tree: c9cea8f431c929f70a9371f4b379ab66c15c5293
Unable to checkout '9692d47896b508d94991ec9dfa75fd6bddd73981' in submodule path 'kernel/ubuntu-trusty/ubuntu-trusty'
Unable to checkout 'c9cea8f431c929f70a9371f4b379ab66c15c5293' in submodule path 'qemu/SnabbCo/qemu'
```

and I know you had something similar once: do you know the solution?

General observation: running commands partly on the host and partly in a container seems to be asking for trouble. I see this on NixOS but I guess this would also bite e.g. with different versions of Ubuntu inside and outside.

lukego:
Zooming out a bit and thinking holistically about the whole set of use cases and whole technology stack and how it relates to this PR...

Use cases:

1. Compile and run Snabb Switch.
2. Deploy a Snabb Lab server.
3. Quickly and repeatably run CI tests on any machine.
4. Quickly experiment with new CI-like tests on any machine.
5. Publish new tests that are easy to share and repeatable.

Compile and run Snabb Switch: currently we don't define a controlled build environment and we just hope that the build infrastructure is simple enough that the user will succeed. They do need a suitable gcc, GNU make, GNU shellutils, etc. This seems to be working basically okay but does have lapses e.g. somebody reported failure to build on a recent Debian release and it was a `find` argument that turned out to be very recently added. The overall situation here seems basically satisfactory to me anyway (if anything I am tempted to move towards making distro-portable binary releases instead of expecting users to compile).

Deploy a Snabb Lab server. I am very happy with NixOS from an administrator point of view. Makes it easy to use config files in a Github repo to manage details like kernel version, IOMMU setting, hugetlb preallocation, user accounts on each machine, etc. Huge improvement over the situation with Ubuntu and a bunch of people independently making well-meaning changes in the root shell. NixOS is a weird user-space environment but containers are the solution here: just run something like 'tmux docker --privileged -t -i ubuntu' and you can forget about NixOS. (This should work fine: I'm writing this on a laptop that runs a full graphical Ubuntu inside a container on ChromeOS.) NixOS is essentially a hypervisor for containers (same idea as CoreOS, etc). On reflection we shold really *not* bother about stuff like `#!/usr/bin/env bash` to make NixOS happy because that is a layering violation - NixOS is the container hypervisor and not the runtime environment for running the CI test suite.

Quickly and repeatably run CI tests on any machine. Covered most excellently by this PR. Containerized test environment is obtained from dockerhub and then an arbitrary version of Snabb Switch can be tested inside. Nits: there is a learning curve for troubleshooting the test environment is a bit steep and also some risks of container-related hygenie problems e.g. tests are likely to fail if you compile snabbswitch on the host and then try to run it inside the container (at least on a NixOS host). For example if you compile `snabb` on NixOS and then run inside an Ubuntu container you will get an error like `snabb: File not found` which is really complaining that libc isn't at the path the linker expects.

Quickly experiment with new CI-like tests on any machine. Like: fire up a tmux session containing two VMs with vhost-user interfaces, or with packetblaster attached to a couple of interfaces, and then be able to manually run a Snabb instance to process traffic. For example the test case in #612 (`shunt` script) could reuse a "blasting 64-byte packets through bump in the wire" test harness instead of the user having to manually work out how to run packetblaster in the right topology. I believe the last time we discussed this we decided to make it explicitly out of scope for the CI but this still seems like a pity to me because we have all the "bits" inside test_env and seem like they could make life easier. I know that people are regularly fighting with understanding QEMU's incomprehensible command-line syntax, as seen in e.g. the Snabb NFV Getting Started Guide, and it would be lovely if test_env provided reusable convenience scripts that could be used instead (topic for the future..)

Publish new tests that are easy to share and repeatable. This PR seems like a huge step forwards in this respect: have a common set of env variables that can be reused when defining tests, have a way to create new versions of the test "assets" like VM images, and have everything hosted on Github and Dockerhub. (Incidentally I noticed that Github has added support for large file storage now.)


... Sorry, mega braindump I know ...


So coming back on topic all we really need to solve with this PR are:

1. Run CI test suite on any docker host (including chur).
2. Make sure mere mortals like me can successfully bootstrap a custom test environment.

then relatedly I want to install NixOS and SnabbBot on all lab servers.


While I am already ranting...

I also think we need to stop manually entering PCI addresses. This seems fine from the SnabbBot perspective since you can pick an address and stick to it. However it is terrible from the perspective of sharing development servers between multiple people because everybody is manually guessing addresses to try and we end up either with very low hardware utilization or with people frequently colliding with each others' tests. We need to fix this somehow. This is challenging because different people have different needs at different times e.g. sometimes you want to run a 1-minute test with any pair of ports, other times you want to run a long test with 6 ports on the same NUMA node, etc. I am not sure how to solve this. Could even be worth throwing hardware at the problem e.g. moving towards having more smaller lab servers each for use by small groups instead of a few beefy ones. NixOS might make this realistic to administrate and though the cost per core+port would look bigger on paper it might be less in practice if it increases utilization i.e. density of people who can work in the lab before it feels crowded.

End mega braindump ! Sorry to hijack this PR thread :)

eugeneia:
> I am struggling with running things a bit though. I am still not sure how you were able to see the error message from DPDK but I didn't find it. It might make life easier if there is an easy way to attach to the tmux and see the processes?

I launched a bash shell inside a `eugeneia/snabb-nfv-test` container, and ran the benchmark script in the background, which will start a VM that listens for telnet on localhost:5000. Then I connect to that VM, realize that DPDK didn't start / crashed. Attemping to launch it manually again yields the error message. The VM could maybe mount a host OS directory and write a log to it? The output of the QEMU processes are generally not useful for debugging  stuff that happens inside the VM (but feel free to `tmux attach-session` inside the container, its all there). There is no easy way for peeking into the VM in place because its rarely what you want to see, and if you do you can telnet in and poke around in any capacity. The point is a lot of hair is already on fire here and every new requirement adds icing to the cake. This whole docker/qemu business feels a ot like this movie, Inception I think was the title. ;)

> I tried building the assets inside an Ubuntu container with a Dockerfile like this:

That can't work since the VM images are built using docker. What you could do is `make bzImage` from within a container that has your working directory mounted as a voulme (`-v /host/dir:/container/dir`).

> but for some reason I can't check stuff out:

The soultion is to manually fetch what you need, in case of qemu:

```
(cd qemu/SnabbCo/qemu; git fetch v2.4.0-snabb:v2.4.0-snabb)
git submodule update qemu/SnabbCo/qemu
```

For the kernel just checkout master and use that it doesn't really matter:

```
(cd kernel/ubuntu-trusty/ubuntu-trusty; git checkout master)
```

> General observation: running commands partly on the host and partly in a container seems to be asking for trouble. I see this on NixOS but I guess this would also bite e.g. with different versions of Ubuntu inside and outside.

I don't really like the idea of having yet another Dockerfile for building the kernel. I think it might be worth getting rid of the `make deb-pkg` routine and install headers manually in the guests instead (its debian specific anyways). I think its sensible to have some dependencies for building images, and keep those small. Having extra images to poduce assets in feels overly complex to me. Generally, I would strongly prefer removing levels of indiection instead of adding more.

> Quickly experiment with new CI-like tests on any machine. Like: fire up a tmux session containing two VMs with vhost-user interfaces, or with packetblaster attached to a couple of interfaces, and then be able to manually run a Snabb instance to process traffic.

[test_env](https://github.com/SnabbCo/snabbswitch/tree/master/src/program/snabbnfv/test_env#test_env---easy-to-use-rewrite-of-bench_env) can totaly do this right now:

```
bash -rcfile program/snabbnfv/tet_env/test_env.sh
```

I use this regularily to test stuff in an ad-hoc way. I remeber you not liking the fact that it uses bash functions, and it could surely be improved upon, but functional it is.

> Run CI test suite on any docker host (including chur).

I take it, if this could be solved with DPDK configure options that would not be good enugh. We want to be able to build the image on NixOS?

lukego:
The code is really getting pleasingly short and orthogonal :). Maybe we could make it more user-friendly though?

Some details are hidden but could be helpful to make more transparent to the user:

- How the docker container is launched. Hidden in `dock.sh` but could be simplified enough to run manually? (also: quietly using a global `--name` that will collide when two instances are started using the same unix account.)
- The software versions that will be used for test. These seem to be quietly stored in `~/.test_env` within the container. Could be simpler for the user if the software and images were stored in more obvious places (e.g. QEMU installed in `$PATH` and test assets given unique names)? Then one could conceivably even run the test scripts independently of the docker env.
- Usage of tmux for managing processes. Could be more obvious if the default behavior would start the tmux session interactively so that the user could easily page between the windows (and potentially see output of commands that have failed unexpectedly in them)?
- Generally there is a lot of indirection: the test scripts don't really spell out what commands they are running but rather call shell functions that make a lot of substitutions. It is necessary to read the shell libraries pretty carefully to work out details like how tmux is used and what the QEMU image filenames will be. I might find the code easier to read and troubleshoot if it were built from more independent parts e.g. a shell script with a documented getopt-style command-line syntax for starting QEMU and perhaps "inlining" the tmux commands instead of hiding them in the library?

There are also some details that the user supplies explicitly that I would love to somehow eliminate:

- PCI addresses. Generally speaking we don't have a solution to the "how do I decide which PCI address to supply?" problem. Could be that this needs to be automated somehow e.g. always take the lowest-numbered cards or probe for a card that is available. See #634. (This might be a separate problem: perhaps we need another method by which the PCI address env vars will be already defined so that the user doesn't have to do it.)
- Telnet ports. This just seems like a leak of an internal detail of the scripts. Can we avoid this? Looks like QEMU supports using a named pipe or unix socket (filename) instead of a tcp socket (port number). Then we would need a telnet/netcat-like program that can talk to a unix socket or named pipe.
- Pcap file for packetblaster. Really we should extend packetblaster to generate packets of a chosen size instead of needing pcap files.

I don't really now how good and relevant all of these ideas are but this is what is coming to mind while poking around and trying to develop a mental model :)

lukego:
> This whole docker/qemu business feels a ot like this movie, Inception I think was the title. ;)

Yeah :). It is surprisingly complicated. Is there a simpler way we could do all of this?

lukego:
Howdy!

I could not resist having another look at methods for storing and sharing big files like VM images in a controlled way. I had a quick read about:

- Github Large File Support (new)
- Dockerhub
- Dropbox
- Bittorrent Sync
- Syncthing
- Tarsnap
- `git-annex`

I actually think that `git-annex` looks very interesting:

- Lets us add big files anywhere e.g. in `program/snabbnfv/test/vmimages`.
- New images can be sent in PRs (together with corresponding bootstrapping script).
- Github only actually transfers/stores the filename and SHA hash of the image (not content).
- Separate commands for syncing content e.g. directly between machines or via third-party hosts.
- Can refer to non-open-source VMs by SHA hash (have to obtain VM file separately).
- Compatible with third party hosting (e.g. for backups) but not dependent on it.

I wonder if that would potentially simplify the workflow somehow?

justincormack:
"I could not resist having another look at methods for storing and sharing big files like VM images in a controlled way."

The advantage of dockerhub is you can do automated builds on it, and then retrieve them by tag. You can use travis to do this too, although the authentication is more complicated.

eugeneia:
Building a kernel on NixOS is non-trivial as it turns out (e.g. using `make targz-pkg` instead of `make deb-pkg` will fail because of missing `/bin/pwd`, really!). My Makefiles break in other surprising ways as well... Not liking NixOS right now.

> I actually think that git-annex looks very interesting:

I like the idea of not supplying the NFV assets in docker images. A monolithic docker image is somewhat convenient but the amount of duplication is insane.

> New images can be sent in PRs (together with corresponding bootstrapping script)

Bootstrapping images is very hard actually. To elaborate what I have tried:

* Bootstrap “natively”: Super complicated and arbitrary shell script that's definitely not portable and can't be run in Docker

* Bootstrap from Docker containers: Fairly straight forward and portable but can also not be run in Docker.

We don't want either of those in the snabbswitch repository.

lukego:
This is a fascinating odyssey through modern and classic build tools. My head does hurt :).

Let's look again at the problems that we want to solve (and apologies for freely revising this list on a whim...):

1. Make it easy ("one-liner") to faithfully reproduce a test case on a new machine.
2. Make it easy to package up a test case for somebody else to run.
3. Simplify test scripts: let them assume they are running in a suitable environment.
4. Make it possible to define many different test environments (qemu, guest, libvirt, openstack, etc).
5. Make it easy to deploy a CI to a Snabb Lab server.
6. Make it easy to deploy a CI as part of a different environment e.g. GitLab/Jenkins.
7. Make everybody comfortable to do day-to-day hacking in the Snabb Lab.

How are we doing? which problems are easy vs hard, and which tools match which problems?

(1) does seem like a good match for Docker. Docker does deployment well if you have the test environment already uploaded to Dockerhub.

(2) seems to depend on the problem. You could share a test case as a shell script, a snsh script, a Dockerfile, etc, depending on how much of its environment it has to capture. I do think that Docker is a valuable addition to this toolbox e.g. if you want to demonstrate that Snabb Switch fails to build on a distro/version, or performs badly with a distro's libc, or you want to run some other random software as part of your test case and not require the user to install it separately.

(3) test script simplification seems to be working out well? In particular the "separation of concerns" of test case vs. test environment has simplified the scripts in snabbnfv/test_env because they don't have to manually build/install QEMU, guest image, etc. This is only possible because we now have separate means to run with a well-defined test environment (e.g. docker container).

(4) is still a problem. We want our test infrastructure to be "parameterized" with many different software versions e.g. hypervisor software (QEMU, Libvirt, OpenStack, etc) and guest kernel (e.g. different Linux versions with different virtio_net behavior) and guest userspace (e.g. different DPDK versions that have to be compiled to match both the running kernel and running CPU). Docker does not seem to solve this problem especially well: its "build" step is annoyingly restricted, Dockerfiles don't compose in any nice way, and generally it does not provide a solution to this problem (only a sandbox where we can fight with distro-specific toolchains like debootstrap). This is still a hard problem.

(5) making it easy to deploy CI to a new Snabb Lab server. Have we solved this? This would be cool to actually do. I suppose that the CI could run either via Docker, NixOS, or simply a shell script. I suppose that step one would be to run a CI that only tests on the host environment that it finds (e.g. QEMU found in $PATH) and second step would be to parameterize it to test multiple environments (once we solve that problem to our satisfaction). Or?

(6) making the CI deployable to third party environments e.g. somebody's private GitLab/Jenkins setup or OpenStack 3rd-party CI, etc. Could be that we can keep it simple here and just use solution (1) i.e. the CI would execute a one-liner that runs the tests in a Docker container. If that were too hard we could alternatively use a dedicated machine with the same software as the Snabb Lab and have CI run tests in that via SSH.

(7) making the Snabb Lab comfy for day-to-day development. Docker seems to *not* be the solution to this actually because it's just a bit awkward to use a docker container as your login shell and if you complain then people will say "don't use Docker like that". Could be that we should just make the default NixOS configuration more friendly e.g. include all editors, compilers, tools, optimal QEMU version, etc that people want to work with? People are also free to use their own favorite container- or VM- based tools to make themselves at home, if they happen to already have one.

eugeneia:
I just added a dockerfile for a build environment to compile the kernel in to `snabbswitch-docker`. I was able to build an image (on NixOS) that runs the DPDK benchmark on chur.

eugeneia:
@lukego I have been testing the new SnabbBot in action (on both grindelwald and chur), here is how it looks: https://github.com/eugeneia/snabbswitch/pulls

eugeneia:
> (5) making it easy to deploy CI to a new Snabb Lab server. Have we solved this? This would be cool to actually do. I suppose that the CI could run either via Docker, NixOS, or simply a shell script. I suppose that step one would be to run a CI that only tests on the host environment that it finds (e.g. QEMU found in $PATH) and second step would be to parameterize it to test multiple environments (once we solve that problem to our satisfaction). Or?

> (6) making the CI deployable to third party environments e.g. somebody's private GitLab/Jenkins setup or OpenStack 3rd-party CI, etc. Could be that we can keep it simple here and just use solution (1) i.e. the CI would execute a one-liner that runs the tests in a Docker container. If that were too hard we could alternatively use a dedicated machine with the same software as the Snabb Lab and have CI run tests in that via SSH.

I think we are pretty close to ticking these two off.  This PR's version of SnabbBot has very limited requirements: Docker, some standard CLI tools (the most exotic being jq). It is run as a cron job and can be pointed to any repository. E.g. anybody could run their own SnabbBot on their FooCompany/snabbswitch fork, or contribute test hardware to SnabbCo/snabbswitch's CI. Creating new test images is not *easy* but doable when you know what you are doing and once done, any SnabbBot instance could run the resulting test image by fetching it from docker hub. Using images to reproduce tests locally is also straight forward and reproducing SnabbBot results given an image should be comfortable.
(Downside: SnabbBot needs to run with root privileges in order to delete files created by Docker run)

What's missing is a way to run the tests without docker. E.g. a clean way to use just the assets outside of the image. I think the simplest way to consolidate this is to add a mechanism to `snabbswitch-docker` that will push just the assets to an alternate repository, e.g. git-lfs or something similar.

I suppose we focus on transitioning to the new SnabbBot in the coming release (hopefully we can make it to that date) and then incrementally tackle the other issues.

eugeneia:
I have evaluated git-annex for our VM storage needs today but I think its overkill: It does many things that we don't actually need (our VM images are not *precious*, we don't care if they are lost). I think a simple directory containing VM images, hosted on Dropbox or something else, will be enough. Maybe store files containing URL and hash in the repository and fetch them manually?

Maybe I am overreacting to the volume of git-annex's features...

eugeneia:
I added a `tarball` task to the `snabbswitch-docker` repository that will create a plain old tar archive of the bzImage and Guest/DPDK images. For now, I propose we distribute those tarballs using Dropbox for instance.

I would like to get this merged and switch the CI to the new SnabbBot over the next release. While not solving all our problems, I think there are some major advancements in here that should be battle-tested. Let's roll this out and further improve on this in incremental steps. I have been sitting on this egg for way too long already. ;)

lukego:
Obstacle to merge: CI is failing on this PR. Can resolve?

lukego:
re: image files, I am fine with leaving them out of scope for now and expecting to find them in a directory.

I am skeptical of Dropbox. The client software is awkward and it has no revision control to speak of. If we would keep an extra backup of the image files and track their SHA hashes with Git (keeping old images available to repeat tests, writing commit messages to describe new images, etc) then we would just be reinventing git-annex. However: let us argue about this later :).

eugeneia:
> Obstacle to merge: CI is failing on this PR. Can resolve?

Not really. There are incompatibilities with “old SnabbBot” and the new SnabbBot. I could run the new SnabbBot on all open PRs and it would fail on everyone but this one. Some assurance that this works is https://github.com/eugeneia/snabbswitch/pulls where I tested the new SnabbBot.

This will be a “one off” manual intervention and “old PRs” (based on commits before `snabb-docker`) will fail on the new SnabbBot.

Should I try to come up with a seamless transition strategy? I am not sure its a viable effort.

lukego:
@eugeneia OK. Please just tell me what actions I should take (e.g. merge this PR onto next) and when :).

eugeneia:
How about you do the regular release routine and I take care of the SnabbBot merge and transition along with the release? Alternatively you could just merge it into `next` at the end.

lukego:
I don't understand what you mean. The `next` branch is ready for release from my perspective. Do you want to include this PR in the v2015.11 release? If so let me know and I will merge.

I am not sure what you mean to take care of the SnabbBot merge. I would like to avoid e.g. ad-hoc merges onto master that are not part of releases. Is anything like that needed?

btw please release with a nice commit message like in 6c0545d that shows the version number on the first line. This is very handy to understand what release you are on from one of git's one-line summaries.

eugeneia:
Yes please merge it.

eugeneia:
How should we handle “old” PRs that fail the new SnabbBot, ask their authors to rebase before pushing new changes?

lukego:
Good question. Have to think.

Merging the new master commits (`git pull master`) would be preferred to rebase. Could even be that the CI should test not the exact submitted change but the master branch with the PR merged? Could be interesting to see what other CIs do e.g. the Rust one.

lukego:
Pushed to next.

eugeneia:
> Could even be that the CI should test not the exact submitted change but the master branch with the PR merged? Could be interesting to see what other CIs do e.g. the Rust one.

I thought about this today as well... instead of checking out HEAD it would then check out BASE and merge with HEAD, if that fails it would fail the CI? That would mean no passing PRs with merge conflict unless I am missing something.

lukego:
This is an interesting topic :-).

There are several different things the CI could tell us:

1. Does the PR branch work?
2. Does the target branch (e.g. master) work with the PR merged?
3. Does an integration branch (e.g. next) work with the PR merged?

These are actually all interesting questions. (1) tells us whether the change works in isolation. (2) tells us whether the author needs to update the PR by pulling new changes from the target branch and resolving a conflict of some kind. (3) tells us whether maintainers of integration branches (e.g. me with next) need to do some manual work before merging.

Naive question: should we test all scenarios? If not why not?

eugeneia:
My (different) view:

> 1 Does the PR branch work?

Checked by the submitter manually before PR.

> 2 Does the target branch (e.g. **master**) work with the PR merged?

That's actually what we want to know during the PR phase.

> 3 Does an integration branch (e.g. **next**) work with the PR merged?

This would be checked along with the PR that wants to merge next into master.

So I think checking the merge instead of what we do now actually makes sense and I am even tempted to hack it into SnabbBot before deploying it since it would relieve the burden from people to update their PRs.

lukego:
Sounds reasonable to me!

lukego:
Good observation that (1) and (3) can both be handled by people testing locally. This is safe because (2) will prevent a badly tested change from landing on master.

eugeneia:
(3) is even done “automatically” by the CI since the “next PR” will trigger a test of master with next merged.

lukego:
> (3) is even done “automatically” by the CI since the “next PR” will trigger a test of master with next merged.

Indeed: though this happens after I have pushed to next. So if somebody is running next, to be helpful and give more user testing before we make a release, then they are not protected from getting broken changes that don't pass CI. However with the test suite being easy to run it seems reasonable for me to be responsible for testing before pushing instead of relying on SnabbBot.

eugeneia:
@lukego Turns out new CI is failing old PRs anyways because (?) GitHub's “PR base branch” points to the commit from when the PR was created (instead of the current). Guess I will need to ask authors to merge master after all.

lukego:
@eugeneia Can we just ignore the "PR base branch" and use the latest release (master) instead?

I am imagining:

```
git checkout master
git merge pr/xxx
```

or alternatively/equivalently:

```
git checkout pr/xxx
git rebase master
```

Then contributors would only need to manually merge master onto their PR branch in cases where this failed with a conflict.

eugeneia:
I am not sure that would be a good idea, since `pr/xxx`+`current-master` is a moving target. E.g. each release would invalidate previous CI results. Without some further logic to retest open PRs after each release the behavior would be the same, except that SnabbBot assumes you always want to PR against master (could be configurable).

The proposed behavior makes sense, but it would also complicate things.

lukego:
This seems important to me. Every release *does* invalidate the CI results if the CI results are supposed to say "will master work after this change is merged". If we don't handle this automatically then people will have to manually update their PRs on every merge to master (including e.g. quick bugfix releases).

Perhaps we can find an easy formulation for this logic? SnabbBot just needs to know whether a given commit-id is already tested, where that commit-id is not literally the one on the PR branch but rather the commit-id that it gets after doing a "git merge"?

eugeneia:
Agreed. I have a branch where SnabbBot works as follows:

* “result id” == sha1 of master + sha1 of PR head, when master changes PRs will be retested
* whats being tested is the PR's head with master merged in
* merge failure leads to build failure

I have tested it and would like to deploy. Any objections?

lukego:
Just tacking another note onto this (closed/merged) PR in order to get it out of my head:

Nix has a built-in solution to the "git-annex" problem i.e. for keeping, sharing, and backing up assets like VM images that are difficult to replace (hard to download or time consuming to build). We would define Nix expressions to create the assets including the location and sha256 hash of all dependencies. These dependencies could be software (e.g. kernel, iperf, etc) or simply binaries on downloaded from some URL (e.g. specific version of a Juniper vMX image). In either case Nix would download them and store the resulting assets in its binary store. Then there are commands for synchronizing the stores e.g. to send copies to other machines, export them as tarballs, etc. So in principle we should be able to refer to assets based on their primary URL but automatically fetch them from a local mirror and also be able to make reliable backups.

There seems to be a lot to like about Nix. I suppose we will have to find out the hard way what there is to dislike :).

PR #628, opened 2015-09-28
intel10g: Enable TX prefetch for 14.8 Mpps

lukego:
Enable the "TX descriptor prefetching" feature. This is required for the 82599 NIC to achieve line-rate transmission with 64-byte packets.

Resolves SnabbCo/snabbswitch#622.

### DIFF
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -135,6 +135,7 @@ function M_sf:init ()
          :init_statistics()
          :init_receive()
          :init_transmit()
+         :init_txdesc_prefetch()
          :wait_enable()
          :wait_linkup()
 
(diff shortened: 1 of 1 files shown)


eugeneia:
From the SnabbBot report:

> OK: packetblaster-64 on c29f053 : 1.12294 of 9.468

Translation: 12% faster than master.

kbara:
Nice one!

lukego:
Progress at least...

The packetblaster benchmark *should* score 14.8 (Mpps) with 64-byte packets if all is well. That would be more like +55%. I suspect the issue is that packetblaster is sensitive to the other end of the connection and we need something there that is actually up and quickly receiving packets -- but not certain about that.

lukego:
... or really should that be with 60 byte packets in the input file? (the other 4 bytes being the automatic/implicit ethernet CRC rather than part of the packetblaster config.)

PR #631, opened 2015-09-30
lib/pmu_pebs: PEBS (Precise Event Based Samping) [sketch]

lukego:
This does not work yet but it is an implementation of the CPU "PEBS" profiling feature. I am not sure whether it is easy or hard to complete this: the hardware is simple but can this really be used from Linux userspace?

The basic idea is to take samples for profiling directly in the CPU (with extra information available) instead of via a signal handler.

More details from the comments in pmu_pebs.lua:

    PEBS is a hardware feature where the CPU collects its own samples
    and writes them into memory. The samples collected by the CPU can
    include interesting details such as the data address that is being
    referenced when a cache miss occurs. Events can be triggered based
    on a subset of performance events (cache misses, etc).

    This module sets up PEBS and enables event logging into memory.

    Refer to Intel's system programming manual (vol.3) for details:
    http://www.intel.com/content/www/us/en/processors/architectures-software-developer-manuals.html

### DIFF
--- a/src/lib/pmu_pebs.lua
+++ b/src/lib/pmu_pebs.lua
@@ -1,0 +1,97 @@
+-- PMU "Precise Event Based Sampling" (PEBS) support
+--
+-- PEBS is a hardware feature where the CPU collects its own samples
+-- and writes them into memory. The samples collected by the CPU can
+-- include interesting details such as the data address that is being
+-- referenced when a cache miss occurs. Events can be triggered based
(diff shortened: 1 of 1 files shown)


lukego:
Encouraging: this seems to be a working example running in userspace on Linux: https://github.com/pyrovski/powertools/blob/master/msr_pebs.c

lukego:

PR #633, opened 2015-09-30
core.engine: Enable selftest() method

lukego:
The selftest method has been suppressed in a previous change. It seems better to run it and for all available features (including both features that are always enabled and ones that are only optionally enabled).

Currently fails in the app restart-on-error part of the tests. This feature is not currently enabled by default but it should still be working and tested.


eugeneia:
Superseded by #697.

PR #636, opened 2015-10-02
lib.lua tweaks

alexandergall:
This changeset covers three issues

* Caching of `bit` in `lib.lua`
* Use of native Lua endian conversion
* Logging facility
  * Rate-limiting of discard messages
  * Expose logger defaults in the API
  * Use logging in `intel10g` and `nd_light`

### DIFF
--- a/src/core/lib.lua
+++ b/src/core/lib.lua
@@ -171,41 +171,6 @@ function bitset (value, n)
    return band(value, lshift(1, n)) ~= 0
 end
 
--- Manipulation of bit fields in uint{8,16,32)_t stored in network
--- byte order.  Using bit fields in C structs is compiler-dependent
--- and a little awkward for handling endianness and fields that cross
--- byte boundaries.  We're bound to the LuaJIT compiler, so I guess
--- this would be save, but masking and shifting is guaranteed to be
--- portable.  Performance could be an issue, though.
-
-local bitfield_endian_conversion =
-   { [16] = { ntoh = C.ntohs, hton = C.htons },
-     [32] = { ntoh = C.ntohl, hton = C.htonl }
-  }
-
-function bitfield(size, struct, member, offset, nbits, value)
-   local conv = bitfield_endian_conversion[size]
-   local field
-   if conv then
-      field = conv.ntoh(struct[member])
-   else
-      field = struct[member]
-   end
-   local shift = size-(nbits+offset)
-   local mask = lshift(2^nbits-1, shift)
-   local imask = bnot(mask)
-   if value then
-      field = bor(band(field, imask), lshift(value, shift))
-      if conv then
-         struct[member] = conv.hton(field)
-      else
-         struct[member] = field
-      end
-   else
-      return rshift(band(field, mask), shift)
-   end
-end
-
 -- Iterator factory for splitting a string by pattern
 -- (http://lua-users.org/lists/lua-l/2006-12/msg00414.html)
 function string:split(pat)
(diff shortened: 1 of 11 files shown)


alexandergall:
I think the SnabbBot failure in c67235b was due to the log rate-limiting in `nd_light`, which suppressed the "Resolved next-hop" message that the selftest was looking for. This is actually fixed by 6a2d0f1.

alexandergall:
Um, no, c67235b is the newest commit :( I guess we need a more robust way to check whether ND has succeeded than parsing the messages.

alexandergall:
Messages were discarded because the token bucket was empty at startup. This should be fixed now.

src/core/lib.lua
```diff
@@ -386,12 +353,64 @@ if ffi.abi("be") then
...
+-- this would be save, but masking and shifting is guaranteed to be
+-- portable.
+
+local bitfield_endian_conversion = 
+   { [16] = { ntoh = ntohs, hton = htons },
+     [32] = { ntoh = ntohl, hton = htonl }
+  }
+
+function bitfield(size, struct, member, offset, nbits, value)
+   local conv = bitfield_endian_conversion[size]
```
eugeneia: Was `bitfield` just moved within the file or is there some significant change I am not seeing here?

eugeneia:
I am not 100% convinced by using the rate limited logger everywhere. I see cases where it doesn't add much value and could complicate things (e.g. the “ND kickoff” message, really no need to rate limit it). If you think its implementation is mature enough so that we will not be bitten by it I will take your word on it though. Generally, you know I am a big fan of code unification and cleanup and this looks like good work to me.

In any case, the changes to public functions of `core.lib` should be documented in `src/README.md.src`!

src/core/lib.lua
alexandergall: Yes, it was just moved after the definition of the endian-conversion stuff to avoid forward declarations.

alexandergall:
The main reason for using the logger is not the rate-limiting but the unification of the logging. My suggestion is to use that for all messages that are operationally relevant to have a common format and a way to write them to well-defined places.  Later on we could use it to interface with a standard syslog mechanism as well. The rate-limiter is just a safeguard.

I've been using it for my own stuff extensively and it has worked well so far (try it by logging a message for every packet at 20Mpps :)

I did think about the documentation but was looking in the wrong place (src/core/README). Do'h!

src/core/lib.lua
```diff
@@ -386,12 +353,64 @@ if ffi.abi("be") then
...
+   function htonl(b) return bswap(b) end
+   function htons(b) return rshift(bswap(b), 16) end
 end
 ntohl = htonl
 ntohs = htons
 
+-- The fact that BitOps return signed integers is irritating, to say
+-- the least.  One counter-intuitive consequence is that, for example,
+--
+--   bit.bswap(bit.bswap(0x80000000)) == 0x80000000
```
dpino: To avoid this problem, constant numbers should be normalized using `bit.tobit`. Example:

```lua
> if bit.bswap(bit.bswap(0x80000000)) == 0x80000000 then print("true") else print("false") end
false
> if bit.bswap(bit.bswap(0x80000000)) == bit.tobit(0x80000000) then print("true") else print("false") end
true
```

According to LuaJIT docs "this function is usually not needed since all bit operations already normalize all of their input arguments" [1], but in the case of constants is necessary to call it explictly.

[1] http://bitop.luajit.org/api.html

src/core/lib.lua
```diff
@@ -5,6 +5,8 @@ local C = ffi.C
 local getopt = require("lib.lua.alt_getopt")
 local syscall = require("syscall")
 require("core.clib_h")
+local band, bor, bnot, lshift, rshift, bswap =
```
dpino: Since you're already doing this change, I think it would be also convenient to add an explicit inclusion of the BitOp module. It's not necessary, but the docs recommend it (and other modules do it too):

_The suggested way to use the BitOp module is to add the following to the start of every Lua file that needs one of its functions [...]This makes the dependency explicit, limits the scope to the current file and provides faster access to the bit.* functions, too._

src/lib/protocol/icmp/header.lua
```diff
@@ -74,11 +74,11 @@ end
 
 function icmp:checksum (payload, length, ipv6)
    local header = self:header()
-   header.checksum = C.htons(checksum(header, payload, length, ipv6))
+   header.checksum = lib.htons(checksum(header, payload, length, ipv6))
```
dpino: Why not caching `htons` and `ntohs` here like in the other files?

```lua
local ntohs, htons = lib.ntohs, lib.htons
```

src/core/lib.lua
alexandergall: Well, the endian-conversion *did* change, of course.

src/core/lib.lua
alexandergall: Ok, thanks. Maybe one day I'll even understand why BitOps using signed integers is considered a good thing in the first place :)

I'll remove this function and call bit.tobit explicitely when it's needed.

dpino: @alexandergall If you think is more convenient to have that function, I think is OK. My note was more like an information note :) I'd also prefer the BitOp module returns unsigned integers, LuaJIT docs say the reason to make it signed is cross-platform compatibility.

_"Defining the result type as an unsigned number would not be cross-platform safe. All bit operations are thus defined to return results in the range of signed 32 bit numbers (converted to the Lua number type)."_

http://bitop.luajit.org/semantics.html

alexandergall: I don't think that a single bit.tobit() justifies a wrapper function any more. My original method was cumbersome enough to make it seem useful.

I've actually known the interoperation argument before, but I haven't really understood why the cross-platform problem can't be solved without exposing a counter-intuitive API. Oh, well :)

src/lib/protocol/icmp/header.lua
alexandergall: I didn't bother because ICMP is not particularly performance critical.

Issue #572, opened 2015-07-21
Idea: Garbage collector for packets


lukego:
Here is an idea: To use a simple garbage collector to put freed packets back onto the freelist. (based on #565 and discussions with Juho Snellman.)

The idea is to introduce a packet garbage collector to give us these benefits during multiprocess execution:
1. Allow any process to free any packet.
2. Allow reclaiming packets that were being processed by a crashed process.
3. Allow reclaiming packets that were on the freelist of a crashed process.
4. Allow unused DMA memory to be detected and freed to the operating system.

Sounds good so far?

Here is the idea, which is not fully fleshed out and may be a little oversimplified.

First we add `owner` and `user` metadata fields to `struct packet`:
```c
struct packet {
  pid_t owner; // process that collects and allocates this packet
  pid_t user;  // process that is currently processing this packet
  int length;
  char data[10240];
}
```

When a process allocates a packet in new DMA memory it initializes the `owner` field with its own PID. Each time a packet is sent over a link the `user` field is set to the PID of the process that will receive from that link. We will need to add a `pid_t receiver` to `struct link` and then do:

```lua
function link.transmit (l, p)
   p.user = l.receiver
   ...
end
```

Now it starts to get interesting:

To free a packet we simply set its `user` field to a process that doesn't exist:

```lua
function packet.free (p)
   p.user = 0
end
```

and we rely on the `owner` process to notice this eventually and recycle the packet. How does that happen? Well, the owner periodically performs a garbage collection by scanning every packet it has ever allocated and checking whether the `user` points to a valid PID or not: if not then the packet is added back onto the freelist. This GC would be done at a regular interval e.g. once per breath or each time the packet freelist runs empty.

How about when a process crashes while using some packets? The packets will be reclaimed automatically: the GC will see that the `user` PID is not a running process and treat it the same as `0` i.e. know that the packet is not in use and put it back onto the free list.

How about reclaiming packets from the freelist of a dead process? That could also be detected by a larger GC sweep: scan every packet ever allocated by any process and recycle them if the `owner` PID is not a running process.

and returning packets to the OS e.g. after all traffic processes have terminated? Scan all packets allocated by all processes and free every HugeTLB memory block that does not contain any packets owned or used by a running process.

That's the idea.

[...]

PR #637, opened 2015-10-02
Lib protocol overhaul

alexandergall:
I was always unhappy with the way the `lib.protocol` system dealt with variable-sized headers. The new mechanism no longer uses subclasses for this. Instead, such a header class pre-defines all possible variants up front and its constructor method uses one of them to create instances efficiently.

The `datagram` abstraction now offers a more efficient method to add multiple headers to a packet by using an intermediate buffer, which reduces the number of copy operations to one (each `push` operation implied a `memmove` of the entire packet since the adoption of the *straighline* design).

Finally, the `class` system now supports a more efficient method for the recycling of instance objects by promoting the standard `new` constructor to an instance method (as well as a class method).


eugeneia:
Not sure why SnabbBot failed here, trying to rerun it now.

src/lib/protocol/README.md
```diff
@@ -742,3 +788,18 @@ datagram payload and its byte size.
 
 If *pointer* and *length* are supplied then *length* bytes starting from
 *pointer* are appended to the datagram's payload.
+
+— Method **datagram:data**
+
+Returns the location and size of the buffer of the underlying packet.
+This is a shortcut to *datagram:packet* followed by calls to
+*packet.data* and *pakcet.length*.
```
eugeneia: There's a typo here:  *pakcet*
Please use `M-x ispell` (or equivalent if you are not using Emacs), I know its a chore but its the only way I know of to avoid this systematically.

eugeneia: Also note that this introduces a new convenient but strictly speaking redundant public function that we will have to support. Its also questionable whether `datagram:data` will make easier or harder to read, one more thing to have to look up. Is it worth it?

```
data, len = d:data()

  vs

p = d:packet()
data, len = p:data(), p:length()
```

eugeneia:
> The datagram abstraction now offers a more efficient method to add multiple headers to a packet by using an intermediate buffer, which reduces the number of copy operations to one (each push operation implied a memmove of the entire packet since the adoption of the straighline design).

The way I see this portion of the PR adds plenty of complexity to the datagram interface but I can not find instances where the new functionality is used. Where is the `options` map of `datagram:new` actually used?

src/lib/protocol/README.md
alexandergall: I guess you're right.

alexandergall:
I use it in my l2vpn code (which I haven't pushed yet). That code is modular in the sense that there can be different combinations of transport and tunnel headers and it makes sense to push the headers individually rather than as a single block. The delayed commit method allows me to keep the code modular and still avoid a performance impact by moving the packet around three times in a row.

kbara:
Thank you. The lack of delayed-commit has made us have to make quite a bit of code uglier to avoid exactly that impact.

lukego:
> The delayed commit method allows me to keep the code modular and still avoid a performance impact by moving the packet around three times in a row.

Mea culpa: the way we decided to update `lib.protocol` for straightline clearly turned out to be less than practical.

I have an idea for an alternative to the delayed commit that might solve a wider class of problems including efficiently adding/removing the virtio-net header in #645. Observation: the operations causing the copies in all of these cases are `packet.shiftleft()` and `packet.shiftright()`. How about if we made a modest extension to `struct packet` so that those operations could be implemented without a copy?

Sketch code:

```
struct packet {
  char buffer[10240+512];  // buffer where packet data can be stored
  char *data;              // pointer to start of packet within the buffer
  int length;
}

function packet.allocate (p)
   ...
   p.data = p.buffer + 256 -- new packets get 256 bytes of "headroom" before the data
   return p
end

function packet.shiftright (p, n)
   p.data = p.data - n
   assert(p.data >= p.buffer)
end

function packet.shiftleft (p, n)
   p.data = p.data + n
   assert(p.data + p.length <= ffi.sizeof(p.buffer))
end
```

This would have a few implications:

1. Packet shift operations would become cheap constant-time operations. That would seem to make life easier for datagram and virtio work.
1. Packet data would not have a fixed alignment in memory. This may be okay for SIMD: newer CPUs seem to care less about alignment and they are the ones we need to optimize for. Have to make sure it would also be okay for other purposes like DMA. (I believe the 82599 is happy with any even-numbered address i.e. 2-byte alignment and that should be quite okay.)
1. The `struct packet` would grow and be less minimal. This seems justified if it will simplify the code that is working with packets. (I also still daydream about the GC for packets in #572 that would require a couple of additional fields in the struct too.)

Make any sense?

andywingo:
The assert on "shiftleft" should be
```lua
assert(p.data + p.length <= p.buffer + ffi.sizeof(p.buffer))
```

but yeah.  What would happen if the result weren't even-numbered?  Error?  Seems a strange restriction; I guess we could check if N is odd and memmove in that case, without cost in the normal case that you're shifting by a known amount.

alexandergall:
`length` would need to be updated as well (and the `assert` in `shiftleft` could simply check `n <= p.length`, I think).

Apart from those details and the non-trivial alignment problem, this would certainly simplify `lib.protocol` significantely. It would also make it straight-forward to "relocate" a protocol header to a packet buffer, like with the old multi-buffer approach. This feature was lost with the current design because headers don't stay in a fixed location (you can still do it but only in a convoluted way by calling `new_from_mem` after all headers have been pushed).

kbara:
I like the sketch; I was pondering similar things a little a month or two ago. It's a little fiddly, but seems like a plausible option to me.

PR #640, opened 2015-10-08
Allow hyphens in program name

dpino:
* Function programname didn't replace '-' for '_'
* Restricted regular expression that get rids of version numbers in
program and module names.
* Added selftest.

### DIFF
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -33,19 +33,20 @@ function main ()
    zone("startup")
    require "lib.lua.strict"
    initialize()
-   local program
    local args = parse_command_line()
    if programname(args[1]) == 'snabb' then
-      -- Print usage when no arguments or -h/--help
-      if #args == 1 or args[2] == '-h' or args[2] == '--help' then
-         usage()
-         os.exit(1)
-      else
-         -- Strip 'snabb' and use next argument as program name
-         table.remove(args, 1)
+      -- Print usage with exit status 0 if help requested
+      if args[2] == '-h' or args[2] == '--help' then
+         usage(0)
       end
+      -- Print usage with exit status 1 if no arguments supplied
+      if #args == 1 then
+         usage(1)
+      end
+      -- Strip 'snabb' and use next argument as program name
+      table.remove(args, 1)
    end
-   program = table.remove(args, 1)
+   local program = table.remove(args, 1)
    if not lib.have_module(modulename(program)) then
       print("unsupported program: "..programname(program):gsub("_", "-"))
       print()
@@ -57,7 +58,7 @@ function main ()
    end
 end
 
-function usage ()
+function usage (status)
    print("Usage: "..ffi.string(C.argv[0]).." <program> ...")
    local programs = require("programs_inc"):gsub("%S+", "  %1")
    print()
@@ -68,6 +69,7 @@ function usage ()
    print()
    print("If you rename (or copy or symlink) this executable with one of")
    print("the names above then that program will be chosen automatically.")
+   os.exit(statis)
 end
 
 function programname (program)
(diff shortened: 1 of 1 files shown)


eugeneia:
Possibly related: https://github.com/SnabbCo/snabbswitch/pull/439#issuecomment-100935622

dpino:
The patch admits program names to have hyphens, but directories still need to be named with underscore.

eugeneia:
Should we also allow hyphens in directory names? Was there a specific reason not to allow them? I feel like it would be useful to formulate the current naming restrictions and document them somewhere.

dpino:
I tested the code again. After the PR, directory names can either have a hyphen or an underscore. It works in both cases, but only one of them can be allowed (in other words, snabb-nfv and snabb_nfv cannot coexist, it triggers a compilation error).

As for me, I think both allowed. However, as a style guide rule I would prefer to favour hyphens to separate compound names over underscores.

With regard to hyphens in module names, I found this information but I think this is old Lua behaviour:

_ "Moreover, if the module name has a hyphen, its prefix up to (and including) the first hyphen is removed. For instance, if the module name is a.v1-b.c, the function name will be luaopen_b_c."_

http://lua-users.org/lists/lua-l/2009-07/msg00074.html

SnS creates modules returning tables, which is the recommended way of creating modules since Lua 5.2, instead of given modules a name using the `module` function.

lukego:
Great to revisit the topic of naming!

Couple of ideas:

Do we want to allow versions like `1.0-alpha2`? The regex in this PR would disallow the alphabetic characters and the hyphen. Could be worth considering something more accepting like `-[0-9].*`?

Does our current method of program invocation making naming awkward and could we fix that? Suppose you are deciding whether to call your program `foo` or `snabb-foo`. On the one hand `snabb foo` is a neat invocation but `foo` may clash with something else on your computer. On the other hand `snabb snabb-foo` is a bit too wordy while `snabb-foo` is pretty okay. Could it be worth adding logic so that if `argv[0]` is `snabb-foo` then we treat the program name as `foo`? That way every program could be neatly invoked either via the `snabb` name or via a dedicated name.

Thinking of this last point in terms of our existing programs I would be happy to install packetblaster as `/usr/bin/packetblaster` but I would prefer to install the gc cleanup program as `/usr/bin/snabb-gc` than simply `/usr/bin/gc`.

I am not sure if this makes things too complicated and creates too many choices though. The requirements are not completely straight in my head. One aspect that I think is really important is that each snabb program should be able to be distributed independently i.e. no requirement that your `snabbnfv` and `packetblaster` and other programs are built from the same source.

dpino:
I like the proposed changes. I also find it redundant to have a program called snabb-XXX and have to type snabb two times in case of running it as a parameter of the snabb binary. Summarizing the proposed changes:

* Allow alphanumerics and '-' in version names.
* If argv[0] has a "snabb-" prefix, remove it a treat the rest of the string as the program name. After that rely on the mechanism for executing programs.

dpino:
Updated PR. This is how program name is selected now:

* snabb foo => foo
* snabb-foo => foo
* snabb snabb-foo => foo

What is still not addressed yet is the question concerning hyphens being converted to underscores. That code is still in the `programname` function and it is necessary so programs such as `example_replay` and `example_spray` can work. If we agreed on using hyphens instead of underscores in program names, I can rename the dirs in a different PR. I don't see any technical limitation why it wouldn't be possible but I'm aware of Max's https://github.com/SnabbCo/snabbswitch/pull/439#issuecomment-100935622

lukego:
This looks excellent to me. Especially appreciate the test cases that really spell out the intended behavior.

There is a bug in one usage:

```
$ ./snabb
...
core/main.lua:82: variable 'statis' is not declared
stack traceback:
	core/main.lua:126: in function <core/main.lua:124>
	[C]: in function 'error'
	lib/lua/strict.lua:37: in function '__index'
	core/main.lua:82: in function 'usage'
	core/main.lua:45: in function <core/main.lua:32>
	[C]: in function 'xpcall'
	core/main.lua:155: in main chunk
	[C]: at 0x0044f770
	[C]: in function 'pcall'
	[C]: in function 'require'
```

PR #641, opened 2015-10-08
Add function find_devices to core/lib.lua

dpino:
This function can be helpful to set a PCI address by a pattern, so either finding a device by a full address "0000:01:00.0" or a shorter address ":01:00.0" are valid.

This code is part of snabb-lwaftr. We think it can be useful for other modules too. Function by @andywingo.

### DIFF
--- a/src/lib/hardware/pci.lua
+++ b/src/lib/hardware/pci.lua
@@ -146,8 +146,7 @@ local function find_devices(pattern)
    if #pci.devices == 0 then scan_devices() end
    local ret = {}
    for _,device in ipairs(pci.devices) do
-      if (device.usable and device.driver == 'apps.intel.intel_app' and
-         device.pciaddress:match(pattern)) then
+      if device.usable and device.pciaddress:match(pattern) then
          table.insert(ret, device.pciaddress)
       end
    end
--- a/src/lib/hardware/pci.lua
+++ b/src/lib/hardware/pci.lua
@@ -142,6 +142,31 @@ function root_check ()
    lib.root_check("error: must run as root to access PCI devices")
 end
 
+local function find_devices(pattern)
+   if #pci.devices == 0 then scan_devices() end
+   local ret = {}
+   for _,device in ipairs(pci.devices) do
+      if (device.usable and device.driver == 'apps.intel.intel_app' and
+         device.pciaddress:match(pattern)) then
+         table.insert(ret, device.pciaddress)
+      end
+   end
+   return ret
+end
+
+-- Return device's PCI address matching pattern
+function find_device(pattern)
+   local devices = find_devices(pattern)
+   if #devices == 0 then
+      error('no devices matched pattern "'..pattern..'"')
+   elseif #devices == 1 then
+      return devices[1]
+   else
+      local devices_str = table.concat(devices, ' ')
+      error('multiple devices matched pattern "'..pattern..'":'..devices_str)
+   end
+end
+
 --- ### Selftest
 ---
 --- PCI selftest scans for available devices and performs our driver's
--- a/src/core/lib.lua
+++ b/src/core/lib.lua
@@ -4,6 +4,7 @@ local ffi = require("ffi")
 local C = ffi.C
 local getopt = require("lib.lua.alt_getopt")
 local syscall = require("syscall")
+local pci = require("lib.hardware.pci")
 require("core.clib_h")
 local bit = require("bit")
 local band, bor, bnot, lshift, rshift, bswap =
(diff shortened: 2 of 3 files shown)


eugeneia:
Public functions of `core.lib` should be documented in `src/README.md.src`. Refer to the [documentation-guide](https://github.com/eugeneia/snabbswitch/blob/documentation-fixups/src/doc/documentation-guide.md) for pointers.

dpino:
Updated PR:
* Moved function to `lib/hardware/pci.lua`. 
* Documented function in `lib/hardware/README.md.src`.

andywingo:
apparently this code is over at https://github.com/dpino/snabbswitch/compare/find_devices currently

dpino:
Sorry, I closed this PR by mistake. I restored the dpino:add_find_devices branch but it seems this PR cannot be reopened. I can create a new PR if necessary.

src/lib/hardware/pci.lua
```diff
@@ -142,6 +142,31 @@ function root_check ()
    lib.root_check("error: must run as root to access PCI devices")
 end
 
+local function find_devices(pattern)
+   if #pci.devices == 0 then scan_devices() end
+   local ret = {}
+   for _,device in ipairs(pci.devices) do
+      if (device.usable and device.driver == 'apps.intel.intel_app' and
```
eugeneia: What about the Solarflare driver?

dpino:
PR rebased to master. I added a patch removing the constraint that limits `find_devices` search to intel devices.

lukego:
Great initiative, Diego! We definitely need a shorter way to refer to PCI devices.

Zooming out to the overall idea rather than the code, I see two alternative ways to implement this:

1. Using Lua patterns to match PCI addresses.
2. Exactly defining a short/canonical form for PCI addresses.

Lua patterns are tempting. We are already using them in packetblaster too. Packetblaster accepts a pattern to name multiple devices, so you can write `0[123]:00.[01]` to potentially match all of `0000:01:00.0 0000:01:00.1 0000:02:00.0 0000:02:00.1 0000:03:00.0 0000:03:00.1`. Or you could simply write `.` to match all available devices. The downside I see is that this is a leaky abstraction (exposing Lua quirks to end-users of Snabb based applications) and it can be surprising too e.g. that `.` will be treated as a wildcard rather than a literal period (awkward since PCI addresses have a significant period character in them).

I feel like the principle of least surprise takes precedence here. I have come to think of the packetblaster command-line as a misfeature.

The alternative is to define (or find and reuse) a canonical shortening of PCI addresses. Like the way IPv4 and IPv6 addresses have lang/short forms via `pton` and `ntop`.

Here is an idea for some isomorphic address transformations:

    1 <-> 01 <-> 01:00 <-> 01:00.0 <-> 0000:01:00.0

and the canonical representation could be the full address but omitting any leading `0000:`.

What do you think? Would that work for lwAFTR too?

kbara:
It's slightly awkward that it takes until the 3rd element of that to have a way to specify addresses like 01:00.1. I'm not sure that 1 and 01 are great abbreviations; they imply .0, making it more awkward to specify .1 in comparison, which I can imagine being annoying for people just starting to use Snabb (they'll get used to using 1, then have to scramble around how to figure out how to refer to a .1 address). And it's not obvious looking at a command line that you're referring to a PCI address at that point. '1' looks like it could equally be an overly-abbreviated non-canonical form of lots of different things.
I think 01.0 is ok (I'm not enthusiastic about it, but it'll save some typing), I like 01:00.0 as the canonical representation, and 0000:01:00.0 should be allowed (but never required), I'd say. (One thing I've wondered about, but never dug into, is whether the 0000 is always 0000, and why; do you know?)
I agree with you that exposing Lua patterns is the wrong approach.
All of the options you've mentioned could work for the lwAFTR.

lukego:
@kbara The full syntax is `DOMAIN:BUS:DEVICE.FUNCTION`. Here is [one](http://wiki.xenproject.org/wiki/Bus:Device.Function_(BDF)_Notation) documented convention.

On reflection I also like the idea of having a canonical form that simply omits any leading `0000:`. So we would write `01:00.0` and expect to see PCI addresses printed in that same shortened format. If the domain is non-zero then the long format would be used. End of story.

kbara:
Sounds fantastic.

dpino:
I implemented the expansion in #660. It's slightly different to what @lukego suggested. IMHO the inconvenient of expanding "1" as "0000:01:00.0" is that there's not short form for device == 1. The only way to write such address is using the full form "0000:00:01.0". I think the schema should be similar to IPv6 abbreviation, where "::1" means "0:0:0:0:0:0:0:1". So, "0000:04:00.0" would be abbreviated as "4:". A single "4" gets expanded to "0000:00:04.0". There are more examples in the PR.

kbara:
Honestly, that feels like way too many options. Documenting it, visually parsing what other people are using, and the increased use of things that look almost IPv6-like strike me as all richer sources of confusion this way. I'm all for convenience, but I think this goes so far that it's less convenient or practical.

lukego:
I agree with @kbara. @dpino, I humbly apologise for suggesting the more complex scheme. The extra code and explanations are really not justified by the convenience (which in itself might even lead to confusing inconsistency in usage).

eugeneia:
Is this obsoleted by #669 as well?

dpino:
@eugenia Right. I close this PR :)

PR #642, opened 2015-10-09
cosmin's public key

capr:
I want in!


lukego:
Key added via SnabbCo/snabblab-nixos#1.

PR #645, opened 2015-10-27
Implementation of virtio net driver

dpino:
This is a first implementation of a virtio net driver.

Code by @javierguerragiraldez. I cleaned up the code a bit and tested it in a virtualized guest. Packets flow from the guest to the host and viceversa. SnabbSwitch runs on the guest.


src/apps/basic/basic_apps.lua
```diff
@@ -95,20 +95,14 @@ function Tee:new ()
...
-                  transmit(output[k], k == #output and p or packet.clone(p))
-               end
+   for _, inport in ipairs(self.input) do
+      while not inport:empty() do
+         local pkt = inport:receive()
+         local used = false
+         for _, outport in ipairs(self.output) do
+            if not outport:full() then
+               outport:transmit(used and pkt:clone() or pkt)
+               used = true
```
lukego: Is this a rewrite of the `Tee` app to preserve the functionality but make the code shorter? That would be very welcome: nothing is better than reducing code size while preserving functionality :-).

Review is a bit hard though when there is no commit message explaining why the change is being made. Since this is included with the commit that adds the Virtio-net driver I am left wondering whether this is actually modifying the `Tee` behavior in some way that the Virito-net driver depends on.

(Just as feedback for the future: I'm not suggesting to go back and rebase it now.)

src/apps/virtio_net/virtio_net_dev.lua
```diff
@@ -0,0 +1,208 @@
...
+      int16_t csum_offset;
+      // int16_t num_buffers;    // only if MRG_RXBUF feature active
+   }
+]])
+
+function VGdev:transmit(p)
+   ffi.fill(pk_header, ffi.sizeof(pk_header))
+   local ethertype = ffi.cast('uint16_t*', p.data+12)[0]
+   if ethertype == 0xDD86 or ethertype == 0x0080 then
+      local startoffset = C.prepare_packet(p.data+14, p.length-14)
```
lukego: Principle of least surprise: How come this app is inspecting the packet payload and requesting that the checksum be recalculated?

I have a few objections to this:

1. How about if the sender has already calculated the checksum? Then it is wasteful to ask for the device to repeat this work. (This will have overhead on the hypervisor in situations where the checksum is not offloaded onto the NIC.)
1. How about if the sender is supposed to pass packet payload verbatim? e.g. a Layer-2 device that should pass packets unmodified regardless of the validity of their Layer-3/Layer-4 checksums.
1. How does this really behave and are there some invariants that need to be taken care of? (For example does the packet need to already have a valid "header pseudo-checksum" stored in its checksum field?)

One way to take care of all these objections would be to skip the checksum offload feature and simply transmit packets verbatim. Then the other apps that are feeding the packets would need to take care of the checksums e.g. preserving the one that is already there, or calculating with our SIMD routines, or calculating a delta when updating headers. That may be unacceptable for other reasons but that would be an interesting discussion in its own right.

src/core/packet.lua
```diff
@@ -109,3 +114,21 @@ function preallocate_step()
...
+
+ffi.metatype(packet_t, {__index = {
+   clone = clone,
+   append = append,
+   prepend = prepend,
+   shiftleft = shiftleft,
+   free = free,
+   physical = physical,
+   dump = dump,
+}})
```
lukego: Is it important to add an FFI metatable to the packet struct in this PR? This looks like an unrelated change that is coming along for the ride.

I do actually think that it is a good time for us to define a unified "Snabb API" and that FFI metatables are probably the right way to dispatch calls to packets, links, etc (instead of full module calls that can be slow and local caching of functions that is verbose). However I am not so wild about doing this piecemeal unless this is solving an important problem somehow.

lukego:
This is fantastic stuff!

Having a Virtio-net driver is extremely valuable and will make a lot of VM-based deployment scenarios easy.

I would like to please see more documentation e.g. a README.md that explains the exact behavior of the app and includes references to relevant specifications (e.g. the new Virtio-net standard from OASIS).

lukego:
Generally also: I would like to keep core modules like `packet` fairly minimal. In the past the packet module in particular had become a bit of a kitchen sink for utility functions and it was very satisfying the strip this away as part of the "straightline" refactoring. I think that PRs adding code to core modules like `packet` and `link` need to be accompanied by a strong rationale for factoring the code in that way.

In the code on this PR I can imagine a case being made for `packet.physical()` being a fundamental thing that we want but I am not so sure about adding `packet.dump()` when we already have a seemingly equivalent `lib.hexdump()`.

src/core/packet.lua
javierguerragiraldez: Yes, this is unrelated.  It's the kind of things that have been in my fork for a long time and I found indispensable.  sometimes I forget to backport to `master` style before PR'ing... or maybe I'm hopeful that it would finally be merged before this one.

The `:dump()` method was there mostly for debugging.  should be removed from the PR.

src/apps/basic/basic_apps.lua
javierguerragiraldez: I forgot that was there... no, there's no new behavior; just a kata of porting generic apps to OOP style.  (and making it more readable, I guess)

src/apps/virtio_net/virtio_net_dev.lua
javierguerragiraldez: AFAICT, it's the protocol for asking checksum offloading, it does the minimum (and mostly constant time) and leaves the bulk of the task to the virtio 'device'.  I guess you're right that a 'switch-like' level 2 device shouldn't care about payload; but equivalent code is in the virtio-net kernel driver, so I ported it here.  Of course, Linux networking isn't the paragon of well-layered structure...

src/core/packet.lua
lukego: Agree this makes sense as part of an improved API. Responded on #605.

src/apps/virtio_net/virtio_net_dev.lua
lukego: This should be explained in the README for the app. I suppose that any behaviour is fine provided that the user understands it.

lukego: Note: On Snabb NFV at least there is overhead for requesting checksum offload from the "device". This will become a SIMD computation on the snabbnfv process and it will impact packet rate. Likewise on packets being received so that checksum validity can be reported.

I would suggest a best practice of not negotiating offloads unless they are actually needed. There will be analogous situations for other hypervisor networking implementations.

eugeneia:
May I suggest interactively rebasing this PR and splitting it into logical commits? Too many unrelated changes in two bulk commits.

kbara:
@eugeneia I was thinking the same earlier today.

src/apps/virtio_net/virtio_net_dev.lua
```diff
@@ -0,0 +1,208 @@
...
+         ct = ct,
+         size = ffi.sizeof(ct),
+         offset = offset,
+      }
+      offset = offset + struct[fld].size
+   end
+   return struct, offset
+end
+
+local function fieldrd(field, fd)
```
nnikolaev-virtualopensystems: This is kind of generic  (and fieldwr) - maybe we want it somewhere in the lib?

src/apps/virtio_net/virtio_net_dev.lua
```diff
@@ -0,0 +1,208 @@
...
+   local bar = open_bar(pci.path(args.pciaddr..'/resource0'), virtio_pci_bar0)
+
+   bar.status = 0 -- Reset device.
+   bar.status = bor(bar.status, 1) -- Acknowledge.
+   -- Check something.
+   bar.status = bor(bar.status, 2) -- Driver.
+   local features = bar.host_features
+   if debug then print('host_features', features) end
+   if band(features, min_features) ~= min_features then
+      bar.status = bor(bar.status, 128) -- Failure.
```
nnikolaev-virtualopensystems: Those bar.status values, definitely deserve an enum ...

src/apps/virtio_net/virtio_net_dev.lua
javierguerragiraldez: This is an adapter to be able to read/write to a non-mmap()'able PCI resource, especially since it requires that the operations are in the exact size of the field.  I don't know if there are other similar cases.

src/core/packet.lua
```diff
@@ -109,3 +114,21 @@ function preallocate_step()
...
+function dump(p, w)
+   w = w or io.write
+   for i = 0, p.length-1 do
+      if i % 16 == 0 then w('\n', bit.tohex(i, -4), ': ') end
+      w(bit.tohex(p.data[i], -2), ' ')
+   end
+   w('\n')
+end
+
+ffi.metatype(packet_t, {__index = {
```
nnikolaev-virtualopensystems: What is the performance impact of this?

src/lib/checksum.c
```diff
@@ -150,3 +150,42 @@ uint32_t pseudo_header_initial(const int8_t *buf, size_t len)
   }
   return 0xFFFF0001;
 }
+
+uint8_t startoffset[2];
+uint8_t *prepare_packet(int8_t *buf, size_t len)
```
nnikolaev-virtualopensystems: Don't we have the protocol library to deal with parsing packets and pushing headers onto payload.
Honestly I like this hackish way better, but still it will be probably harder to maintain in the future.

Actually, could it be that the checksum calculation belongs to the protocol library?

src/apps/virtio_net/virtio_net_dev.lua
```diff
@@ -0,0 +1,208 @@
...
+   if debug then print('got features: ', bar.host_features, bar.guest_features) end
+   if band(bar.status, 8) ~= 8 then
+      bar.status = bor(bar.status, 128) -- Failure.
+      bar:close()
+      return nil, "feature set wasn't accepted by device"
+   end
+
+   if debug then print("enumerating queues...") end
+   local vqs = {}
+   for qn = 0, 16 do
```
nnikolaev-virtualopensystems: So you try to get 16 queues here, and then go just vqs[0]/vqs[1] in the other part of the code?
Is it a multiqueue implementation or not?

src/apps/virtio_net/virtio_net_dev.lua
```diff
@@ -0,0 +1,208 @@
...
+   local ethertype = ffi.cast('uint16_t*', p.data+12)[0]
+   if ethertype == 0xDD86 or ethertype == 0x0080 then
+      local startoffset = C.prepare_packet(p.data+14, p.length-14)
+      if startoffset ~= nil then
+         pk_header.flags = 1 -- VIRTIO_NET_HDR_F_NEEDS_CSUM
+         pk_header.csum_start = startoffset[0]+14
+         pk_header.csum_offset = startoffset[1]
+      end
+   end
+   p:prepend(pk_header, ffi.sizeof(pk_header))
```
nnikolaev-virtualopensystems: This is going to be sloooow .... :) Is there a smarter way?

src/apps/virtio_net/virtio_net_dev.lua
```diff
@@ -0,0 +1,208 @@
...
+   end
+end
+
+function VGdev:can_receive()
+   return self.vqs[0]:more_used()
+end
+
+function VGdev:receive()
+   local p = self.vqs[0]:get()
+   p:shiftleft(ffi.sizeof(pk_header))
```
nnikolaev-virtualopensystems: ... performance again ...
And what happens to that pk_header that you have received? Maybe we want to do something with it?

src/apps/virtio_net/virtio_net_app.lua
```diff
@@ -0,0 +1,68 @@
...
+   end
+   dev:sync_transmit()
+end
+
+function virtio_net:pull()
+   local dev = self.device
+   local l = self.output.tx
+   if not dev or not l then return end
+
+   dev:sync_receive()
```
nnikolaev-virtualopensystems: but sync_receive is NYI?

src/apps/virtio_net/virtio_net_app.lua
```diff
@@ -0,0 +1,68 @@
...
+
+   engine.configure(config.new())
+   local c = config.new()
+   config.app(c, 'source', pcap.PcapReader, input_file)
+   config.app(c, 'virtio_net', virtio_net, {pciaddr=pcidev})
+   config.app(c, 'sink', basic_apps.Sink)
+   config.link(c, 'source.output -> virtio_net.rx')
+   config.link(c, 'virtio_net.tx -> sink.input')
+   engine.configure(c)
+   engine.main({duration = 1, report={showlinks=true, showapps=true}})
```
nnikolaev-virtualopensystems: I believe on every selftest there should be at least one assert condition where you say what you expect to happen.
Relying on "it ended and it didn;t crash" approach is probably not the best verification strategy.

src/apps/virtio_net/virtio_net_dev.lua
javierguerragiraldez: the virtio specs allow for multipart packets, but the added complexity is significant (much worse than what hardware NICs do, IMHO).  Currently this line doesn't stand out in profiler runs, the queue notification alone (a single 16-bit IO) takes 90%

src/apps/virtio_net/virtio_net_app.lua
javierguerragiraldez: on a pure polling driver, there seems to be no need for `:sync_receive()`; it's still there mostly as a placeholder for some notification suppression schemes that might help significantly with performance.  so far, none have yielded any consistent advantage, unfortunately.

nnikolaev-virtualopensystems:
Overall a very good and long awaited patch(es). Still it might need a spin or two before it gets the right shape, me thinks.

src/apps/virtio_net/virtio_net_dev.lua
nnikolaev-virtualopensystems: Are you testing with 60 byte packets?

src/apps/virtio_net/virtio_net_dev.lua
javierguerragiraldez: the header is used mostly to reassemble multipart packets and a few other flags for other kinds of devices (like virtio-block)

nnikolaev-virtualopensystems: then at least assert on the multipart flag. THis could also help future developers to note the header here.

src/apps/virtio_net/virtio_net_dev.lua
javierguerragiraldez: mostly, yes; but 1500 bytes don't make much difference either.  of course, all benchmarks must be repeated for any significant code change.

lukego: Have to be careful about benchmarking. I think you will see completely different performance and profiler results depending on what hypervisor networking you are using on the QEMU level.

If the Linux kernel is doing the hypervisor networking then the packet rate will probably be pretty lousy and I can well believe that interrupt signalling can be the bottleneck.

If the hypervisor is doing kernel-bypass networking (e.g. snabbnfv) then it should be possible to achieve good results and it should be possible to suppress all interrupts so the bottleneck will not be there.

It would be interesting to know which scenario(s) this driver is optimized for. Personally I am much more interested in the scenario with kernel-bypass networking on the hypervisor (e.g. snabbnfv) because that is the only setup where I can envision practical end-to-end performance. In that sense I would not be tempted to make optimizations based on profiling with the kernel underneath: but maybe you guys care about that scenario aswell or instead.

lukego: I had missed the `packet.prepend()` to add the virtio-net header on the front. I do expect that this will be a performance issue if you were pushing a few million packets per second. If this turns out to be the case then it would be interesting to find a way to eliminate that operation. Could potentially be done by using two vring descriptors per packet (this is a common/legacy mode of operation, I think, Nikolay?) or more drastically we could reserve a little scratch area at the start of the `struct packet` where such metadata can be written.

lukego:
This looks it is is going to be a really popular feature! Lots of people interested in reviewing the code :-)

Regarding rebasing and respinning: I would suggest that the optimal way to do that would be to rebase onto a new branch called e.g. `virtio_net_v2` and submit a new PR from that (referencing this one).

The alternative would be to force-push a rebase onto this branch but that would make a mess of this discussion on Github.

Since there is so much interest in this feature it would seem especially valuable to be verbose in documentation, comments, commit messages, and PR text so that everybody gets a common understanding.

src/apps/virtio_net/virtio_net_dev.lua
lukego: (Could also be that the `packet.prepend()` doesn't turn out to be so expensive. It will be interesting to check that. There is AVX2 support landing in DynASM so we will have the option to try optimizing these SIMD copy routines ourselves before we go changing data structures.)

dpino:
@lukego I'm interesting in benchmarking the driver using the approach you commented:

_"If the hypervisor is doing kernel-bypass networking (e.g. snabbnfv) then it should be possible to achieve good results and it should be possible to suppress all interrupts so the bottleneck will not be there."_

I'm not sure of what this means. To clarify, would be the idea to launch a virtual machine using the vhost-user driver plus snabbnfv instead of using a tap interface? I have tried that but once inside the guest I cannot see the NIC registered at `/sys/class/net`, which makes sense since it's bypassing the kernel. The virtio-net driver needs a NIC with a PCI address on the guest side [1].

This is the command I'm using to launch the VM:

```
sudo $QEMU \
    -m 1024 -kernel $KERNEL \
        -append 'earlyprintk root=/dev/vda rw console=ttyS0' \
    -numa node,memdev=mem \
        -object memory-backend-file,id=mem,size=1024M,mem-path=/hugetlbfs,share=on \
    -netdev type=vhost-user,id=net0,chardev=char0 \
        -chardev socket,id=char0,path=/home/igalia/dpino/socket/vhost_A.sock,server \
        -device virtio-net-pci,netdev=net0,mac=52:54:00:00:00:01 \
    -M pc -smp 1 -cpu host --enable-kvm -serial stdio \
    -drive if=virtio,file=$IMG,format=raw -curses
```

[1] https://github.com/dpino/snabbswitch/commit/e9ce5d90a10ca02275c78a62c6020bec9acd01f1#diff-369c9508a86f051bb8fb8bed85095bfdR63

lukego:
@dpino Yes the benchmarking setup I mean is to have snabbnfv on the hypervisor side with a setup like the one described in the [Snabb NFV Getting Started Guide](https://github.com/SnabbCo/snabbswitch/blob/master/src/program/snabbnfv/doc/getting-started.md).

This setup with kernel-bypass on the hypervisor should be almost identical from the guest perspective. The difference is within QEMU that it will use a unix domain socket to setup vring memory mappings and interrupts (rather than `/dev/vhost-net` towards the kernel). You should still have a Virtio-net PCI device available in the guest and it should behave in the same way because this is implemented with the same code in QEMU in both cases. The difference should just be that other software is processing the descriptors you put into the vrings because the memory mapping is exported to a Snabb Switch process instead of the kernel.

I'd recommend checking `lspci` to see if you have a Virtio-net PCI device. I am not sure exactly what presence/absence in `/sys/class/net` means (may say more about the guest kernel than the guest hardware?)

dpino:
@lukego Thanks, it worked. I managed to run virtio_net/selftest using snabbnfv as hypervisor. Now I need a better way of measuring performance, plus running the lwaftr on the guest side and see how it performs.

lukego:
I would reuse the performance test we already have and replace DPDK in the guest with Snabb Switch. Check out Max's new CI infra!

src/apps/virtio_net/virtio_net_dev.lua
lukego: If you want to experiment/benchmark with removing the `packet.prepend()` then I have a proposal for a simple way to do that:

We could add a "scratch" area of free space before the packet payload and this app could write the virtio-net header at the end of this space (without moving the payload).

The new definition of the packet struct would be:

```c
struct packet {
  unsigned char scratch[64]; // new
  unsigned char data[...];
  int length;
}
```

This seems like a simpler solution for this use case compared with extending the packet struct with an extra indirection to allow repositioning the start of the payload (as discussed on #637). That would be a bigger change and I am not quite sure how to weight the pros/cons of that yet. I really like having a simple canonical representation of packets like we do today and wouldn't like to rush into giving that up.

dpino:
@lukego With regard to the PCI address needed by the driver, I refer to the PCI address that it needs when initializing the driver, same as the Intel82599 driver does. Relevant lines:

https://github.com/SnabbCo/snabbswitch/pull/645/files#diff-6752da38bfe8e57abc0421a80fd9c5d3R63
https://github.com/SnabbCo/snabbswitch/pull/645/files#diff-c74e02ebb8719147b6734d961b00ca20R84

The address is used to access the PCI Base Address Register and setup accordingly http://wiki.osdev.org/Virtio.

eugeneia:
> I would reuse the performance test we already have and replace DPDK in the guest with Snabb Switch. Check out Max's new CI infra!

I would be more than happy to see DPDK replaced with Snabb Switch. :)

eugeneia:
Obsoleted by #681.

PR #652, opened 2015-11-03
Replace all-zero MAC address for 02:00:00:00:00:00

dpino:
The motivation of this patch is to avoid potential problems because of using MAC addresses with special meanings. For instance, I recall a performance bug because of using a MAC address starting with '01:', which means multicast.

An all zero MAC address is a unicast globally administered address. According to [1], addresses with starting with '00:00:00' are reserved. I think is better to use one of the four family types of unicast locally administered MAC address available in the MAC address space: X2-XX-XX, X6-XX-XX, XA-XX-XX or XE-XX-XX. In this patch I picked 02:00:00:00:00:00.

[1] http://www.iana.org/assignments/ethernet-numbers/ethernet-numbers.xhtml

### DIFF
--- a/src/apps/socket/raw.lua
+++ b/src/apps/socket/raw.lua
@@ -40,8 +40,8 @@ function selftest ()
    local ethernet = require("lib.protocol.ethernet")
    local ipv6 = require("lib.protocol.ipv6")
    local dg_tx = datagram:new()
-   local src = ethernet:pton("00:00:00:00:00:01")
-   local dst = ethernet:pton("00:00:00:00:00:02")
+   local src = ethernet:pton("02:00:00:00:00:01")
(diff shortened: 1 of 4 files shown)


justincormack:
You can apply for an official MAC prefix, probably better than using random
addresses.

lukego:
@alexandergall do you have strong feelings on this patch (affecting `vpws.lua`)?

alexandergall:
The destination address in `vpws.lua` is just a dummy value that should never appear on the wire. It will be overweritten either by a statically configured address or the address discovered by `nd_light`.

Anyway, `vpws.lua` will be replaced by the `l2vpn` program that I'll push to the `vpn` branch soon.

PR #653, opened 2015-11-03
Use %x wildcard where possible

dpino:
Some regular expressions use a customized regular expression [0-9a-fA-F] to match hexidecimal numbers. It is possible to use Lua's %x wildcard to do the same.

http://www.lua.org/pil/20.2.html


eugeneia:
:+1: Makes the regular expressions less nasty. Could you please merge master to make SnabbBot pass the branch?

eugeneia:
I was unable to reproduce the CI failure and don't think its related to this PR. See #663.

PR #659, opened 2015-11-12
Cast T->mcode when passed as argument in setintfield

dpino:
So it gets rid of the warning:

```
lib_jit.c: In function ‘lj_cf_jit_util_traceinfo’:
lib_jit.c:303:32: warning: passing argument 4 of ‘setintfield’ makes integer from pointer without a cast
     setintfield(L, t, "mcode", T->mcode);
```

`T->mcode` is a _void*_, size of _long_ is wide enough to hold a _void*_.

### DIFF
--- a/lib/luajit/src/lib_jit.c
+++ b/lib/luajit/src/lib_jit.c
@@ -300,7 +300,7 @@ LJLIB_CF(jit_util_traceinfo)
     setintfield(L, t, "link", T->link);
     setintfield(L, t, "nexit", T->nsnap);
     setintfield(L, t, "szmcode", T->szmcode);
-    setintfield(L, t, "mcode", T->mcode);
+    setintfield(L, t, "mcode", (int32_t)(intptr_t)T->mcode);
     setintfield(L, t, "mcloop", T->mcloop);
     setstrV(L, L->top++, lj_str_newz(L, jit_trlinkname[T->linktype]));
     lua_setfield(L, -2, "linktype");


lukego:
Awesome !

kbara:
I disagree with this patch (more details later today).

kbara:
Short version: let's be a bit more careful when we're squashing compiler warnings.
I propose ```setintfield(L, t, "mcode", (int32_t)(intptr_t) T->mcode);```

-----

We're dealing with a 32-bit pointer, even on 64-bit systems, in this case. (As mentioned on http://lua-users.org/lists/lua-l/2009-08/msg00202.html and confirmed by printing the pointer addresses). So, semantically, we're putting a 32-bit pointer into a 32-bit int:
```static void setintfield(lua_State *L, GCtab *t, const char *name, int32_t val) ```.
So, it is legitimate to get rid of the warning.

You may rightfully object that int32_t cannot hold arbitrary pointers. This is true. However:
a) Casting it to long and then having it be downcast to int32_t has the same problem; see 'pedantic note 3'.
b) We know it's a 32-bit pointer; this is apparently a key luajit abstraction.

I really do not like casting to long here. As an alternative, I propose:
```setintfield(L, t, "mcode", (int32_t)(intptr_t) T->mcode);```

Why two casts? It most clearly reflects the semantics, while getting rid of the warning.
The int32_t cast shows what the value is actually truncated to, but is not sufficient by itself:
```
lib_jit.c: In function ‘lj_cf_jit_util_traceinfo’:
lib_jit.c:304:32: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
     setintfield(L, t, "mcode", (int32_t)T->mcode);
```
We know that it's a 32-bit pointer (potentially on a 64-bit system), but that information is not available to GCC.

The intptr_t cast is sufficient to get rid of the warning, but hides that the resulting value will then be truncated, and that this is only valid because it's actually a 32-bit pointer.

Pedantic note 1:

```The type of T->mcode is not void*, it's MCode*```. MCode is uint8_t or uint32_t.

```
#if LJ_TARGET_X86ORX64
typedef uint8_t MCode;
#else
typedef uint32_t MCode;
#endif
```


Pedantic note 2:

'long' is not big enough to hold pointers on all platforms. For instance, it's 32 bits on 64-bit Windows, which has 64-bit pointers.
You'd want intptr_t if this was your goal. It's part of C99, and used elsewhere in the codebase:
```
lib_jit.c:    setintptrV(L->top++, (intptr_t)(void *)T->mcode);
```

Pedantic note 3:
Downcasting is legal in C - and doesn't give warnings.

```
void foo(int a) {/*code elided*/}
int *somevar;
foo((long) somevar);
```
This is actively misleading code. Even if 'long' happens to be large enough to hold a pointer on the current platform, foo still only takes an int, and will silently downcast; if the pointer doesn't fit in an int, it will be silently truncated.

A runnable example (run under 64 bit x86 Linux)

```
#include <stdio.h>

long long foo(int x) {
    return (long long) x;
}

int main(int argc, char **argv)
{
    int a = 3;
    int *b = &a;
    printf("a: %d,  b: %p, foo-b: %llx, foo-long-b: %llx\n", a, b, foo(b), foo((long) b));
    printf("%llx\n", (long long) 0x123456789ab);
    return 0;
}

```
and the results:
a: 3,  b: 0x7ffe29d02034, foo-b: 29d02034, foo-long-b: 29d02034
123456789ab
```

dpino:
I like this more. Thanks for detailed explanation and links. I updated the PR. In this case I squashed the previous commit as it's just one line of code.

kbara:
Thank you.

eugeneia:
I think this is ready except for the to-be-resolved merge conflict. @dpino can you merge with master?

dpino:
@eugeneia I rebased the PR and push forced again. I was able to merge the PR into current master:

```
commit adbb255b66ba08b30e4b417813dbffff62864d50
Merge: e285507 145bdec
Author: Max Rottenkolber <max@mr.gy>

    Merged PR #670 (v2015.12 release) onto master
```

dpino:
It seems it can be merged now :)

Issue #657, opened 2015-11-11
PCI address short version


lukego:
Snabb Switch should accept the short version of PCI addresses like `09:00.0` instead of insisting on `0000:09:00.0`. The name could be expanded internally in `pci.lua`.

dpino:
There's a PR for this https://github.com/SnabbCo/snabbswitch/pull/641

eugeneia:
Closing this because #669 landed.

PR #660, opened 2015-11-12
Implement PCI address expansion

dpino:
Adds function `expand_pciaddr` which expands a PCI address to BDF format http://wiki.xenproject.org/wiki/Bus:Device.Function_%28BDF%29_Notation

Example:

* "1:"                   => 0000:01:00.0
* "01:"                 => 0000:01:00.0
* "01:00.0"          => 0000:01:00.0
* ":01:"                => 0000:01:00.0
* "0000:01:"        => 0000:01:00.0
* "0000:01:00.0" => 0000:01:00.0
* ":1"                   => 0000:00:01.0
* "1"                    => 0000:00:01.0

A single number is interpreted as a device and all the empty elements are filled up with zeroes. Semicolons tell what position an element occupies in the BDF pattern. For instance, ":1" tells everything to the left is empty and the element ends up occupying the device position. On the other hand, "1:" tells everything to the right is empty, the element ends up occupying the bus position and PCI is filled up with zeroes. This pattern is the same as ":1:".

This PR solves #657.

### DIFF
--- a/src/lib/hardware/pci.lua
+++ b/src/lib/hardware/pci.lua
@@ -31,9 +31,50 @@ function scan_devices ()
    end
 end
 
+-- Expands a PCI address to BDF format: '0000:00:00.0'
+function expand_pciaddr (pciaddress)
+   pciaddress = assert(tostring(pciaddress))
+   local ret = {}
+   -- Iterate on each part.
+   local each, pos, i = "", 0, 0
+   while true do
+      each, i = pciaddress:match("(%x+):()", pos)
+      if not each then break end
+      table.insert(ret, each)
+      pos = i
+   end
+   -- Last part.
+   if pos < #pciaddress then
+      table.insert(ret, pciaddress:sub(pos):match("([%x.]+)"))
+   end
+   -- Empty part?
+   if pos == #pciaddress + 1 then
+      table.insert(ret, "0")
+   end
+   -- Rearrange parts according to total size.
+   local pci, bus = "0000", "00"
+   if #ret == 1 then
+      ret = {pci, bus, ret[1]}
+   end
+   if #ret == 2 then
+      ret = {pci, ret[1], ret[2]}
+   end
+   local device, func = ret[3]:match("(%x%x).(%x)")
+   if not device then
+      device, func = ret[3], "0"
+   end
+   -- Format each part and return.
+   ret = {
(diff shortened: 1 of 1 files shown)


eugeneia:
All the PCI addresses I ever needed to type when working with Snabb Switch have the form `0000:XX:00.Y`. So a far as I can tell I only need a pattern `XX:.Y := 0000:XX:00.Y`. Is this coincidental? When are the domain/device fields relevant, does it depend on the specific hardware?

kbara:
@eugeneia I asked basically the same in https://github.com/SnabbCo/snabbswitch/pull/641 , and Luke pointed me to http://wiki.xenproject.org/wiki/Bus:Device.Function_%28BDF%29_Notation , which doesn't fully answer this but gives some interesting context.

dpino:
Before I was working with the virtio_net driver I thought the only use case for PCI address simplification would be the pattern you describe "XX:00.Y" which could be conveniently write as "04:00.1", for instance. 

The virtio_net driver takes a PCI address from devices at _/sys/class/net_, which in the VM I use for testing have the form "00:0X.Y" (many devices in my laptop at _/sys/bus/pci/devices_ follow this pattern too). A PCI address were device is the only significant part could be simplified as ":04.0"  or "4.0" or even "4". I think most of the time we will be dealing with this two type of addresses. The multiple examples are there to show the function works well and it's consistent. Basically what it does is to fill up with zeros what's left unwritten.

lukego:
@eugeneia I believe it is hardware dependent, yes. The network cards we are working with are pretty simple e.g. one ethernet controller chip with two ports attached to it. There could be others that e.g. combine multiple ethernet chips multiplexed behind a PCIe switching chip. Companies like [Silicom](http://silicom-usa.com/] make a lot of interesting cards in this direction. I don't know exactly how the PCIe addresses will end up looking but I wouldn't want to make too many assumptions based on the cards we have in the lab today.

@dpino I am surprised that the virtio-net driver doesn't identify the device by a PCI address exactly like the Intel10G one? (Why does `/sys/class/net` come into the picture? Can discuss this on that PR branch though.)

eugeneia:
Is this PR obsoleted by #669  and can be closed?

dpino:
@eugeneia Right. Closed PR.

Issue #643, opened 2015-10-14
Solarflare driver segfaults on lib.nfv.config selftest


eugeneia:
Solarflare driver segfaults on  `lib.nfv.config` selftest. To reproduce run `make lib.nfv.config` with a Solarflare NIC. Result:

```
selftest: lib.nfv.config
testing:	program/snabbnfv/test_fixtures/nfvconfig/switch_nic/x
engine: start app id0_Virtio
engine: start app id0_NIC
Opened SolarFlare interface p10p1 (MAC address 00:90:f5:ef:aa:67, MTU 1514)
testing:	program/snabbnfv/test_fixtures/nfvconfig/switch_filter/x
engine: reconfig app id0_NIC
Opened SolarFlare interface p10p1 (MAC address 00:90:f5:ef:aa:67, MTU 1514)
engine: start app id0_Filter_in
load: time: 0.30s  fps: 0         fpGbps: 0.000 fpb: 0   bpp: -    sleep: 100 us
testing:	program/snabbnfv/test_fixtures/nfvconfig/switch_qos/x
engine: stop app id0_Filter_in
engine: reconfig app id0_NIC
Opened SolarFlare interface p10p1 (MAC address 00:90:f5:ef:aa:67, MTU 1514)
load: time: 0.30s  fps: 0         fpGbps: 0.000 fpb: 0   bpp: -    sleep: 100 us
testing:	program/snabbnfv/test_fixtures/nfvconfig/switch_tunnel/x
engine: reconfig app id0_NIC
Opened SolarFlare interface p10p1 (MAC address 00:90:f5:ef:aa:67, MTU 1514)
engine: start app id0_Tunnel
engine: start app id0_ND
Sending neighbor solicitation for next-hop 2::2
load: time: 0.28s  fps: 3         fpGbps: 0.000 fpb: 0   bpp: 86   sleep: 100 us
testing:	program/snabbnfv/test_fixtures/nfvconfig/scale_up/y
engine: stop app id0_Tunnel
engine: stop app id0_ND
engine: reconfig app id0_NIC
Opened SolarFlare interface p10p1 (MAC address 00:90:f5:ef:aa:67, MTU 1514)
engine: start app id1_Virtio
engine: start app id1_NIC
Opened SolarFlare interface p10p1 (MAC address 01:90:f5:ef:aa:67, MTU 1514)
[...]

PR #664, opened 2015-11-17
nd_light: Fix use after free bug in neighbor solicitation.

eugeneia:
This fixes a “use after free” bug (#643) in nd_light. The bug only surfaced when using the Solarflare driver but is indeed unrelated. I want to highlight a pattern that I think is close to a whole class of bugs:

```
 function nd_light:stop ()
    packet.free(self._next_hop.packet)
+   self._next_hop.packet = nil
    packet.free(self._sna.packet)
+   self._sna.packet = nil
 end
```

So what I did here previously (I wrote the bug in the first place) is to free “static” packets when the app is stopped. In itself this is important as to not leak memory. The nd_light app specifically launches a timer that will send `self._next_hop.packet` periodically. What happens when you start nd_light and stop it right away? The timer gets activated, the packet is freed, the timer runs, attempts to clone a packet which was already freed: wham, segmentation fault.

As a response I added explicit statements unbinding the slots that hold pointers to freed packets. I think this is an important lesson learned: **Do not keep invalid pointers around.** If you free a packet held in a variable that can be accessed outside the lexical scope (for instance an object slot), null it right away. That way you will get a meaningful error message instead of a segmentation fault.

### DIFF
--- a/src/apps/ipv6/nd_light.lua
+++ b/src/apps/ipv6/nd_light.lua
@@ -136,6 +136,9 @@ function nd_light:new (arg)
    -- Timer for retransmits of neighbor solicitations
    nh.timer_cb = function (t)
                     local nh = o._next_hop
+                    -- If nh.packet is nil the app was stopped and we
+                    -- bail out.
+                    if nh.packet then return nil end
                     o._logger:log(string.format("Sending neighbor solicitation for next-hop %s",
                                                 ipv6:ntop(conf.next_hop)))
                     link.transmit(o.output.south, packet.clone(nh.packet))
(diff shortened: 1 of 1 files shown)


lukego:
This fixes a bug so I am merging it.

However, I disagree on the plan for handling this class of error. My position is that timers in apps should be implemented locally inside the app and not with a global timer whose lifetime is independent of the app instance.

This is the timer implementation that makes sense to me, potentially with some library support:

```
local deadline
function push ()
   if deadline == nil or deadline <= engine.now() then
      ... timer code ...
      deadline = (deadline or engine.now()) + conf.interval
   end
   ... other push method app logic ...
end

This is simple and local and has no risk of the timer callback outliving the app.

Or?

PR #668, opened 2015-11-30
Add 128, 512, 1500 and 9000 byte packet test dumps

eugeneia:
This adds some more PCAP dumps (128, 512, 1500 and 9000 byte packet) for use with packetblaster.


lukego:
This seems sensible but it does highlight a limitation of the packetblaster UI i.e. the lack of a simple method to describe the packet(s) that should be sent. I suppose this could be done either inside packetblaster or with a utility to generate the pcap files programatically rather than as opaque binaries. (See also [Scapy](http://www.secdev.org/projects/scapy/)).

I suppose that improving packetblaster is a whole topic of its own though. One more feature that we are sorely lacking is to perform basic handshaking with a router (e.g. ARP or IPv6 neighbor discovery) so that we can generate load when there is a router between packetblaster and the system under test. (Have needed this in the past and worked around e.g. with a static configuration on the router.)

Lots of unrealized potential in the packetblaster program...

eugeneia:
@lukego I am thinking about making my winter intern extend packetblaster to generate its own packages. It seems like a well-sized task for an intern. :)

Issue #671, opened 2015-12-03
Fixed-function PMU counters not working?


andywingo:
For this lua file:

```lua
local pmu = require('lib.pmu')
local has_pmu_counters, err = pmu.is_available()
if not has_pmu_counters then
   error('No PMU available: '..err)
end

local function count_cycles(f, iterations)
   pmu.setup({"cycles"})
   local set = pmu.new_counter_set()
   pmu.switch_to(set)
   local res = f(iterations)
   pmu.switch_to(nil)
   pmu.report(pmu.to_table(set))
   return pmu.to_table(set).cycles, res
end

local function print_loop(count)
   for i=1,count do print(i) end
end

print(count_cycles(print_loop, 10))
```

If I run it like `sudo taskset -c 0 ./snabb snsh /tmp/test.lua` I get:

```
$ sudo taskset -c 0 ./snabb snsh /tmp/test.lua
1
2
3
4
5
6
7
8
9
10
EVENT                                             TOTAL
cycles                                                0
ref_cycles                                            0
instructions                                          0
cpl_cycles.ring0                                     13
cpl_cycles.ring0_trans                               13
cpl_cycles.ring123                               34,448
cycle_activity.cycles_l1d_pending                     0
cycle_activity.cycles_l2_pending                  5,442
cycle_activity.cycles_ldm_pending                42,706
cycle_activity.cycles_no_execute                 73,052
dsb2mite_switches.penalty_cycles                    374
0	nil
```

The cycles/ref_cycles/instructions are borked.  What is going on, I wonder?

I should note that I have this local modification:

```diff
diff --git a/src/lib/pmu.lua b/src/lib/pmu.lua
index 9255c33..68fd5f4 100644
--- a/src/lib/pmu.lua
+++ b/src/lib/pmu.lua
@@ -199,7 +199,7 @@ function setup (patterns)
       table.sort(set)
    end
    local ndropped = math.max(0, #set - pmu_x86.ngeneral)
-   if ndropped > 0 then set[pmu_x86.ngeneral+1] = nil end
+   while #set > pmu_x86.ngeneral do table.remove(set) end
    local cpu = cpu_set()[1]
    -- Enable all fixed-function counters (IA32_FIXED_CTR_CTRL)
    writemsr(cpu, 0x38d, 0x333)

```

I think just setting the element to nil doesn't truncate the length unless that's the last element.

andywingo:
[...]

PR #672, opened 2015-12-04
Fix bugs in PMU code

andywingo:
There were two bugs in the PMU code.  The first was that if the events
you select matched more than pmu_x86_ngeneral counters, the set of
events was being incorrectly truncated.  This would later manifest as an
error when we try to do defs[set[j+1]], as some random element in the
middle of `set' could be nil.

The other bug was that the global IA32_PERF_GLOBAL_CTRL register wasn't
being initialized.  One symptom could be that any given counter wasn't
counting -- for me it was the cycles fixed-function counter.  We didn't
observe this in the past because this is global state that could have
gotten initialized by any other program -- perhaps a previous invocation
of `perf`.

### DIFF
--- a/src/lib/pmu.lua
+++ b/src/lib/pmu.lua
@@ -199,8 +199,13 @@ function setup (patterns)
       table.sort(set)
    end
    local ndropped = math.max(0, #set - pmu_x86.ngeneral)
-   if ndropped > 0 then set[pmu_x86.ngeneral+1] = nil end
+   while (#set - pmu_x86.ngeneral) > 0 do table.remove(set) end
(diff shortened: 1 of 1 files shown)


andywingo:
Fixes #671

lukego:
Thanks! Looks like I manage to sneak this one into the `v2015.12` release. Indeed I had probably indirectly initialized the PMU with `perf` before running tests.

btw: I met an Intel engineer at a conference recently and he reminded me of their own stand-alone pmu software effort (separate from kernel perf framework): https://software.intel.com/en-us/articles/intel-performance-counter-monitor

PR #675, opened 2015-12-07
first approximation tap app

petebristow:
This app provides access to linux Tap devices. It uses ljsyscall rather than the old C stubs and removes those stubs.


src/apps/tap/tap.lua
```diff
@@ -0,0 +1,99 @@
...
+end
+
+function Tap:push ()
+   local l = self.input.input
+   while not link.empty(l) do
+      local p = link.front(l)
+      local len, err = S.write(self.sock, p.data, p.length)
+      if not len and err.errno ~= const.E.AGAIN or len and len ~= p.length then
+         -- How should this be bubbled up?
+         -- What does a partial write actually mean?
```
eugeneia: I think its safe to “die” upon a write error. If Linux breaks, we can't do anything about it. :) `assert` and `error` are fine!

src/apps/tap/tap.lua
```diff
@@ -0,0 +1,99 @@
...
+   end
+end
+
+function Tap:stop()
+   S.close(self.sock)
+end
+
+function selftest ()
+   -- How should this be implemented. 
+   -- Should it rely on RawSocket?
```
eugeneia: You can definitely use other apps in the selftest! E.g. other tests use the [pcap](https://github.com/SnabbCo/snabbswitch/tree/master/src/apps/pcap#pcapreader-and-pcapwriter-apps-appspcappcap) apps to test predefined packet streams and compare the output to a known result.

src/apps/tap/tap.lua
```diff
@@ -0,0 +1,99 @@
...
+      local p = packet.allocate()
+      local len, err = S.read(self.sock, p.data, C.PACKET_PAYLOAD_SIZE)
+      if not len and err.errno == const.E.AGAIN then 
+         packet.free(p)
+         return
+      end
+      if not len then
+         -- How should this be bubbled up?
+         assert(0, "something broke")
+      end
```
lukego: Good question! Handling unexpected I/O errors is not something we handle very sophisticatedly today.

Raising an error (like this assert) is probably a reasonable strategy. Really we need to restore the engine code that catches these and restarts individual apps. Restarting the app does seem like the most reasonable response to an unrecognized error (e.g. maybe somebody manually deleted the tap device with `tunctl` and it needs to be recreated). This would be the normal [crash-only software](https://en.wikipedia.org/wiki/Crash-only_software) or [let it crash](http://stackoverflow.com/questions/4393197/erlangs-let-it-crash-philosophy-applicable-elsewhere) design applied at the app level.

For non-fatal errors like the `EAGAIN` we should also bump `core.counter` counters. Today we are pretty slack with counters and that is a pity because they are crucial for understanding what is going on with a running system. I think that each app should have a well-defined set of counters (`tx_dropped`, etc) as part of its interface and documentation on equal footing to its configuration items. Maybe the Tap app could be an early adopter :-) up to you.

lukego: This is off at a tanget, but just as an example of the sort of counters we should be keeping in our apps in the future: [YANG model for IP router statistics](https://tools.ietf.org/html/draft-baill-netmod-yang-ip-stats-01).

If somebody builds a generic IP router on Snabb Switch then they will want to be able to derive precisely those counters from the apps they are using. So it is good for us to start having the standard IETF/YANG counter definitions in the back of our minds at least.

(The other important counters are from the standard SNMP MIBs but the YANG people are already onto that an have well-defined mappings e.g. in Section 3 "Relationship to IP MIB" of the internet draft linked above.)

petebristow:
I've removed the ability to set the MAC address, if it needs to be done it should be done with the standard linux commands, including it leads to a slippery slope of supporting all ip2 commands which would be crazy. 
There is a selftest which replays pcaps. 
TEST_SKIPPED is exported by the Makefile so it can be used in tests.
Additional comments have been added to make the code clearer.

Is using a network namespace reasonable to isolate tests? Where/how is the test suite run?

kbara:
It's handy to be able to set the mac address for things like test suites, without messing around with external commands; that's something I'll miss.

eugeneia:
@kbara The question is, is it worth adding the complexity in order to avoid a `selftest.sh` script like included in this PR?

kbara:
@eugeneia I'm in favour of having a selftest.sh. I'm just not convinced by removing under a dozen lines of code that provide a useful feature, which is replaced by having to run another program to do the same configuration.

With the lwaftr, we have a bunch of end-to-end tests that take and emit packet captures, and we can compare them really easily because the output is expected to always be bit-identical. Part of this is that we set mac addresses. As we explore more complicated test setups (potentially including tap devices), setting mac addresses remains functionality I'd be glad to have in anything we're using or potentially using that has mac addresses.

Test suites, including selftest() functions, routinely use other Snabb apps. Having to call out to a system-specific tool in the middle of those is doable, but...

lukego:
Somewhat zoomed-out thoughts stimulated by the PR:

- I am glad to drop the dependency on `lib.macaddress`. Could be that it is worth removing that module entirely to further our goal of having as little code and as few abstractions as possible? (Or is it important? Maybe we need to decide more broadly what qualifies for an abstract data type.)
- Network namespaces seem like a good solution for `selftest.sh` to avoid interfering with host networking. This does highlight the general problem that we don't have a clean/well-defined/isolated environment where the tests are executed. I wonder: should we define one and if so how?
- Promiscuous mode worth supporting, even as the default? To avoid the "ohhhhh, right, the packets were dropped because I forgot to put it into promiscuous mode" troubleshooting phenomenon :-)

petebristow:
@lukego 
* packet.free issue is fixed
* I'll open another CR to harmonize the macaddress parts and we can debate that there. 
* As a strawman how about 
```docker run -w /snabb -u root --privileged=true --rm=true -it -v $PWD:/snabb lts /snabb/apps/tap/selftest.sh```
  * Ubuntu LTS can be swapped out for other targets as required
  * A docker image of the target OS is a prerequisite but this aids testing against the target in future.
  * Builds need to be done inside the target container
* Can you expand a little on this. I'm not seeing the use case where packets will arrive with the 'wrong' mac address so promiscuous mode is required.

@kbara If other tests are using Tap I think they should have supporting selftest.sh that setup the environment before hand. For some of the other apps I'm working on it would be 'nice' to have support to set ipv4/v6 addresses and setting the MTU which would warrant inclusion if setting the mac makes it in. Then Tap turns into a probably buggy ifconfig reimplementation.

eugeneia:
@petebristow regarding docker: you can simply `scripts/dock.sh "(cd ..; make) && make test"`.

lukego:
@petebristow On reflection I don't have a use case for promiscuous mode. I suppose it makes sense for raw sockets and not for tap devices. If promiscous mode is needed on the tap device, e.g. because it is put into a bridge with `brctl`, then this should be done by the kernel and no the snabb process.

lukego:
@petebristow Great! This is merged onto the `next` branch for the next release now.

Please revise PRs by pushing new commits instead of rebasing. This is to make it safe for people to merge the PR branches into their own long-lived branches without waiting for a final version. (I also find it easier to follow what is happening on the branch when changes are made with new commits vs rewriting history but I realize that many wise people are the opposite.) More in [git-workflow.md](https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/git-workflow.md). If the workflow is suboptimal then people could also send a PR proposing changes to that document :-)

PR #677, opened 2015-12-11
next: Collected changes for the v2016.01 release

lukego:
Let the testing commence!


eugeneia:
How are we on the first 2016 release? I think there are some PRs that should be merged as well such as the little bug fixes ala #679

eugeneia:
A rich looking release candidate! @lukego unless you oppose I would do the release now.

PR #681, opened 2015-12-22
Virtio net

nnikolaev-virtualopensystems:
This is an implementation of the virtio-net driver part of the [virtio specification] (http://docs.oasis-open.org/virtio/virtio/v1.0/csprd04/virtio-v1.0-csprd04.html). The driver uses indirect descriptor tables and preallocated virtio headers for better performance. Checksum offload packet preparation is available as an optional feature.

It obsoletes #645.


src/lib/virtio/net_driver.lua
```diff
@@ -0,0 +1,171 @@
...
+local ETHERTYPE_IPv4 = C.htons(0x0800)
+local ETHERTYPE_IPv6 = C.htons(0x86DD)
+local ETHERTYPE_OFF = 12
+local ETHERLEN = 14 -- DST MAC | SRC MAC | ethertype
+local VIRTIO_NET_HDR_F_NEEDS_CSUM = 1
+
+local min_features =   C.VIRTIO_RING_F_INDIRECT_DESC + C.VIRTIO_NET_F_CSUM
+local want_features =  C.VIRTIO_F_ANY_LAYOUT +
+                       C.VIRTIO_RING_F_INDIRECT_DESC +
+                       C.VIRTIO_NET_F_CSUM +
```
andywingo: We don't actually want to negotiate checksum offloading, do we?

src/lib/virtio/net_driver.lua
```diff
@@ -0,0 +1,171 @@
...
+   virtio_pci:driver()
+
+   local error = virtio_pci:set_guest_features(min_features, want_features)
+   if error then
+      virtio_pci:free()
+      return nil, error
+   end
+
+   if debug then print("enumerating queues...") end
+   local vqs = {}
```
andywingo: seems that if this table starts from 0, it should be an ffi table; dunno, just a thought

src/lib/virtio/virtio_pci.lua
```diff
@@ -0,0 +1,155 @@
...
+   uint16_t queue_notify;
+   uint8_t status;
+   uint8_t isr;
+   uint16_t config_vector;
+   uint16_t queue_vector;
+]]
+
+local
+function open_bar (fname, struct)
+  local fd, err = S.open(fname, 'rdwr')
```
andywingo: can we indent by 3 spaces? :)

src/core/lib.lua
```diff
@@ -686,6 +686,38 @@ function getenv (name)
    else return nil end
 end
 
+function fstruct(def)
```
andywingo: My inner Max cringes at the thought of more undocumented functions here :)  To me, it would seem an easier merge to not add these to core.lib.

src/core/packet.lua
```diff
@@ -109,3 +114,21 @@ function preallocate_step()
    packet_allocation_step = 2 * packet_allocation_step
 end
 
+function dump(p, w)
```
andywingo: Likewise the dump function is probably not needed at all in core.packet, and we probably shouldn't be adding the metatype here -- or if we do, we should document it.  I believe these changes are from @javierguerragiraldez originally, and though they are likely very nice changes it will simplify our lives to not include them in this PR.  Seems to me, anyway; wdyt?

src/lib/checksum.lua
```diff
@@ -68,6 +68,62 @@ function verify_packet (buf, len)
    return ipsum(buf+headersize, len-headersize, initial) == 0
 end
 
+local
+function prepare_packet_l4( buf, len, csum_start, csum_off)
```
andywingo: I believe Luke's convention is "local function foo (a, b)".  Note local and function on the same line, and space before the paren, not after.  Also 2-space indent.

src/lib/checksum.lua
```diff
@@ -68,6 +68,62 @@ function verify_packet (buf, len)
...
+
+  local hwbuf =  ffi.cast('uint16_t*', buf)
+  local proto = buf[9];
+
+  local csum_start = lshift(band(buf[0], 0x0F),2)
+  local csum_off
+
+  -- Update the IPv4 checksum (use in-place pseudoheader, by setting it to 0)
+  hwbuf[5] = 0;
+  hwbuf[5] = C.htons(C.cksum_generic(buf, csum_start, 0));
```
andywingo: This would appear to call cksum_generic on all transmitted packets.  AFAIU we should not be doing this -- it's the app's responsibility.  We don't do this on intel cards and it seems there is no reason to do it on virtio interfaces.  Besides that, this calls the generic checksum routine, not the avx2 optimized one; at the least we should be using lib.checksum.ipsum instead.

src/lib/checksum.lua
```diff
@@ -98,7 +154,102 @@ end
...
+   local s = [[
```
andywingo: AFAIU this change is unrelated to the patch; feel free to submit separately, or not at all if you don't feel like submitting it.

src/lib/checksum.lua
```diff
@@ -68,6 +68,62 @@ function verify_packet (buf, len)
...
+  if proto == 6 then
+    csum_off = 16
+  -- UDP
+  elseif proto == 17 then
+    csum_off = 6
+  end
+
+  return prepare_packet_l4( buf, len, csum_start, csum_off)
+end
+
```
andywingo: A final note here -- I'm skeptical about the value of these functions being in lib.checksum.  I mean, it also seems that we don't need them, but if we *do* need them, to me they don't seem to be generic enough to warrant a place here, and instead we should define them closer to their use.

src/lib/virtio/net_driver.lua
nnikolaev-virtualopensystems: Right, this should depend on the `use_checksum` argument.

src/lib/virtio/virtio_pci.lua
nnikolaev-virtualopensystems: Of course :)

src/core/packet.lua
nnikolaev-virtualopensystems: Agree, the minimum requirement is the `physical` function. Everything else can go away.

andywingo:
In general, looking pretty good to me!  Some comments above.  Especially looking forward to a version that avoids checksumming :)

I realize this might not be a thing you can do quickly, but what are your thoughts about testing and benchmarking?  We should put some measure in place to make sure the functionality doesn't regress, and if we can, it would be good to ensure that performance doesn't regress either.  /Cc @lukego for his thoughts there.

src/lib/checksum.lua
nnikolaev-virtualopensystems: Will do, thanks.

src/core/lib.lua
nnikolaev-virtualopensystems: Maybe we just need the right documentation?
I mean, for sure these are generic functions, not specific to virtio or PCI.

src/core/lib.lua
```diff
@@ -686,6 +686,38 @@ function getenv (name)
...
+
+function fieldrd(field, fd)
+   local buf = ffi.typeof('$ [1]', field.ct)()
+   local r, err = fd:pread(buf, field.size, field.offset)
+   if not r then error(err) end
+   return buf[0]
+end
+
+function fieldwr(field, fd, val)
+   local buf = ffi.typeof('$ [1]', field.ct)()
```
andywingo: This function is a source of a trace abort.

e.g. in a -jv output there are these lines

```
[TRACE --- (20/0) zone.lua:38 -- NYI: unsupported variant of FastFunc ffi.typeof at lib.lua:713]
[TRACE --- (20/0) zone.lua:38 -- NYI: unsupported variant of FastFunc ffi.typeof at lib.lua:713]
[TRACE --- (20/0) zone.lua:38 -- NYI: unsupported variant of FastFunc ffi.typeof at lib.lua:713]
[TRACE --- (20/0) zone.lua:38 -- NYI: unsupported variant of FastFunc ffi.typeof at lib.lua:713]
[TRACE  22 (20/0) zone.lua:38 -- fallback to interpreter]
```

AFAIU This means that a trace 22, coming from trace 20, met something it couldn't compile, so it is falling back to the interpreter.  I don't know why it printed it like 4 times, did it try 4 times?  I don't know.  Probably so, thinking maybe if it tried again the trace would go somewhere else.  It means that this allocation won't be sunk, the whole thing will go on the interpreted path, etc.  It's only used in one place; let's just inline it into its caller.  Same with fstruct and fieldrd.

src/lib/virtio/virtq_driver.lua
```diff
@@ -0,0 +1,247 @@
...
+
+local debug = _G.developer_debug
+
+local ffi    = require("ffi")
+local C      = ffi.C
+local memory = require('core.memory')
+local band   = require('bit').band
+
+local VirtioVirtq = {}
+VirtioVirtq.__index = VirtioVirtq
```
andywingo: what does this do?

aperezdc: It sets the base table itself as metatable for created instances. It is a common Lua idiom in many projects, though in Snabb I have seen the `return setmetatable(obj, { __index == BaseObject })` used more often (if not always).

It can be argued that setting the base object as metatable (`Base = {}; obj.__index = Base`) as done in the patch saves the creation of one table per instance created. In the particular case of Snabb applications, there will never be huge amounts of them being created, so the gain is marginal.

src/lib/checksum.lua
nnikolaev-virtualopensystems: The checksum functionality is now optional.
The functions are meant to calculate partial checksums, and for me, this is where they belong.
For sure they are not virtio specific.

src/core/lib.lua
nnikolaev-virtualopensystems: The virtio PCI BAR0 access is done on initialization. At runtime, only the device notification is done. And with this approach it will be slow.
Unfortunately, the mmap of rersource0 for virtio-pci is not failing. It is not entirely clear why,though. 

By inlining - do you mean to just replace `ffi.typeof('$ [1]', field.ct)` with `ffi.typeof('struct pci_whatever')`?

src/apps/virtio_net/README.md
```diff
@@ -0,0 +1,27 @@
...
+With `VirtioNet` SnabbSwitch can be used as a virtual ethernet interface
+by *QEMU virtual machines*. When connected via a UNIX socket, packets can
+be sent to the virtual machine by transmitting them on the `rx` port and
+packets send by the virtual machine will arrive on the `tx` port.
+
+![VirtioNet](.images/VirtioNet.png)
+
+## Configuration
+
+The `VhostUser` app accepts a table as its configuration argument. The
```
eugeneia: Shouldn't it say “VirtioNet” here?

src/core/lib.lua
eugeneia: At least add comments that describe usage and purpose of these functions, as of now I have no idea what they even do. And functions that are not `local` to `core.lib` indeed require documentation in `src/README.md.src`. Without the documentation I can hardly tell where the functions belong.

@andywingo Undocumented public functions in `core.lib` are a bug, I'd appreciate any tips on which functions are inadequately documented.

src/core/packet.lua
```diff
@@ -94,6 +94,11 @@ end
 -- Return pointer to packet data.
 function data (p) return p.data end
 
+-- Return physical address of packet data
```
eugeneia: I don't think this function justifies extending the packet API. [virtual_to_physical](https://github.com/SnabbCo/snabbswitch/tree/master/src#memory-corememory) is public, what's wrong with:

```
local memory = require("core.memory")

phys = memory.virtual_to_physical(packet.data(p))
```

eugeneia:
Merged into `max-next`.

lukego:
I read somewhere - can't find where now - that @wingo said this virtio-net implementation has now been verified to meet the performance targets of the `snabb-lwaftr` application? In this case a hearty congratulations to all concerned!

wingo:
Yes!  When we were testing, we had a test setup in which one host process was providing the vhost-user interface for *two* NICs, not via Snabb-NFV but written in Snabb and using the Snabb vhost-user bits.  Turns out the performance was limited by the host and not the guest.  When running under snabb-nfv, we were able to hit our 4MPPS target for the lwaftr.  Excellent work @nnikolaev-virtualopensystems and team!

The test application in the guest, which was similar to the DPDK's l2fwd, was also seeing some trace aborts.  It turned out this was something to do with the guest's configuration.  I don't remember the details; @dpino knows more.  Anyway with a different guest program or even with the lwaftr we don't spend any time in the interpreter.

wingo:
Clarification: switching to snabb-nfv in the host provided the speedup because you have to run it in a core-per-NIC configuration, so the host side had two cores to service traffic instead of one.

lukego:
Super cool.

I would love to have a CI benchmark that is equivalent to the DPDK-VM one we already have but running Snabb with Virtio-net inside the VM. The entrypoint to that code is [packetblaster_bench.sh](https://github.com/SnabbCo/snabbswitch/blob/max-next/src/program/snabbnfv/packetblaster_bench.sh) if you guys are interested. That way we could keep track of VirtioNet performance both to catch regressions and also to see whether it can take advantage of optimizations that we make on the host underneath.

lukego:
Once more: This is a great piece of work!

Just reading the code now on the `max-next` branch and really enjoying it. Really nice touch with the diagram of the vrings :-)

Issue #648, opened 2015-10-29
Packet copies: Expensive or cheap?


lukego:
Just wanted to capture some thoughts about whether packet copies are fundamentally expensive or cheap.

Packet copies are a recurring theme in networking in general and in Snabb Switch in particular. In the olden days packet copies were prohibitively expensive and had to be avoided. In the future it looks like Intel's SIMD work should make packet copies trivially cheap. Currently we are somewhere in between - but where exactly?

Thought experiment: Consider the extreme case of processing 64-byte packets.

#### AVX512 (2017)

Here is the assembler code to copy a 64-byte packet on a CPU supporting AVX512 (now expected on Xeon CPUs in 2017) as an "unrolled loop":

```
vmovdqu zmm0, [src]
vmovdqu [dst], zmm0
```

That is two instructions: load from memory into the 64-byte `zmm0` register and then store the same value back to memory. These instructions are cheap: modern Intel CPUs can typically execute 2-3 such instructions *per cycle*. Note that the `u` in the instruction name says that these are *unaligned* memory accesses: Intel's latest engineering work has been to make unaligned SIMD operations fast so that programmers don't have to worry about it.

This looks like a really small and cheap bit of code. I find it hard to imagine trying to optimize this away even from an inner loop.

#### AVX2 (2014)

On an AVX2 CPU (Haswell) the situation is similar except that we need twice as many instructions because we only have 32-byte `ymm` registers:

```
vmovdqu ymm0, [src]
vmovdqu [dst], ymm0
vmovdqu ymm0, [src+32]
vmovdqu [dst+32], ymm0
```

This is four instructions now but these are also cheap (Haswell can execute more than one such instruction per cycle - see [Agner Fog's instruction tables](http://www.agner.org/optimize/instruction_tables.pdf)).

This also seems like a really tiny amount of work for a CPU to do. Hard to imagine this being a performance hotspot during packet processing.

#### Sandy Bridge (2009)

On AVX CPUs (Sandy Bridge) the situation is a little more complex. Certain instructions only support 16-byte registers and unaligned memory accesses are more expensive. However: these CPUs are dying out and it seems reasonable to expect that a new networking product developed today will rarely be deployed on a Sandy Bridge CPU.

#### Summary

Here is how it seems to me:

- Small packet copies should be extremely cheap today on Haswell and even cheaper in the future with AVX512.
- Added overhead of a "real" copy routine *should* be insignificant e.g. adding a loop to copy a variable amount of data.
[...]

PR #682, opened 2015-12-24
Fix build with Musl libc

justincormack:
Minor portability fixes that let Snabb build under Musl libc, tested on Alpine Linux.

### DIFF
--- a/gcc-preinclude.h
+++ b/gcc-preinclude.h
@@ -2,7 +2,12 @@
 // See: http://www.win.tue.nl/~aeb/linux/misc/gcc-semibug.html
 //      https://rjpower9000.wordpress.com/2012/04/09/fun-with-shared-libraries-version-glibc_2-14-not-found/
 
+#define _GNU_SOURCE 1
+
+#include <features.h>
+
 #ifndef __ASSEMBLER__
+#ifdef __GLIBC__
 __asm__(".symver memcpy,memcpy@GLIBC_2.2.5");
 #endif
-
+#endif
--- a/src/apps/solarflare/ef_vi.h
+++ b/src/apps/solarflare/ef_vi.h
@@ -1,3 +1,4 @@
+#include <time.h>
 
 /* etherfabric/base.h */
 
--- a/src/apps/vhost/vhost_user.c
+++ b/src/apps/vhost/vhost_user.c
@@ -3,6 +3,7 @@
 #include <fcntl.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <string.h>
 #include <sys/mman.h>
 #include <sys/socket.h>
 #include <sys/un.h>
--- a/src/core/lib.c
+++ b/src/core/lib.c
@@ -2,7 +2,7 @@
 #include <time.h>
 #include <sys/stat.h>
 #include <sys/time.h>
-#include <sys/unistd.h>
+#include <unistd.h>
 
 /* Return the current wall-clock time in nanoseconds. */
 uint64_t get_time_ns()
--- a/src/core/memory.c
+++ b/src/core/memory.c
@@ -9,7 +9,7 @@
 // virtual addresses (remove the tag bits) and to test addresses for
 // validity (check the tag bits).
 
-#define _GNU_SOURCE
+#define _GNU_SOURCE 1
 
 #include <assert.h>
 #include <fcntl.h>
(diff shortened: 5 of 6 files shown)


lukego:
Thinking about the different ways we could approach musl support in Snabb Switch:

1. Import musl as a subtree and make the master branch always build a statically-linked executable.
2. Support musl as our second first-class platform with full test coverage (including performance regression tests compared with glibc).
3. Support musl as a second-class platform with only basic test coverage e.g. CI uses docker to check that Snabb Switch compiles on Alpine Linux.
4. Just merge patches that look like they would help musl but without doing any upstream testing.

How about we celebrate the new year by being wild and crazy with option 1?

That would be to add musl as our fourth dependency (alongside luajit, ljsyscall, and pflua) and truly ship a statically linked executable that will run on any Linux/x86-64 machine.

The first step would be to send a PR that adds musl as a subtree and uses it for the build. This will probably trigger test failures e.g. I'm expecting the NFV application to have a performance regression because the inner loop depends on the libc memcpy to be optimized for packet copy workload (glibc memcpy is, musl memcpy is not). This probably should be resolved in our own code i.e. by having a suitable packet-copy routine of our own instead of depending on non-standard properties of the libc one.

I have had a quick look at musl. I appreciate the coding style i.e. really focusing on short and simple code. Compilation time is about 1 minute serially but only 5 seconds in parallel (measured on chur) so we would still hit our compile time targets if we assume `make -j`.

I believe there will be more issues e.g. that we can't use `dlsym` with a static link and this will break LuaJIT's FFI? This is something we would need to resolve before merge too.

Thoughts for/against? anybody willing to take the lead on such an integration e.g. to create a branch that compiles a static executable with musl imported as a git subtree?

justincormack:
A few thoughts

a. A standalone statically linked executable would be really nice, I really appreciate the ease of install of for example Go projects that do this now.
b. I would still recommend doing 4 first, ie merging this patch, as it is generally helpful, and makes it easier to experiment with how to get to 1.
c. What size and alignment for memcpy is most important? Guessing around 1k, maximally aligned would be typical. This is really a separate issue, maybe we should have our own memcpy code to be in control anyway, will look at performance.
d. I did do some work ages ago on dealing with the dlsym issue, by generating a dlsym function which looks like the normal one but just returns pointers to the static functions. See https://github.com/justincormack/ljsyscall/blob/master/examples/dl.c but it was only semi scripted as a proof of concept. I think this needs more thought in terms of a generic simple solution. Also in terms of generated code size, you lose some of the advantage of static linking if you simply link all symbols in, rather than just the ones you use, so it needs a bit of fine tuning. I think I might try a generic script to work with ljsyscall statically linked as a first pass on this.

So I think it might still be best to split this up into a few different activities even if aim is 1.

lukego:
Sounds like a plan! Merged this PR onto next.

Here is some data about the packet copy routine for us:

1. Primarily copying ~ 32 - 1500 bytes that is in cache.
1. Safe to round up copies to whole cache lines (either guarded with a predictable branch or actually guaranteed).
1. *May* have nice alignment. (Can potentially ensure this for `struct packet` but less certain for virtio-net buffers allocated by VMs.)
1. Caller will know whether src/dst overlap in memory and whether forwards or backwards copy is needed.

Question then is whether we should write a `memcpy()` or write one or more special purpose `packetcopy()` routines.

I have browsed memcpy implementations around the internet (dpdk, libc, musl, etc). The shortest path to a relatively simple and efficient one that matches our needs seems to be to take Agner Fog's design and remove the special cases that are not relevant for us. He seems to have struct a balance between performance and simplicity (well, I say that not having benchmarked it for our use case).

See also [memcpy odyssey](https://groups.google.com/forum/#!topic/snabb-devel/rbz7ipRAhlw) threads on snabb-devel and #648 (Packet copies: Expensive or cheap?).

lukego:
Good news re: memcpy is that we do have CI coverage for this. A bad memcpy should impact the performance regression tests for forwarding packets through a DPDK VM. So if that test passes then we probably have a winner.

(One weakness of that test is that it is using fixed-size power-of-2 packets. To be more confident of the memcpy we would want to vary the packet sizes. I have done this manually a few times with glibc memcpy and it seems to behave well.)

justincormack:
One thing that is different is that we write the tests for special cases in Lua so they are inlineable, so we don't need to run the tests at all in a tight loop if they are static. We can also use the built in luajit memcpy (which we could change if it is non optimal). So we need a slightly different test setup.

lukego:
LuaJIT only uses its own memcpy for fixed-size objects e.g. structs or constant-size arrays. Otherwise it will emit a call to libc memcpy. (This is much like GCC behavior.) The calls that we have to libc memcpy today are actually `ffi.copy()` with variable size data in the Lua sources.

PR #683, opened 2015-12-27
intel1g: Intel 1G driver (i350, i210)

hb9cwp:
This is an update to Luke's PR #561 
selftest() is now working with i350 NICs on chur, and with i210 NICs on new APU2 boards from PC Engines. It still requires more work and testing. But MAC loopback works without loosing too many frames at 1.47 Mpps of 60 Bytes per packet :-)

```
[hb9cwp@chur:~]$ sudo SNABB_SELFTEST_INTEL1G_0="0000:0c:00.1" ./snabb snsh -t apps.intel.intel1g
selftest: intel1g
table: 0x40bf9f60       table: 0x40bfa318       table: 0x410e89b8
link report:
           1,475,419 sent on nic.tx -> sink.rx (loss rate: 0%)
           1,476,506 sent on source.tx -> nic.rx (loss rate: 0%)
apps report:
selftest: ok
Stats from NIC registers:
  Rx Packets=        1476082  Octets= 94469248
  Tx Packets=        1476082  Octets= 94469248
  Rx Good Packets=   1476082
  Rx No Buffers=     0
  Rx Packets to Host=1476082
Stats from counters:
  rxPackets=         1475551  rxBytes= 88533060
  txPackets=         1476266  txBytes= 88575960
input link:  txpackets=         1476506   rxpackets=    1476266   txdrop=       0
output link: txpackets=         1475419   rxpackets=    1475419   txdrop=       132
```


hb9cwp:
More selftest() with i350 on chur:
```
[hb9cwp@chur:~/snabbswitch/hb9cwp/snabbswitch/src/program/top]$ ps ax|grep snabb
 1610 pts/0    S+     0:00 sudo SNABB_SELFTEST_INTEL1G_0=0000:0c:00.1 ./snabb snsh -t apps.intel.intel1g
 1611 pts/0    R+     1:00 ./snabb snsh -t apps.intel.intel1g

[hb9cwp@chur:~/snabbswitch/hb9cwp/snabbswitch/src/program/top]$ sudo snabb top 1611
Kfrees/s        freeGbytes/s    breaths/s      
 2979.00         0.18            159700         

 Links (rx/tx/txdrop in Mpps)    rx      tx      rxGb    txGb    txdrop 
 nic.tx -> sink.rx               1.49    1.49    0.09    0.09    0.00   
 source.tx -> nic.rx             1.49    1.49    0.09    0.09    0.00   

```

```
$ uname -a
Linux chur 4.2.5 #1-NixOS SMP Thu Jan 1 00:00:01 UTC 1970 x86_64 GNU/Linux

$ less /proc/cpuinfo
...
vendor_id       : GenuineIntel
cpu family      : 6
model           : 45
model name      : Intel(R) Xeon(R) CPU E5-2650 0 @ 2.00GHz
stepping        : 7
microcode       : 0x70d
cpu MHz         : 2799.687
cache size      : 20480 KB
...
bogomips        : 3999.98
```


selftest() with i210 on APU2b4 with Ubuntu 14.04.3 LTS:

```
r@a2b:~/snabbswitch/hb9cwp/snabbswitch/src/program/top$ sudo snabb top 1450
 Kfrees/s        freeGbytes/s    breaths/s      
 1451.50         0.09            3000           

 Links (rx/tx/txdrop in Mpps)    rx      tx      rxGb    txGb    txdrop 
 nic.tx -> sink.rx               0.68    0.68    0.04    0.04    0.02   
 source.tx -> nic.rx             0.75    0.75    0.05    0.05    0.00   
```

```
$ uname -a
Linux a2b 3.13.0-71-generic #114-Ubuntu SMP Tue Dec 1 02:34:22 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

$ less /proc/cpuinfo
...
vendor_id       : AuthenticAMD
cpu family      : 22
model           : 48
model name      : AMD GX-412TC SOC                               
stepping        : 1
microcode       : 0x7030105
cpu MHz         : 1000.000
cache size      : 2048 KB
...
bogomips        : 1996.09
TLB size        : 1024 4K pages
...
```

lukego:
Great work, Rolf! Merged onto `next`.

How are you finding the driver hacking experience? :-)

There is a bunch more that we can do with Intel drivers. I would like to use this simple 1G driver as the base to generalize up to the new features that we will want. Thinking along the lines of:

1. Support RSS to spread traffic between queues.
2. Support VMDq to spread traffic between queues (already works on intel10g driver).
3. Support X710/i40e (10G/40G) once available in the lab.
4. Support FM10K (100G) once available in the lab.
5. Support VF (SR-IOV) for multiple NIC families.

but one step at a time :-) and having I350 support is really great now! Supporting 1G for I/O is a hard requirement for many applications and even at this speed it is important to have kernel-bypass drivers.

Interesting to see some test results on another x86 platform than Xeon too...

hb9cwp:
Thanks Luke for merging and asking!

Digging into the intel1g app turned out to be a good learning exercise into both the internal workings of Snabb Switch, and Lua at the same time (but supposedly much less steep than DPDK would be :-) 
It took me a while to convince myself that your intel1g code was a draft that required more work. Initially, I could not understand from the code how the (shadow) descriptor ring buffers work. In particular, ```(can_)add_receive_buffer()``` were misleading until I figured out that they were "dead code", probably just pseudo-code to illustrate what the NIC does in hardware. And that the receiver was still lacking the handling of packet buffers. The last struggle was to figure out why RDH index was never advancing although the receiver received frames from the MAC loopback. The solution was to initialize RDT to (ndesc-1) instead of 0 in order to trigger the NIC to re-load descriptors into its cache...

Now, I want to add some input parameter checks, for ex. the i350 supports 8 rings each whereas the i210 only 4 each. Then I intend to add support for initializing the PHY in order to send and receive frames between interfaces/hosts over cables, instead of MAC loopback.
Yet, I am not sure how to do that in line with your intention with this new simple 1GE driver. First, I will study the intel10g_app, and I appreciate any guidance, hints and corrections to get that working.
My objective is to have a working intel1g app that can serve as basis for further experiments with L2TPv3 over IPsec not only on 10GE machines, but with 1GE NICs (i210, i350) as well.

lukego:
Sounds great! and sounds like an authentic driver-hacking experience :).

I think it is fine to focus on the immediate needs of 1G in that module. I would like to converge our various Intel drivers in the near future, and perhaps deprecate the existing intel10g one in favour of another one more like this 1G driver, but I think it is fine to write them as separate drivers to start with (1G, 10G, 40G, 100G) and unify them when the similarities emerge.

The similarities I have in mind are that each Intel NIC will do transmit/receive in basically the same way (TDT, RDT, etc), have <n> RX and TX queues that can be used independently, and have some means of dispatching packets between queues (e.g. RSS and VMDq). So we should be able to provide a fairly consistent interface to all the drivers and also share some code (e.g. transmit/receive but not PHY init).

I also think that reusable test suites will be really key. If the drivers have a uniform interface then we could reuse the test suites that check for basic TX/RX, PHY init (e.g. 100 times in a loop to catch timing bugs), RSS/VMDq dispatching onto the right queus, etc.

hb9cwp:
Great, so I push ahead with some input validation, elimination of possible race condition while reading RDH or TDH, and setting up PHY.

Is is worth to spend a few cycles on trying to apply some of my lessons learned to the intel10g app (see my footnote which I edited into my previous comment), or are the better spent on writing a new 10GE driver based on the intel1g skeletton, once this becomes more feature-complete and settles down?

hb9cwp:
Showing the 'model' is user-friendlier than hex codes of 'vendor' and 'device ID' (which 'lspci -nn' shows as well): 
```
[hb9cwp@chur:~/snabbswitch/hb9cwp/snabbswitch/src]$ ./snabb snsh -t lib.hardware.pci             
selftest: pci
pciaddress  model              interface  status  driver               usable
01:00.0     Intel 82599 SFP    -          -       apps.intel.intel10g  yes
01:00.1     Intel 82599 SFP    -          -       apps.intel.intel10g  yes
03:00.0     Intel 82599 SFP    -          -       apps.intel.intel10g  yes
03:00.1     Intel 82599 SFP    -          -       apps.intel.intel10g  yes
05:00.0     Intel 82599 SFP    -          -       apps.intel.intel10g  yes
05:00.1     Intel 82599 SFP    enp5s0f1   up      apps.intel.intel10g  no
07:00.0     Intel 82599 SFP    -          -       apps.intel.intel10g  yes
07:00.1     Intel 82599 SFP    enp7s0f1   up      apps.intel.intel10g  no
09:00.0     Intel 82599 SFP    enp9s0f0   up      apps.intel.intel10g  no
09:00.1     Intel 82599 SFP    enp9s0f1   up      apps.intel.intel10g  no
0c:00.0     Intel 350          enp12s0f0  up      apps.intel.intel1g   no
0c:00.1     Intel 350          -          -       apps.intel.intel1g   yes
82:00.0     Intel 82599 SFP    enp130s0f0 up      apps.intel.intel10g  no
82:00.1     Intel 82599 SFP    enp130s0f1 up      apps.intel.intel10g  no
86:00.0     Intel 82599 SFP    enp134s0f0 down    apps.intel.intel10g  yes
86:00.1     Intel 82599 SFP    enp134s0f1 down    apps.intel.intel10g  yes
88:00.0     Intel 82599 SFP    enp136s0f0 up      apps.intel.intel10g  no
88:00.1     Intel 82599 SFP    enp136s0f1 up      apps.intel.intel10g  no
```

lukego:
> Is is worth to spend a few cycles on trying to apply some of my lessons learned to the intel10g app

It would be interesting to know what you have in mind. Perhaps propose via an Issue or draft PR anyway?

Overall though I like the idea of evolving a new driver style separately and then once this is mature we can reconcile it with the intel10g driver. I will open an issue to talk about another proposed new requirement on I/O drivers that may be hard to retrofit onto the existing intel10g driver.

lukego:
This branch seems to have lost the documentation that I had on mine. Can you reintroduce it please?

lukego:
I wrote up the idea of requiring these I/O drivers to support multiple Snabb processes separately attaching to transmit/receive queues on the same device (with one process being responsible for initialization) in #687. What do you think: doable?

hb9cwp:
I've now pulled in and pushed your README.*. Is that all that was missing?

Also, I observe that CI check fails repeadly. But it looks as if it is related to SnabbNFV which I have not touched?

lukego:
@eugeneia do you see why the CI is failing on this latest commit?

hb9cwp:
Backed out my renaming of the intel10g driver name which breaks CI...

lukego:
@hb9cwp What do you think about #689? Is that design realistic? I mean for two different processes to attach to the same card, and somehow initialize it correctly, without undue synchronization (e.g. avoiding complicated dependency on which process starts first or IPC or ...).

I am wondering how to neatly handle the case when for example one process attaches to a TX queue on the NIC and then another process reset the NIC afterwards. Can the TX process cleanly detect the reset and restore the queue for example? I am not sure if this is realistic. It would surely be nice and simple.

hb9cwp:
@lukego Still feeling as a newcomer and still lacking enough insight to give you a good answer right now. 
However, at this very moment, I am working on building separate transmitter and receiver programs to test drive MAC/PHY parts of the intel1g app, e.g. do simplex one-way transmissions, adding sequence numbers to the packet payloads of an extended basic.Source, and evtl. some packet loss and bit errors. Also, I might try to split in half intel1g into tx and rx paths somewhere along your outline. Perhaps I will have an opinion about your idea after that experience.

eugeneia:
@lukego I think we merged this at an unfortunate point, the next release will have to follow up with the unmerged commits in this PR following 09ac90a (including the documentation besides other things).

hb9cwp:
@lukego Paraphrasing your design idea using your new triangle notation from #684 for sending minimal packets with 60 Byte dummy payload over a cable between two hosts that have (at least) one NIC each, and eventually maps two processes onto two cores on each host: 
snabbhalfduplex
So the key question is if there is a reasonable implementation which splits a NIC triangle into two separate tx and rx sub-triangles along the shagged lines in red.

lukego:
Great picture!

Yes. To make it more complicated too:

- Can be that you have many instances of those triangles e.g. if the NIC supports 128 transmit and receive queues then you may have <N> processes each wanting to attach to <M> queues.
- Can be that setting up the queues in the right way (e.g. telling the NIC to dispatch a specific MAC address to a specific queue) requires some synchronization (can't have two processes doing this at once because they would be touching the same registers).

I think that initialization is the hard part, at least for Intel NICs, because once that is done they do have a separate copy of all the TX/RX registers (TDH, TDT, etc) for each queue at a different address and so processes can access them directly/separately/safely.

eugeneia:
Merged into `max-next` (up to f19c5c1).

hb9cwp:
@eugeneia Just backed out that debug code in ```apps.basic.basic_apps```. 
Sorry for the delay, because last week I had focus on getting ```l2vpn``` from Alex to work over ```ipsec.esp``` with  over ```intel1g```... 
During that work, I have spotted another performance degradation in the combination of ```l2vpn``` with ```intel1g``` (it uses almost 100% CPU of one core, even if the bridge is idling). I search for the culprit now. Thus consider ```intel1g```as "working", but still work-in-progress :-)

src/apps/basic/basic_apps.lua
```diff
@@ -4,6 +4,7 @@ local app = require("core.app")
 local freelist = require("core.freelist")
 local packet = require("core.packet")
 local link = require("core.link")
+--local struct = require("struct")
```
eugeneia: @hb9cwp You missed this line. :-)

hb9cwp: @eugeneia Done, thanks.

PR #690, opened 2015-12-31
Add test sink a basic pcap replay testing framework

petebristow:
Rather than reimplement pcap replay in every selftest provide a helper function to do it.
Does this look useful to others?


eugeneia:
Looks polished and I understand the motivation for `TestSink`. On the other hand I have never really longed for `TestSink` (as writing these kind of tests is straight forward) and I am averse to fixing a problem that does not exist.

One way of disproving me would be to convert the test cases to which `TestSink` applies to use `TestSink` (you could amend the conversions to this PR). The way I see it, if the result is less code in total and prettier to look at, then my skepticism is refuted, and if we want to adopt `TestSink` we'd want to convert the applicable selftests anyways. I tried it for `keyed_ipv6_tunnel` to get a feeling:

```diff
diff --git a/src/apps/keyed_ipv6_tunnel/tunnel.lua b/src/apps/keyed_ipv6_tunnel/tunnel.lua
index 9bf92ea..e638deb 100644
--- a/src/apps/keyed_ipv6_tunnel/tunnel.lua
+++ b/src/apps/keyed_ipv6_tunnel/tunnel.lua
@@ -239,11 +239,11 @@ end
 prepare_header_template()
 
 function selftest ()
+   local testsink = require("apps.testsink.testsink")
    print("Keyed IPv6 tunnel selftest")
    local ok = true
 
    local input_file = "apps/keyed_ipv6_tunnel/selftest.cap.input"
-   local output_file = "apps/keyed_ipv6_tunnel/selftest.cap.output"
    local tunnel_config = {
       local_address = "00::2:1",
       remote_address = "00::2:1",
@@ -253,19 +253,20 @@ function selftest ()
    } -- should be symmetric for local "loop-back" test
 
    local c = config.new()
+   config.app(c, "test", testsink.TestSink, { fuzzy = false })
    config.app(c, "source", pcap.PcapReader, input_file)
+   config.app(c, "tee", basic_apps.Tee)
    config.app(c, "tunnel", SimpleKeyedTunnel, tunnel_config)
-   config.app(c, "sink", pcap.PcapWriter, output_file)
-   config.link(c, "source.output -> tunnel.decapsulated")
+   config.link(c, "source.output -> tee.input")
+   config.link(c, "tee.input -> tunnel.decapsulated")
    config.link(c, "tunnel.encapsulated -> tunnel.encapsulated")
-   config.link(c, "tunnel.decapsulated -> sink.input")
+   config.link(c, "tunnel.decapsulated -> test.rx")
+   config.link(c, "tee.output -> test.comparator")
    app.configure(c)
 
    app.main({duration = 0.25}) -- should be long enough...
    -- Check results
-   if io.open(input_file):read('*a') ~=
-      io.open(output_file):read('*a')
-   then
+   if #engine.app_table.test.errs > 0 then
       ok = false
    end
```

In the above case (which I chose randomly) I feel that `TestSink` is detrimental: more abstraction to know of, requirement for `Tee` to reproduce the particular case, one more line in total. The question is if it is worth grokking the `TestSink` and `harness` APIs for the provided functionality. Maybe I am not using it properly though...

Could it be that the main functionality here is the `fuzzy` option? Generally, I like the idea of an app to match/compare packet streams, and writing a file to disk to compare seems backwards. Could this PR be stripped down to a simple, orthogonal app that frees us of the “compare pcap dumps” pattern?

petebristow:
@eugeneia 
Great review thank you.

Harness doesn't play well if you need more supporting apps to correctly test the candidate app. 
Is there or should there be a wider framework?

TestSink was a response to the following
* selftest() in socket/raw.lua looks painful for a replay type test
* selftest() in socket/raw.lua ignores the race condition of interacting with linux, `fuzzy` helps here
* selftest() in tap/tap.lua doesn't work as it miscalls `C.memcmp`
* selftest() in tap/tap.lua tries to reimplement `engine.main()` badly
* not calling `engine.configure(config.new())` between tests can leave debris from previous test runs, counters for example
* A pcapfile1 == pcapfile2 doesn't given any indication which packets are broken or how.
* The required boilerplate leads to shortcuts, along the lines of  `foo == encode(decode(foo))`, which was used in `selftest` in tunnel.lua. Which has caught me out in the past where both encode and decode have the same error. `harness` doesn't support directly messing with `config`
* There are at least 2 approaches to test failure, os.exit(), error(). It might be nice to have something a bit more structured. 

```lua
-- I've removed the simple benchmark as it doesn't really test anything.
function selftest ()
   print("Keyed IPv6 tunnel selftest")
   local tunnel_config = {
      local_address = "00::2:1",
      remote_address = "00::2:1",
      local_cookie = "12345678",
      remote_cookie = "12345678",
      default_gateway_MAC = "a1:b2:c3:d4:e5:f6"
   } -- should be symmetric for local "loop-back" test

   local testsink = require("apps.testsink.testsink")
   assert(testsink.harness(SimpleKeyedTunnel, tunnel_config,
      { "dencapsulated" = "apps/keyed_ipv6_tunnel/selftest.cap.input" }, 
      { "encapsulated"  = "apps/keyed_ipv6_tunnel/selftest.cap.output" }
   ))

   assert(testsink.harness(SimpleKeyedTunnel, tunnel_config
      { "encapsulated"  = "apps/keyed_ipv6_tunnel/selftest.cap.output" },
      { "dencapsulated" = "apps/keyed_ipv6_tunnel/selftest.cap.input" }
   ))
end
```
```lua
-- tap.lua
function selftest()
-- tapsrc and tapdst are bridged together in linux. Packets are sent out of tapsrc and they are expected
-- to arrive back on tapdst. Linux may create other control-plane packets so to avoid races if a packet 
-- doesn't match the one we just sent keep looking until it does match. 
-- The linux bridge does mac address learning so some care must be taken with the preparation of 
-- selftest.cap A mac address should appear only as the source address or destination address
-- This test should only be run from inside apps/tap/selftest.sh
   if not os.getenv("SNABB_TAPTEST") then os.exit(engine.test_skipped_code) end

   local testsink = require("apps.testsink.testsink")
   assert(testsink.harness(Tap, {},
      { "input" = "apps/tap/selftest.cap" }, 
      { "output"  = "apps/tap/selftest.cap" },
      { fuzzy = true }
   )
end
```

eugeneia:
Great list of pain-points!

> selftest() in socket/raw.lua looks painful for a replay type test

I actually think its OK looking. It uses a handful of well-known Snabb-isms (lib.protocol.datagram, core.link, push/pull API), mind that maybe I am too used to reading this already but consider this: it's a white-box test with a transparent implementation, it doesn't hide what's going on behind abstractions, yet it reuses core Snabb concepts effectively.

> selftest() in socket/raw.lua ignores the race condition of interacting with linux, fuzzy helps here

Strongly agree. I think a packet matching app like TestSink could be a great tool for the chest.

> selftest() in tap/tap.lua doesn't work as it miscalls `C.memcmp`

I don't think it does? Anyways this danger would indeed be eliminated by a packet matching app.

> selftest() in tap/tap.lua tries to reimplement `engine.main()` badly

I don't think that's the case, its a white-box test that transparently exercises the app API.

> not calling `engine.configure(config.new())` between tests can leave debris from previous test runs, counters for example

Usually not an issue because `make test` runs each test in its own instance. It shouldn't be necessary to call `engine.configure(config.new())` either as any call to `configure` will delete apps that are not in the new configuration.

> A pcapfile1 == pcapfile2 doesn't given any indication which packets are broken or how.

Agree, good argument for a specialized packet matching app.

> The required boilerplate leads to shortcuts, along the lines of `foo == encode(decode(foo))`, which was used in selftest in tunnel.lua. Which has caught me out in the past where both encode and decode have the same error. harness doesn't support directly messing with config

`foo == encode(decode(foo))` tests if encode/decode preserve the input and nothing else that's true. I disagree with the notion that its the boilerplate's fault for not having written a more elaborate test, it is certainly trivial to extend it. While `harness` might simplify this particular case, I am worried about future tests being forced to fit into the `harness` API. In free form tests like we have now, nothing keeps you from writing types of tests that haven't been thought of before. Once `harness` is widespread on the other hand, people will think in its terms. Thinking about how to extend the `harness` API for a new use-case is imho more mental overhead than just writing the (well-understood) boilerplate.

> There are at least 2 approaches to test failure, os.exit(), error(). It might be nice to have something a bit more structured.

I would advise to always use `error` and `assert` when possible as the stack trace is handy, but sometimes `os.exit` is more fitting because you want to run tests even if previous tests failed and exit with a final status or skip tests.

I am pro `TestSink` (maybe rename it to just `Match`) and contra `harness`. I am personally a fan of “no curtains”, bottom-up software. Abstractions should be small and orthogonal as I want to build more complex programs from small, easy to understand, pieces. I already know that I will need to read the source code of `harness` 100 times because every time I debug a test I have to recap how `harness` works because *it hides how the test works*, and then forget about it again. Furthermore I fear that it will need to be extended with new requirements for yet unknown tests and I bet that will lead to either a monster or throwing it away again and replacing it with something else.

`Match` could be a very generic and useful app that fits in with the spirit of `basic_apps`. Its documentation has to be proofread and improved though. Maybe its interface should be more extendable (for other possible matching modes).

```lua
config.app(c, "m", test.Match , { mode = "strict"|"fuzzy"|... }) -- important to define a default for `mode'.
```

In any case, “fuzzy” and “strict” have to be better defined in the documentation so that ots crystal clear to use.

As to `harness`: Maybe we can find smaller, orthogonal solutions for the problems that it solves which then can be composed with existing apps. I am not against what `harness` does but how it does it: it does not compose.

PR #692, opened 2016-01-03
Basic benchmarking tool with json output.

petebristow:
Further to [688](https://github.com/SnabbCo/snabbswitch/issues/688)
This is a discussion point rather than a 'pull request'.

This provides go like benchmark functions that can be embedded in apps with json output. 
They are run from snabbmark. The work flow is envisaged to be something like

`make && sudo ./snabb snabbmark lib apps.basic.basic_apps >> test.json`
Make changes 
`make && sudo ./snabb snabbmark lib apps.basic.basic_apps >> test.json`

Then use jq to extract and compare benchmark runs.
jq -s 'sort_by(.date) | .[] | select(.hostname == "dock" and .test == "benchmarkSourceToSink" ) | "(.date) (.links.input.packets)"' tests.json

Import version 1.0.0 of json4lua and bring back encode
Add benchmarks for Tee and Source
Allow snabbmark to run benchmarks

### DIFF
--- a/src/apps/basic/basic_apps.lua
+++ b/src/apps/basic/basic_apps.lua
@@ -33,6 +33,12 @@ function Source:stop ()
    packet.free(self.packet)
 end
 
+function benchmarkSourceToSink (c)
+   config.app(c, "source", Source)
+   config.link(c, "source.output -> sink.input")
+   return c
+end
+
 --- # `Join` app: Merge multiple inputs onto one output
 
 Join = {}
@@ -113,6 +119,16 @@ function Tee:push ()
          end
       end
    end
+end
+
+function benchmarkTee (c)
+   config.app(c, "source", Source)
+   config.app(c, "t1", Tee)
+   config.app(c, "t2", Tee)
+   config.link(c, "source.output -> t1.input")
+   config.link(c, "t1.output -> t2.input")
+   config.link(c, "t2.output -> sink.input")
+   return c
 end
 
 --- ### `Repeater` app: Send all received packets in a loop
--- a/src/lib/benchmark.lua
+++ b/src/lib/benchmark.lua
@@ -1,0 +1,83 @@
+module(..., package.seeall)
+
+--[[
+	Things that could be in the json blob
+	-- cpuid
+	-- git commit id
+	-- packet size
+	-- src pcap file
+	-- pcap md5/sha1
+	-- uname
+	-- libc version
+	-- description from benchmark definiton
+	-- comments for benchmark run, eg 'measure impact of adding 1/10/40 counters to link structure'
+]]
+
+function runbenchmarks(module, bname)
+	local vl = require(module)
+	local basic = require("apps.basic.basic_apps")
+	local counter = require("core.counter")
+	local pcap = require("apps.pcap.pcap")
+	local ffi = require("ffi")
+	local C = ffi.C
+
+	local rets = { }
+
+	local blaster = function(self, name, pcapfile)
+		config.app(self,  ("%s_pcap"):format(name), pcap.PcapReader, pcapfile)
+		config.app(self, name, basic.Repeater)
+		config.link(self, ("%s_pcap.output -> %s.input"):format(name,name))
+	end
+	local senv = {}
+	local S = require("syscall")
+	for k,v in pairs(S.environ()) do
+		if k:match('^SNABB_') then 
+			senv[k] = v
+		end
+	end
+	if not bname then
+		bname = "benchmark"
+	end
+	for i,v in pairs(vl) do
+		if i:find( ("^%s"):format(bname) ) then
+			engine.configure(config.new())
+			local c = config.new()
+			c.dur = 5
+			c.src = blaster
+			c.busywait = false
+			c.Hz = false
+			config.app(c, "sink", basic.Sink)
+			c = v(c)
+			engine.configure(c)
+			engine.busywait = c.busywait
+			engine.Hz = c.Hz
+			local start = C.get_monotonic_time()
+			engine.main({ duration = c.dur, no_report = true })
+			local finish = C.get_monotonic_time()
+			stats = {}
+			for i,v in pairs(engine.link_table) do
+				local l = i:match('-> sink%.(.*)')
+				if l then
+					stats[l] = {}
+					stats[l].packets = tonumber(counter.read(v.stats.rxpackets))
+					stats[l].bytes = tonumber(counter.read(v.stats.rxbytes))
+				end
+			end
+			table.insert(rets, {
+				duration = {
+					requested = c.dur,
+					actual = finish - start
+				},
+				busywait = c.busywait,
+				hz = c.Hz,
+				module = module,
+				test = i,
+				links = stats,
+				env = senv,
+				hostname = S.gethostname(),
+				date = os.date("!%Y-%m-%dT%TZ")
+			})
+		end
+	end
+	return rets
+end
--- a/src/lib/json.lua
+++ b/src/lib/json.lua
@@ -1,20 +1,45 @@
+-----------------------------------------------------------------------------
 -- JSON4Lua: JSON encoding / decoding support for the Lua language.
 -- json Module.
 -- Author: Craig Mason-Jones
--- Homepage: http://json.luaforge.net/
--- Version: 0.9.40
+-- Homepage: http://github.com/craigmj/json4lua/
+-- Version: 1.0.0
 -- This module is released under the MIT License (MIT).
+-- Please see LICENCE.txt for details.
 --
--- NOTE: This is only the decode functionality ripped out from JSON4Lua.
--- See: https://github.com/craigmj/json4lua
+-- USAGE:
+-- This module exposes two functions:
+--   json.encode(o)
+--     Returns the table / string / boolean / number / nil / json.null value as a JSON-encoded string.
+--   json.decode(json_string)
+--     Returns a Lua object populated with the data encoded in the JSON string json_string.
+--
+-- REQUIREMENTS:
+--   compat-5.1 if using Lua 5.0
+--
+-- CHANGELOG
+--   0.9.20 Introduction of local Lua functions for private functions (removed _ function prefix).
+--          Fixed Lua 5.1 compatibility issues.
+--      Introduced json.null to have null values in associative arrays.
+--          json.encode() performance improvement (more than 50%) through table.concat rather than ..
+--          Introduced decode ability to ignore /**/ comments in the JSON string.
+--   0.9.10 Fix to array encoding / decoding to correctly manage nil/null values in arrays.
+-----------------------------------------------------------------------------
 
-module(..., package.seeall)
-
+-----------------------------------------------------------------------------
+-- Imports and dependencies
+-----------------------------------------------------------------------------
 local math = require('math')
 local string = require("string")
 local table = require("table")
 
-local base = _G
+-----------------------------------------------------------------------------
+-- Module declaration
+-----------------------------------------------------------------------------
+local json = {}             -- Public namespace
+local json_private = {}     -- Private namespace
+
+-- Public functions
 
 -- Private functions
 local decode_scanArray
@@ -24,6 +49,65 @@ local decode_scanNumber
 local decode_scanObject
 local decode_scanString
 local decode_scanWhitespace
+local encodeString
+local isArray
+local isEncodable
+
+-----------------------------------------------------------------------------
+-- PUBLIC FUNCTIONS
+-----------------------------------------------------------------------------
+--- Encodes an arbitrary Lua object / variable.
+-- @param v The Lua object / variable to be JSON encoded.
+-- @return String containing the JSON encoding in internal Lua string format (i.e. not unicode)
+function json.encode (v)
+  -- Handle nil values
+  if v==nil then
+    return "null"
+  end
+
+  local vtype = type(v)
+
+  -- Handle strings
+  if vtype=='string' then
+    return '"' .. json_private.encodeString(v) .. '"'     -- Need to handle encoding in string
+  end
+
+  -- Handle booleans
+  if vtype=='number' or vtype=='boolean' then
+    return tostring(v)
+  end
+
+  -- Handle tables
+  if vtype=='table' then
+    local rval = {}
+    -- Consider arrays separately
+    local bArray, maxCount = isArray(v)
+    if bArray then
+      for i = 1,maxCount do
+        table.insert(rval, json.encode(v[i]))
+      end
+    else  -- An object, not an array
+      for i,j in pairs(v) do
+        if isEncodable(i) and isEncodable(j) then
+          table.insert(rval, '"' .. json_private.encodeString(i) .. '":' .. json.encode(j))
+        end
+      end
+    end
+    if bArray then
+      return '[' .. table.concat(rval,',') ..']'
+    else
+      return '{' .. table.concat(rval,',') .. '}'
+    end
+  end
+
+  -- Handle null values
+  if vtype=='function' and v==null then
+    return 'null'
+  end
+
+  assert(false,'encode attempt to encode unsupported type ' .. vtype .. ':' .. tostring(v))
+end
+
 
 --- Decodes a JSON string and returns the decoded value as a Lua data structure / value.
 -- @param s The string to scan.
@@ -31,10 +115,10 @@ local decode_scanWhitespace
 -- @param Lua object, number The object that was scanned, as a Lua table / string / number / boolean or nil,
 -- and the position of the first character after
 -- the scanned JSON object.
-function decode(s, startPos)
+function json.decode(s, startPos)
   startPos = startPos and startPos or 1
   startPos = decode_scanWhitespace(s,startPos)
-  base.assert(startPos<=string.len(s), 'Unterminated JSON encoded object found at position in [' .. s .. ']')
+  assert(startPos<=string.len(s), 'Unterminated JSON encoded object found at position in [' .. s .. ']')
   local curChar = string.sub(s,startPos,startPos)
   -- Object
   if curChar=='{' then
@@ -77,14 +161,14 @@ end
 -- @param startPos The starting position for the scan.
 -- @return table, int The scanned array as a table, and the position of the next character to scan.
 function decode_scanArray(s,startPos)
-  local array = {}        -- The return value
+  local array = {}  -- The return value
   local stringLen = string.len(s)
-  base.assert(string.sub(s,startPos,startPos)=='[','decode_scanArray called but array does not start at position ' .. startPos .. ' in string:\n'..s )
+  assert(string.sub(s,startPos,startPos)=='[','decode_scanArray called but array does not start at position ' .. startPos .. ' in string:\n'..s )
   startPos = startPos + 1
   -- Infinite loop for array elements
   repeat
     startPos = decode_scanWhitespace(s,startPos)
-    base.assert(startPos<=stringLen,'JSON String ended unexpectedly scanning array.')
+    assert(startPos<=stringLen,'JSON String ended unexpectedly scanning array.')
     local curChar = string.sub(s,startPos,startPos)
     if (curChar==']') then
       return array, startPos+1
@@ -92,8 +176,8 @@ function decode_scanArray(s,startPos)
     if (curChar==',') then
       startPos = decode_scanWhitespace(s,startPos+1)
     end
-    base.assert(startPos<=stringLen, 'JSON String ended unexpectedly scanning array.')
-    object, startPos = decode(s,startPos)
+    assert(startPos<=stringLen, 'JSON String ended unexpectedly scanning array.')
+    object, startPos = json.decode(s,startPos)
     table.insert(array,object)
   until false
 end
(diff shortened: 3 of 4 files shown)


lukego:
Great initiative!

> This provides go like benchmark functions

I am not familiar with how benchmarking is done in Go. Could you write a README that summarizes the idea and links to relevant references e.g. to the Go feature?

I like the idea of including engine parameters in the `config` object. Really we should unify all of our configuration - engine, programs, apps, links - into a tree structure that can be modelled in [YANG](https://tools.ietf.org/html/rfc6020). Then we could have quite complete external representations e.g. load/store fragments as JSON. (I'll file a separate issue with that idea.)

petebristow:
# Benchmarking
Solid benchmark coverage means we can make changes with confidence there aren't negative impacts. This benchmark module is inspired by the [golang](https://golang.org/pkg/testing/) testing module. It supports placing benchmark specifications along side the code being benchmarked. If benchmarking / testing is made sufficiently painless it will be more widely adopted.

Json was chosen as it gives flexibility that csv doesn't. 
* `SNABB_DPDK` is important to NFV but irrelevant to benchmarkTee
* Stats on multiple links may be of interest to some benchmarks

## snabbmark
`./snabb snabbmark lib modulename [funcRegex]` loads `modulename` and benchmarks all functions matching `funcRegex` which defaults to `bechmark.*`
It is expected that benchmark specific configuration be passed through environment variables. Environment variables should also contain any other information pertinent to the benchmark such as DPDK version. Whilst this information is not used by snabbmark directly it needs to be included in the resultant json result.
```
SNABB_PCI0=0000:01:00.0
SNABB_TELNET1=5001
SNABB_DPDK=1.7
```
Benchmark results are output as json objects, 1 object per result.  
```json
{
  "duration": {
    "actual": 5.000008995994,
    "requested": 5
  },
  "hostname": "dock",
  "date": "2016-01-01T03:05:51Z",
  "hz": false,
  "module": "apps.basic.basic_apps",
  "links": {
    "input": {
      "bytes": 18651403800,
      "packets": 310856730
    }
  },
  "test": "benchmarkSourceToSink",
  "busywait": false,
  "env": { 
  }
}
```

## Embedded benchmarks
Any function of the form
`function benchmarkXXX(c) return c end`
will be timed by snabbmark. `c` is a `core.config` object that describes the benchmark with the following additions.
* c.duration is passed to engine.main
* c.Hz is passed to engine.Hz
* c.busywait is passed to engine.busywait
* c:src("appName", "pcapfile") will create a PcapReader / basic_apps.Repeater app pair such that `appName.output` is a continuous loop of the packets from pcapfile
* `config.app(c, "sink", basic_apps.Sink)` has already been created. Packet and byte counts are reported for the input links of "sink". 

```lua
function benchmarkTee (c)
   config.app(c, "source", Source)
   config.app(c, "t1", Tee)
   config.app(c, "t2", Tee)
   config.link(c, "source.output -> t1.input")
   config.link(c, "t1.output -> t2.input")
   config.link(c, "t2.output -> sink.input")
   return c
end
```

## (lib.benchmark)
`lib.benchmark.runbenchmarks`

lukego:
Thanks for writing this up!

I am definitely sold on having a low-ceremony way to define benchmarks that can be run in a generic way and produce data in a uniform format that we can track.

Can we try the simplest thing that could possibly work though? How about if the benchmark functions take no parameters and return no values, but just do some work and we record how long it takes? If they need to run an app network or tweak engine parameters then they can do that directly (or via library functions).

petebristow:
What about something like the following. I couldn't resist passing something into the benchmark function.

```lua
local ffi = require('ffi')
local C = ffi.C
local lib = require('core.lib')
local S = require("syscall")

local B = {}
function B:new()
	local cpuinfo = lib.readfile("/proc/cpuinfo", "*a")
	return setmetatable({
	        -- use data from issue 478
		build = lib.readcmd('git rev-parse HEAD', '*l'),
		hostname = S.gethostname(),
		sse2 = (cpuinfo:match('sse2') == "sse2"),
		avx2 = (cpuinfo:match('avx2') == "avx2"),
                date = os.date("!%Y-%m-%dT%TZ"),
		cpumodel = cpuinfo:match('model name%s:%s([^\n]*)')
	}, { __index = B })
end
function B:start()
	self.start_time = C.get_monotonic_time()
end
function B:finish()
	local finish = C.get_monotonic_time()
	if self.duration then return end
	self.duration = finish - self.start_time
	self.start_time = nil
end
function benchmark(f)
	local b = B:new()	    
	b:start()
 	f(b)
 	b:finish()
    return b
end
function jbenchmark(benchmarks)
	local enc = require("lib.json").encode
	for _, f in pairs(benchmarks) do
		print(enc(benchmark(f)))
	end
end

function benchmarkChecksums()
   	local initial = 0
	local n = 1000000
	local array = ffi.new("char[?]", n)
	local res = ffi.new("char[?]", 5001)
   	for i = 0, n-1 do array[i] = i end
   	require("lib.checksum_h")

   	assert(lib.csum(array, 20) == 42395, "lib.csum does not calculate the correct csum")
   	assert(C.cksum_generic(array, 20, initial) == 42395, "cksum_generic does not calculate the correct csum")
   	assert(C.cksum_sse2(array, 20, initial) == 42395, "cksum_sse2 does not calculate the correct csum")
   	assert(C.cksum_avx2(array, 20, initial) == 42395, "cksum_avx2 does not calculate the correct csum")

   	jbenchmark{function(b)
   		b.name = "csum"
   		b.runs = 100000 * 5000
   		local csum = lib.csum
   		for i=1,100000 do
   			for j=1,5000 do
   				res[j] = csum(array + 20 * j, 20)
   			end
   		end
   	end}
   	jbenchmark{function(b)
   		b.name = "generic"
   		b.runs = 100000 * 5000
   		local csum = C.cksum_generic
   		for i=1,100000 do
   			for j=1,5000 do
			   	res[j] = csum(array + 20 * j, 20, initial)
   			end
   		end
   	end}
   	jbenchmark{function(b)
   		b.name = "sse2"
   		b.runs = 100000 * 5000
   		local csum = C.cksum_sse2
   		for i=1,100000 do
   			for j=1,5000 do
				res[j] = csum(array + 20 * j, 20, initial) 
   			end
   		end
   	end}
   	jbenchmark{function(b)
   		b.name = "avx2"
   		b.runs = 100000 * 5000
   		local csum = C.cksum_avx2
   		for i=1,100000 do
   			for j=1,5000 do
				res[j] = csum(array + 20 * j, 20, initial)
   			end
   		end
   	end}
end
function benchmarkPairs() 
	local tab = {}
	for i=1,1000000 do
		table.insert(tab, i)
	end
	jbenchmark{function(b)
		b.name = "pairs"
		b.runs = 2000
		for i=1,b.runs do
			local i = 0 
			for l,w in pairs(tab) do
				i = w
			end
		end
	end}
	jbenchmark{function(b)
		b.name = "ipairs"
		b.runs = 2000
		for i=1,b.runs do
			local i = 0 
			for l,w in ipairs(tab) do
				i = w
			end
		end
	end}
end
function selfbenchmark()
	benchmarkPairs()
	--benchmarkChecksums()

	local apps = require("apps.basic.basic_apps")
	local counter = require("core.counter")
	for _,v in pairs({10,20,30,40,50,60,100}) do
		jbenchmark{
			function(b)
				b.name = "Sink -> Source" .. tostring(v)
				b.links = v
				engine.configure(config.new())
				local c = config.new()
				config.app(c, "source", apps.Source)
				config.app(c, "sink", apps.Sink)
				local l = "source.output -> sink.input"
				config.link(c, l)
				for i=1,v do
					config.link(c, "sink.output -> source.input" .. tostring(i))
				end
				engine.configure(c)
				b:start()
				engine.main({ duration = 1, no_report=true })
				b:finish()

				b.Hz = engine.Hz
				b.busywait = engine.busywait
				b.runs = tonumber(counter.read(engine.link_table[l].stats.rxpackets))
			end
		}
	end
end

selfbenchmark()
```

lukego:
This is looking very much like the kind of benchmarking that I want to do!

Can we define a really simple interface between the benchmark definitions and the framework that is running them? The benchmark definitions could be as simple and minimalist as possible (so that people will write them) and the benchmarking framework could have multiple implementations (depending on what people need e.g. simple wall-clock time, or count cycles, or track with PMU, or track with external profiler, or collect JSON, or ...).

In terms of minimalist definitions I mean something like what we do with the `selftest` framework. This is really trivial. To be testable a module defines a function `selftest()` that returns on success (otherwise raises exception or terminates process). The "framework" that runs this is actually just one line of code in snsh that does `require(moduletotest).selftest()`. This has been pretty successful in terms of adoption and I think the simplicity of the framework is a part of that.

Then we also have a more elaborate bit of framework, like a Makefile target that detects what selftests exist and runs them all and logs results, but nobody really has to care about that because it is separate.

So could we make the interface for this test framework similarly minimal?

Here is one idea. Suppose that a module that is benchmarkable would have a table called `selfbenchmarks` like this:

```
selfbenchmarks = {}

-- Benchmark how long it takes to execute an empty loop 1 billion times
function selfbenchmarks.loop_1e9 ()
   for i = 1, 1e9 do end
end

-- Benchmark how long it takes to execute nested loops 1 billion times
function selfbenchmarks.loop_1e6_1e3 ()
   for i = 1, 1e6 do
      for j = 1, 1e3 do end
   end
end
```

and then a simple test framework could execute these with something like:

```
function benchmark (modulename)
   for name, fn in pairs(require(modulename).selfbenchmarks) do
      local start = now()
      fn() -- run the benchmark function
      local end = now()
      print(name, end-start)
   end
end
```

... and then this would be the starting point for more fancy extensions e.g. fancier test frameworks that measure and report in more elaborate ways, libraries that can be called to generate interesting test data, ways to setup engine parameters, etc.

What do you think?

andywingo:
Regarding the return value from the benchmark routine -- I think we are getting into dangerous microbenchmark weirdness territory here.  If the compiler notices that a loop does nothing visible, it's free to remove it entirely, if it can prove that the loop terminates.  Alternately if a statement in a loop doesn't contribute to any output value or side effect, it can be removed or hoisted or any such thing.  These are real transformations that happen in real optimizing compilers.

The loops that you test Luke are approximately useless -- their performance will change as LuaJIT changes, and they have no value on their own.  They *are* useful if you know what you're testing, i.e. you've looked at the traces and the assembly and you know what the code maps to and you're interested in testing some low-level code generation thing -- and I'm sure you know this! -- but that mapping can change over time.  Semantically they mean nothing and so the optimizer is free to remove them entirely.

There is also the concern that a benchmarking loop can have a bug.  If you don't check that the function being benchmarked computes the right value, optimizations can break the property being tested.

In summary I think it's important for a benchmark to define its expected return value in some way :)  In particular if the return value gets written into the JSON data point, there's no way the compiler can remove the test under consideration.  It has to compute the value, so it has to run the test.

lukego:
@andywingo I would like to resolve this by saying that the individual selfbenchmark functions (e.g. `loop_1e9`) are responsible for executing the code that should be measured. If they fail to do this then it's a bug in the selfbenchmark function.

In practice I would expect selfbenchmark functions to have to be written very carefully depending on exactly what they intend to measure. If the compiler eliminating important work is a concern they may need to pass certain values to C functions or store them in FFI structures and force a compiler barrier; if the cache is a concern they may need to choose their workload size and access pattern very carefully; if they are sensitive to the alignment of data in memory then they need to allocate objects deliberately; etc. There are so many of these concerns that I don't think we can outsource them to the framework above us on the stack very effectively.

I may be overstating the case against return values from test functions, but there it is anyway :).

There is actually one value that I would like to return but did not mention for simplicity :-) and that is the "auxiliary" table used in the PMU. For example then you declare some metrics like how many iterations you executed and this can be used to normalize the time spent e.g. into nanoseconds per iteration. This would be useful for interpreting data e.g. on a checksum routine I would be interested in comparing the cycles-per-byte metric for various conditions such as input size and alignment. For example see the `/breath` and `/packet` columns in the [lib.pmu example](https://github.com/SnabbCo/snabbswitch/blob/master/src/lib/pmu.lua#L81-L90).

LuaJIT does support multiple return values so we could in principle make both of us happy :-).

petebristow:
It's not clear to me what benefit `selfbenchmarks = {}` brings over and above calling `function selfbenchmark()` and expecting it to call `jbenchmark()` or a wrapper as appropriate.

Timing the functions in `selfbenchmarks` means
* You can't do expensive setup
* You can't easily run a series of benchmarks. I hit a couple of subtleties with closures which broke things, why make it easier to get wrong?
* It's more difficult to run a series of benchmarks independent of the `selfbenchmarks={}` whole, without naming hacks and regexes. 
* Taking no options and returning nothing means you will only ever know the duration and potentially the function name. In terms of returning the table from lib.pmu. What makes the particularly more interesting than any other data the benchmark might generate?
* Passing in b means you can affect the benchmark run, at the moment start / finish it. If you have startup / cleanup costs. Perhaps turn on pmu or something else. 
* Reporting values because b is pass by ref is a bit tacky, but I had imagined that `b:recordLinkStats()` or `b:recordPmuStats()` might be added and lukego didn't like return values much.
* Is returning a value materially better than pass by ref for stopping the compiler a) optimizing b) optimizing too much?

lukego:
@petebristow There is a lot going on in your snippet and I am probably missing the forest for the trees. The coding style is unfamiliar to me and my brain is asking things like: what is `b`? what does the "j" in `jbenchmarks` mean? is the camelCase significant? It would help me to have a little more commentary or to break the code into more distinct bits and say what the interfaces between them are.

I do start to come around to your way of thinking though. My example is weak in terms of not allowing expensive setup or easily running many variations in a loop (without awkwardly factoring them into separate named functions). Could be nicer to have one function that makes a series of calls to e.g. `start(name)` and `stop()` for the parts that it wants to profile more like your example.

The reason I like the idea of returning the table of "aux" metadata (which I left out of the initial proposal because it is really a non-essential added detail) is not specific to the PMU but simply because it provides some ways to normalize the data e.g. to convert from "total time to run the benchmark" to "nanoseconds per packet" that may be easier to interpret and compare. Like if the benchmark performs a loop N times then it can tell the framework the value of N to help with reporting on per-loop instead of total-for-all-loops.

lukego:
@petebristow okay, reading the code closer I see that you have this stuff covered with fields in the `b` struct. I'd still like to find the simplest possible way to write the code in the modules being benchmarked, with the simplest interface to the testing framework that is worrying about JSON etc (or not, depending on what the user is interested in e.g. what arguments passed to snsh/snabbmark).

lukego:
@petebristow So what do you think of this example that is moving in the direction of the one you posted:

```lua
-- Run a series of checksum benchmarks with different alignment, size,
-- and architecture-specific code.
function selfbenchmark (b)
   for _,align in ipairs({0, 1, 2, 3, 4, 5, 6, 7, 8}) do
      for _, size in ipairs({32, 64, 100, 128, 200, 256, 500, 512, 1000, 1024, 9000}) do
         for _, code in ipairs({"generic", "sse", "avx"}) do
            local name = 'checksum-align:'..align..'-size:'..size
            local data = allocate_packets(size, align, 1e3)
            -- Start benchmark, or skip if b:start() returns false for this benchmark name
            if b:start(name, {checksum=1e6}) then
               -- Perform 1e6 checksums (1e3 loops over 1e3 packets)
               for i = 1, 1e3 do
                  for j = 1, #data do
                     functions[code](data[j])
                  end
               end
               b:stop()
            end
         end
      end
   end
end
```

This one has a simple `b` object with two methods:

- `start(name, aux) => boolean` return true if the benchmark called `name` should be executed. Start recording and remember the `aux` values for later reporting.
- `stop()` stop recording.

This interface is small but hopefully generic so that programs like `snabbmark` could implement arbitrarily complex profiling and reporting in the `b` argument that they pass.

lukego:
@petebristow Daydreaming about orthogonality...

Perhaps we could have a global `bench` module instead of the local `b` callback parameter?

So any code could run a benchmark like:

```
bench.start("my-benchmark-1000", {...properties...})
... work ...
bench.stop()
```

and fancy features like JSON capture, PMU recording, etc could be handled separately e.g. with:

```
bench.set_mode('pmu')
bench.set_json_output('/tmp/%s.json')
```

etc. Then the `selfbenchmark()` function would work exactly like `selftest()` i.e. simply a function that gets called.

(This starts to sound a lot like the existing interface to the `jit.p` profiler in LuaJIT.)

petebristow:
@lukego Right back at you.
```lua
-- Run a series of checksum benchmarks with different alignment, size,
-- and architecture-specific code.
function selfbenchmark ()
   local bench = require("lib.benchmark")

   -- work out what versions make sense to test on this cpu.
   -- Other similar benchmark selection operations should happen here not 
   -- hidden in b:start
   local instructions
   local cpuinfo = lib.readfile("/proc/cpuinfo", "*a")
   for _, code in ipairs({"sse", "avx", "avx512"})
      if cpuinfo:match(code) then 
         table.insert(instructions, code)
      end
   end

   for _,align in ipairs({0, 1, 2, 3, 4, 5, 6, 7, 8}) do
      for _, size in ipairs({32, 64, 100, 128, 200, 256, 500, 512, 1000, 1024, 9000}) do
         for _, code in ipairs(instructions) do
            local name = 'checksum-align:'..align..'-size:'..size.."code:"..code
            local data = allocate_packets(size, align, 1e3)
            bench.run(name, function(b)
               for i = 1, 1e3 do
                  for j = 1, #data do
                     functions[code](data[j])
                  end
               end
   -- I prefer the pass in a function to run approach. It makes it very clear
   -- the boundaries of what is being benchmarked. It stops you starting the 
   -- benchmark in one function and stopping it in another entirely. 

   -- If we want to support such approaches, for example to benchmark (or profile)
   -- every thousand'th breath then I'd want to make that intention explicit.
   -- Particularly as you as the core/app.lua maintainer might want to profile x
   -- whilst at the same time I as the apps/widget Ops guy want to profile my app
   -- on packets > 1500 bytes.

   -- start / finish might be more appropriate, finish is more final, start/stop 
   -- suggests to me that a you can start / stop / start / stop the same 
   -- benchmark run. Is that a good thing? Isn't that 2 benchmarks, bodged 
   -- together.

   -- Return from the bottom of the benchmark, there are some things you might not
   -- know until the end of the run. Namely how much work your doing. 
   -- For example it's not possible to send exactly
   -- 1e6 packets through an app network. engine.main({ duration = about1second }) 
   -- and then see how many packets have been processed. Adding packet counts 
   -- to basic.Source and basic.Repeater gets around that but I'm not sure it's
   -- the answer.

   -- If we decide to support 'profiling' this becomes more important, you have 
   -- no idea at the begining of the breath how much stuff your going to process.
   -- b.stop({ vars }) could be another approach.

   -- I don't like the idea of burying parameters just in the benchmark name.
   -- Having them 'machine readable' from the start means they can be ignored
   -- in pretty 'tabled' output that's for immediate human consumption. When
   -- we log the benchmark results and start to find trends or CI walk for local
   -- minima then having keyvalue is nicer.
               return { name = name, align = align, code = code, 
                        size = size, checksum = 1e6, bytes = 1e6*size }
            end)
         end
      end
   end
end

-- Very simple selfbenchmark()
function selfbenchmark
   local bench = require("lib.benchmark")
   bench.run("simpletest", function()
        -- do something
        return { packets = 1e4 }
   end)
end
```
I'm generally against adding stuff to the global namespace because once it's there
it's very hard to remove. If we start by requiring a `require("benchmark")` then we can always
promote it at a later date.


```
checksum-align:0-size:32-code:generic = 312312312ns (312ns/checksum, 10ns/byte)
checksum-align:0-size:32-code:sse = ...
```

How do we communicate what the 'interesting' keys are for formatting purposes. 
For your benchmark of csum maybe ns per (packet, byte) are the interesting things.
If you have pmu then you might want per cycle or cache miss per packet and byte.
For my benchmark of pairs/ipairs none of those are really meaningful. It's an 
issue I side stepped when I wrote the JSON outputs as it was assumed that something
else would take the json and do the 'RightThing' with it.

Perhaps, a (pre)process step happens on the data before display. The default 
process would look for keys of packets / bytes and divide those out by duration
(and if pmu is around cycles/misses ect ect). But we provide a means to add others
```
function foo(tab)
   return {
      nsPerChecksum = tab.duration / tab.checksum,
      -- bytes is automatically tried
      -- nsPerByte  = tab.duration / tab.bytes
   }
end
function selfbenchmark
   local bench = require("lib.benchmark")
   bench.run("simpletest", function()
      b.process_with(foo)
   end)
end
```

This would also allow filtering of which columns to display in the human readable
output, we just display name,duration,bytes,packets,nsPerPacket,nsPerByte + anything 
in the table returned from process ?

I'm not convinced by `bench.set_json_output('/tmp/%s.json')` to me the output 
presentation has nothing to do with the benchmark specification and this should
be hidden by benchmark.run. If benchmark is being run interactively then a human
readable table should be output as it's meant for immediate interpretation. If 
it's being run by a CI or by `make sendBenchmarksToSnabbDevelopersPrivacyOff` 
then it would output json or to a rest endpoint directly.

lukego:
@petebristow So we are edging closer together :)

One issue I have with using functions is that Emacs will indent your code much more "aggressively" with default settings.

Hand-indented example:

```
bench.run("simpletest", function()
   b.process_with(foo)
end)
```
Emacs default settings:
```
bench.run("simpletest", function()
                           b.process_with(foo)
end)
```
which is a lousy objection but it is annoying to fight the tooling.

lukego:
Here is another idea that I will just throw out there:

```
-- Return a table of benchmark functions that can be run.
-- The functions take no arguments, consume time that should be measured, and return metadata.
function selfbenchmarks ()
   do_expensive_setup() -- not counted
   local benchmarks = {} -- table that we will add our benchmark functions to (and return)
   -- Simple benchmark
   function benchmarks.simpletest ()
      do_work()
      return { iterations = 1000 }
   end
   -- Complex collection of related benchmarks
   for _,align in ipairs({0, 1, 2, 3, 4, 5, 6, 7, 8}) do
      for _, size in ipairs({32, 64, 100, 128, 200, 256, 500, 512, 1000, 1024, 9000}) do
         for _, code in ipairs(instructions) do
            local name = 'checksum-align:'..align..'-size:'..size.."code:"..code
            local data = allocate_packets(size, align, 1e3)
            -- Create a closure that runs this benchmark variant
            local fn = function () for i = 1, 1e3 do functions[code](data[i]) end
            -- Add it to the benchmarks table with a name
            benchmarks[name] = fn
         end
      end
   end
   return benchmarks
end
```

PR #693, opened 2016-01-03
Add public key for pete.bristow@gmail.com

petebristow:


lukego:
Added here: https://github.com/SnabbCo/snabblab-nixos/commit/0f809f44782113b356e8bd334cdcb58ba2f7ace6

You should be able to login to chur with `ssh -p 2020 petebristow@lab1.snabb.co` and have sudo access there.

Your account will be on the new lab servers too: I'll post an update on those to the list tomorrow.

PR #679, opened 2015-12-17
Fixes for various issues against 2015.12

eugeneia:
### Issue #666

Print meaningful error message when unable to create `/var/run/snabb`:

```
$ ./snabb
core/shm.lua:111: Unable to create /var/run/snabb: Permission denied
stack traceback:
	core/main.lua:126: in function <core/main.lua:124>
	[C]: in function 'assert'
	core/shm.lua:111: in function 'mkdir'
	core/shm.lua:84: in function 'map'
	core/counter.lua:48: in function 'open'
	core/app.lua:28: in main chunk
	[C]: in function 'require'
	core/main.lua:116: in function 'initialize'
	core/main.lua:35: in function <core/main.lua:32>
	[C]: in function 'xpcall'
	core/main.lua:155: in main chunk
	[C]: at 0x0044fa80
	[C]: in function 'pcall'
	core/startup.lua:1: in main chunk
	[C]: in function 'require'
	[string "require "core.startup""]:1: in main chunk
```

### Issue #658

* Print useful error message when unable to open PCI device:

(When the target device of `device_info` does not exist)

```
$ sudo SNABB_PCI0="0000:05:00.0" SNABB_PCI1="0000:05:00.1" ./snabb snsh -t apps.intel.intel_app
selftest: intel_app
lib/hardware/pci.lua:38: No such device: 0000:05:00.0
stack traceback:
	core/main.lua:126: in function <core/main.lua:124>
	[C]: in function 'assert'
	lib/hardware/pci.lua:38: in function 'device_info'
	apps/intel/intel_app.lua:150: in function 'selftest'
	program/snsh/snsh.lua:29: in function <program/snsh/snsh.lua:29>
	core/lib.lua:405: in function 'dogetopt'
	program/snsh/snsh.lua:63: in function 'run'
	core/main.lua:58: in function <core/main.lua:32>
	[C]: in function 'xpcall'
	core/main.lua:155: in main chunk
	[C]: at 0x0044fa80
	[C]: in function 'pcall'
	core/startup.lua:1: in main chunk
	[C]: in function 'require'
	[string "require "core.startup""]:1: in main chunk
```

(When the selected device exists but is not suitable)

```
$ sudo SNABB_PCI0="0000:04:00.0" SNABB_PCI1="0000:04:00.1" ./snabb snsh -t apps.intel.intel_app
selftest: intel_app
SNABB_PCI_INTEL[0|1]/SNABB_PCI[0|1] not set or not suitable.
```

[...]

PR #695, opened 2016-01-04
core.shm: Better error on permission denied to /var/run/snabb

### LINKED ISSUE #666: Confusing error when running non-root on first execution
The first time you run `snabb` on a machine it will create `/var/run/snabb`. This fails with a confusing error message if the process is not running as root:

```
[luke@chur:~]$ snabb
core/shm.lua:86: shm open error (7724/engine/breaths):No such file or directory
stack traceback:
	core/main.lua:116: in function <core/main.lua:114>
	[C]: in function 'error'
	core/shm.lua:86: in function 'map'
	core/counter.lua:48: in function 'open'
	core/app.lua:28: in main chunk
	[C]: in function 'require'
	core/main.lua:106: in function 'initialize'
	core/main.lua:35: in function <core/main.lua:32>
	[C]: in function 'xpcall'
	core/main.lua:121: in main chunk
	[C]: at 0x0044f740
	[C]: in function 'pcall'
	[C]: in function 'require'
```

The root cause is a permissions problem without adequate error checking:

```
mkdir("/var/run/snabb", 01777)          = -1 EACCES (Permission denied)
```

This should either be a fatal error or the process should find a way to tolerate this problem and execute normally.

lukego:
Fixes #666.

### DIFF
--- a/src/core/shm.lua
+++ b/src/core/shm.lua
@@ -105,7 +105,11 @@ end
 function mkdir (name)
    -- Create root with mode "rwxrwxrwt" (R/W for all and sticky)
    local mask = S.umask(0)
-   S.mkdir(root, "01777")
+   local ok, err = S.mkdir(root, "01777")
+   if not ok and err.ACCES then
+      print("failed to create "..root..": "..tostring(err).."; need to run as root")
+      os.exit(1)
+   end
    S.umask(mask)
    -- Create sub directories
    local dir = root


eugeneia:
Seems we duplicated effort here, I had already fixed this in #679 

I think I like [my version](https://github.com/SnabbCo/snabbswitch/pull/679/files#diff-16474a0a41e8099e043053353c6af273L106) better because it yields a more specific error messages (for potentially more error cases).

lukego:
Closed: duplicate of #679.

PR #697, opened 2016-01-07
core.engine: Enable selftest() method, fix app reports with use_restart = true

eugeneia:
This PR supersedes #633 and fixes a regression that broke app reports when `use_restart` is `true`.


eugeneia:
Merged into `max-next`.

PR #699, opened 2016-01-10
make PREFIX work like expected, and use DESTDIR for that

comotion:
PREFIX is usually used to indicate installation in /, /usr or /usr/local, while DESTDIR is used when creating packages etc. This is how autocrap does it and a lot of packagers expect this behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ LUASRC = $(wildcard src/lua/*.lua)
 LUAOBJ = $(LUASRC:.lua=.o)
 CSRC   = $(wildcard src/c/*.c)
 COBJ   = $(CSRC:.c=.o)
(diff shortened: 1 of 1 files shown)


lukego:
Thanks! Having a nonstandard `make install` behavior is unfortunate.

One more alternative we should consider: should we drop our `install` target? I do like the idea of Snabb Switch being a single file that could be (but needn't be) distributed as a binary. If we would want to create symlinks, manpages, etc then this could perhaps be done by a command `snabb install [dest]` rather than a Makefile.

eugeneia:
Assuming `make install` works otherwise, I think its safe to accept this patch for now and possibly replace `make install` in another PR. Merged into `max-next`.

PR #700, opened 2016-01-10
page pointers portable to 32bit platforms

comotion:
Use the appropriate pointer type `uintptr_t`.

A 32bit build will fail with the following error due to -Werror:
```
core/memory.c: In function 'virtual_to_physical':
core/memory.c:36:16: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
   virt_page = ((uint64_t)ptr) / 4096;
                ^
...
core/memory.c:51:5: error: format '%lx' expects argument of type 'long unsigned int', but argument 4 has type 'uint64_t' [-Werror=format=]
core/memory.c: In function 'allocate_huge_page':
core/memory.c:93:26: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
   realptr = shmat(shmid, (void*)virtual_address, 0);
                          ^
```
This patch is split out from #483, and fixes the problem by using an int type that changes size with the page size.

### DIFF
--- a/src/core/memory.c
+++ b/src/core/memory.c
@@ -29,18 +29,18 @@
 
 // Convert from virtual addresses in our own process address space to
 // physical addresses in the RAM chips.
-uint64_t virtual_to_physical(void *ptr)
+uint64_t virtual_to_physical(uintptr_t *ptr)
 {
-  uint64_t virt_page;
+  uintptr_t virt_page;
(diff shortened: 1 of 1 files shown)


eugeneia:
I am merging this into `max-next` as its a change regarding general correctness of C code.

Cc @lukego

Issue #665, opened 2015-11-18
Low DPDK throughput in NFV benchmark


eugeneia:
`program/snabbnfv/packetblaster_bench.sh` performs badly due to DPDK dropping packets. See #588.

lukego:
@eugeneia Could you please update the issue text to say exactly what we know about the differences between the fast and slow test environments and how to run both tests? (In the discussion on the other issue I lose track a bit of exactly what code is being tested and what its results are.)

Standard question whenever a mysterious 30% performance difference appears: Can the NUMA affinity have changed? (Could check e.g. in `htop` which node's cores are being used in each run of the benchmark.)

eugeneia:
@lukego I started testing from a clean slate, to be honest I didn't *know* anything and that's why I did not want to pollute the issue with previous speculations.

I am quite sure that its unrelated to NUMA affinity, as nothing changed there and the other benchmarks are not affected.

So here we go, fresh start. There are two different versions of DPDK being tested:

* [“legacy”](https://github.com/eugeneia/dpdk/tree/legacy-snabb) - this is the version we used previously in the opaque image, and we are using [now](https://hub.docker.com/r/eugeneia/snabb-nfv-test/) for the CI
* [v2.1.0-snabb](https://github.com/eugeneia/dpdk/tree/v2.1.0-snabb) - the current version of DPDK with our patches applied.

I also tested on two different machines:

* grindelwald - Intel(R) Xeon(R) CPU E5-2697 v2 @ 2.70GHz
* davos - Intel(R) Xeon(R) CPU E5-2603 v2 @ 1.80GHz

Finally I tested with two different versions of QEMU (2.4.0 and 2.1.0). Now to the results:

1. On grindelwald, it performs worse (~25%) when using v2.1.0-snabb instead of legacy DPDK (4 Mpps vs 3 Mpps), but QEMU versions do not make a difference
2. On davos using legacy DPDK, it performs worse (~30%) when using QEMU 2.4 instead of 2.1 (2.4 Mpps vs 3.5 Mpps)

So while there is a performance hit when upgrading to DPDK 2.1, the reason for the decrease in performance on davos seems to be indeed the new QEMU (2.4.0) version.

@lukego On a side node, legacy DPDK does indeed seem to work with 1000 byte packets.

lukego:
Can you tell me how to reproduce please?

We should not apply any patches to DPDK.

9000 byte is what I believe won't work in the older DPDK.

[...]

PR #707, opened 2016-01-13
Tail calls do not contribute to unroll counter

wingo:
See https://github.com/Igalia/snabbswitch/pull/196 and https://github.com/Igalia/snabbswitch/pull/198 for some discussion.  We were seeing trace aborts due to too many tail calls, and those trace aborts ended up being permanent, dumping us down to the interpreter.  But our tail calls were really just functions that happened to be called in tail position, not real loops, so it was a bit of a bogus limit.  Given that Lua doesn't encourage loops via tail calls I think it's appropriate to remove this bit of code in LuaJIT itself.  WDYT?


andywingo:
https://github.com/LuaJIT/LuaJIT/pull/123 for the upstream PR.

nnikolaev-virtualopensystems:
@andywingo that is interesting, do you think this could have some nice effect on the virtio-net driver performance? Or it is completely different ball?

andywingo:
@nnikolaev-virtualopensystems I don't know; I think the virtio-net issues are in a different place, but I am not sure.  @dpino has been the one looking into that.  I'll let you know if I find something though!

dpino:
@nnikolaev-virtualopensystems the patch had the effect of reducing the amount of aborted traces due to "loop unroll limit reached", however the reason for LuaJIT to blacklist several traces which eventually made it to run in interpreted mode was due to a misconfiguration in the testing script in the guest (the cards were cross sending packets with each other).

In any case, the main reason for the poor performance we observed was that the host was handling two cards with one core, when actually how we are would like to run the lwaftr virtualized is by bringing up two cards where each of them is managed by a different snabbnfv process running in a different core. With this setting performance improves. There are additional comments in https://github.com/Igalia/snabbswitch/issues/199

nnikolaev-virtualopensystems:
Thank you, @dpino !
I didn't quite get the "cross sending packets with each other" part. Can you give a link to the "proper" configuration of the guest?
Regarding the host - should we conclude that SnabbSwitch (being a framework) is not able to handle 2xNICs and 2xVhostUser interfaces in a single process? Because with the described configuration (two `snabb-nfv` processes) you map a single CPU core to read packets from the NIC and send then to the VM, and vice versa. I mean - 2 cores on a NUMA node are spent just to provide the connectivity to the VM.

Anyway - it is great news, now that we know that the `virtio-net` is going upstream and with a decent performance. Great times for SnabbSwitch!

dpino:
@nnikolaev-virtualopensystems Sure, I can give more details on the configuration part. To make a recap the settings were the following:

There are two cards connected as this:

```
81:00.0 <-> 02:00.0
81:00.1 <-> 02:00.1
```

In the testing environment, packetblaster sends packets on 02:00.0. A script sets up 81:00.0 and 81:00.1 to be managed by vhost-user, so QEMU can use these cards and expose them to the guest as 00:08.0 and 00:09.0. These are devices that are going to be managed by the VirtioNet driver inside the guest. They are set as 'eth0' and 'eth1':

```
config.app(c, "eth0", virtio_net.VirtioNet, {pciaddr=pci1})
config.app(c, "eth1", virtio_net.VirtioNet, {pciaddr=pci2})
```

In the original script that I tested these cards were cross sending packets with each other:

```
config.link(c, "eth0.tx -> eth1.rx")
config.link(c, "eth1.tx -> eth0.rx")
```

A measurement script in the host listens to traffic received on 02:00.1. Instead of sending back what is received from eth1 to eth0, what I did was to pipe it to a sink. With that change profiles on LuaJIT do not report switched to interpreted mode and all the backlisted traces are gone. Traffic can still be measured by the measurement script.

```
config.link(c, "eth0.tx -> eth1.rx")
config.link(c, "eth1.tx -> sink.rx")
```

Regarding the host, I don't know if that could be concluded for every Snabb Switch application. The original script that I used for testing brought up 2 NICs managed by VhostUser on the same program. Performance of the lwAFTR was around 2.5 MPPS under that setting. I didn't tried though bringing up the NICs with SnabbNFV and run each process on the same core. That would be an interesting experiment (check whether there's any difference of running that setting as two processes running on a core, or just one process).

nnikolaev-virtualopensystems:
Thanks @dpino, however, this still confuses me. It turns out that if I want to pass the traffic bidirectional, I'll have to put up with low performance? I mean in the real world `config.link(c, "eth1.tx -> sink.rx")` does not make much sense.

dpino:
@nnikolaev-virtualopensystems I understand what you mean, sinking one of the links is like sending half of the traffic. I can comment on what I observed:

In the testing code, the performance sending traffic bidirectionally or only in one direction was actually the same, although when traffic was sent bidirectionally LuaJIT reported 45% time execution in interpreted mode (many traces were aborted).

When I first started to run the lwAFTR using the virtio_net driver (with traffic generated by 'transient' from the host) I realized that LuaJIT didn't report any % time executed in interpreted mode and there was no trace aborts, so I concluded this was an issue in the testing apps code.

Maybe there's a different app layout for benchmarking a bidirectional flow of traffic, I don't know, I'd need to look into that. But even running the lwAFTR using 2xVirtioHost cards with one process LuaJIT didn't report switching to interpreted mode.

lukego:
There has been a performance regression observed when using indirect descriptors (especially) and mergeable RX buffers. The root cause is not known yet but it does not seem to be anything "obvious". (Could be additional L3 cache misses due to touching multiple guest memory locations?). You might want to test without indirect descriptors if this is convenient. See #665.

wingo:
Hi :) Discussion veered off into the weeds a bit here on this issue, but I would like to merge this PR to Snabb as a first step towards merging the lwaftr (https://github.com/SnabbCo/snabbswitch/pull/722#issuecomment-179098565).  What's the right way to do that?  Perhaps a special lwaftr integration branch?  I want to make sure this patch is acceptable though before relying on it.  If it is acceptable I see no reason to delay it landing in -next; but I understand if you want to use this as a testing ground for some theory about git branches and network topologies :)

lukego:
@wingo Thank you for suffering through these long winded workflow discussions :).

I am in favor of bringing in this change to support merge of the lwaftr. Merge from @eugeneia would make it official. I am impressed at your compiler skills to make this change in LuaJIT :-).

eugeneia:
Since this was rejected from LuaJIT with some explanation by Mike Pall I would had assumed this is not the right way to go and there would be an iteration of the solution. We can merge this and it will be reasonably simple to deal the resulting LuaJIT divergence because its tracked in Git but let me provide an anecdote so we all know what this means:

The performance regression described in https://github.com/SnabbCo/snabbswitch/pull/707#issuecomment-174639201 took quite some effort to track down, one of the reasons for this was that it is only visible without our patches to QEMU and DPDK. When upgrading QEMU and DPDK our patches did no longer apply uncovering the mentioned regression. One might argue that this could have been avoided if we had not ingested the patches, and avoided diverging from upstream.

So imho there should be a cost/benefit analysis to this change that includes the cost of diverging from LuaJIT. It would be nice to avoid it altogether, but at the very least we have to document this explicitly. E.g. “ah, yes, there was this patch we applied and we forgot all about it” is a realistic scenario that has happened before.

wingo:
I hear what you're saying @eugeneia, perhaps it isn't clear that I definitely made this analysis for our purposes in the lwaftr.  I quote from the comment that I linked in my last note:

> [The patch from #707] is not merged yet to our [lwaftr] branch but I was going to merge today. If I don't merge it I have to avoid merging some other things I had planned for today. Upstream discussions are complicated: https://github.com/LuaJIT/LuaJIT/pull/123#issuecomment-172511232. To me in the Snabb context it's an obviously good change; in a more general LuaJIT context I can see that Mike can consider it a somewhat lateral move (better for some cases, worse for others), but unfortunately he didn't comment on an idea for improving it; a bit irritating but oh well.

Do you want to add to this analysis?  Do you want it recorded somewhere?  What's the next step?

For me a suitable next step would be just to merge this patch, given that as Luke says, to reproduce upstream issues we need to build from upstream anyway.  But perhaps you have a concrete suggestion.

eugeneia:
I suggest the following: I merge this onto `max-next` and create a bug issue “LuaJIT upstream divergence” in which we can track the process of working with LuaJIT upstream to obsole this patch. Does that sound reasonable?

lukego:
@eugeneia I don't think we need to single out this patch. There is a diff between our luajit and the upstream one and some of those changes will probably never e accepted. Good idea to try and minimize the diff over time but sometimes we will just agree to disagree with them.

lukego:
One day we can have a "luajit" subsystem maintainer who will volunteer to worry about this stuff...

eugeneia:
@lukego Maybe not single out, but also track the existing divergences. I believe there is only the DynASM Lua-mode changes besides this one? Would be nice to have a definitive list. (Is there a git command to effectively diff `lib/luajit` with e.g. `LuaJIT:v2.1`?)

lukego:
@eugeneia Good question. The quickest way to get an overview seems to be:

```
[luke@lugano-1:~/git/snabbswitch]$ git log --oneline --graph lib/luajit
* 8dd093b Synchronize lib/luajit with LuaJIT/LuaJIT:v2.1
* 782e9b3 Add pointer to int casting of T->mcode in setintfield call
* ef49da4 jit.util.traceinfo(): Include mcode, mcloop, szmcode
*-.   863854c Merge PRs #619 #620 #621 (LuaJIT extensions) into next
|\ \  
| | * a212176 jit.p: Enhance 'v' mode of profiler
| | * 065979c jit.p: Add support for profiling at trace granularity
| |/  
|/|   
| * 53aedce jit.p: Add CPU performance counters as profiler source
|/  
* 2d8ba7c Merge commit '532f9f4dcb1a8fbccb3c766be96f72b995834b03' as 'lib/luajit'
```

and for full details it seems to be necessary to export the subtree with `git subtree split` or `git subtree push` e.g. into a repository of its own. However, those commands don't seem to actually work when I try them for the first time:

```
[luke@lugano-1:~/git/snabbswitch]$ git subtree split --prefix lib/luajit
fatal: bad object 55c3b29f7b20f3801848e0ab71e9de1d22207b95
[luke@lugano-1:~/git/snabbswitch]$ git subtree push --prefix=lib/luajit ~/tmp master
git push using:  /home/luke/tmp master
fatal: bad object 55c3b29f7b20f3801848e0ab71e9de1d22207b95
```

and I am not immediately sure if that is due to user error, software error, repository error, etc.

lukego:
Here is at least a middle ground for exporting the changes as individual patch files:
```
[luke@lugano-1:~/git/snabbswitch]$ git format-patch 2d8ba7c lib/luajit/
0001-jit.p-Add-support-for-profiling-at-trace-granularity.patch
0002-jit.p-Add-CPU-performance-counters-as-profiler-sourc.patch
0003-jit.p-Enhance-v-mode-of-profiler.patch
0004-jit.util.traceinfo-Include-mcode-mcloop-szmcode.patch
0005-Add-pointer-to-int-casting-of-T-mcode-in-setintfield.patch
```

which look like this:

```diff
[luke@lugano-1:~/git/snabbswitch]$ cat 0004-jit.util.traceinfo-Include-mcode-mcloop-szmcode.patch 
From ef49da4219100a39e8f285609f7e6645fb317651 Mon Sep 17 00:00:00 2001
From: Luke Gorrie <luke@snabb.co>
Date: Wed, 23 Sep 2015 13:25:27 +0200
Subject: [PATCH 4/5] jit.util.traceinfo(): Include mcode, mcloop, szmcode

---
 lib/luajit/src/lib_jit.c | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/lib/luajit/src/lib_jit.c b/lib/luajit/src/lib_jit.c
index 9e0e01d..975063c 100644
--- a/lib/luajit/src/lib_jit.c
+++ b/lib/luajit/src/lib_jit.c
@@ -299,6 +299,9 @@ LJLIB_CF(jit_util_traceinfo)
     setintfield(L, t, "nk", REF_BIAS - (int32_t)T->nk);
     setintfield(L, t, "link", T->link);
     setintfield(L, t, "nexit", T->nsnap);
+    setintfield(L, t, "szmcode", T->szmcode);
+    setintfield(L, t, "mcode", T->mcode);
+    setintfield(L, t, "mcloop", T->mcloop);
     setstrV(L, L->top++, lj_str_newz(L, jit_trlinkname[T->linktype]));
     lua_setfield(L, -2, "linktype");
     /* There are many more fields. Add them only when needed. */
-- 
2.7.0
```

lukego:
> fatal: bad object 55c3b29f7b20f3801848e0ab71e9de1d22207b95

On reflection I am pretty sure this is user error. I haven't added LuaJIT upstream as a remote and fetched its commits, so there is no way that `git subtree` could reconstruct the history based on the squashed version in the snabbswitch repo.

eugeneia:
Merged into `max-next`.

PR #708, opened 2016-01-13
Add a Match app to apps/basic. 

petebristow:
This PR replaces PR#690. It's been reimplemented, pared down and moved into apps/basic. Match takes 2 inputs and compares them, providing some feedback as to whether they match.


andywingo:
Neat.  I guess this is for test suites and such.  Initial thought: I'm not sure how useful it is, given that when there's a problem you usually want to be able to see the output of a link as a pcap file, and then compare it to an expected output as a pcap file.  In your patch I don't think you get that kind of context.  For that reason I don't think I'd be tempted to use it.

Cosmetics:
Perhaps it should be named Check or something; "match" implies that we match packets and then do something with them.  Keep lines to within the 80-char line limits.  Compare to nil instead of "nil", or just test for truthiness.

eugeneia:
@andywingo This `Match` has support for fuzzy matching and generally gives you more information than comparing Pcap files (see `Match:log`). I guess we could add a method to write errors to a Pcap file on demand.

petebristow:
@andywingo 
Match was @eugeneia suggestion, I'm happy with whatever. 
I can certainly dump exceptions to a pcap if that's seen as useful, personally I found hexdumps more helpful.
I started by comparing to nil, however core/config.lua explicitly uses "nil" which means it doesn't work.
It seemed weird but I assumed it was an intentional, known thing. 
```
-- Example: config.app(c, "nic", Intel82599, {pciaddr = "0000:00:01.00"})
function app (config, name, class, arg)
   arg = arg or "nil"
   assert(type(name) == "string", "name must be a string")
   assert(type(class) == "table", "class must be a table")
   config.apps[name] = { class = class, arg = arg}
end
```

src/apps/basic/basic_apps.lua
```diff
@@ -86,6 +87,90 @@ function Sink:push ()
...
+   return setmetatable({
+         cfg = cfg,
+         seen = 0,
+         errs = { },
+         done = false
+      },
+      {
+         __index=Match
+      })
+end
```
eugeneia: `cfg` is being converted to a string here because Apps are supposed to be able to accept their configuration as strings. The common pattern to read an app config is this:

```lua
function App:new (arg)
      local conf = config.parse_app_arg(arg) or {} -- The default
      ....
end
```

src/apps/basic/basic_apps.lua
```diff
@@ -86,6 +87,90 @@ function Sink:push ()
    end
 end
 
+--- ### `Match` app: Compare packets recieved on rx and comparator
+Match = {}
+
```
eugeneia: I think `Match` is a bit too fancy for `apps/basic/basic_apps.lua` and would move it two its own module `apps/match/match.lua`.

src/apps/basic/basic_apps.lua
eugeneia: Actually, the above breaks when `arg` is an empty table, this would be more correct:

```lua
function App:new (arg)
      local conf = config.parse_app_arg(arg) or {}
      conf.foo = conf.foo or "default"
end
```

petebristow:
@eugeneia I'm happy to move it to apps/match but what's changed since the PR690 where you suggested TestSink should be moved to basic_apps as Match? 

Also I pulled in the changes from max-next so that I got the fix for config.parse_app_config so the my tests passed. That's added lots of files to the PR that probably don't make sense and probably not what I wanted to do? Can you give me a pointer on what I should of done?

eugeneia:
>  I'm happy to move it to apps/match but what's changed since the PR690 where you suggested TestSink should be moved to basic_apps as Match?

My bad, I said “in the spirit of basic_apps” but did not mean to imply it should be in the module basic_apps. Misleading communication on my side, sorry!

> Also I pulled in the changes from max-next so that I got the fix for config.parse_app_config so the my tests passed. That's added lots of files to the PR that probably don't make sense and probably not what I wanted to do? Can you give me a pointer on what I should of done?

I think you don't really need that bug fix, you can just pass in `{}` as the empty configuration to avoid triggering the bug. As to your branch, if you are not happy with it, you can always close this PR and supersede it with a new branch, just reference #708 to indicate it to be a continuation of this PR. We try to avoid rebasing “public” branches to preserve history, one way to do this is to just leave the branch you are unhappy with as is and start  a new branch (PR branches should be based on `master`).

eugeneia:
@petebristow What is the status here? I am somewhat betting on this landing sooner or later. Can I help you with anything regarding this PR?

eugeneia:
Closing in favor of #861.

PR #709, opened 2016-01-14
LuaJIT and DynASM updates from upstream

capr:
- LuaJIT was updated to LuaJIT/LuaJIT:v2.1
- DynASM was updated to capr/LuaJIT:v2.1-dynasm-luamode which is the new upstream[1] for DynASM-luamode

NOTE: you can pull this update directly from capr/snabbswitch:master which is the branch I made this PR from.

[1] the old upstream was luapower/dynasm:master but that has the "weird" luapower directory structure, plus now you can patch-in luamode over any up-to-date luajit fork by applying the patch at https://github.com/capr/LuaJIT/commit/403f386372adb9080594091f485b3441a6e8eda8.patch
 with `patch -p1 | curl <above url>` from the LuaJIT root directory.


andywingo:
if merged, we should close https://github.com/SnabbCo/snabbswitch/pull/705.

lukego:
Looks like the LuaJIT Makefile has a bug that it misses some dependencies that need to be rebuilt after a DynASM upgrade? (If I'm reading the CI error correctly.)

lukego:
@eugeneia @capr: The "make clean" compile issue seems to be a bug in upstream LuaJIT. I filed an issue as luajit/luajit#126. I suggest we merge this code but also pull any fix that happens to materialize before the release.

Issue #710, opened 2016-01-17
Idea: Double the performance of Snabb NFV


lukego:
Here is a fun idea to play with in the background: How about doubling the performance of Snabb NFV?

### Why

Our original performance target was to handle 10 Gbps per core for realistic ISP workloads based on a reference processor that Intel use in their NFV case studies (e.g. Xeon E5 2690V2 w/ 12 cores @ 3.0 GHz). Such performance was unheard of with Virtio-net at that time and we had to write the QEMU code to make it possible. There were few if any virtual machines available that were optimized to keep up with millions of packets per second on their Virtio-net interfaces.

Time is moving forwards though. High-speed Virtio-net is an established idea now. Optimized applications like Igalia's `snabb-lwaftr` can use the available capacity and more. The clock speeds on Intel's latest high-end CPUs are lower than the previous generation.

Doubling the performance of Snabb NFV would be valuable. On the one hand it would enable efficient deployments with only one reserved core per 20G of traffic. On the other hand it would provide a performance buffer for maintaining 10G per core performance in the presence of complications like slow processors, NUMA mismatches, and performance-affecting bugs.

### How

Snabb NFV processing cost is split fairly evenly between:

- Virtio-net vring processing.
- Data copies to/from guest memory (Virtio-net "DMA").

and also for client/server applications like iperf/apache/postgresql/etc:

- Checksum computations to offload work from the guest.

So one optimization strategy would be to double the performance of each item on that list.

Here is an initial sketch of how that might play out:

- Virtio-net processing could be optimized by reviewing and improving the JIT'ed inner loop code.
- Data copies could be optimized with a special purpose assembler routine (sese #648).
- Checksum computations are bound by the SIMD performance of the CPU and that is scheduled to double in the near future with CPUs supporting AVX512 instructions.

### Progress

Here is a placeholder list to see how we are doing:

- [ ] Double virtio-net vring processing performance
- [ ] Double virtio-net copy performance.
- [ ] Double SIMD checksum performance.

And interesting performance milestones:

- [ ] Achieve 10 Gbps full-duplex packet forwarding with 128-byte packets.
- [ ] Achieve 20 Gbps full-duplex packet forwarding with 256-byte packets.
- [ ] Achieve 20 Gbps full-duplex TCP performance with iperf.

kbara:
That would be amazing.

mwiget:
[...]

PR #711, opened 2016-01-17
[draft] lib.blit: Introduce API for optimized "blitter" for Virtio-net DMA

lukego:
Just wanted to share this fun hack for a software [blitter](https://en.wikipedia.org/wiki/Blitter) that I am working on in connection with #710.

The idea is to isolate the virtio-net memory copy operations from the rest of the vring processing so that they can be optimized separately. See the individual commit comments for more details.

I have started to experiment with writing an optimized batch-mode version of the blitter in assembler. The initial results are encouraging but it is early days yet and I don't know which aspects of the design are important. See [gist](https://gist.github.com/lukego/450e108c9c9cf420252d) for rough notes and code.

I started looking at this in the context of debugging #665.

If there is a suitable online betting site then I am willing to wager that a suitable blit subroutine will speed up the copy operations for Virtio-net by 2x - 4x :smile: because the performance we see with `memcpy` seems nowhere near the capabilities of the processor.

### DIFF
--- a/src/lib/blit.lua
+++ b/src/lib/blit.lua
@@ -1,0 +1,33 @@
+-- blit.lua - offload engine for memory operations
+
+module(..., package.seeall)
+
+local ffi = require("ffi")
+
+-- The blit module provides "blitter" operation to offload
+-- performance-critical memory operations. The API allows scheduling a
+-- series of operations, that can be performed at any time and in any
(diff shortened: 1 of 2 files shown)


lukego:
Closed: action has moved to #719.

Issue #701, opened 2016-01-10
Confusing error message when snabb is started on unsupported platforms


comotion:
I'm having issues installing. plaform is linux-musl-i486 and build completes correctly, but launching `snabb` results in:

```
  snabb
BINARY    2.0M snabb
make[1]: Leaving directory `/tmp/snabbswitch/src'
/tmp/snabbswitch $ ./src/snabb
startup: unhandled exception
syscall/linux/ffi.lua:16: module 'syscall.linux.x86.ffi' not found:
        no field package.preload['syscall.linux.x86.ffi']
        no file './syscall/linux/x86/ffi.so'
        no file '/usr/local/lib/lua/5.1/syscall/linux/x86/ffi.so'
        no file '/tmp/snabbswitch/lib/luajit/usr/local/lib/lua/5.1/syscall/linux/x86/ffi.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
        no file './syscall.so'
        no file '/usr/local/lib/lua/5.1/syscall.so'
        no file '/tmp/snabbswitch/lib/luajit/usr/local/lib/lua/5.1/syscall.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
```

eugeneia:
@justincormack Does this ring any bells to you? Seems to be ljsyscall related on first look.

lukego:
The fundamental issue here is that Snabb Switch only supports Linux/x86-64. That is, i486 is not a supported platform.

Are you interested in porting to i486? This should be reasonably straightforward and I can sketch the work likely involved. More challenging if your target CPU does not support huge pages.

(Curious: what are you working on here?)

lukego:
More specifically to this PR:

The top-level `Makefile` is [cherry picking](https://github.com/SnabbCo/snabbswitch/blob/master/Makefile#L15-L21) the relevant modules from ljsyscall. It takes x64 (supported platform) but not x86 (unsupported).

lukego:
(Sorry @justincormack for doing violence to ljsyscall in our downstream packaging like this. Perhaps we should change that to avoid creating support problems for you. And we should really print a message at startup to complain about running on an unsupported platform instead of hitting a random error.)

eugeneia:
Does anyone mind if I change the title to something reflecting “we should really print a message at startup to complain about running on an unsupported platform instead of hitting a random error”?

comotion:
[...]

PR #712, opened 2016-01-17
porting.md: Add notes about Snabb Switch porting & portability

lukego:
This is an attempt to set realistic expectations for users and developers about Snabb Switch portability both now and in the future. This is to avoid causing confusion as we have in #701 for example.

Summary:
- Added `doc/porting.md` to summarize the situation.
- Link from top-level README.
- Print a warning at startup when running on unsupported platform.

Feedback welcome!

### DIFF
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -32,6 +32,10 @@ debug_on_error = false
 function main ()
    zone("startup")
    require "lib.lua.strict"
+   -- Warn on unsupported platforms
+   if ffi.arch ~= 'x64' or ffi.os ~= 'Linux' then
+      io.stderr:write("Warning: "..ffi.os.."/"..ffi.arch.." is not a supported platform\n")
+   end
    initialize()
    local args = parse_command_line()
    local program = programname(args[1])
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -34,7 +34,7 @@ function main ()
    require "lib.lua.strict"
    -- Warn on unsupported platforms
    if ffi.arch ~= 'x64' or ffi.os ~= 'Linux' then
-      io.stderr:write("Warning: "..ffi.os.."/"..ffi.arch.." is not a supported platform\n")
+      error("fatal: "..ffi.os.."/"..ffi.arch.." is not a supported platform\n")
    end
    initialize()
    local args = parse_command_line()
--- a/README.md
+++ b/README.md
@@ -32,9 +32,9 @@ Snabb Switch is written using these main techniques:
 
 Snabb Switch compiles into a stand-alone executable called
 `snabb`. This single binary includes multiple applications and runs on
-any modern Linux distribution. (You could think of it as a
-[busybox](http://en.wikipedia.org/wiki/BusyBox#Single_binary) for
-networking.)
+any modern [Linux/x86-64](src/doc/porting.md) distribution. (You could
+think of it as a [busybox](http://en.wikipedia.org/wiki/BusyBox#Single_binary)
+for networking.)
 
 ## How is it being used?
 
--- a/src/doc/porting.md
+++ b/src/doc/porting.md
@@ -1,0 +1,67 @@
+# Porting Snabb Switch
+
+## Background
+
+Snabb Switch is a low-level program. The source code includes device
+drivers, assembler code, and optimizations for specific CPU families.
+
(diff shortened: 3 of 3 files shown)


lukego:
One vague point in the document is: when is a port a "net benefit" and desirable to merge onto the main branch?

I imagine that we would want to weight the trade-offs. Some ports are more desirable than others.

Good things might be:

- Bring more of our target users (network operators of all shapes and sizes) into the community.
- Bring more developers into the community to work with.
- Rewrite ugly non-portable C code as neat portable Lua/ljsyscall code.
- Better define interfaces towards primitive functions (e.g. PCI device access and DMA memory allocation).

Neutral things might be:

- Support platforms that aren't relevant to network operators (e.g. Raspberry Pi, OSX laptop).

Bad things might be:

- Create many more lines of code to maintain.
- Globally complicate our programming style: make everybody spend brain cycles thinking about POSIX correctness, race conditions due to relaxed/weak cache coherence protocols, additional memory alignment restrictions, etc.
- Create technical/practical/social barriers to writing awesome optimizations for important platforms (e.g. Xeon).
- Create pressure to replace simple/snabby tools (like DynASM) with complex/unsnabby tools (like LLVM).
- Make our code uglier e.g. `#ifdef`.
- Make our build process uglier e.g. autoconf.

Quite a list, eh?

The example to follow would seem to be LuaJIT itself. LuaJIT has managed to retain its character and style while supporting very many platforms: Windows, Linux, BSD, OSX, Playstation, Xbox, i386, x86-64, ARM, PPC, E500, MIPS. This has required extreme dedication to the maintenance -- for each architecture Mike has written an assembler and used that to write the virtual machine in assembly language by hand. This, in my mind, is the kind of work that we are signing up for when we start accepting new platform ports into Snabb Switch.

eugeneia:
As of now we do not have the resources to support more platforms. By resources I mean neither the infrastructure (hardware, CI system) nor the people (that know the details of a platform). So currently I can only imagine a third party maintained, long lived fork that supports a new architecture.

The good news is that we have monthly releases which can be merged downstream, and anybody can run their own SnabbBot on their own hardware. So at least there are some tools that support the process.

I think a hierarchy where upstream merges from platform specific forks, but the forks stay autonomous is most practical. In the case when a platform fork “dies” we can simply stop merging with it and put “no longer supports $platform, sorry” into the release notes. Of course, the code will already be in upstream, and no longer actively maintained. I am wary of a situation where we gather substantial unmaintained code in upstream. On the other hand we have had that situation previously as well, and it wasn't a big problem.

My point is, in order to support a new platform, we have to have “a guy for that”.

andywingo:
s/guy/person/, but yes :)

lukego:
Great feedback!

I rewrote `porting.md` to be less wishy-washy and defend our lovely x86-64 geekout fest. How does it look now?

src/core/main.lua
```diff
@@ -32,6 +32,10 @@ debug_on_error = false
 function main ()
    zone("startup")
    require "lib.lua.strict"
+   -- Warn on unsupported platforms
+   if ffi.arch ~= 'x64' or ffi.os ~= 'Linux' then
+      io.stderr:write("Warning: "..ffi.os.."/"..ffi.arch.." is not a supported platform\n")
```
andywingo: Why not just error out here?  It would be the porter's job to fix this.

lukego:
@wingo Good idea to error rather than warn on unsupported platforms. Adopted.

I am happy with this branch now and would like to start upstreaming it now please.

wingo:
For me, LGTM, I'm happy to merge into `wingo-next`.  Procedural question: you OK with me resolving the merge conflicts?

lukego:
> For me, LGTM, I'm happy to merge into wingo-next.

Thanks! I set you as *Assignee* as a proof-of-concept to see that this is really possible.

> Procedural question: you OK with me resolving the merge conflicts?

Yep. I'm thinking that merges need to be resolved at the point they occur. I imagine that mentionbot will likely poke everybody involved in the conflict when you PR your branch to the next hop and so we will have a chance to check the resolution if we like.

wingo:
I merged this onto `wingo-next` and was waiting for github to mention this PR but I guess I have to mention the PR by name, somehow :)  Anyway, merged, will get in next time wingo-next goes to next.

I agree with your general anti-portability vibe these days btw! :fistbump:

PR #714, opened 2016-01-18
next: Collected changes for the v2016.01 release

lukego:
The next branch is open!

Starting this cycle the plan is for feature/fix PRs to first be merged into an integration branch and then later PR'd to `next`. Currently there is one such integration branch: `max-next`.


lukego:
@eugeneia your `max-next` branch is now automatically being synced to the snabbco repo so that it is visible in the [branch list](https://github.com/SnabbCo/snabbswitch/branches/all).

This is done by a shell script of mine. That's a bit ad-hoc. Can be that we could push to these branches directly instead of having them indirectly sync'd now that Github supports "protected branches" that block force pushes. (The main reason I setup the script to synchronize is so that people can rely on these well-known branches to not be rebased without notice.)

eugeneia:
@lukego Are we ready for release?

lukego:
Yep!

eugeneia:
Ok done: https://github.com/SnabbCo/snabbswitch/releases/tag/v2016.02

Mistake on my side: the commit message mistakenly says v2016.03... :no_good: not sure what to do now, any amendment would clobber the tag (and history).

PR #705, opened 2016-01-11
Dynasm all the vregs

andywingo:
This merges the LuaJIT updates from https://github.com/eugeneia/snabbswitch/commits/esp into next.  The PR is based on master. Cc @eugeneia.

PR #715, opened 2016-01-18
LuaJIT upgrade from v2.1 branch

lukego:
Pull changes from upstream LuaJIT v2.1 branch using the command:

```
git subtree pull --prefix=lib/luajit https://github.com/luajit/luajit v2.1 --squash
```

The contents should be equivalent to commit 23a301cf32a9ccfb12a86d0fd9523cdfa2c3cdc5 from #709 except that by using `git subtree pull` we get automatically generated commit messages to keep track of which version we pulled and which changes we got.

@eugeneia One idea would be to merge this branch onto `max-next` and then cherry-pick the DynASM update from #709 on top of it?


lukego:
Closing in favor of #705.

PR #717, opened 2016-01-19
Fixed a bug where string couldn't be app arguments

aequabit:
`lib.load_string` failed to properly call Lua's `loadstring` built-in.


eugeneia:
Merged into `max-next`.

PR #719, opened 2016-01-19
Optimized "blitter" routine written in assembler [wip]

lukego:
This is an experiment towards doubling the performance of virtio-net copies (#710) with an optimized blitter routine (#711) written in AVX assembler (compatible with Sandy Bridge onwards).

Caveats and notes:
- The blitter is used only for Virtio-net "DMA" routines i.e. guest memory data copies.
- This version only supports copies that are multiples of 32 bytes.
- This works with the DPDK-VM performance test suite (packet sizes are always multiple of 32).
- This does _not_ work with the Linux-VM functional test suite (packets of all shapes and sizes).
- Could easily be undetected bugs that invalidate all results, not confident until the full test suite runs.

To sanity check the performance I tried three scenarios: master branch (baseline), hack to skip data copies entirely (maximum possible speedup), and then the actual asm blitter code.

Testing with 128-byte packets on `lugano-1` I see these results:
- 5.6 Mpps baseline from the master branch.
- 8.1 Mpps (+45%) maximum when copies are skipped.
- 7.0 Mpps (+25%) with the assembler blitter.

I interpret this to mean that the copy performance is more than doubled i.e. with this optimization we are achieving more than half of the maximum possible speedup.

The challenges I see now are:
- Get the full test suite running without losing this performance boost.
- Use the PMU to understand what is really going on and how dependable this speedup will be.
- Try some further optimization ideas e.g. streaming multiple packets in parallel.

Likely we also need to update the DPDK-VM benchmark to test with a more interesting variation of packet sizes so that we don't accidentally optimize for the "packet size is always a power of 2" special case that we would never see in real life :-).

### DIFF
--- a/src/lib/blit.dasl
+++ b/src/lib/blit.dasl
@@ -1,0 +1,132 @@
+-- blit.lua - offload engine for memory operations
+
+module(..., package.seeall)
+
+local dasm = require("dasm")
+local ffi = require("ffi")
+
+|.arch x64
+|.actionlist actions
+
+-- The blit module provides "blitter" operation to offload
+-- performance-critical memory operations. The API allows scheduling a
+-- series of operations, that can be performed at any time and in any
+-- order, and then executing a "barrier" to wait for completion.
+
+ffi.cdef[[
+  struct blit_queue_entry {
+    void *src;
+    void *dst;
+    uint64_t len;
+  };
+]]
+
+-- Queue of memory operations
+local maxqueue = 10000
+local queue = ffi.new("struct blit_queue_entry[?]", maxqueue+2)
+local nqueued = 0
+
+function copy (dst, src, len)
+   assert(len%32==0)
+   if nqueued == maxqueue then barrier() end
+   queue[nqueued].src = src
+   queue[nqueued].dst = dst
+   queue[nqueued].len = len
+   nqueued = nqueued + 1
+   -- Sentinel
+   queue[nqueued].src = nil
+end
+
+-- Assembler code for the barrier operation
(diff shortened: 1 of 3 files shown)


src/lib/blit.dasl
```diff
@@ -0,0 +1,138 @@
...
+   -- Load parameters for the next copy
+   | mov rsi, [r8]     -- source
+   | mov rdi, [r8+8]   -- destination
+   | mov rcx, [r8+16]  -- length
+
+   -- Copy 32 bytes at a time
+   | xor rax, rax
+   |->copy:
+   | vmovdqu ymm0, [rsi+rax]
+   | vmovdqu [rdi+rax], ymm0
```
andywingo: You might want to experiment unrolling this manually.  I got some significant speedups by having more loads in flight.

andywingo:
Looks nice.  See also https://github.com/Igalia/snabbswitch/pull/204.

src/lib/blit.dasl
lukego: It is quite delicate :-). I started out with an unrolled version of the inner loop and then found that the looping version delivered the same performance. There have been other very innocent code variations that were much slower though. I want to use the PMU to explore these differences.

I would like to try unrolling the outer loop though to see if coping several packets in parallel could help.

andywingo: Yes, delicate indeed :)  One thing to try is instead of doing load, store, load, store, to do load, load, store, store.  That was what worked best for me.  Good luck :)

Issue #718, opened 2016-01-19
Standard tables for fast lookups


plajjan:
I think it would be useful for Snabb to have a couple of standard libs of table implementation that are well suited for network forwarding, e.g. very fast and deterministic lookup and update times, that can be included from an app.

We need equal matching (ARP, ethernet forwarding, potentially firewall stuff) as well as longest prefix matching for routing. While flexible key length would be sweet, I suppose it might be easier to optimise if we keep to a set of standard key lengths. 16 (TCP / UDP ports), 32 (IPv4), 48 (MAC) and 128 (IPv6) comes to mind.

It would also be nice if all tables would behave in roughly the same way, ie the methods they expose and what data can be stored, like if we allow a bit of extra data or a reference to some other data structure where more information can be stored, think next-hop, action or whatever for a key.

First candidate for inclusion is @andywingo's PodHashMap in lwAFTR that implements equal matching for 32 bit keys.

lukego:
YES!

See also luajit/luajit#115 which has a [bounty](https://www.bountysource.com/issues/28692239-feature-request-support-large-tables-with-ffi-data-keys) attached.

andywingo:
Note that the PodHashMap supports keys of any length.   It has optimized equality routines for keys that are numbers and keys that are 4 and 8 bytes long.  Adding more optimized comparators is easy.

For longest-prefix matching, you could look up a number of masked key lengths in parallel using a lookup streamer.  Or you could use binary search, depending on how many routes there are.  Probably parallel lookups of N key lengths are better.

I suspect the PodHashMap fulfills most requirements.  It's on my list to submit to Snabb core soonish.  Incidentally if you have any opinions about names, they are welcome :)  cdatamap, cdatahash, ffimap, ffihash, podmap, etc...

plajjan:
And speaking of LPM, I find it very interesting that Juniper seem to have opted for a combination of bloom filters and equal match tables for implementing LPM. I don't think it's explicitly stated but it can kind of be derived from information, see http://forums.juniper.net/t5/Data-Center-Technologists/Juniper-QFX10002-Technical-Overview/ba-p/270358

My interpretation is that they have as many tables as there are bit lengths, so 32 for IPv4 and then use Bloom filters to quickly evaluate if it's probable that there is a match in a table for the destination address. 
```
for pl in 32..0:
    if bloom_check(dst_address, table[pl]):
        nh = lookup(dst_address, table[pl])
        if nh:
            return nh
```
[...]

PR #722, opened 2016-01-20
Add lib.ctable module.

wingo:
Add a standard hash table for use when dealing with FFI keys and values.  See https://github.com/SnabbCo/snabbswitch/issues/718.  For the road to this implementation, see https://github.com/Igalia/snabbswitch/pull/97, where this table was called "PodHashMap".

### DIFF
--- a/src/lib/README.ctable.md
+++ b/src/lib/README.ctable.md
@@ -16,7 +16,7 @@ robin-hood hash table with linear probing.  This means that to look up a
 key in the table, we take its hash value (using a user-supplied hash
 function), map that hash value to an index into the table by scaling the
 hash to the table size, and then scan forward in the table until we find
-an entry whose hash value that is greater than or equal to the hash in
+an entry whose hash value is greater than or equal to the hash in
 question.  Each entry stores its hash value, and empty entries have a
 hash of `0xFFFFFFFF`.  If the entry's hash matches and the entry's key
 is equal to the one we are looking for, then we have our match.  If the
--- a/src/lib/README.ctable.md
+++ b/src/lib/README.ctable.md
@@ -4,12 +4,12 @@ A ctable is a hash table whose keys and values are instances of FFI
 data types.  In Lua parlance, an FFI value is a "cdata" value, hence the
 name "ctable".
 
-A ctable is parameterized the specific types for its keys and values.
-This allows for the table to be stored in an efficient manner.  Adding
-an entry to a ctable will copy the value into the table.  Logically, the
-table "owns" the value.  Lookup can either return a pointer to the value
-in the table, or copy the value into a user-supplied buffer, depending
-on what is most convenient for the user.
+A ctable is parameterized for the specific types for its keys and
+values.  This allows for the table to be stored in an efficient manner.
+Adding an entry to a ctable will copy the value into the table.
+Logically, the table "owns" the value.  Lookup can either return a
+pointer to the value in the table, or copy the value into a
+user-supplied buffer, depending on what is most convenient for the user.
 
 As an implementation detail, the table is stored as an open-addressed
 robin-hood hash table with linear probing.  This means that to look up a
--- a/src/lib/README.ctable.md
+++ b/src/lib/README.ctable.md
@@ -1,0 +1,198 @@
+### `ctable` (lib.ctable)
+
+A ctable is a hash table whose keys and values are instances of FFI
+data types.  In Lua parlance, an FFI value is a "cdata" value, hence the
+name "ctable".
+
+A ctable is parameterized the specific types for its keys and values.
+This allows for the table to be stored in an efficient manner.  Adding
+an entry to a ctable will copy the value into the table.  Logically, the
+table "owns" the value.  Lookup can either return a pointer to the value
+in the table, or copy the value into a user-supplied buffer, depending
+on what is most convenient for the user.
+
+As an implementation detail, the table is stored as an open-addressed
+robin-hood hash table with linear probing.  This means that to look up a
+key in the table, we take its hash value (using a user-supplied hash
+function), map that hash value to an index into the table by scaling the
+hash to the table size, and then scan forward in the table until we find
+an entry whose hash value that is greater than or equal to the hash in
+question.  Each entry stores its hash value, and empty entries have a
+hash of `0xFFFFFFFF`.  If the entry's hash matches and the entry's key
+is equal to the one we are looking for, then we have our match.  If the
+entry's hash is greater than our hash, then we have a failure.  Hash
+collisions are possible as well of course; in that case we continue
+scanning forward.
+
+The distance travelled while scanning for the matching hash is known as
+the /displacement/.  The table measures its maximum displacement, for a
+number of purposes, but you might be interested to know that a maximum
+displacement for a table with 2 million entries and a 40% load factor is
+around 8 or 9.  Smaller tables will have smaller maximum displacements.
+
+The ctable has two lookup interfaces.  One will perform the lookup as
+described above, scanning through the hash table in place.  The other
+will fetch all entries within the maximum displacement into a buffer,
+then do a branchless binary search over that buffer.  This second
+streaming lookup can also fetch entries for multiple keys in one go.
+This can amortize the cost of a round-trip to RAM, in the case where you
+expect to miss cache for every lookup.
+
+To create a ctable, first create a parameters table specifying the key
+and value types, along with any other options.  Then call `ctable.new`
+on those parameters.  For example:
+
+```lua
+local ctable = require('lib.ctable')
+local ffi = require('ffi')
+local params = {
+   key_type = ffi.typeof('uint32_t'),
+   value_type = ffi.typeof('int32_t[6]'),
+   hash_fn = ctable.hash_i32,
+   max_occupancy_rate = 0.4,
+   initial_size = math.ceil(occupancy / 0.4)
+}
+local ctab = ctable.new(params)
+```
+
+— Function **ctable.new** *params*
+
+Create a new ctable.  *params* is a table of key/value pairs.  The
+following keys are required:
+
+ * `key_type`: An FFI type for keys in this table.
+ * `value_type`: An FFI type for values in this table.  (In the future,
+   `value_type` will be optional; a nil `value_type` will create a
+   set).
+ * `hash_fn`: A function that takes a key and returns a hash value.
+
+Hash values are unsigned 32-bit integers in the range `[0,
+0xFFFFFFFF)`.  That is to say, `0xFFFFFFFF` is the only unsigned 32-bit
+integer that is not a valid hash value.  The `hash_fn` must return a
+hash value in the correct range.
+
+Optional entries that may be present in the *params* table include:
+
+ * `initial_size`: The initial size of the hash table, including free
+   space.  Defaults to 8 slots.
+ * `max_occupancy_rate`: The maximum ratio of `occupancy/size`, where
+   `occupancy` denotes the number of entries in the table, and `size` is
+   the total table size including free entries.  Trying to add an entry
+   to a "full" table will cause the table to grow in size by a factor of
+   2.  Defaults to 0.9, for a 90% maximum occupancy ratio.
+ * `min_occupancy_rate`: Minimum ratio of `occupancy/size`.  Removing an
+   entry from an "empty" table will shrink the table.
+
+#### Methods
+
+Users interact with a ctable through methods.  In these method
+descriptions, the object on the left-hand-side of the method invocation
+should be a ctable.
+
+— Method **:resize** *size*
+
+Resize the ctable to have *size* total entries, including empty space.
+
+— Method **:insert** *hash* *key* *value* *updates_allowed*
+
+An internal helper method that does the bulk of updates to hash table.
+*hash* is the hash of *key*.  This method takes the hash as an explicit
+parameter because it is used when resizing the table, and that way we
+avoid calling the hash function in that case.  *key* and *value* are FFI
+values for the key and the value, of course.
+
+*updates_allowed* is an optional parameter.  If not present or false,
+then the `:insert` method will raise an error if the *key* is already
+present in the table.  If *updates_allowed* is the string `"required"`,
+then an error will be raised if *key* is /not/ already in the table.
+Any other true value allows updates but does not require them.  An
+update will replace the existing entry in the table.
+
+Returns the index of the inserted entry.
+
+— Method **:add** *key* *value* *updates_allowed*
+
+Add an entry to the ctable, returning the index of the added entry.  See
+the documentation for `:insert` for a description of the parameters.
+
+— Method **:update** *key* *value*
+
+Update the entry in a ctable with the key *key* to have the new value
+*value*.  Throw an error if *key* is not present in the table.
+
+— Method **:lookup_ptr** *key*
+
+Look up *key* in the table, and if found return a pointer to the entry.
+Return nil if the value is not found.
+
+An entry pointer has three fields: the `hash` value, which must not be
+modified; the `key' itself; and the `value`.  Access them as usual in
+Lua:
+
+```lua
+local ptr = ctab:lookup(key)
+if ptr then print(ptr.value) end
+```
+
+Note that pointers are only valid until the next modification of a
+table.
+
+— Method **:lookup_and_copy** *key* *entry*
+
+Look up *key* in the table, and if found, copy that entry into *entry*
+and return true.  Otherwise return false.
+
+— Method **:remove_ptr** *entry*
+
+Remove an entry from a ctable.  *entry* should be a pointer that points
+into the table.  Note that pointers are only valid until the next
+modification of a table.
+
+— Method **:remove** *key* *missing_allowed*
+
+Remove an entry from a ctable, keyed by *key*.
+
+Return true if we actually do find a value and remove it.  Otherwise if
+no entry is found in the table and *missing_allowed* is true, then
+return false.  Otherwise raise an error.
+
+— Method **:selfcheck**
+
+Run an expensive internal diagnostic to verify that the table's internal
+invariants are fulfilled.
+
+— Method **:dump**
+
+Print out the entries in a table.  Can be expensive if the table is
+large.
+
+— Method **:iterate**
+
+Return an iterator for use by `for in`.  For example:
+
+```lua
+for entry in ctab:iterate() do
+   print(entry.key, entry.value)
+end
+```
+
+#### Hash functions
+
+Any hash function will do, as long as it produces values in the right
+range.  In practice we include some functions for hashing byte sequences
+of some common small lengths.
+
+— Function **ctable.hash_32** *number*
+Hash a 32-bit integer.  As a `hash_fn` parameter, this will only work if
+your key type's Lua representation is a Lua number.  For example, use
+`hash_32` on `ffi.typeof('uint32_t')`, but use `hashv_32` on
+`ffi.typeof('uint8_t[4]')`.
+
+— Function **ctable.hashv_32** *ptr*
+Hash the first 32 bits of a byte sequence.
+
+— Function **ctable.hashv_48** *ptr*
+Hash the first 48 bits of a byte sequence.
+
+— Function **ctable.hashv_64** *ptr*
+Hash the first 64 bits of a byte sequence.
--- a/src/lib/README.md
+++ b/src/lib/README.md
@@ -1,6 +1,8 @@
-### `ctable` (lib.ctable)
+## Specialized data structures
 
-A ctable is a hash table whose keys and values are instances of FFI
+### Ctable (lib.ctable)
+
+A *ctable* is a hash table whose keys and values are instances of FFI
 data types.  In Lua parlance, an FFI value is a "cdata" value, hence the
 name "ctable".
 
@@ -25,7 +27,7 @@ collisions are possible as well of course; in that case we continue
 scanning forward.
 
 The distance travelled while scanning for the matching hash is known as
-the /displacement/.  The table measures its maximum displacement, for a
+the *displacement*.  The table measures its maximum displacement, for a
 number of purposes, but you might be interested to know that a maximum
 displacement for a table with 2 million entries and a 40% load factor is
 around 8 or 9.  Smaller tables will have smaller maximum displacements.
@@ -55,15 +57,13 @@ local params = {
 local ctab = ctable.new(params)
 ```
 
-— Function **ctable.new** *params*
+— Function **ctable.new** *parameters*
 
-Create a new ctable.  *params* is a table of key/value pairs.  The
+Create a new ctable.  *parameters* is a table of key/value pairs.  The
 following keys are required:
 
- * `key_type`: An FFI type for keys in this table.
- * `value_type`: An FFI type for values in this table.  (In the future,
-   `value_type` will be optional; a nil `value_type` will create a
-   set).
+ * `key_type`: An FFI type (LuaJIT "ctype") for keys in this table.
+ * `value_type`: An FFI type (LuaJT "ctype") for values in this table.
  * `hash_fn`: A function that takes a key and returns a hash value.
 
 Hash values are unsigned 32-bit integers in the range `[0,
@@ -71,7 +71,7 @@ Hash values are unsigned 32-bit integers in the range `[0,
 integer that is not a valid hash value.  The `hash_fn` must return a
 hash value in the correct range.
 
-Optional entries that may be present in the *params* table include:
+Optional entries that may be present in the *parameters* table include:
 
  * `initial_size`: The initial size of the hash table, including free
    space.  Defaults to 8 slots.
@@ -93,7 +93,7 @@ should be a ctable.
 
 Resize the ctable to have *size* total entries, including empty space.
 
-— Method **:insert** *hash* *key* *value* *updates_allowed*
+— Method **:insert** *hash*, *key*, *value*, *updates_allowed*
 
 An internal helper method that does the bulk of updates to hash table.
 *hash* is the hash of *key*.  This method takes the hash as an explicit
@@ -104,18 +104,18 @@ values for the key and the value, of course.
 *updates_allowed* is an optional parameter.  If not present or false,
 then the `:insert` method will raise an error if the *key* is already
 present in the table.  If *updates_allowed* is the string `"required"`,
-then an error will be raised if *key* is /not/ already in the table.
+then an error will be raised if *key* is *not* already in the table.
 Any other true value allows updates but does not require them.  An
 update will replace the existing entry in the table.
 
 Returns the index of the inserted entry.
 
-— Method **:add** *key* *value* *updates_allowed*
+— Method **:add** *key*, *value*, *updates_allowed*
 
 Add an entry to the ctable, returning the index of the added entry.  See
 the documentation for `:insert` for a description of the parameters.
 
-— Method **:update** *key* *value*
+— Method **:update** *key*, *value*
 
 Update the entry in a ctable with the key *key* to have the new value
 *value*.  Throw an error if *key* is not present in the table.
@@ -126,7 +126,7 @@ Look up *key* in the table, and if found return a pointer to the entry.
 Return nil if the value is not found.
 
 An entry pointer has three fields: the `hash` value, which must not be
-modified; the `key' itself; and the `value`.  Access them as usual in
+modified; the `key` itself; and the `value`.  Access them as usual in
 Lua:
 
 ```lua
@@ -137,7 +137,7 @@ if ptr then print(ptr.value) end
 Note that pointers are only valid until the next modification of a
 table.
 
-— Method **:lookup_and_copy** *key* *entry*
+— Method **:lookup_and_copy** *key*, *entry*
 
 Look up *key* in the table, and if found, copy that entry into *entry*
 and return true.  Otherwise return false.
@@ -148,7 +148,7 @@ Remove an entry from a ctable.  *entry* should be a pointer that points
 into the table.  Note that pointers are only valid until the next
 modification of a table.
 
-— Method **:remove** *key* *missing_allowed*
+— Method **:remove** *key*, *missing_allowed*
 
 Remove an entry from a ctable, keyed by *key*.
 
@@ -178,21 +178,25 @@ end
 
 #### Hash functions
 
-Any hash function will do, as long as it produces values in the right
-range.  In practice we include some functions for hashing byte sequences
-of some common small lengths.
+Any hash function will do, as long as it produces values in the `[0,
+0xFFFFFFFF)` range.  In practice we include some functions for hashing
+byte sequences of some common small lengths.
 
 — Function **ctable.hash_32** *number*
+
 Hash a 32-bit integer.  As a `hash_fn` parameter, this will only work if
 your key type's Lua representation is a Lua number.  For example, use
 `hash_32` on `ffi.typeof('uint32_t')`, but use `hashv_32` on
 `ffi.typeof('uint8_t[4]')`.
 
 — Function **ctable.hashv_32** *ptr*
+
 Hash the first 32 bits of a byte sequence.
 
 — Function **ctable.hashv_48** *ptr*
+
 Hash the first 48 bits of a byte sequence.
 
 — Function **ctable.hashv_64** *ptr*
+
 Hash the first 64 bits of a byte sequence.
--- a/src/lib/ctable.lua
+++ b/src/lib/ctable.lua
@@ -11,6 +11,9 @@ local max, floor, ceil = math.max, math.floor, math.ceil
 CTable = {}
 
 local HASH_MAX = 0xFFFFFFFF
+local uint16_ptr_t = ffi.typeof('uint16_t*')
+local uint32_ptr_t = ffi.typeof('uint32_t*')
+local uint64_ptr_t = ffi.typeof('uint64_t*')
 
 local function make_entry_type(key_type, value_type)
    return ffi.typeof([[struct {
@@ -38,13 +41,20 @@ local function make_equal_fn(key_type)
       return function (a, b)
          return a == b
       end
+   elseif size == 2 then
+      return function (a, b)
+         return cast(uint16_ptr_t, a)[0] == cast(uint16_ptr_t, b)[0]
+      end
    elseif size == 4 then
-      local uint32_ptr_t = ffi.typeof('uint32_t*')
       return function (a, b)
          return cast(uint32_ptr_t, a)[0] == cast(uint32_ptr_t, b)[0]
       end
+   elseif size == 6 then
+      return function (a, b)
+         return (cast(uint32_ptr_t, a)[0] == cast(uint32_ptr_t, b)[0] and
+                 cast(uint16_ptr_t, a)[2] == cast(uint16_ptr_t, b)[2])
+      end
    elseif size == 8 then
-      local uint64_ptr_t = ffi.typeof('uint64_t*')
       return function (a, b)
          return cast(uint64_ptr_t, a)[0] == cast(uint64_ptr_t, b)[0]
       end
@@ -360,7 +370,7 @@ end
 -- TODO: Switch to a hash function with good security properties,
 -- perhaps by using the DynASM support for AES.
 local uint32_cast = ffi.new('uint32_t[1]')
-function hash_i32(i32)
+function hash_32(i32)
    i32 = tobit(i32)
    i32 = i32 + bnot(lshift(i32, 15))
    i32 = bxor(i32, (rshift(i32, 10)))
@@ -377,6 +387,25 @@ function hash_i32(i32)
    return uint32_cast[0]
 end
 
+function hashv_32(key)
+   return hash_32(cast(uint32_ptr_t, key)[0])
+end
+
+function hashv_48(key)
+   local hi = cast(uint32_ptr_t, key)[0]
+   local lo = cast(uint16_ptr_t, key)[2]
+   -- Extend lo to the upper half too so that the hash function isn't
+   -- spreading around needless zeroes.
+   lo = bor(lo, lshift(lo, 16))
+   return hash_32(bxor(hi, hash_32(lo)))
+end
+
+function hashv_64(key)
+   local hi = cast(uint32_ptr_t, key)[0]
+   local lo = cast(uint32_ptr_t, key)[1]
+   return hash_32(bxor(hi, hash_32(lo)))
+end
+
 function selftest()
    print("selftest: ctable")
 
@@ -385,7 +414,7 @@ function selftest()
    local params = {
       key_type = ffi.typeof('uint32_t'),
       value_type = ffi.typeof('int32_t[6]'),
-      hash_fn = hash_i32,
+      hash_fn = hash_32,
       max_occupancy_rate = 0.4,
       initial_size = ceil(occupancy / 0.4)
    }
@@ -419,26 +448,19 @@ function selftest()
 
    -- A check that our equality functions work as intended.
    local numbers_equal = make_equal_fn(ffi.typeof('int'))
-   local four_byte = ffi.typeof('uint32_t[1]')
-   local eight_byte = ffi.typeof('uint32_t[2]')
-   local twelve_byte = ffi.typeof('uint32_t[3]')
-   local four_byte_equal = make_equal_fn(four_byte)
-   local eight_byte_equal = make_equal_fn(eight_byte)
-   local twelve_byte_equal = make_equal_fn(twelve_byte)
    assert(numbers_equal(1,1))
    assert(not numbers_equal(1,2))
-   assert(four_byte_equal(ffi.new(four_byte, {1}),
-                          ffi.new(four_byte, {1})))
-   assert(not four_byte_equal(ffi.new(four_byte, {1}),
-                              ffi.new(four_byte, {2})))
-   assert(eight_byte_equal(ffi.new(eight_byte, {1,1}),
-                           ffi.new(eight_byte, {1,1})))
-   assert(not eight_byte_equal(ffi.new(eight_byte, {1,1}),
-                               ffi.new(eight_byte, {1,2})))
-   assert(twelve_byte_equal(ffi.new(twelve_byte, {1,1,1}),
-                            ffi.new(twelve_byte, {1,1,1})))
-   assert(not twelve_byte_equal(ffi.new(twelve_byte, {1,1,1}),
-                                ffi.new(twelve_byte, {1,1,2})))
+
+   local function check_bytes_equal(type, a, b)
+      local equal_fn = make_equal_fn(type)
+      assert(equal_fn(ffi.new(type, a), ffi.new(type, a)))
+      assert(not equal_fn(ffi.new(type, a), ffi.new(type, b)))
+   end
+   check_bytes_equal(ffi.typeof('uint16_t[1]'), {1}, {2})         -- 2 byte
+   check_bytes_equal(ffi.typeof('uint32_t[1]'), {1}, {2})         -- 4 byte
+   check_bytes_equal(ffi.typeof('uint16_t[3]'), {1,1,1}, {1,1,2}) -- 6 byte
+   check_bytes_equal(ffi.typeof('uint32_t[2]'), {1,1}, {1,2})     -- 8 byte
+   check_bytes_equal(ffi.typeof('uint32_t[3]'), {1,1,1}, {1,1,2}) -- 12 byte
 
    print("selftest: ok")
 end
--- a/src/lib/ctable.lua
+++ b/src/lib/ctable.lua
@@ -1,0 +1,444 @@
+module(..., package.seeall)
+
+local ffi = require("ffi")
+local C = ffi.C
+local S = require("syscall")
+local bit = require("bit")
+local bxor, bnot = bit.bxor, bit.bnot
+local tobit, lshift, rshift = bit.tobit, bit.lshift, bit.rshift
+local max, floor, ceil = math.max, math.floor, math.ceil
+
+CTable = {}
+
+local HASH_MAX = 0xFFFFFFFF
+
+local function make_entry_type(key_type, value_type)
+   return ffi.typeof([[struct {
+         uint32_t hash;
+         $ key;
+         $ value;
+      } __attribute__((packed))]],
+      key_type,
+      value_type)
+end
+
+local function make_entries_type(entry_type)
+   return ffi.typeof('$[?]', entry_type)
+end
+
+-- hash := [0,HASH_MAX); scale := size/HASH_MAX
+local function hash_to_index(hash, scale)
+   return floor(hash*scale + 0.5)
+end
+
+local function make_equal_fn(key_type)
+   local size = ffi.sizeof(key_type)
+   local cast = ffi.cast
+   if tonumber(ffi.new(key_type)) then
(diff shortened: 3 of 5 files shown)


andywingo:
Probably needs some documentation.  What kind?

This is ported from the PodHashMap that we use in the lwaftr.  We have performance tests but I don't know how to support them in Snabb yet.

I also removed the streaming lookup bits from this table.  We can add those in a separate PR, as they rely on a couple of dynasm modules.  As you like.

andywingo:
Another thing I left out was support for saving the table to disk and loading it back again.  Will do that in a future commit.

lukego:
@alexandergall what would it take for you to adopt this for the learning bridge (#638) as a step towards consolidating our table structures?

lukego:
(See also #718)

eugeneia:
> Probably needs some documentation. What kind?

@andywingo What do you mean by “What kind?” Generally any public API has to be accompanied by a complete [API documentation](https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/documentation-guide.md). Also useful would be to update this PR's description to tell:

* what this code does (motivation)
* how it does it (unless trivially obvious)
* if there are any related discussions/references, e.g. #718

andywingo:
@alexandergall Here's a test MAC table:

```lua
function mac_table_check()
   local uint32_ptr_t = ffi.typeof('uint32_t*')
   local uint16_ptr_t = ffi.typeof('uint16_t*')
   local key_t = ffi.typeof('struct { uint8_t mac[6]; }')
   local value_t = ffi.typeof('struct { uint16_t lo; uint16_t hi; }')
   local cast = ffi.cast
   local bor = bit.bor

   local function hash_mac(key)
      local hi = cast(uint32_ptr_t, key.mac)[0]
      local lo = cast(uint16_ptr_t, key.mac + 4)[0]
      -- Extend lo to the upper half too so that the hash function isn't
      -- spreading around needless zeroes.
      lo = bor(lo, lshift(lo, 16))
      return hash_i32(bxor(hi, hash_i32(lo)))
   end

   -- 14-byte entries
   local occupancy = 2e5
   local params = {
      key_type = key_t,
      value_type = value_t,
      hash_fn = hash_mac,
      max_occupancy_rate = 0.4,
      initial_size = ceil(occupancy / 0.4)
   }
   local ctab = new(params)

   -- Fill with { i, 0 } => { bnot(i) }.
   do
      local k = key_t()
      local v = value_t()
      for i = 1,occupancy do
         cast(uint32_ptr_t, k.mac)[0] = i
         cast(uint16_ptr_t, k.mac+4)[0] = 0
         cast(uint32_ptr_t, v)[0] = bnot(i)
         ctab:add(k, v)
      end
   end   

   local pmu = require('lib.pmu')
   local has_pmu_counters, err = pmu.is_available()
   if not has_pmu_counters then
      print('No PMU available: '..err)
   end

   if has_pmu_counters then pmu.setup() end

   local function measure(f, iterations)
      local set
      if has_pmu_counters then set = pmu.new_counter_set() end
      local start = C.get_time_ns()
      if has_pmu_counters then pmu.switch_to(set) end
      local res = f(iterations)
      if has_pmu_counters then pmu.switch_to(nil) end
      local stop = C.get_time_ns()
      local ns = tonumber(stop-start)
      local cycles = nil
      if has_pmu_counters then cycles = pmu.to_table(set).cycles end
      return cycles, ns, res
   end

   local function check_perf(f, iterations, max_cycles, max_ns, what)
      require('jit').flush()
      io.write(tostring(what or f)..': ')
      io.flush()
      local cycles, ns, res = measure(f, iterations)
      if cycles then
         cycles = cycles/iterations
         io.write(('%.2f cycles, '):format(cycles))
      end
      ns = ns/iterations
      io.write(('%.2f ns per iteration (result: %s)\n'):format(
            ns, tostring(res)))
      if cycles and cycles > max_cycles then
         print('WARNING: perfmark failed: exceeded maximum cycles '..max_cycles)
      end
      if ns > max_ns then
         print('WARNING: perfmark failed: exceeded maximum ns '..max_ns)
      end
      return res
   end

   local function test_lookup(count)
      local result
      local k = key_t()
      for i = 1, count do
         cast(uint32_ptr_t, k)[0] = i
         result = ctab:lookup_ptr(k).value.lo
      end
      return result
   end

   check_perf(test_lookup, 2e5, 300, 100, 'lookup (40% occupancy)')
end
```

Add that to ctable.lua and test on your machines.  Here on this old i7-3770 sandy bridge desktop I get about 50ns/lookup.  This CL doesn't have streaming lookups yet though, so I can't give those numbers.

andywingo:
Thanks @eugeneia, that's what I was asking for.  I am surprised a bit by the conventions tho -- to me a PR shouldn't say what the code does, the code should say it, augmented by comments if needed.  This module is lacking the documentation block of course.  I also don't really get the purpose of out-of-line per-function documentation; it is bound to drift out of sync.  We're always spelunking in the code anyway, to me the right place to document is there.  Obviously we disagree about how to do things :)

kbara:
Frankly, I'm extremely glad to have the out-of-line per-exposed-API documentation.
a) It can be skimmed a lot faster than the code
b) It gives a clue as to what's public API vs implementation detail in the implementor's head

I pretty much always end up reading the code after, but that serves a different set of purposes.

eugeneia:
> Thanks @eugeneia, that's what I was asking for. I am surprised a bit by the conventions tho -- to me a PR shouldn't say what the code does, the code should say it, augmented by comments if needed. This module is lacking the documentation block of course. I also don't really get the purpose of out-of-line per-function documentation; it is bound to drift out of sync. We're always spelunking in the code anyway, to me the right place to document is there. Obviously we disagree about how to do things :)

Basically, what @kbara said, but to elaborate a bit:

* I need to know the intention to be able to do efficient code review. I can only tell that the code is incorrect if I know what its intentions are. 

* We need complete and correct API documentation to efficiently use the code. The whole point of an API is that I *do not* need to read the code.

> I also don't really get the purpose of out-of-line per-function documentation; it is bound to drift out of sync.

Documentation and actual API will *not* drift out of sync because I do code review and will not accept PRs that break the API.

andywingo:
I feel like I'm being lectured-to here a bit and I don't enjoy it.  Respectfully @eugeneia I would have thought that as a person who does code review that you would have followed some of the long, long discussions about ffi hash tables that have taken place over the last 3 months and for that reason we would start from implementation concerns rather than process.  Anyway, thank you for the documentation link, and I look forward to your comments on the code.

alexandergall:
@andywingo @lukego I would certainly appreciate if I can replace my MAC table hack with this. Unfortunately, I don't have the time right now (need to finish the NixOS packaging first). I haven't looked at the code either, but curious how this affects the issue I was having with "branchy loops" in my bridge app. If someone volunteered to write a performant bridge app for me, I wouldn't mind :)

andywingo:
@alexandergall if you use the streaming lookup, there are no branches at all if you have no hash collisions.  That's because it fetches the entire bucket, so to speak, into cache, then uses a branchless binary search for the hash in that bucket.  (I use the terminology you use in your mac table.)  If you use the API in this PR you can get side traces though because of the linear probe.  On the other hand this API fetches less memory than a streaming lookup, which can make it competitive.  In practice we haven't experienced the branchy loop problem (yet?), and this table is integrated in the lwaftr.

Question for you though: what is your performance target?  Is 50 ns/lookup OK?

andywingo:
On an interlaken-alike haswell E5-2620v3, 20 runs of that benchmark I pasted above:

```
lookup (40% occupancy): 109.41 cycles, 34.24 ns per iteration (result: 62143)
lookup (40% occupancy): 108.10 cycles, 34.07 ns per iteration (result: 62143)
lookup (40% occupancy): 109.62 cycles, 34.31 ns per iteration (result: 62143)
lookup (40% occupancy): 109.20 cycles, 34.17 ns per iteration (result: 62143)
lookup (40% occupancy): 109.52 cycles, 34.28 ns per iteration (result: 62143)
lookup (40% occupancy): 109.35 cycles, 34.22 ns per iteration (result: 62143)
lookup (40% occupancy): 109.16 cycles, 34.16 ns per iteration (result: 62143)
lookup (40% occupancy): 109.44 cycles, 34.53 ns per iteration (result: 62143)
lookup (40% occupancy): 109.65 cycles, 34.57 ns per iteration (result: 62143)
lookup (40% occupancy): 107.28 cycles, 34.04 ns per iteration (result: 62143)
lookup (40% occupancy): 109.63 cycles, 34.31 ns per iteration (result: 62143)
lookup (40% occupancy): 109.41 cycles, 34.24 ns per iteration (result: 62143)
lookup (40% occupancy): 109.85 cycles, 34.63 ns per iteration (result: 62143)
lookup (40% occupancy): 101.95 cycles, 31.91 ns per iteration (result: 62143)
lookup (40% occupancy): 109.46 cycles, 34.26 ns per iteration (result: 62143)
lookup (40% occupancy): 109.28 cycles, 34.20 ns per iteration (result: 62143)
lookup (40% occupancy): 102.32 cycles, 32.27 ns per iteration (result: 62143)
lookup (40% occupancy): 109.40 cycles, 34.24 ns per iteration (result: 62143)
lookup (40% occupancy): 109.42 cycles, 34.24 ns per iteration (result: 62143)
lookup (40% occupancy): 109.40 cycles, 34.24 ns per iteration (result: 62143)
```

lukego:
@eugeneia @andywingo Github Pull Requests are unfortunately ambiguous. People open a PR with a specific intention, but what is that intention? (To get code accepted upstream ASAP? To get early feedback on a design before going forward? Just to be social and show the code they are working on?). So it is easy to rub each other the wrong way e.g. don't bother to merge something that you assume is a draft, or give "I can't merge this until X,Y,Z" feedback on code that is not prepared for merge.

I'm not sure how to reduce this. I tend to put hints in my PR titles like `[draft]` to mean "I'm probably going to throw this away and rewrite it" or `[wip]` to mean "here's something I am working on that I don't want to be merged yet".

eugeneia:
@andywingo No lecturing intended at all. I am sorry if my response came across as harsh, it was not intended to, my sole intent was to answer the questions asked.

@lukego raises some points that apply here regarding PR intents (probably a good idea to “standardize” the `[draft]` and `[wip]` tags, actually we could use labels for that if the work out the permissions).

E.g. my perspective was “what is required for this to land in the next release” but maybe that isn't the intent at all here. I have indeed not followed the ffi hash table discussion, so some context in the PR description would help me to get into the code. Where is that discussion by the way, is #718 a follow up?

alexandergall:
@andywingo Thanks for the explanation. That sounds terrific and should eliminate a big part of the problem. There was a branchy loop in the bridge code as well (determining whether to unicast or flood a packet). I might come back to you for advice how to eliminate that one, too. And, yes, 50ns/lookup is nothing I would complain about :) My performance goal for the entire VPN app is fairly moderate, something like 6-8Mpps. But getting faster is always nice.

lukego:
@andywingo @alexandergall you guys might want to talk cycles because I think you're testing with much different CPU clock speedw

alexandergall:
@lukego The clock on my system is 3.5GHz.

lukego:
@wingo Thank you for doing this work and sending it upstream!

The "need a big table for FFI keys/values" problem has been with us since pretty much the beginning of Snabb Switch, way back when @plajjan did his DDoS protection prototype. Since then we have tried a lot of different things but never found a really general solution: until now it would seem.

I think the code is extremely stylish btw :-) both on this PR and also the asm code for searching the table.  I would not have guessed it would be possible to have a sophisticated optimized hashtable in so few lines of simple code.

For my part I had experimented with bindings to a few off-the-shelf libraries ("Judy arrays in 1250 lines", khash, a patricia trie) but was not satisfied with any of those and gradually coming to the conclusion that what we really wanted was something almost exactly like what you have written here.

In the back of my mind I wonder if some day we could unify our hashtables with the builtin table support in LuaJIT -- without mixing up FFI data with the Lua heap -- but I have no idea if that is practical and it seems like something for later if at all.

wingo:
Hi :)  Picking up this thread.  The reason I initially asked about the documentation format was because there was no obvious place to which to add documentation, and the lib.foo modules in general are undocumented, with some exceptions, and the modules that are documented have their own subdirectories.  There was no example.  I will instead add a lib/README.ctable.md and include it in genbook.sh.  Let me know if there is a better way, @eugeneia :)

eugeneia:
@wingo Eh, please no `README.foo.md`, any reason why `ctable` couldn't be in its own subdirectory? Alternatively I suggest just using `src/lib/README.md` with a subsection for `ctable`. I would prefer the subdirectory.

wingo:
@eugeneia Anything is possible :)  However it seems to me that it would be worse to require the user to do

```lua
local ctable = require('lib.ctable.ctable')
```

than

```lua
local ctable = require('lib.ctable')
```

For that reason I think a subdirectory is not the right thing.  Thoughts on naming are still welcome though :)

eugeneia:
@wingo I agree that the repeating module names are not pretty but plenty of other Snabb Switch modules already look like that (for the same reason). Its just a cosmetic issue that does not impede on functionality so I tend to not consider this an argument against the sub directory.

wingo:
I feel strongly enough about the unnecessary module hierarchy that I will resubmit a patch without the additional subdir.  Though it's a factor to consider, to me the documentation build system should not dictate the module hierarchy.

Feel free to NAK if you also feel strongly about this.

eugeneia:
The documentation build system does not dictate the module hierarchy, but rather the general build system. E.g. in order to have a module that consists of N files and having a directory that contains all N files the module hierarchy has to reflect this. Imho this issue is not severe enough to justify changing the build system.

I personally think there is value to having documentation (in form of README.md) and code in a dedicated directory (GitHub plays nicely with this, as it will display the README.md when browsing a directory). See [/SnabbCo/snabbswitch/tree/master/src/lib/hardware](https://github.com/SnabbCo/snabbswitch/tree/master/src/lib/hardware) for an example where this works nicely (in both ways, e.g. `hardware/` contains two modules that share a README.md).

That being said, I am not against putting `ctable`'s docs into `src/lib/README.md`.

wingo:
So, probably the documentation needs to move; but its final location depends on how it's integrated into the document, so please see the genbook change, @eugeneia.

These commits went ahead and added README.ctable.md even after you expressed some disagreement there.  I was going to change, but then looking on how to integrate the docs into the main book I hesitated and left them where they are for now.  I didn't add lib/README.md because I didn't know how to section the documentation for this module, relative to the other modules that are in lib/ but already in the genbook.sh.  To have a comprehensible documentation of lib/ means that to me perhaps we need to *not* have a src/lib/README.md.

I anticipate adding a few more of these specialized data structures in other commits, so let's set the pattern correctly here.  @eugeneia for me this PR is ready to go, please tell me the changes I must make.  Thanks! :)

src/lib/README.ctable.md
```diff
@@ -0,0 +1,198 @@
+### `ctable` (lib.ctable)
+
+A ctable is a hash table whose keys and values are instances of FFI
+data types.  In Lua parlance, an FFI value is a "cdata" value, hence the
+name "ctable".
+
+A ctable is parameterized the specific types for its keys and values.
```
plajjan: Syntax error, I can't read this sentence ;)

src/lib/README.ctable.md
```diff
@@ -0,0 +1,198 @@
...
+table "owns" the value.  Lookup can either return a pointer to the value
+in the table, or copy the value into a user-supplied buffer, depending
+on what is most convenient for the user.
+
+As an implementation detail, the table is stored as an open-addressed
+robin-hood hash table with linear probing.  This means that to look up a
+key in the table, we take its hash value (using a user-supplied hash
+function), map that hash value to an index into the table by scaling the
+hash to the table size, and then scan forward in the table until we find
+an entry whose hash value that is greater than or equal to the hash in
```
plajjan: s/that//

lukego:
re: `lib.ctable.ctable` this makes me crazy too. There is actually a standard Lua idiom for avoiding this i.e. when you `require("foo")` it first looks for `foo.lua` and then for `foo/init.lua` so that you can transparently implement a module in several files. However, I don't believe this nice feature is available in LuaJIT when working with Lua modules that are linked as object code. I hope that we can clean this up in #734.

@wingo Rebasing open PRs creates problems for me. I can't really follow the review on Github, e.g. Kristian's comments have disappeared, and it would also burn me if I had merged this branch onto one of my own branches and now wanted to pull in the latest updates. This is why [git-workflow.md](https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/git-workflow.md) requests that branches not be rebased after the Pull Request is made. (I do think we should make an exception for PRs named with the prefix `[sketch]` though... perhaps all of these tips/requests should be migrated into CONTRIBUTING.md?)

plajjan:
My comments are still visible. There's a "show outdated diff"  next to the "plajjan commented on an outdated diff". I really think rebasing is the right thing to do here. Why would we want Snabb git history to reflect spelling mistakes. I don't think you should base your work on random branches on the Internet. First read what it's about and understand the intent. If it's long lived (like the next branches) then go ahead but like this, where it's obviously just to get things in order to quickly merge on master, then expect it to move. Publishing a branch for a PR is no contract...  :)

lukego:
@plajjan This seems like the "centralized rebase-oriented workflow" vs "distributed merge-based workflow" debate. If people want to switch Snabb to a rebase-oriented workflow then a PR to `git-workflow.md` would be the way to propose it.

However, I would ask people to accept that I am trying to establish a specific workflow here, the one that has been very successful in the Linux kernel world, and to help me by embracing this and figuring out how to make it work well for us. I am sure we will want to make a lot of adjustments over time but they should all be in the same spirit.

Just a few notes based on my mental model of the merge-based workflow:

> Why would we want Snabb git history to reflect spelling mistakes.

The same reason you would want your bank statement to show if you paid the wrong amount for something and then corrected that with additional transaction(s). In a merge-based workflow the git history is a ledger keeping track of which commits are in each branch and git uses this information directly when you ask it queries like "do I have all the changes from branch FOO?" or run merge operations like "catch me up with the latest changes in branch FOO".

There should be ways to elide this information when you are not interested in it e.g. run `git log` with the `--merges` option to compress the history into a higher-level view. If people want to see the Snabb history in a different way then I am certainly up for helping to find the right git incantations.

> I don't think you should base your work on random branches on the Internet.

Right, you need to know what to expect from the branch. I expect that branches PR'd to the main Github repo will be developed with a consistent forward-moving history (unless clearly marked otherwise). Being able to rely on this is powerful, for example Alex could have merged this `ctable` into his ALX branch and shipped it in a product without waiting for consensus on the API and for documentation to be available.

> where it's obviously just to get things in order to quickly merge on master

This is a red flag to me. There should not be any urgency to landing changes on master, that would be a bottleneck and a scalability problem. Changes should already be available to users directly from the feature branches where they are maintained. The master branch represents the "trailing edge" of changes that are mature enough to propagate to everybody.

> Publishing a branch for a PR is no contract... :)

People need to know what to expect from each other. I have tried to spell out the expectations for the master branch in `git-workflow.md`. These may need to be written better, or migrated into a CONTRIBUTING.md, but I certainly expect downstream branches to carefully consider the expectations of their upstream branches in order to make things run smoothly.

lukego:
@plajjan ... oh now I can see your comments. I had clicked on the "outdated diff" link but didn't realize that the text of your comment was there if I scroll down below the diff. That is nice i.e. information is not being lost.

Can be that rebasing makes sense in these situations but I still want to understand how it fits into the broader merge-based workflow so that the master branch does not become a bottleneck.

wingo:
For what it's worth I didn't rebase in this PR -- I added commits.  Since they changed the things that @plajjan commented on, github folded thae comments.  I'm OK rebasing or not, doesn't matter much to me, but even in a situation where we do decide rebasing is good, it shouldn't happen while review is underway :)

lukego:
@wingo @plajjan Coming back on topic, here is the sort of Git workflow that I would ideally like to have for ctable:

screen shot 2016-02-03 at 09 05 42

That is:

- The `ctable` feature is developed on a feature branch that lives on a feature branch while the first version of the code is being developed.
- The applications that care about this functionality, for example `lwaftr` and `ALX`, merge the change directly as soon as it is working well enough for their use cases. This is when optimizations and fixes are made.
- Once the code has been battle-tested sufficiently a subsystem maintainer merges it onto the `data-structures` branch for broader integration. This is when the code is fit into the rest of the system e.g. replacing existing code that is rendered obsolete such as [mac_table.c](https://github.com/SnabbCo/snabbswitch/blob/master/src/apps/bridge/mac_table.c), adopting `ctable` to backend the `conntrack` module instead of Lua tables, etc. This is when we cleanup the code and make sure the net result is simplification (less special-purpose table code, clearer solution for the next person who needs a table, etc) rather than obfuscation (yet another hashtable that lives in parallel with the existing ones, confusing picture for the next person who may decide to play it safe and write their own again, etc).
- Once the code base has been updated sufficiently to have "internalized" the `ctable` feature then this can be merged from the `data-structures` branch onto `next` for general release.

I see a lot to like about this picture. The missing pieces I see now are (a) we don't have a `data-structures` branch that somebody is lovingly maintaining and (b) it is not clear whether it is safe for an application like ALX to merge this `ctable` branch i.e. whether this is a stable branch that will be maintained for the lifetime of the initial version of the feature or whether it is a throw-away branch that will be rebased at will.

lukego:
@wingo oh sorry for flaming you for my own lack of understanding of the Github PR UI :-).

wingo:
For the lwaftr, things are a bit more complicated.  There is some functionality that we need that is not implemented in this PR:

 * Loading and saving to disk
 * Streaming lookups, which eneeds
   * binary search dynasm stuff
   * streaming copy dynasm stuff

All of these things exist in the PodHashMap already and have been pretty well tested for our use case and for performance, FWIW.  I have this feeling again that master is receding before my eyes :)

I have a non-interest in maintaining a general-purpose long-lived data-structures branch -- it's too broad a remit :)  I could do a short-lived branch to change other ctable-like things to use it, though.  Like your data-structures branch but more lowly falutin'.  The risk I see is that I don't want to have to adapt the lwaftr twice: once to the ctable API, and then once again when/if we change it.  That makes me nervous :)

lukego:
@wingo Maybe it is time to take the discussion: Could we merge the whole lwAFTR branch onto master? 

You would have the `program/lwaftr/` folder for keeping all code that you currently consider private to your application e.g. load/save of tables to disk and dynasm lookup routines. If other people want to use this code they could make a PR to "promote" it up to `core/` or `lib/` e.g. when refactoring the code so that they can use these libraries in their own application. You could chime in there with your feedback ("Awesome! Great to see it adopted!") or ("We want to keep this private! We have lots more application-specific things we want to try out!").

I don't think this would require much changes to the maintenance workflow that you are already using today. One thing is that you would need to keep application-specific code in `program/lwaftr/` (including stuff you just can't be bothered arguing in review about at the moment) and another thing is that people could propose changes to the `program/lwaftr/` code that is on master (e.g. as part of a major refactoring effort or just to add a feature) and the maintenance workflow would need to make sure you sign off on these before they are accepted.

The benefits I see are:

- Get lwaftr development closely in sync with the master branch.
- Put lwaftr code in plain sight where people can borrow ideas, propose promoting certain code, update lwaftr code as part of sweeping refactorings (like the "straight line" change we did a year ago).
- Take these "upstreaming" discussions in the context of refactoring (move code from `lwaftr/` up to `lib/`) rather than "from scratch" based on code that is in limbo on a feature branch that nobody has merged.
- Make it easier for the upstream CI to run functional and performance regression tests for the lwaftr application.

wingo:
@lukego Sure :)  At one early point I had a vision of landing the thing nicely, in a dozen commits or so, all of which making sense on their own: but maybe that is an unattainable thing :)

Another question, and it's a current one: we have one change to LuaJIT, https://github.com/SnabbCo/snabbswitch/pull/707.  Not merged yet but I was going to merge today.  If I don't merge it I have to avoid merging some other things I had planned for today.  Upstream discussions are complicated: https://github.com/LuaJIT/LuaJIT/pull/123#issuecomment-172511232.  To me in the Snabb context it's an obviously good change; in a more general LuaJIT context I can see that Mike can consider it a somewhat lateral move (better for some cases, worse for others), but unfortunately he didn't comment on an idea for improving it; a bit irritating but oh well.

So, complicated situation, but we felt comfortable making that change while we had a separate branch.  WDYT about that patch in Snabb itself?  There is a PR here: https://github.com/SnabbCo/snabbswitch/pull/707

plajjan:
Assumption is the mother of all evil. I wrote my previous comment from a mobile device and didn't see whether Andy had rebased or not and I simply jumped on the bandwagon, assuming it was rebased, when Luke said he did :P 
I suppose my comments are moot given that he hasn't rebased and I feel I spurred the derailing off this issue from its original topic :/

@wingo how would I use ctable with IPv6? I see src/lib/protocol/ipv6 using char[16] to represent an IPv6. I would instinctively go with uint64_t[2] - either way it's an array. Does this make a difference to how I would use ctable or will it happily eat that as well? Forgive my novice C/FFI knowledge.

wingo:
@plajjan Either way is fine.  There is no pre-supplied hash function for 16-byte values, so you will have to build one, and you can use ffi.cast to access the key in any way you like.  Likewise for equality it will fall back to use C.memcmp instead of inlining the equality check.  Both of these are probably areas for potential improvement, and the dependency is just on the size in bytes and not how it's divided up.

I would use uint8_t[16] because it works better with other parts of snabb, like lib.protocol.

lukego:
> At one early point I had a vision of landing the thing nicely, in a dozen commits or so, all of which making sense on their own: but maybe that is an unattainable thing :)

I know what you mean. For what it is worth, this is not so important from my perspective. I find it most productive to think of Git as a tool for keeping track of *content* and the commit history as a means towards this end. I suspect that this is the way Linus intended it ("git - the stupid content tracker"). So if the first version of lwaftr would come as 500 commits, or as one squashed commit, etc, that is not a big deal for me personally. I have accepted that I will need to learn to query the git history like a graph database in order to make sense of what is going on anyway :-).

The LuaJIT patch is a very interesting one. I would be tempted to bring that onto master because it's written and maintained by people who understand compilers, it is directly needed by one application (so far), and because the LuaJIT upstream situation is still in flux and some degree of self-sufficiency in the Snabb community seems pragmatic. The main cost I see is that if we want to report bugs to LuaJIT upstream we probably need to separately make sure they are reproducible without our own changes (but creating separate minimal test cases seems to be normal practice for reporting bugs upstream anyway).

I do see that there will be situations where an application genuinely does need to adopt and ship changes that are not immediately acceptable to the master branch. This will disrupt the merge workflow. I am not sure to what extent we can anticipate and avoid these problems vs deal with them as they arise. So long as landing code on the master branch is not on the critical path for applications to make their releases I feel like this should be manageable.

plajjan:
@wingo ok!

As part of #718 I would like to see that we have the complete set of functions needed for lookups of common data types, including 16 byte / IPv6. This doesn't mean this PR necessarily needs to include a hash function for IPv6 but I don't think we should put it on the user to write one.

I imagine myself being the target audience of Snabb, as a framework, to implement simple forwarding functions. I understand what a hash function does and I suppose I could convert the supplied hash_32 function to something else but I wouldn't really *understand* what I'm doing. I don't speak ASM. This is why I think Snabb should ship with ready-to-use classes/functions for this :)

lukego:
@plajjan Thank you for patiently sharing your perspective and stimulating discussions on the Git workflow :). I do want to observe that in the Linux community there is actually a *lot* of rebasing done at the edges of the network before changes are accepted by the first hop and so it may be that we will have more rebasing in the Snabb future too. I mostly want to avoid rushing into this too early in case it disrupts building a distributed workflow (which seems like a pretty delicate operation to begin with).

lukego:
@plajjan @wingo Surely LuaJIT must have some nice hashing functions in the C code somewhere. I wonder if we could call those via FFI as a fall-back for cases that don't have a hand-written hash?

wingo:
@lukego LuaJIT hashes object by identity, not by content.  The exception are strings, which have an embedded hash:

https://github.com/SnabbCo/snabbswitch/blob/master/lib/luajit/src/lj_str.c#L132

It's a Jenkins hash like the one in ctable.lua, but it's not available as a C function.  However writing these functions in LuaJIT is faster than calling out to C AFAIU.  The assembly ends up quite nice, and you avoid the optimization barrier.  Maybe 16-byte values are on the edge of what can be done nicely in an inline way.  We could certainly generate better hash functions though for arbitrary key sizes I think.

lukego:
@wingo makes sense, thanks for clearing that up!

lukego:
@wingo I would like to try using `ctable` to backend the `packet_filter.conntrack` module.

There are two challenges. One is that it mixes two sizes of keys in the same table (IPv4 vs IPv6) but it should be straightforward to resolve this by using separate tables. The other is that the keys are larger (37 bytes for IPv6 connections and 13 bytes for IPv4 connections) and I am not sure the most straightforward way to hash those? (The hash is actually over multiple separate fields in the packet header but the code currently copies those into a linear structure and then hashes that like a string.)

These are the definitions of the keys:

```lua
ffi.cdef [[
   typedef struct {
      uint32_t src_ip, dst_ip;
      uint16_t src_port, dst_port;
      uint8_t protocol;
   } __attribute__((packed)) conn_spec_ipv4;

   typedef struct {
      uint64_t a, b;
   } __attribute__((packed)) ipv6_addr;

   typedef struct {
      ipv6_addr src_ip, dst_ip;
      uint16_t src_port, dst_port;
      uint8_t protocol;
   } __attribute__((packed)) conn_spec_ipv6;
]]
```

There is some very basic performance test available that could be sanity-checked for impact on lookup speeds i.e. how many packets are processed in one second:

```
[luke@lugano-1:~/git/snabbswitch/src]$ sudo ./snabb snsh -t apps.packet_filter.pcap_filter
selftest: pcap_filter
Run for 1 second (stateful = false)...
link report:
             757,532 sent on pcap_filter.output -> sink.input (loss rate: 0%)
          20,327,070 sent on repeater.output -> pcap_filter.input (loss rate: 0%)
                 161 sent on source.output -> repeater.input (loss rate: 0%)
ok: accepted 3.7267% of inputs (within tolerance)
Run for 1 second (stateful = true)...
link report:
             724,611 sent on pcap_filter.output -> sink.input (loss rate: 0%)
           9,721,875 sent on repeater.output -> pcap_filter.input (loss rate: 0%)
                 161 sent on source.output -> repeater.input (loss rate: 0%)
ok: accepted 7.4534% of inputs (within tolerance)
selftest: ok
```

That code uses Lua dictionaries with string keys and I don't think that will ever provide such robust performance as ctable (but maybe competitive when tested with few different keys).

eugeneia:
> I don't think this would require much changes to the maintenance workflow that you are already using today. One thing is that you would need to keep application-specific code in program/lwaftr/ (including stuff you just can't be bothered arguing in review about at the moment)

I have to voice strong discomfort here since I need to review every changed file to weed out forgotten debug statements etc. Going through a PRs “changed files” and categorizing them between “do care about” and “do not care about” based on directory prefix is not something I would like to do. Generally I also like the parallel branches model way better than the parallel module prefixes that I think are being proposed here: would it not be an annoyance for the “lwaftr subtree” to be forcefully updated every month? Either we ignore compatibility and lwaftr has to accept “master's” pace or we ensure no breakage and “lwaftr” glues master to the floor.

I feel very good about the workflow described in https://github.com/SnabbCo/snabbswitch/pull/722#issuecomment-179082817 because it gives a very clear model of what I have to care about (a single edge) and what I don't have to care about (a ton of edges). Imho its a good sign when lots of stuff can happen concurrently without synchronization.

Regarding rebasing vs merging. I propose a radical no-rebase policy for public branches, and I consider any branch public that has a PR attached to it. My motivation is as follows: Whenever a PR's branch is rebased I have to read *all of it* again because I have no way of telling what changed. With merges and non-destructive pushes I have a very clear (even supported by git) way of telling whats new since I last looked at it (it will be the difference between my local prXXX branch and the PR's branch). In the case where a rebase can not be avoided at all I recommend closing the original PR, and open a new one with a new branch referencing the old one (this is important, always reference any information/discussion/PRs related in the description as it will save a lot of detective work). 

Another thing I feel very strongly about is to divide and conquer. E.g. this thread is now about roughly four different issues of which three have very little to do with `ctable`. GitHub does not allow for nested comments (which I think is a great thing, nested comments are the enemy of focused discourse) and as such we now have very long thread that doesn't add much in terms of PR review but that I still have to fully read when I want to review this PR. It would be better if every implicit sub thread would be a new issue or PR referencing this one.

The above also applies to code, let me give an example: from my perspective, changes to the Makefile hurt this PR and should be in its own PR. Imagine the Makefile changes break something, and while debugging a month later you find out PR #XXX “add something completely unrelated” was the culprit. That's non intuitive and can be easily avoided by staying on topic. Of course what “on topic” means is relative, but how the documentation is built and the Makefile is imho off topic for a PR about a small well-encapsulated hashable module.

Once PRs become these intertwined balls of loosely related things that yet manage to choke each other become a lot less “attractive”. They become unwieldy and it's more fun to kill them off and start fresh than to merge them, simply because merging a complicated diff is much less satisfying than merging a simple isolated diff. The oversight is lost, I can't reliably account for it any more. Now these emotions might be related to my very limited multitasking capabilities, but I know that by dividing concerns it is possible to reduce a lot of what I like to dub ”brain freeze”, and I think everybody profits from this. This thread is becoming a black hole sucking in precious attention and this elaborate comment is not helping at all, which imho proves my point. This thread has a lot of good discussion and code going but I think every sub discussion and unrelated change would profit from being in its own thread and PR.

wingo:
I don't see how your comment helps, @eugeneia.  I know you're frustrated but I think we're getting somewhere; all discussion has been relevant to the PR at hand.  If you want to review the code, I would be delighted; it's over on the code tab :)

wingo:
I also see how it's easy to get the impression that there was rebasing in this PR, and that would be bad.  But there wasn't!  That was just a confusion.

eugeneia:
As per my elaborate campaign: My recommendation for the documentation for `ctable` would be to just put it into `lib/README.md` and forget about it for now. I don't like to see Makefile changes in this PR, I would prefer if these issues were addresses separately.

Regarding the documentation itself, some things I noticed:

* The headline should not use the monospace tag, e.g. read “Ctable” instead of “`ctable`”
* Introduce concepts such a “a *ctable*” as described in https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/documentation-guide.md#markup-conventions
* Separate implementation details (these should be comments in the source) from user documentation
* Prepend method definitions with the class name. E.g. **ctable:new** instead of **:new**
* Why is `ctable:new` a function? Shouldn't it be a method?
* An “FFI type” is a “ctype” in LuaJIT terminology, be explicit
* “In the future, value_type will be optional; a nil value_type will create a set”, if it is not so it should not be in the documentation
* “[0, 0xFFFFFFFF)” appears to be a typo
* “params” its preferable to spell out words completely
* I would make the configuration keys a single listing and prefix optional keys with *Optional.*
* I think `/not/` is supposed to be `**not**`
* “Method :lookup_ptr”'s description has some formatting issues (generally, proofreading the rendered document will avoid these issues)
* Comma separate function/method arguments
* “Hash functions”: what range? Formatting broken

wingo:
Thanks for the review.  In the future, would it be possible to make specific comments as line notes?  That way I can be sure I am addressing your comments.  I will try to respond to some comments in the source and fix the rest.

src/lib/README.ctable.md
```diff
@@ -0,0 +1,198 @@
...
+name "ctable".
+
+A ctable is parameterized for the specific types for its keys and
+values.  This allows for the table to be stored in an efficient manner.
+Adding an entry to a ctable will copy the value into the table.
+Logically, the table "owns" the value.  Lookup can either return a
+pointer to the value in the table, or copy the value into a
+user-supplied buffer, depending on what is most convenient for the user.
+
+As an implementation detail, the table is stored as an open-addressed
```
wingo: >Separate implementation details (these should be comments in the source) from user documentation
We can certainly remove implementation details from here, but the performance characteristics of this table are user-visible features and it's hard to explain them without describing how the table works.  WDYT?

src/lib/README.ctable.md
```diff
@@ -0,0 +1,198 @@
...
+   key_type = ffi.typeof('uint32_t'),
+   value_type = ffi.typeof('int32_t[6]'),
+   hash_fn = ctable.hash_i32,
+   max_occupancy_rate = 0.4,
+   initial_size = math.ceil(occupancy / 0.4)
+}
+local ctab = ctable.new(params)
+```
+
+— Function **ctable.new** *params*
```
wingo: > Prepend method definitions with the class name. E.g. ctable:new instead of :new

> Why is ctable:new a function? Shouldn't it be a method?

This is a function exported by the module.  It is not a method on an object.  I believe this documentation is correct;  LMK if there are more concerns.

eugeneia:
> I don't see how your comment helps, @eugeneia. I know you're frustrated but I think we're getting somewhere; all discussion has been relevant to the PR at hand. If you want to review the code, I would be delighted; it's over on the code tab :)

I am sorry you feel that way. Why would you think I am frustrated? I was simply replying to the different issues that where touched on. I think you missed my point: yes, the different discussions all do relate to this PR but they do not have to be in a single thread and we can use GitHub's cross referencing to make it more easily digestible.

>  In the future, would it be possible to make specific comments as line notes? 

Generally this is how I prefer it too, but the case of proofreading documentation it tends to involve creating a LOT of commits. Maybe comments on source are cheap and I shouldn't be afraid of the overhead.

src/lib/README.ctable.md
```diff
@@ -0,0 +1,198 @@
...
+following keys are required:
+
+ * `key_type`: An FFI type for keys in this table.
+ * `value_type`: An FFI type for values in this table.  (In the future,
+   `value_type` will be optional; a nil `value_type` will create a
+   set).
+ * `hash_fn`: A function that takes a key and returns a hash value.
+
+Hash values are unsigned 32-bit integers in the range `[0,
+0xFFFFFFFF)`.  That is to say, `0xFFFFFFFF` is the only unsigned 32-bit
```
wingo: This isn't a typo: [] indicate inclusive bounds, () indicate exclusive bounds.  I think it's explained in the next sentence.

lukego:
@eugeneia I see it as a high priority to bring programs onto the master branch as much as possible. This is something that would be done voluntarily by people developing those programs to make their own lives easier. It is our job as upstream to "sell" this concept so that people actually decide to do it.

The reason this is important to me is that I think it is very powerful to have a common code base. This makes it possible to make "global" changes to the software like the "straightline" redesign or rewriting the standard hashtable or ... lots of interesting things to keep the software moving forward. I see lots of this kind of work on the horizon e.g. making programs work with any I/O interface instead of depending on an 82599 NIC, making programs parallelize in consistent ways, and so on. I would much prefer to tackle these problems with all of the relevant code "on the table" rather than tucked away in branches and without upstream test coverage.

Now, if this concept is going to succeed, people will need to actually want to host their software on the master branch. This will mean staying in control of their code and not having to worry about undue interference from gatekeepers (i.e. us) between themselves and their users. This will require establishing definite expectations about who is responsible for what.

I have discussed this privately with @alexandergall in the past in the context of his VPWS/ALX application. The basic model that we came up with was for applications to have their own private directory trees (`program/foo/`) where they have a large degree of autonomy. There they can make their own rules: use camelCase, ALL_CAPS_MODULE_NAMES, comments in Chinese, etc. They only need to avoid negatively impacting the build e.g. blowing up the size of the binary or the compilation time (which could also be solved by creating separate Makefile targets for them).

The expectation is that the code *outside* those directory trees would be shared between all the programs and we would maintain that together in accordance with the evolving community standards.

lukego:
@eugeneia really we are running the major risk of blowing that whole scheme here. Andy is trying to submit some valuable code and I am holding it up by talking about elaborate Git workflows and you are holding it up by cataloguing mistakes in the README. If the downstream branches see engaging with upstream as having net-negative value then people will not bother sending us code and we will become irrelevant.

I think that we need to re-style ourselves as humble servants rather than gatekeepers :-)

src/lib/README.ctable.md
```diff
@@ -0,0 +1,198 @@
...
+ * `min_occupancy_rate`: Minimum ratio of `occupancy/size`.  Removing an
+   entry from an "empty" table will shrink the table.
+
+#### Methods
+
+Users interact with a ctable through methods.  In these method
+descriptions, the object on the left-hand-side of the method invocation
+should be a ctable.
+
+— Method **:resize** *size*
```
wingo: > Prepend method definitions with the class name. E.g. ctable:new instead of :new

So, `new` is a function, but there are many methods like `resize`.  I was a bit worried though that there would be confusion between the module name and the class name, and for that reason left off the LHS on these definitions.  In fact partly for that reason we don't mention the name "class" in these docs, as it doesn't really exist.  What's the right solution here?  Is this acceptable as I've described it above?

eugeneia: Alex has made the precedent for OO style in Snabb and I have accepted/adopted his conventions because I felt they were good. In these terms the module `ctable` is a class, which would have a class method `new` to instantiate `ctable` objects. You can find examples for this in `lib.protocol` and its documentation. (Generally this is how I do it and what I recommend: look for prior examples in existing code and try to avoid breaking conventions where possible.)

wingo: I do the same, but in this particular case I have been really weirded out by the OO conventions of the protocol libraries; I find them distinctly non-snabby.  Compare to `core.packet` for example.  Sometimes I avoid using the protocol libraries just because I find them too weird!

eugeneia: To add to that I just noticed the hash functions of the `ctable` module, from an OO perspective these would probably be class methods as well? There is a precedent in `lib.protocol` where header classes have class and instance methods as well but the distinction is not made in the documentation. I guess this is a FIXME, I am somewhat out of answers.

eugeneia: Yes they are definitely distinct from `core.*` (but I don't think that is necessarily an issue), imho the question is if the style works well for the API. But, we are digressing again. I say do as you see fit for now.

plajjan: I don't think the hash_32 are class methods as they don't expect to be operating on a class.

At least in a different language (like python), new() would be a class method, returning an instance of the ctable class. But the hash functions don't return an instance of a class, just a value, so they are just functions that happen to be residing in this module.

wingo: Just one more note:

> Edit: but please add the prefixes, e.g. “Function ctable.new ... returns a ctable.” and “Method ctable:foo ...”

In `Function ctable.new`, the left hand side `ctable` is the module:

```lua
local ctable = require('lib.ctable')
```

But in “Method ctable:foo ...”, the left-hand-side is just *a* ctable.  This method is not available otherwise!  For example you can't get to this function except as a property of a ctable.  Notably it is not an export of the `ctable' module.  You sure you want to document it in these two ways?

lukego:
This ctable change is looking *extremely* polished to me. I am looking forward to getting it on a feed to `next` :).

wingo:
Okeysmokes, I have addressed the review comments, I think.  Please take a look @eugeneia when you can.  Thanks!

The open question that remains is which feed to `next` to choose or create.  Should I create a branch to do this and convert the conntrack module, @lukego?  I'd also be happy to do that as a followup, if you like.

lukego:
@wingo You have already done so much. Somebody should update the conntrack module but I think we can take this asynchronously and it need not be you if you don't feel inspired to do so.

I'm assuming that this change is targeted to `max-next` and then from there to `next` but that's really up to you guys.

wingo:
Heh, the code isn't all that much, and having a use in-tree would help.  I'll handle the conntrack thing in a followup.  Probably some other things will come first though.

wingo:
What is the status here?

eugeneia:
Merged into `max-next`.

src/lib/README.md
```diff
@@ -0,0 +1,202 @@
+## Specialized data structures
+
+### Ctable (lib.ctable)
+
+A *ctable* is a hash table whose keys and values are instances of FFI
+data types.  In Lua parlance, an FFI value is a "cdata" value, hence the
```
eugeneia: Since there are some FIXMEs in the code I assume this will be iterated upon. I will leave some notes of things I thought about while going through this PR:

Most of the introduction (from “In Lua parlance...” to line 35) are implementation details. I would avoid talking about them in the documentation. The reader most likely only cares about how to use it, not how it works (the how can be source code comments). This case is especially questionable because implementation details are mixed with usage information, so you have to read the whole lot even if you are only looking for either one.

src/lib/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+Create a new ctable.  *parameters* is a table of key/value pairs.  The
+following keys are required:
+
+ * `key_type`: An FFI type (LuaJIT "ctype") for keys in this table.
+ * `value_type`: An FFI type (LuaJT "ctype") for values in this table.
+ * `hash_fn`: A function that takes a key and returns a hash value.
+
+Hash values are unsigned 32-bit integers in the range `[0,
+0xFFFFFFFF)`.  That is to say, `0xFFFFFFFF` is the only unsigned 32-bit
```
eugeneia: Do we need a “Notation” chapter in the documentation that lists and explains the notations used throughout the docs? While the `[...,...)` notation might be obvious to 99% of readers, we still need to either avoid it or document it. There has to be a cost/value calculation for each special notation, e.g. how often is it useful vs how many notations does the reader have to know about to read the docs.

src/lib/README.md
```diff
@@ -0,0 +1,202 @@
...
+ * `min_occupancy_rate`: Minimum ratio of `occupancy/size`.  Removing an
+   entry from an "empty" table will shrink the table.
+
+#### Methods
+
+Users interact with a ctable through methods.  In these method
+descriptions, the object on the left-hand-side of the method invocation
+should be a ctable.
+
+— Method **:resize** *size*
```
eugeneia: I still think it should say **ctable:resize**. It even says “resize the ctable” in the description. If its an object of type ctable then the method should be `ctable:resize`.

src/lib/ctable.lua
```diff
@@ -0,0 +1,466 @@
...
+local required_params = set('key_type', 'value_type', 'hash_fn')
+local optional_params = {
+   initial_size = 8,
+   max_occupancy_rate = 0.9,
+   min_occupancy_rate = 0.0
+}
+
+function new(params)
+   local ctab = {}   
+   local params = parse_params(params, required_params, optional_params)
```
eugeneia: I think `parse_params` and `set` are over engineered. Current Snabb idiom would suggest this way more obvious (and less) code:

```
assert(params.required1, "error blah")
...
params.optional1 = params.optional1 or "default-value"
...
```

wingo: I would do this, but it has three drawbacks:
 * It doesn't allow for boolean values that default to true.
 * It modifies the incoming parameters.
 * It doesn't give an error if the user mis-types a parameter name.

The last one is the most important to me.

eugeneia: I can relate to the last point, might be worth to spin this out into `core.lib` in the future then. We already have some common code for app configurations which then should be unified with this.

lukego:
Glad to have this change land on next. Note that there are infinitely many opportunities for everybody to further improve this code and documentation with follow-on PRs.

src/lib/README.md
plajjan: I quite like the introduction, including both *use* and *how*, but maybe I'm more interested in the details of the *how* than the average user.

eugeneia: My concern is that non-normative sections in the documentation can lead to other relying on a specific behavior that is not actually part of the API. But I might be wrong, feedback in #753 would be appreciated,

PR #723, opened 2016-01-20
core.memory: get the page size from sysconf

nnikolaev-virtualopensystems:
The page size is hardcoded to 4K, which is a problem on systems which
define different value for it. Use sysconf(_SC_PAGESIZE) to determine
the actual page size.

base_page_size is removed from memory.lua since is not used.

### DIFF
--- a/src/core/memory.c
+++ b/src/core/memory.c
@@ -27,13 +27,24 @@
 
 #include "memory.h"
 
+static long page_size = 0;
+
+// global initalizer before running main()
+// GCC specific:
+// https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html
+void __attribute__((constructor, used)) initializer(void)
+{
+  // get the system page size
+  page_size = sysconf(_SC_PAGESIZE);
+}
+
 // Convert from virtual addresses in our own process address space to
 // physical addresses in the RAM chips.
 uint64_t virtual_to_physical(void *ptr)
 {
   uint64_t virt_page;
   static int pagemap_fd;
-  virt_page = ((uint64_t)ptr) / 4096;
+  virt_page = ((uint64_t)ptr) / page_size;
   if (pagemap_fd == 0) {
     if ((pagemap_fd = open("/proc/self/pagemap", O_RDONLY)) <= 0) {
       perror("open pagemap");
@@ -51,7 +62,7 @@ uint64_t virtual_to_physical(void *ptr)
     fprintf(stderr, "page %lx not present: %lx", virt_page, data);
     return 0;
   }
-  return (data & ((1ULL << 55) - 1)) * 4096;
+  return (data & ((1ULL << 55) - 1)) * page_size;
 }
 
 // Map a new HugeTLB page to an appropriate virtual address.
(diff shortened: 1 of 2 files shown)


lukego:
What systems are these? See also #712.

nnikolaev-virtualopensystems:
The system in question is AArch64 compiled with CONFIG_ARM64_64K_PAGES.
But this is also valid for PowerPC which can have 16K, 64K and even 256K page size.

This patch does not support those platforms, it just uses a more portable. POSIX compliant and non-hardcoded way to determine the page size value.

lukego:
The `master` branch does not accept this kind of change. There need to be separate `posix`, `arm`, `ppc` branches where somebody is maintaining ports. If ports work well and have users then we can plan for merging them onto master.

There could also be a branch for collecting general portability patches but again this is not `master`.

More on #712.

kbara:
As far as I can tell, this isn't entirely irrelevant even on x86_64: https://en.wikipedia.org/wiki/X86-64#Page_table_structure

I think this patch is worth merging, for what it might be worth.

nnikolaev-virtualopensystems:
I don't think that replacing `4096` with `sysconf(_SC_PAGESIZE)` is architecture specific.
It's just a more generic way to determine the value.

It is also a better programming style than having arbitrary and unexplained numbers all over the code.

I understand #712, but this is not porting, although the patch makes this particular part of the code a little bit more portable. If this is not suitable for `master` then so be it.

lukego:
I agree that the magic number is bad and it would be better to avoid that.

I am not wild about using a GCC extension to do this though. I want to rewrite as much of the C code we have as possible in Lua based on ljsyscall and adding new C code that is potentially hard to rewrite seems like a step in the opposite direction.

lukego:
@nnikolaev-virtualopensystems Thank you for your understanding :). I hope that we will have a portability branch in the future for collecting these changes (and that once ports are working well we could also merge them onto master).

In the short term I'd like to come up with a tagging scheme on Github to make these portability fixes easy to find for the day when somebody decides to collect and integrate them on a branch. There have been about half a dozen such changes submitted to master. Working this out on #725.

lukego:
> I think this patch is worth merging, for what it might be worth.

@kbara My beef about merging this as an improvement for x86-64 is: shouldn't we rewrite this C code in Lua? I want to use C very conservatively so that Snabb hackers need only basic C skills. So I am not enthusiastic with replacing a magic number (okay, lame) with a new GCC trick that I don't really understand using an advanced compiler feature that I have never seen before.

lukego:
@kbara Historical context is that most of the C code in Snabb Switch should never have been written: we should have written it in Lua with ljsyscall. Unfortunately, we didn't use ljsyscall back in the beginning so every function that made non-trivial use of system calls or the standard C library had to be written in C and called via FFI. This C code then has a natural tendency to spread e.g. because an extension is needed and it is inconvenient to do in Lua because of the FFI barrier so we add it to C instead.

I really want to weed out this code. I don't feel that there is a place in Snabb Switch for C code that could have been written in Lua. I dislike the way the smattering of C code that we have is consuming brain cycles on talking about POSIX compliance and so on. I'd much prefer to work with tools that don't require such a level of language-lawyering to use effectively.

eugeneia:
I will close this as it doesn't seem to be going anywhere. Please reference this PR in case an ljsyscall based alternative is found.

PR #726, opened 2016-01-22
engine: Detect bad return value from app start() method

lukego:
Raise a specific error instead of allowing a bad value to propagate in the engine (leading to a more confusing error message).


eugeneia:
Merged into `max-next`.

PR #727, opened 2016-01-23
packetblaster synth: Generate synthesized packets

aequabit:
Tests are missing, but it works as it should. 

Hope you're happy now, @lukego!


lukego:
Great feature!

Often we have wanted to test with a different packet size, or specific mixes of packet sizes (e.g. to have variation but maintain a definite average size), and it is really primitive to do this by creating new pcap files all the time. Much better to be able to do it on the command line with a series of packet sizes.

The Synth `selftest` method looks like it may depend on the app outputting exactly the same number of packets when executed for a very short time period? That seems like it could break in one of two ways: first that we could change the size of `struct link` so that it would output a different number of packets per breath, and second that the engine might run the app for multiple breaths. Can we make this more robust somehow? (Just to avoid confusing the next person who changes the size of a link or the engine event loop behavior and sees this test start failing.)

eugeneia:
@lukego Yeah we are aware of this, but didn't come up with an elegant solution yet, I think the `Match` app from #708 could solve this.

eugeneia:
@lukego What do you think, is the flaky `Synth:selftest` a deal breaker? If not I would merge this into `max-next` and fix the selftest when `Match` is ready.

petebristow:
Is synth envisaged to be a general purpose packet crafter? 
So far I've been using scapy to generate pcaps on the fly and /dev/stdout and /dev/stdin as hacks to run it as a single command. That is then combined with `PcapReader` and `Repeater`

```
'wrpcap("/dev/stdout", [ Ether()/Dot1Q(vlan=x)/IP(src="127.0.0.1")/UDP(sport=10)/Raw(open("/dev/urandom","rb").read(10)) for x in range(5,70) ])' | scapy | snabb snsh vlan.lua
```
lib.protocol doesn't really have the brevity of scapy.
Would you see synth eventually replacing this use case?
If it doesn't what's the line of reasoning to not include this in Source?
Is the use case for synth particularly wide spread as it currently stands?
Does having Match join Synth under test make sense?
`Match` doesn't help where tests haven't run for long enough but would help where more than 1 breath is taken.

eugeneia:
> Is synth envisaged to be a general purpose packet crafter? 
>  Would you see synth eventually replacing this [scapy] use case?

`Synth` was born mainly with this PR in mind, e.g. to support a synthesizing `packetblaster`. It could certainly be extended, or replaced by a more sophisticated app for that matter.

>  If it doesn't what's the line of reasoning to not include this in Source?

`Synth` is Ethernet aware so I felt it was too specific for `basic_apps`.

>  Is the use case for synth particularly wide spread as it currently stands?

I don't think so / I don't know.

>  Does having Match join Synth under test make sense?

I thought the same thing. I think so, yes.

> `Match` doesn't help where tests haven't run for long enough but would help where more than 1 breath is taken.

Why is that? That seems like an arbitrary limitation. I imagined this would be a perfect case for “fuzzy” matching, e.g. ensure at least one packet of each specified size is sent.

lukego:
@eugeneia I would prefer to see `max-next` passing tests so that I can merge the changes that are already there. I'm not sure what the optimal workflow is for releases but I feel like I should be merging `max-next` onto `next` at least once per week or so. Then we would be using `next` as a place to land changes as quickly as they can be reviewed by a subsystem maintainer and also by me and other people could test them there.

lukego:
@eugeneia oh it is passing :-) Cool! I merged `max-next` into `next` now. Sure, merging this onto `max-next` seems okay to me, though it seems likely we will forget about fixing the tests if we don't do it now.

eugeneia:
@lukego I will merge it then and take personal responsibility for remembering to fix the test. :-)

PR #731, opened 2016-02-02
New Intel X520 2-port Adapter detection

mwiget:
$ lspci -nn|grep X520
03:00.0 Ethernet controller [0200]: Intel Corporation Ethernet 10G 2P X520 Adapter [8086:154d] (rev 01)
03:00.1 Ethernet controller [0200]: Intel Corporation Ethernet 10G 2P X520 Adapter [8086:154d] (rev 01)


eugeneia:
@mwiget I assume we indeed support this model, e.g. the test suite passes for these NICs?

mwiget:
Tried to run selftest via SFP+ fibre loopback cable, but the card had a hard time bringing the link up, even when managed by the kernel/ethtool. No twinax cable at hand. But I got more info's about the card, in case someone else has one to give the test a shot: Dell X520-DP 10GB DA/SFP+[http://accessories.euro.dell.com/sna/productdetail.aspx?c=uk&l=en&s=dhs&cs=ukdhs1&sku=540-11141] 

Despite not having been able to run it in loopback mode, we successfully used both ports heavily during a full week of test (lwaftr), connected via fibre to a router.

eugeneia:
@lukego I would merge this since the card does work for lwaftr but I feel like we need to track that we need to document that the card is not supported by the CI, maybe a source comment would be enough. What do you think?

mwiget:
No rush merging. I'm sure I'll be back in Greece  soon, carrying a twin ax loop back cable to retest.

Marcel

Sent from Outlook Mobile<https://aka.ms/qtex0l>




On Tue, Feb 9, 2016 at 8:46 AM -0800, "Max Rottenkolber" <notifications@github.com<mailto:notifications@github.com>> wrote:


@lukego<https://github.com/lukego> I would merge this since the card does work for lwaftr but I feel like we need to track that we need to document that the card is not supported by the CI, maybe a source comment would be enough. What do you think?

-

lukego:
I am not 100% sure but I think this is just another OEM packaging of the same ethernet controller that we are testing. I would be comfortable merging the patch and considering the card to be supported.

If we can establish that it really is significantly different to the NICs we already have then I would be happy to acquire one for the CI.

eugeneia:
I'll merge this then, a bug should be filed when problems with this card become apparent.

PR #736, opened 2016-02-02
Implement l2fwd-like app

dpino:
Basic implementation of l2fwd app (like DPDK's). This is how I understand it should work:

Given the following conf:
- 01:00.0 <-> 02:00.0
- 01:00.1 <-> 02:00.1

A packetblaster sends packets into 01:00.0.
l2fwd forwards packets received on 02:00.0 to 01:00.1.
- The app reports Gbps of link.
  Alternatively, a separate program could measure packets received on 02:00.1.


lukego:
Neat :).

I appreciate that you have made a simple solution for choosing between different I/O sources (tap, virtio, 10G) in `select_nic_driver()`. This is a nice baby-step towards solving that problem more generally in the future i.e. making Snabb programs more flexible about their I/O sources.

Aside: I would really like to have a variant of our standard CI benchmark that uses this `l2fwd` instead of the DPDK one. Then we could make sure that we have the same performance with both guest Virtio-net drivers. @eugeneia would be the person to talk with about how to set this up.

src/program/l2fwd/l2fwd.lua
```diff
@@ -0,0 +1,84 @@
...
+function L2Fwd.new(conf)
+   local o = {}
+   return setmetatable(o, { __index = L2Fwd })
+end
+
+function L2Fwd:push()
+   local input, output = assert(self.input.input), assert(self.output.output)
+
+   while not link.empty(input) do
+      link.transmit(output, link.receive(input))
```
nnikolaev-virtualopensystems: The DPDK L2FWD app actually does some processingon each packet. See here: http://dpdk.org/browse/dpdk/tree/examples/l2fwd/main.c#n239

The main idea is that if I send traffic with a destination MAC A and I don't change it in the L2FWD, then all possible switches between the sender and the L2FWD will get messed. Like if we use a VMDq port then the integrated switch will learn the MAC A on a ceratin port and won't forward the returned packet to the external port.

So what DPDK does is to replace each packet's destination MAC address with `02:00:00:00:00:xx`, where `xx` is the number for the port it came from (DPDK actually assigns numbers to each port it recognizes). And then also changes the source with the MAC of the port it send it through. So it is more or less a real forwarding.

But SnabbSwitch is in an even better position since it actually has a real switching app - the learning bridge. I guess it is the most suitable in this situation.

src/program/l2fwd/l2fwd.lua
```diff
@@ -0,0 +1,84 @@
...
+
+function run(args)
+   local opts, pciaddr1, pciaddr2 = parse_args(args)
+   local c = config.new()
+
+   local driver1 = select_nic_driver(pciaddr1)
+   local driver2 = select_nic_driver(pciaddr2)
+
+   config.app(c, "nic1", driver1, {pciaddr = pciaddr1})
+   config.app(c, "nic2", driver2, {pciaddr = pciaddr2})
```
nnikolaev-virtualopensystems: There DPDK L2FWD actually accepts and an arbitrary number of ports. There is a nice explanation of how it works here: http://dpdk.org/doc/guides/sample_app_ug/l2_forward_real_virtual.html 

Of course, this one is not exactly matching the learning bridge that I mentioned before.
My suggestion is that we decide which one is better - to expose the learning bridge through this app, or to mimic the DPDK's L2FWD behavior.

nnikolaev-virtualopensystems: And I vote for the learning bridge :)

lukego: Question is what is the high-level intention of this program? If the idea is to have absolute minimum overhead for benchmarking purposes then learning bridge could be counter productive.

wingo: One use case for this program is for when you ship a program to a customer and they report that nothing is working :)  In that case it's useful to have a simple snabb application they can try that doesn't depend on your application code, and which they can use to benchmark performance as well.

I think we should steal ideas from the dpdk but we don't necessarily need to match the particular feature set of any given application.  In this case I think a minimal counterpart to packetblaster is a good feature set.

nnikolaev-virtualopensystems: Still you need to, at least, swap source/destination MAC addresses. I personally very often have to do some magic tricks when other switching components are involved in the datapath because they "learn" MAC addresses from the traffic that I sent, and then there is no way that I can get my traffic back.

Also, I would like to be able to do the loopback on a single interface, like: `config.link(c, "nic1.tx -> nic1.rx")`
As you mention - this is the simplest way that the customer can test - 1 port reflecting the packets back.

eugeneia:
@dpino Without having looked too closely, I am absolutely in *love* with the the idea that we could use this as an alternative to DPDK's `l2fwd`. It wouldn't replace DPDK as we need that for interop testing but it could make benchmarking much simpler. :+1:

src/program/l2fwd/l2fwd.lua
eugeneia: @dpino I want to merge this, do you want to amend anything to this before I do? E.g. swapping Ethernet source and destination would be a low hanging fruit. It would make sense to test if this `l2fwd` works in the context of `packetblaster->snabbnfv<->vm(snabb+l2fwd)`

src/program/l2fwd/l2fwd.lua
```diff
@@ -0,0 +1,84 @@
...
+
+   config.link(c, "nic1.tx -> nic2.rx")
```
nnikolaev-virtualopensystems: Could we get also `config.link(c, "nic2.tx -> nic1.rx")` here.

eugeneia: @nnikolaev-virtualopensystems Am I right in assuming that would make l2fwd a two-way street?

nnikolaev-virtualopensystems: Exactly, this makes it full featured 2-way "forwarder".

dpino:
@eugenia Sorry, I was a bit away lately. I will tackle these issues in the coming days and verify it works in the context of `packetblaster->snabbnfv<->vm(snabb+l2fwd)`. Thanks everyone for the review and comments.

eugeneia:
@dpino Hey, no rush at all. I am super supportive of taking it easy.

dpino:
I think this is working now. Here is how it works:

Let's assume the following scheme:

Host1          Host 2
01:00.0 <-> 01:00.0
01:00.1 <-> 01:00.1

L2Fwd creates a full-duplex softwire in Host2 between 01:00.0 and 01:00.1.

Packetblaster blasts packets into Host 1 01:00.0 which eventually reach Host 2 01:00.0, because both cards are wired together. The packets are forwarded to Host 2 01:00.1 and transmitted. Finally the packets arrive to Host1 01:00.1

To make packets come back the other way, it's necessary to create a packet bouncer in Host 1 01:00.1. I added a new packetblaster mode called 'bounce' which takes only two PCI params, the NIC where to blast packets to and the NIC of the bouncer, which sends back any traffic it receives.

With regard to mimic the MAC addresses scheme DPDK's l2fwd app uses, I didn't follow it as packets get through anyway. I think what's interesting of this app is to have a tool which can be a counterpart of packetblaster and can help verify NIC links are working.

Example of use:

**Host 1**

```
$ sudo ./snabb packetblaster bounce my.pcap 0000:02:00.0 0000:03:00.0
Transmissions and receptions (last 1 sec):
02:00.0	TXDGPC (TX packets)	433,968	GOTCL (TX octets)	240,418,272
02:00.0	RXNFGPC (RX packets)	7,205	GORCL (RX octets)	3,991,570
Transmissions and receptions (last 1 sec):
02:00.0	TXDGPC (TX packets)	1,848,822	GOTCL (TX octets)	1,024,116,090
02:00.0	RXNFGPC (RX packets)	1,573,190	GORCL (RX octets)	871,551,692
Transmissions and receptions (last 1 sec):
02:00.0	TXDGPC (TX packets)	2,177,641	GOTCL (TX octets)	1,206,404,804
02:00.0	RXNFGPC (RX packets)	2,177,631	GORCL (RX octets)	1,206,408,128
```

**Host 2**

```
$ sudo ./snabb l2fwd -v 0000:02:00.0 0000:03:00.0
Report (last 1 sec):
link report:
              95,052 sent on nic1.tx -> nic2.rx (loss rate: 0%)
               2,944 sent on nic2.tx -> nic1.rx (loss rate: 0%)
load: time: 0.98s  fps: 100,419   fpGbps: 0.449 fpb: 6   bpp: 550  sleep: 35  us
Report (last 1 sec):
link report:
             942,990 sent on nic1.tx -> nic2.rx (loss rate: 0%)
             225,032 sent on nic2.tx -> nic1.rx (loss rate: 0%)
load: time: 1.00s  fps: 1,070,037 fpGbps: 4.785 fpb: 12  bpp: 550  sleep: 0   us
Report (last 1 sec):
link report:
           3,120,653 sent on nic1.tx -> nic2.rx (loss rate: 0%)
           2,402,703 sent on nic2.tx -> nic1.rx (loss rate: 0%)
load: time: 1.00s  fps: 4,355,314 fpGbps: 19.477 fpb: 22  bpp: 550  sleep: 0   us
Report (last 1 sec):
```

I also tested the l2fwd app running within a guest using the virtionet driver. It works fine too.

dpino:
Apparently, I don't see any reason why CI failed cc @eugeneia

eugeneia:
Non-blocking questions:

1. I think I previously missed the fact that l2fwd operates on two duplex ports. Unless I am missing something l2fwd is still fundamentally incompatible with the  `packetblaster->snabbnfv<->vm(snabb+l2fwd)` scenario since l2fwd needs two duplex ports but the VM only has one (`vhost_user`)? This is not a problem with this PR, just want to clear up my understanding.
2. To be honest I do not think I really understand the motivation behind l2fwd/bounce. I see no other use case for bounce than in conjunction with l2fwd, and to me it seems the scope of l2fwd is very limited: to test/benchmark two duplex ports. Why two? Why not test each NIC individually? Why are bounce and l2fwd separate programs when they only work together (my assumption)?

src/apps/intel/loadgen.lua
```diff
@@ -14,9 +14,10 @@ local can_transmit, transmit
 
 LoadGen = {}
 
-function LoadGen:new (pciaddress)
-   local o = { pciaddress = pciaddress,
-               dev = intel10g.new_sf({pciaddr=pciaddress}) }
+function LoadGen:new (conf)
+   local o = { pciaddr= conf.pciaddr,
+               dev = intel10g.new_sf({pciaddr=conf.pciaddr}),
+               report_rx = conf.report_rx, }
```
eugeneia: What exactly do the changes to `LoadGen` do? Does the `LoadGen` documentation need an update to reflect `conf.report_rx`?

dpino: I modified LoadGen so it reports on received packets as well as transmitted packets.  As traffic gets bounced to its originator and the NIC is locked by the program running the packetblaster, this was the only possible way I thought of of checking traffic gets actually bounced back.

I missed updating LoadGen docs, if the change finally makes it I will update the docs.

dpino:
@eugeneia My understanding after reading http://dpdk.org/doc/guides/sample_app_ug/l2_forward_real_virtual.html is:

* There's a traffic originator running in one side (host A).
* Packets reach a destination host (host B) and get forwarded between two ports on the same host. That's what the l2fwd app does AFAIU.
* Packets finally reach host A on a different port and go back the same way.

Host A actually needs two packet generators, an app that creates synthetic packets as the packetblaster does and something that will bounce traffic.  Checking whether packets are being received in the packetblaster originator app tells traffic has gone all the way through the links.

With regard to the packetblaster-bounce/l2fwd tandem, I agree with your comment.  The point is that as the traffic generators run in a host independent of the host that has created the port forwarder, l2fwd/bounce cannot be on the same app.

OTOH, packetblaster-bounce (NIC1 <-> NIC2) can be used to measure throughput of NIC2, received packets and egress link load towards NIC1.  Currently it doesn't print those reports but that would be easy to fix.

cc @lukego @wingo @nnikolaev-virtualopensystems

eugeneia:
I would like other opinions on this as well. I have a very different picture of what `l2fwd` should do (e.g. run on a single NIC, receive on rx, swap ethernet src and dst, and send them back on tx), and this would imho simplify it while obsoleting bounce altogether. Note that this view is very centered on my own contact with DPDK's l2fwd and might be completely uneducated/ignorant.

dpino:
@eugenia Totally understand, I also want to hear other opinions.  With regard to MAC addresses and swapping them, I forgot to mention that I obvious that part because it's not necessary. NICs when instantiated as SingleFunction, which is the case of packetblaster, work in promiscuous mode.

nnikolaev-virtualopensystems:
As explained before once you have a HW switch between the l2fwd and the packet blaster then you need MAC swapping. I understand that this is not your use case, but that is what a forwarder should do - otherwise it is just a bridge, or softwire or something else but not L2FWD IMHO.

eugeneia:
@dpino Would the model I described cover your use case as well?

dpino:
@eugenia I'm not an user of DPDK's l2fwd app so it's hard for me to tell how much of this app can replace it. If the use case you described is enough for ditching DPDK's l2fwd in favour of this app I can implement it that way. I see your description more like the bouncer mode I added to packetblaster, but I can put that functionality in its own app. It will be handy too for me to test a VirtioNet link. I think I will create a brand new PR.

eugeneia:
@dpino We use `l2fwd` to test a duplex port, with the slight ugliness that our “softwire“ (Snabb NFV) did the benchmarking. What if `packetblaster` was extended to measure incoming bandwidth, and `l2fwd` behaved like in DPDK? Then we could set up the following topology:

```
packetblaster<->(virtio|NIC|snabbnfv)<->l2wfd)
```

Then we could benchmark/test any duplex “softwire” with `packetblaster`, given `l2fwd` sits on the other end. We would have two orthogonal parts: a benchmarking tool `packetblaster` and a simple forwarder `l2fwd` that would hopefully be generic enough to serve a bulk of benchmarking scenarios.

dpino:
@eugeneia I'm retaking this task. Your suggestion sounded good to me, however there's something I don't understand yet and it's how DPDK's l2fwd is currently being used.

AFAIU, DPDK is only used in the packetblaster benchmark test. The test uses `qemu-dpdk.img` VM which has DPDK on it.

For testing purposes, I replaced qemu-dpdk.img in the packetblaster benchmark for `qemu.img`. The benchmark works but it's much slower. I was thinking about why it was working and my conclusion is that when SnabbNFV runs on its most basic form all the traffic that is received on the VM is sent back to the host because the configuration goes like this:

```lua
      config.link(c, NIC..".tx -> "..VM_rx)
      config.link(c, VM_tx.." -> "..NIC..".rx")
```

The packets are received on the NIC running packetblaster as both cards are wired together. Is this the reason why the packets in a VM are sent back to the host and traffic can be benchmarked?

As for the qemu-dpdk.img scenario I understand the benchmark goes faster as DPDK skips the VM kernel, what I don't see is how DPDK or DPDK's l2fwd is run or who starts it :/ I grepped for "l2fwd" in the code base and I didn't find any results. OTOH, I see the source code of DPDK is in the /root folder, although I'm not sure if it's built as the `build` folder is empty.

eugeneia:
`qemu-dpdk.img` contains the following `/etc/rc.local`:

```
#!/bin/sh
mount -t hugetlbfs nodev /hugetlbfs
echo 64 > /sys/devices/system/node/node0/hugepages/hugepages-2048kB/nr_hugepages
modprobe uio
insmod /root/dpdk/x86_64-native-linuxapp-gcc/kmod/igb_uio.ko
/root/dpdk/tools/dpdk_nic_bind.py --bind=igb_uio 00:03.0
screen -d -m /root/dpdk/examples/l2fwd/x86_64-native-linuxapp-gcc/l2fwd -c 0x1 -n1 -- -p 0x1
exit 0
```

I can't really tell you what the cryptic l2fwd invocation means though.^^

dpino:
@eugeneia Thanks, I was missing that part :)

wingo:
It would be really great to have this app :)  Right now we are landing changes to the virtio-net interface without the benefit of CI or benchmarking, at least in core; sub optimal!

PR #694, opened 2016-01-04
README.md: Fix and update packetblaster example

lukego:
Fix: had missed the 'replay' argument.

Update: use shorter canonical PCI device name (without leading 0000:).

PR #737, opened 2016-02-04
Globally use '#!/usr/bin/env <shell>' hashbang line

lukego:
Convert all shell scripts from having path dependencies like:

    #!/bin/bash
to

    #!/usr/bin/env bash

The immediate utility of this change is to fix annoying breakage on NixOS. For example, now `make doc/snabbswitch.html` works on lab servers.

Exception is that I did not descend into subdirectories of dependencies e.g. ljsyscall and pflua.

I made the changes with an Emacs macro in the hope of avoiding an unfortunate typo that will break something.

### DIFF
--- a/src/apps/solarflare/selftest.sh
+++ b/src/apps/solarflare/selftest.sh
@@ -1,3 +1,3 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 ./snabb snabbmark solarflare 10e6 128 10
--- a/src/apps/tap/selftest.sh
+++ b/src/apps/tap/selftest.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 sudo ip netns add snabbtest || exit $TEST_SKIPPED
 sudo ip netns exec snabbtest ip link add name snabbtest type bridge
 sudo ip netns exec snabbtest ip link set up dev snabbtest
--- a/src/bench/basic1-100e6
+++ b/src/bench/basic1-100e6
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 out=$(./snabb snabbmark basic1 100e6)
 # Extract floating point Mpps number from output.
--- a/src/bench/packetblaster-64
+++ b/src/bench/packetblaster-64
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 [ ! -z "$SNABB_PCI_INTEL0" ] || exit 1
--- a/src/bench/packetblaster-synth-64
+++ b/src/bench/packetblaster-synth-64
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e 
 
 [ ! -z "$SNABB_PCI_INTEL0" ] || exit 1
--- a/src/bench/snabbnfv-iperf-1500
+++ b/src/bench/snabbnfv-iperf-1500
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 out=$(program/snabbnfv/selftest.sh bench)
 # Extract floating point Gbits number from output.
--- a/src/bench/snabbnfv-iperf-jumbo
+++ b/src/bench/snabbnfv-iperf-jumbo
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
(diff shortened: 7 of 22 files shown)


lukego:
@domenkozar does this look sensible to you?

wingo:
Annoyingly, this fix doesn't work on GuixSD, where they don't have /usr at all, not even for /usr/bin/env.  They do have /bin/sh as the only thing in /bin though; I don't know why they make that exception.  Anyway this fix LGTM.

lukego:
Aside: I wonder if there is any toolkit for writing "shell scripts" directly in Lua. This could conceivably make the code more homogeneous and avoid obscure dependencies.

I see a major problem with both C and shell that using them "professionally" requires having a sixth sense for when you might be using something that is non-POSIX or otherwise taboo. I much prefer to use tools that are more WYSIWYG. These days using ljsyscall seems *vastly* preferable to writing system code in C... is there a candidate for something vastly preferable to shell too?

lukego:
@wingo @eugeneia Can you tell me which of you two is the "next hop" who I am negotiating with to merge this PR? (I understand that everybody in the community is welcome to review this but I hope that one person will be assigned to merge it when they determine that it is ready according to their own criteria.)

lukego:
SnabbBot is flagged a CI error but this looks spurious. It reports a performance regression on the NFV iperf benchmark that I can't account for based on the content of this PR.

wingo:
Hah, I wasn't thinking, sorry :)  Will take some time to get my merging hat on.  Luke does the PR policy apply to you too, or do you also have a next branch?  I am happy to take this one as a first merge.

lukego:
Here is what I would ideally like as a downstream contributor:

1. I can create a feature branch based on master and send a Pull Request to SnabbCo:master branch.
2. Responsible "upstream" maintainer is identified i.e. the person who will merge this (or not) onto a branch that feeds to master.
3. The maintainer guides me through the review process e.g. tells me what I need to fix now, tells me what I should do differently next time, tells me if feedback from some subject-matter expert is needed, etc.
4. The maintainer merges the change onto their branch. Now from my perspective the rest of the upstreaming flow is "somebody else's problem" at least to a first approximation. (I still may want to watch the progress, see what edits are made upstream, be available to help with problems e.g. that might cause the change to be reverted, etc).

I would hope that any confusion and misunderstandings in this process will at least contribute towards making things go smoother the next time i.e. I am developing a useful functioning relationship with my upstream maintainer(s).

And yes please I would like to send my changes through the same upstreaming process as everybody else. I see this as a value-adding process and that skipping it by merging directly onto `next` I would be missing out on that.

domenkozar:
If `/bin/sh` also works on guix, we could use that. But unfortunately that will break if we use bash features and folks have a different shell set.

lukego:
> If /bin/sh also works on guix, we could use that. But unfortunately that will break if we use bash features and folks have a different shell set.

Thanks for clarifying. This would be a bigger change i.e. porting from bash to vanilla bourne shell and I am not motivated to do that on this branch.

wingo:
OK, to be clear in the future for this commit I should have said "LGTM, I will merge onto wingo-next."

To reply to @domenkozar, having to think about whether a script is bash or portable bourne shell is a bigger change I wouldn't feel comfortable doing.  (fwiw /bin/sh is bash on guix but that is not relevant.)  I think I was just griping, not adding value to the discussion :)

lukego:
> OK, to be clear in the future for this commit I should have said "LGTM, I will merge onto wingo-next."

Github can also help with this process btw. Here is how I try to involve it when accepting changes:

1. Make a non-fast-forward merge (`git merge --no-ff origin/pr/nnn`) into my branch with a commit message summary line of the form `Merged PR #nnn (very short summary) into my-branch`.
2. Push the change to my branch e.g. `lukego/next`.

Then Github will automatically add a cross-reference to this PR page that links to the merge commit so that I can see whether the change has landed. See for example the note on (random example) #694.

Could also make sense to close PR branches once they have been merged onto their upstream. I am not sure. My habit has been to leave PRs open until they land on master so that when you list the PRs on Github you see the code that has not landed yet (and is perhaps extra useful to review). That practice is perhaps a crutch that I used because without changes passing through multiple maintainers I was not always sure that I am doing the right thing by landing them on `next`.

wingo:
Thanks for the note, I would have done that except that I was in the middle of another change.

lukego:
@wingo Roger that. Thanks for the info! I am a happy downstream contributor :).

lukego:
Closing this PR because I see that it has landed upstream on `wingo-next`. Thanks @wingo!

PR #739, opened 2016-02-04
Simplify core.memory API

lukego:
Streamline the core.memory API:

- Rewrite the blurb.
- Keep dma_alloc() and virtual_to_physical().
- Remove dma_min_addr and dma_max_addr (unused).
- Deprecate huge_page_size (used but should it be?)

Backstory here is that the core.memory module has exported low-level information to other modules: the VFIO module (since removed) that needed to whitelist all DMA memory with the kernel, a /dev/vhost-net accelerator for tap devices (since removed) that also needed to whitelist all packet memory, and to a lesser extent the vhost-user implementation (that is still alive and well but may be overusing the memory API e.g. in dead code that should be removed).

Adding the deprecation note on "huge_page_size" may or may not be a good idea. Have to look more closely at the usage and clarify with a later commit.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -333,48 +333,23 @@ Allocate packet and fill it with the contents of *string*.
 
 ## Memory (core.memory)
 
-Snabb Switch does two things specially when it comes to memory: It
-runs with a fixed physical memory map and it allocates *huge pages*
-from the operating system.
+Snabb Switch allocates special
+[DMA](https://en.wikipedia.org/wiki/Direct_memory_access) memory that
+can be accessed directly by network cards. The important
+characteristic of DMA memory is being located in contiguous physical
+memory at a stable address.
 
-Running with a fixed memory map means that every virtual address in the
-Snabb Switch process has a fixed *physical address* in the RAM
-chips. This means that we are always able to convert from a virtual
-address in our process to a physical address that other hardware (for
-example, a network card) can use for DMA.
+— Function **memory.dma_alloc** *bytes*
 
-Huge pages (also known as *HugeTLB pages*) are how we allocate large
-amounts of contiguous memory, typically 2MB at a time. Hardware devices
-sometimes require this, for example a network card's *descriptor ring*
-may require a list of pointers to available buffers in physical memory.
+Returns a pointer to *bytes* of new DMA memory.
 
-— Variable **memory.chunks**
+— Function **memory.virtual_to_physical** *pointer*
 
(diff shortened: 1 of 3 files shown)


lukego:
Too bad the filename extension is `.md.src` instead of `.md`. This seems to suppress Github's quite awesome capability to [show diffs in the rendered versions of markdown files](https://help.github.com/articles/rendering-differences-in-prose-documents/).

lukego:
Removing `[wip]` tag. This branch is supposed to be a straightforward removal of dead code and shortening of long-winded documentation. I'd like to land it upstream.

lukego:
@eugeneia Small rant about keeping documentation build artefacts in the source tree, which you are free to ignore:

- I accidentally pushed a change to README.md.src without updating README.md. Seems easy mistake to make and easy to miss.
- Running `make markdown` has not worked on lab servers ever since they switched to NixOS. That is due to the hashbang issue (#737). This seems like a sign that people are not routinely working with updating the docs.
- Running `make markdown` regenerated the PNG image files and git shows these as changed when I don't believe that they really are, i.e. `ditaa` seems to be non-deterministic. I simply ignored adding these files on the assumption that the changes are not significant.
- To install ditaa on my laptop required a 120MB download of Java, etc. This is one of the most "unsnabby" things that I can imagine i.e. downloading a massive piece of software from Oracle in order to update a readme file. (I think somebody has actually ported Ditaa to Go because elsewhere in the world people also hate this dependency like I do.)
- I don't know if I have the same version of Ditaa as you do anyway so the images may be changing randomly depending on who builds and commits them i.e. this is an uncontrolled dependency in the software environment for maintaining our *source code* which seems to me kind of extreme.

Thank you for listening :). I feel that we are quite far from nirvana in terms of the documentation maintenance/build process but I don't have any actionable ideas right now.

I am the person who introduced Ditaa into the build chain in the first place so I feel that I an entitled to rant about it like this :).

eugeneia:
I am all for using a non-Java drop in replacement of ditaa; its not like I am a fan of Java its just that ditaa is super convenient. Regarding its non-determinism: yep I noticed that too, while its not critical it sure is confusing. Regarding `make markdown` not working in the lab: I haven't noticed because I do that locally, and suspect others do the same. I think you are theexception when it comes to doing anything beyond `git pull` and running snabb in the lab. Anyhow we should make sure that the snabb development dependencies are installed in the lab.

PR #742, opened 2016-02-04
Map `snabb-lwaftr' argv[0] to programs.lwaftr

wingo:
Rework the parsing of the Snabb command line to allow us to ship
a nicely named snabb-lwaftr binary, but have it map internally to the
also nicely-named 'programs.lwaftr' module tree.  Gets rid of the
nasty "snabb" name duplication in the module tree.

### DIFF
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -33,39 +33,34 @@ function main ()
    zone("startup")
    require "lib.lua.strict"
    initialize()
-   local args = parse_command_line()
-   local program = programname(args[1])
-   if program == 'snabb' then
-      -- Print usage with exit status 0 if help requested
-      if args[2] == '-h' or args[2] == '--help' then
-         usage(0)
-      end
-      -- Print usage with exit status 1 if no arguments supplied
-      if #args == 1 then
-         usage(1)
-      end
-      -- Strip 'snabb' and use next argument as program name
-      table.remove(args, 1)
-   end
-   program = select_program(program, args)
+   local program, args = select_program(parse_command_line())
    if not lib.have_module(modulename(program)) then
       print("unsupported program: "..program:gsub("_", "-"))
-      print()
-      print("Rename this executable (cp, mv, ln) to choose a supported program:")
-      print("  snabb "..(require("programs_inc"):gsub("\n", " ")))
-      os.exit(1)
+      usage(1)
    else
       require(modulename(program)).run(args)
    end
 end
 
--- If program stars with prefix 'snabb_' removes the prefix
--- If not, use the next argument as program name
-function select_program (program, args)
-   if program:match("^snabb_") then
-      return program:gsub("^snabb_", "")
+-- Take the program name from the first argument, unless the first
+-- argument is "snabb", in which case pop it off, handle any options
+-- passed to snabb itself, and use the next argument.
+function select_program (args)
+   local program = programname(table.remove(args, 1))
+   if program == 'snabb' then
+      while #args > 0 and args[1]:match('^-') do
+         local opt = table.remove(args, 1)
+         if opt == '-h' or opt == '--help' then
+            usage(0)
+         else
+            print("unrecognized option: "..opt)
+            usage(1)
+         end
+      end
+      if #args == 0 then usage(1) end
+      program = programname(table.remove(args, 1))
    end
-   return programname(table.remove(args, 1)):gsub("^snabb_", "")
+   return program, args
 end
 
 function usage (status)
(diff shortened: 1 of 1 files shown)


lukego:
Interesting! I thought we had already adopted this behavior but looking at the diff it doesn't seem so :).

src/core/main.lua
```diff
@@ -33,39 +33,34 @@ function main ()
...
-function select_program (program, args)
-   if program:match("^snabb_") then
-      return program:gsub("^snabb_", "")
+-- Take the program name from the first argument, unless the first
+-- argument is "snabb", in which case pop it off, handle any options
+-- passed to snabb itself, and use the next argument.
+function select_program (args)
+   local program = programname(table.remove(args, 1))
+   if program == 'snabb' then
+      while #args > 0 and args[1]:match('^-') do
```
eugeneia: Why is there a while loop here? I see three cases and no loop:

* `-h`/`--help`—print usage
* `-x`/`--foo`—error
* `foo`—program

eugeneia:
@lukego I think we are thinking of #640 .

src/core/main.lua
wingo: It's for if Snabb gets any more options.  LMK if you want s/while/if/ and s/do/then/; it will save a character ;)  By only checking for --help / -h the existing code won't detect invalid options, instead treating them as program names; for that reason I thought being more general would be more reliable.

eugeneia: Its a confusing piece of code that I had to read a couple times to know what its doing. I expected a case where it actually loops but since `usage` always exits it never does.

> It's for if Snabb gets any more options.

That explains the motivation behind this and I get it. I recommend avoiding coding for the future, as its confusing for readers.

wingo: Excuse my denseness, I have trouble knowing what your comments mean for me.  I prefer to leave it in, you prefer to leave it out, however what I don't know is whether you as reviewer require me to change the word "while" to "if" and "do" to "then".  If so please let me know.  Tx :)

eugeneia: Yeah, the two of us tend to talk past each other. FIXME ;-)

I generally do not set the requirements for accepting a PR, so I will usually not tell you or anybody else to “do X”. Instead, the project is developed in consensus of the contributors, and I can not truthfully represent that consensus on my own. So what I do is point out issues I see and give arguments as to why I see them as issues. E.g. I supply a single “vote” to the consensus, which is not itself a veto. Deciding what to do with my view is up to you. If you are convinced by my argument you might change the code or try to avoid similar issues in the future. If you are not convinced I would ideally get back an argument that explains why. Unless anybody else takes issue with the code I will merge it, and I will usually wait for a while to give others the chance to look at the PR, discuss it and maybe agree on changes.

So in this particular case I am simply making you aware of the fact, that as a reader of the code I found a particular part of it confusing (its just feedback). I couldn't tell from your reply if you understood my confusion, so I explained it in depth in the hope that you become aware of the general issue. (I also pointed out which part of your comment I found helpful, to give you a feel of what I would like to hear.) My goal (and I think this benefits everybody) is to reduce workload by increasing empathy with the consumers of code: when code is harder that necessary to read for reviewers, it will be more work for all. So in order to reduce that work I am letting you know how it would be easier for me (the reviewer in this case). Same goes for comments I guess, we need to learn what we actually want to hear from each other.

So to sum it up: You don't need to ask me what to do because I won't tell you what to do. It's your call, not mine. What I will tell you is my perspective on your code/comments/... and maybe you can use this second pair of eyes  to make it better.

lukego: @eugeneia I think there are two different hats that one can wear when reviewing code:

Community member: just giving feedback, nothing is binding, no responses are required.

Upstream maintainer: sole person responsible for accepting/merging the branch or spelling out explicitly what needs to be done before that can happen.

If you are the person who is resposible for upstreaming the PR and you don't press "Merge" then this is implicitly strong feedback "this code is not currently acceptable to me and so I have declined to merge it." In that case the contributor needs to clearly understand what are the deal-breakers so that they can fix them and the branch can be merged. The maintainer might say "You have to change A to B and remove C" or "we need to get an ack from @somebody or @someonelse" and then once that is done the contributor will expect the branch to merge.

This is why I have been stressing the importance of unambiguously dispatching incoming PRs between upstream maintainers. Then everybody knows who is wearing which hat and what is the path forward.

Overall there will be agreement between O(log(numberofmaintainers)) people who each merge the change onto their upstream branch as it makes its way to master.

someonelse:
My user name is @someonelse, I don't think I was meant to be included in this coding haha.

lukego:
@someonelse thank you for letting me borrow your name for the purpose of that example :-)

eugeneia:
Merged into `max-next`.

wingo:
Thanks!

PR #750, opened 2016-02-10
Merge max-next for 2016.03 release

eugeneia:
@lukego ready for review! :smiley:

### DIFF
--- a/lib/luajit/src/lj_record.c
+++ b/lib/luajit/src/lj_record.c
@@ -719,9 +719,14 @@ void lj_record_tailcall(jit_State *J, BCReg func, ptrdiff_t nargs)
   /* Move func + args down. */
   memmove(&J->base[-1], &J->base[func], sizeof(TRef)*(J->maxslot+1));
   /* Note: the new TREF_FRAME is now at J->base[-1] (even for slot #0). */
-  /* Tailcalls can form a loop, so count towards the loop unroll limit. */
-  if (++J->tailcalled > J->loopunroll)
-    lj_trace_err(J, LJ_TRERR_LUNROLL);
+
+  J->tailcalled++;
+
+  /* Although it's true that tail calls can form a loop, the Lua programming
+  ** idiom does not encourage this.  Instead of counting tail calls towards the
+  ** unroll limit and potentially causing important traces without loops to
+  ** abort, eventually blacklist, and fall back to the interpreter, just rely on
+  ** JIT_P_maxrecord to catch runaway tail-call loops. */
 }
 
 /* Check unroll limits for down-recursion. */
--- a/src/core/memory.c
+++ b/src/core/memory.c
@@ -29,18 +29,18 @@
 
 // Convert from virtual addresses in our own process address space to
 // physical addresses in the RAM chips.
-uint64_t virtual_to_physical(void *ptr)
+uint64_t virtual_to_physical(uintptr_t *ptr)
 {
-  uint64_t virt_page;
+  uintptr_t virt_page;
   static int pagemap_fd;
-  virt_page = ((uint64_t)ptr) / 4096;
+  virt_page = ((uintptr_t)ptr) / 4096;
   if (pagemap_fd == 0) {
     if ((pagemap_fd = open("/proc/self/pagemap", O_RDONLY)) <= 0) {
       perror("open pagemap");
       return 0;
     }
   }
-  uint64_t data;
+  uintptr_t data;
   int len;
   len = pread(pagemap_fd, &data, sizeof(data), virt_page * sizeof(uint64_t));
   if (len != sizeof(data)) {
@@ -80,7 +80,7 @@ uint64_t virtual_to_physical(void *ptr)
 void *allocate_huge_page(int size)
 {
   int shmid = -1;
-  uint64_t physical_address, virtual_address;
+  uintptr_t physical_address, virtual_address;
   void *tmpptr = MAP_FAILED;  // initial kernel assigned virtual address
   void *realptr = MAP_FAILED; // remapped virtual address
   shmid = shmget(IPC_PRIVATE, size, SHM_HUGETLB | IPC_CREAT | SHM_R | SHM_W);
--- a/src/doc/genbook.sh
+++ b/src/doc/genbook.sh
@@ -53,6 +53,8 @@ $(cat ../apps/socket/README.md)
 
 # Libraries
 
+$(cat ../lib/README.md)
+
 ## Hardware
 
 $(cat ../lib/hardware/README.md)
--- a/src/lib/README.md
+++ b/src/lib/README.md
@@ -1,0 +1,202 @@
+## Specialized data structures
+
+### Ctable (lib.ctable)
+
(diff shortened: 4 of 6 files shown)


lukego:
Thanks! Merged onto `next`.

I respect your decision to merge #700 although I am opposed to the content of that change. Snabb Switch does not actually work on 32-bit platforms and so I consider it a feature if compilation fails and a misfeature of compilation succeeds (deferring the inevitable problem to runtime). I realize that I am the minority here though so I will try to propose an alternative solution in the future rather than pushing back on this PR.

Onward!

lukego:
NB: I think it will work better next time if you make `next` the target of the PR instead of `master`. That way it will only show the commits that are not already on next and Github will detect when I have merged and close the PR.

eugeneia:
Ack! Regarding #700, the way I parsed the conversation/code is that the change was not architecture specific but really about correctness. There was C code and it used `void *` for a value that was not a `void *`, even though its a long shot (and I am no expert in the field of C typing) I see the possibility that this added strictness will help with a future debugging session.

lukego:
I believe that is the majority view. I am the minority who sees this as using C language-lawyering to land a broken untested port of i386.

lukego:
(I am glad that you merged it. That puts pressure on me to propose an alternative solution rather than just complaining and gatekeepering.)

kbara:
I'm of the majority view on this one, with the caveat that I gave the specific patch a couple of minutes of thought rather than a deep review. I am in agreement with your stance that we should minimize various types of cruft in the main code. Avoiding autoconf and a bunch of system-dependent conditionals have advantages. Avoiding dark corners where language lawyering is a necessity is good to the extent it's feasible.

Increasing the correctness of code on the primary platform while not adding anything, though, is something I'm entirely in favor of. I view the pre-patch C as wrong stuff that happened to work on our platform.

lukego:
@kbara I will concede that it would be a good change to land on an integration branch with a goal such as "port Snabb to i386" or "make Snabb independent of machine word size." Such a branch could then be merged onto master if it achieves its objective without undue complexity and is maintainable (e.g. has test case based on strict compiler warnings or running Snabb inside a 32-bit QEMU VM.)

The master branch is not the place to accumulate and integrate such changes piecemeal though. This makes work for the upstream maintainers and it doesn't benefit users. (It may actually harm users if it is motivated by helping distro maintainers to do things that upstream don't want e.g. to ship broken i386 binaries or to link with a LuaJIT that is missing critical fixes or ... etc.)

To me this is the same as if e.g. we had some Guile code and somebody wanted to make it more R5RS compliant by rewriting a few functions to not depend on the order of procedure argument evaluation. This is "more correct" with respect to some well-known specification but it is creating work for maintainers without utility for users (unless taken all the way and actually made portable to other schemes).

lukego:
I suppose the unstated assumptions in my position are:
1. That it *is* correct to cast between `uint64_t` and `void*` in a program that assumes `sizeof(uint64_t) == sizeof(void*)`.
2. That we do make this assumption on today on the master branch and this is a legitimate design decision.
This could be debated on #752.

kbara:
I agree that we shouldn't collect half-made ports on master. Assuming anything about the size of void* with regards to integer sizes horrifies me, though, even in non-portable code. I'm ok with the project loving x86_64 above all other platforms, but the previous code strikes me as gratuitously un-portable, rather than something that made the x86_64 implementation actively better/simpler/shorter.

My general heuristic is that unportable code is more likely to have various subtle bugs in it that porting makes obvious but which are active bugs even on the main platform. I'd also like to see the work of porters made marginally easier rather than marginally more difficult.

I consider the extra work for maintainers to go 'ok' to such patches an ok thing - I'm happy to have them go to kbara-next, anyhow.

lukego:
@kbara How do you feel about the "compromise" of rewriting the code in Lua (#751)?

Personally I like explicit stable ABIs. For example the Intel NIC always represents addresses as 64-bit values whether the host is running in 32-bit or 64-bit mode. Just zero the bits you are not using. I appreciate the simplicity of this and wish that more interfaces were so simple.

I dislike making ABIs implicitly dependent on compiler settings particularly in the Snabb context where we are doing things like mapping memory from virtual machines that may not even have the same word size as the host.

lukego:
I do think that the code only looks "gratuitously" unportable to a serious C hacker who is accustomed to writing portable code. I have the opposite reaction: `intptr_t` seems like a gratuitously high-brow concept to introduce into a program that only targets one architecture.

On a higher-level though the gratuitous thing here is using C at all when this code can so easily be written in Lua. I am much more willing to talk about what makes good Lua programming style for manipulating FFI pointers because that seems much more Snabby to me.

The C code should really be expunged from the codebase IMO. If we had a time machine we would have adopted ljsyscall from the beginning and we could have avoided antagonising the more sophisticated C hackers in the community.

lukego:
> My general heuristic is that unportable code is more likely to have various subtle bugs in it that porting makes obvious but which are active bugs even on the main platform

Give me some time and I will come around to your way of thinking on this. I have already had the realization that it is worth inserting "NOP" memory barrier operations with a rationale for why they do not apply to x86 rather than simply omitting them.

My excuse for digging in my heels on this pointer size issue is that the PR is coming from somebody who is attempting to create a `snabb-i386` package for a Linux distribution and in this situation the most important thing is to clearly communicate that this will not work and should not be done before a genuine port is implemented. I am trying to avoid making it easier for porters to shoot themselves and their users in their feet here.

lukego:
One last comment in this volley :-)

I also want to acknowledge that it is actually really *hard* to write portable code without the benefit of test suites and so on. I feel much more inhibited in my programming when I feel like I need to cross-reference everything I write with abstract specifications.

In support of this thesis I submit into evidence [Exhibit A](https://people.csail.mit.edu/gregs/ll1-discuss-archive-html/msg00612.html): The time when I had to debug a portability problem in a Lisp program written by Guy Steele where he had illegally depended on the order of elements returned by `PAIRLIS`. GLS is the editor of the ANSI standard that this violates and probably the person least likely to make such a mistake in the whole world... and yet it happens to him too.

I prefer to use subsets of these languages that have well-defined behavior.

kbara:
I've just taken a long weekend... now that I'm back:
a) I am *totally* in favor of moving as much stuff from C to Lua as possible, and explicit stable ABIs.
b) My bias is massively on the "be super-portable" side; I ran NetBSD for quite a few years. I see using intptr_t and similar even on one platform as a good thing. (Note that I do share your dismay at many of the methods used for portability, and agree that Snabb should look more like LuaJIT than a typical autoconf C program!)
c) I agree that creating packages that don't work is not good. It'd be cool to see a 32-bit port (but if I were to scratch the itch of porting Snabb, I'd probably port it to aarch64 instead).
d) It's definitely hard to write truly portable code, especially without test suites. All the more reason to improve Snabb's test suite over time. Referring to abstract specs isn't sufficient, either - it's surprising how often specifications and implementations disagree.
e) I fully agree about well-defined behaviour. A bunch of my work on libav was about undefined behaviour, I've read all of Regehr's posts about it, and spent many hours discussing them with many people. (See point a!)

PR #751, opened 2016-02-10
[wip] core.memory: Rewrite memory.c in Lua using ljsyscall

lukego:
Rewrite `memory.c` as part of `memory.lua` using ljsyscall. This is a good deed: reducing the amount of C glue code in Snabb.

Switch also to allocating HugeTLB pages as mmaped files on hugetlbfs instead of POSIX shared memory. The advantage of POSIX shared memory was that we did not need to rely on having a hugetlbfs filesystem mounted -- but now we auto-mount one at /var/run/snabb/hugetlbfs. I have been considering this design for quite some time.

This design is a bit simpler and should also make it easier for one Snabb process to map a page that was allocated by a different process.

The main motivation for rewriting memory.c is to make it easier to maintain. Lua code is more in keeping with the rest of Snabb and it doesn't require people to be expert C programmers and debate the merits of POSIX compliance and using GCC extensions and so on. (It also puts a bit more distance between Snabb and monstrosities like autoconf which cannot be a bad thing.)

TODO:
- [ ] Review style of ljsyscall usage (@justincormack?)
- [ ] _any requirements from upstream will go here_


lukego:
@xrme There is some code here you might be interested in btw: an optional (unused) `persistent` argument that will leave DMA/packet memory with its file backing so that another process could map it (e.g. via SEGV handler).

This is a bit simpler than with the POSIX share memory implementation because you can directly calculate the right filename from the pointer value, it's something like:

```
/var/run/snabb/dma.0000deadbeef
```

where the hex number is the physical address of the memory (which you can get from the virtual address by removing the tag).

xrme:
I agree that autoconf deserves to be given a very wide berth.  However, it seems to me that avoiding C in this case is not a very great simplification.  Nevertheless, if the C code smells funny and you just want it out of the house, that's an understandable reason.

The only technical argument I might make against replacing the C code is that it is easy to call C functions from assembly.  On the other hand, I don't know that assembly language code has any business calling these particular functions.

I prefer this design (mmap and hugetlbfs) to the current System V shared memory approach. It's a real win to be able to easily derive the appropriate name for a huge page directly from a pointer value. This makes it straightforward for multiple processes to allocate packet memory dynamically and share it with other processes.

A handler for SIGSEGV can get info->si_addr, see if it's properly tagged, and if so, construct the appropriate file name, open and mmap it, and proceed.  I like it.

PR #752, opened 2016-02-11
snabbswitch.c: Compile error on non-64bit platforms

lukego:
Provide a clear compilation error on platforms that don't have a 64bit integer size, for example i386. Include a link to porting information so that the user will understand that this is not an arbitrary
restriction.

This does not work: I did not formulate the C preprocessor directive correctly. Can somebody tell me how to write it in a way that works?

This depends on the `porting.md` document from #712 to land upstream.

See also #700 #750 cc @kbara.

### DIFF
--- a/src/core/snabbswitch.c
+++ b/src/core/snabbswitch.c
@@ -4,6 +4,10 @@
 #include "lualib.h"
 #include "lauxlib.h"
 
+#if sizeof(uint64_t) != sizeof(uintptr_t)
+#error "Snabb only supports 64-bit platforms. See https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/porting.md"
+#end
+
 int argc;
 char** argv;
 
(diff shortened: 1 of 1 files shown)


kbara:
I like @xmre 's solution. 

https://gcc.gnu.org/onlinedocs/gcc-3.0.2/cpp_4.html mentions a couple of reasons the original approach wouldn't work (#if ends with #endif, and more seriously, it can't include sizeofs). It's possible to work around those limits, but horrifying: see http://www.pixelbeat.org/programming/gcc/static_assert.html for a rough idea.

lukego:
Thanks for the tips! I have adopted @xrme's suggestion and it does indeed seem to work fine. I'd like to land this code upstream.

kbara:
Merged into kbara-next - https://github.com/kbara/snabbswitch/tree/kbara-next . Note that you'll want to land kbara-next after wingo-next, since the porting.md patch is there.

PR #743, opened 2016-02-04
Rename all README.md.src files to README.src.md

lukego:
This is intended to make Github match the ".md" suffix and treat the files as markdown format. This should enable better support in the web interface e.g. showing diffs with the rendered markdown.

Link to relevant Github feature:
https://help.github.com/articles/rendering-differences-in-prose-documents/

PR #753, opened 2016-02-11
Update documentation guide

eugeneia:
This updates the documentation guide according to #743 and also adds a section that talks about general writing style. I came up with this section today based on my own experiences with reading and writing documentation, so it might be controversial and not what we want for Snabb Switch documentation. Feedback welcome!

### DIFF
--- a/src/doc/documentation-guide.md
+++ b/src/doc/documentation-guide.md
@@ -95,6 +95,72 @@ that concatenates the various `README.md` files we have.
 
 # Stylistic Conventions
 
+## General Considerations
+
+Documentation is most effective when it is precise, concise and uniform.
+Documentation is not a novel, or a blog post. It should not be fun to read.
+Documentation should be normative, and as such be an especially unsurprising
+and bare read. Described below are the cornerstones of boring writing. Embrace
+them!
+
+ - **Preciseness**. Vague descriptions do more harm than good as different
+   readers will derive different interpretations. Make sure you use the most
+   specific phrasing possible. Ideally, there should be no room for
+   interpretation. If it is difficult to come up with a precise definition of
+   an API, chances are the API needs work.
+
+ - **Conciseness**. Every sentence strains the human attention span and risks
+   contradicting others. Express everything the API user needs to know with as
+   few sentences as possible. Unless a phrase is essential in order to
+   correctly *use* an API it does not belong in the documentation.
+   Documentation should only contain information the reader can *depend on*:
+   implementation details that might change are dangerous information, and do
+   not belong in the documentation. Readers might be inclined to write programs
+   with these details in mind. If you *must* include implementation
+   details—e.g. because they are required knowledge to use a module—separate
+   API definition from implementation detail clearly so that it is obvious to
+   the reader what kind of information he is digesting.
+
+ - **Uniformity**. Irregular documentation wastes reader attention by
+   prominently featuring irrelevant quirks. Surprises are reserved for
+   important information. By using simple, unspectacular language you save
+   attention that is better applied to the implications of a description rather
+   than its interpretation. Additionally, documentation becomes more accessible
+   for non-native speakers who might have to look up rarely used words. The
+   same applies for specialized notations. If you can express a definition
+   using either a notation or a simple sentence, go for the latter. Chances are
+   the reader is unfamiliar with the notation, and notations are often very
+   hard to “google”. Ideally, a reader able to read one chapter of the
+   documentation should be able to read every other chapter without having to
+   look up new words or concepts.
+
+
+## Markup Conventions
+
+We markup source code literals in `code` font. E.g.: "The `foobar` module
+is nice" and "`mod.fun(bla)` will make your dreams come true". Parameter
+identifiers are marked up in *italic* font. E.g.: "`mod.foo` takes an
+argument *bar*".
+
+UNIX system calls should be mentioned like so: `usleep(3)`.
+
+We markup specific *concepts* we introduce in italic font the first time
+they are mentioned in order to signify to the reader that a specific
+concept has a well defined meaning.
+
+## Terminology And Normalized Language
+
+The parameter names used in method and function description do not need
+to reflect the names used in the source code. Instead use long,
+descriptive names made out of full words when sensible.
+
+Symbol definition are written in third person, e.g.: "Returns a number"
+instead of "Return a number". When describing default behavior we say
+"The default is..." instead of "Defaults to..." etc.
+
+When in doubt, turn to the [Lua Reference Manual](http://www.lua.org/manual/5.1/)
+for linguistic and stylistic reference.
+
 ## Anatomy Of A Module Section
 
 Every module has its own subsection in the SnabbSwitch manual, and all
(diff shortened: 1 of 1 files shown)


eugeneia:
On thing I think we might want to add is a notation dictionary. E.g. pick a set of useful notations such as `[range_start,range_end)` and describe them in a “Notations” chapter. On the one hand these notations can be very useful on the other hand they can be cryptic if you are unfamiliar with them, and how do you google for `[a,b)`?

src/doc/documentation-guide.md
```diff
@@ -94,6 +95,72 @@ that concatenates the various `README.md` files we have.
 
 # Stylistic Conventions
 
+## General Considerations
+
+Documentation is most effective when it is precise, concise and uniform.
+Documentation is not a novel, or a blog post. It should not be fun to read.
+Documentation should be normative, and as such be an especially unsurprising
```
wingo: There are a lot of things that are great about better doc standards but I don't share the opinions in this "general considerations" section.  Concretely, my differing opinions:

 * Documentation can be boring but it does not have to be.
 * Documentation can be both good and fun.  Good is a requirement, fun is a nice-to-have.
 * 11 lines on being concise is too much
 * It feels like more of a blog post than documentation ;)  Or rather: too much opining for the guidelines.
 * It is a backhanded commentary on my latest doc merge without @-tagging me, but expressed as a PR instead of an issue to discuss.
 * A lost opportunity to use the word "concision", which sounds like scissors and also rhymes with "precision"!  A travesty!

kbara:
I think a notation dictionary is a good idea, as well as a glossary of acronyms.

lukego:
I still feel like the documentation is irrelevant unless/until people are actually using it on a daily basis and I wonder to what extent they are.

src/doc/documentation-guide.md
eugeneia: @wingo While the notation you used serves as an example this is in no way personally about you or the documentation you wrote (other people wrote documentation too, even before last week). Could we agree on not taking these kind of things to a personal level? I find this extremely annoying. If I have a personal issue I will contact you personally in private.

> It feels like more of a blog post than documentation ;) Or rather: too much opining for the guidelining.

Absolutely agree.

kbara:
@lukego I found the documentation on how the Intel driver's 'vlan' argument to be extremely valuable yesterday. I'd say I routinely look at the documentation of Snabb code I'm using, if it has any. It gives an idea of what's intended API vs implementation detail, what's intentional or a bug, etc.

lukego:
@kbara great to hear! How do you consume it -- via Github, build HTML manual, or read the markdown source locally?

wingo:
@kbara @lukego I was going to say exactly the opposite!  That documentation was incorrect, incomplete, and misleading!!! https://github.com/SnabbCo/snabbswitch/issues/749

wingo:
I would have been much better off reading the source code than the documentation.

eugeneia:
I use the documentation frequently on GitHub as well as by reading the Markdown source to “load in” context when working with modules and also to track and identify API changes in code.

src/doc/documentation-guide.md
```diff
@@ -94,6 +95,68 @@ that concatenates the various `README.md` files we have.
 
 # Stylistic Conventions
 
+## General Considerations
+
+Documentation is most effective when it is precise, concise and uniform.
+
+ - **Preciseness**. Vague descriptions do more harm than good as different
```
wingo: for me, lines 102 to 131 can be removed.  They add no more meaning than the words "precise", "concise", and "uniform".

kbara:
I read the markdown locally, *as well* as the source.

lukego:
I would find it valuable to cite/link parts of the documentation that we especially like and want to emulate in the future. Then these rules could be used more as a reference and to fill in details.

I find it noteworthy that @kbara is the number one fan of the documentation and that she is consuming it in source format in her text editor. This reinforces my feeling that we have an accessibility problem for the formatted documentation.

I wonder how `snabb doc man packet` or `snabb doc apropos append` commands might be implemented? (Compile the docs into the executable and print them to the console? Or compile in an index and link to the online docs? Or just not a good idea in the first place?)

kbara:
I don't mind reading various markup formats as source, as long as they're not as annoying as XML with a ton of attributes, fwiw.

src/doc/documentation-guide.md
eugeneia: I don't think stating three words would be enough, and I think that explaining the concepts with reasoning and specific examples is what communicates the intent clearly.

lukego: How about a link to some widely accepted technical writing style guide or a list of pithy quotes? I really do think that "preachy" is a tone that we absolutely want to avoid.

I know that I am a broken record but I still feel like the core problem we need to fix re: documentation is to make it more usable/accessible. Otherwise it just feels like a tax.

lukego:
I reflect that we are all giving our $0.02 here but we have not identified an upstream maintainer to actually guide this change towards merge. How to solve that?

kbara:
I would nominate @eugeneia , on the grounds that he's shown the most enthusiasm/energy on the topic.

wingo:
I don't think reviewing our own patches is a pattern we should be encouraging :)  @kbara that would mean you or me would be the next hop for this one.  I do not like the change as it stands, which made me hesitant to take responsibility for getting it in (or keeping it out of course).  OTOH if you would like to take it @kbara I will not NAK it.

kbara:
Oops, I'd managed to forget that it was his patch; excellent point.
I'm ok with this change as-is, though I think both you and Luke have made reasonable points; a patch that addresses those is also welcome. Reducing the 'general considerations' to a pointer to the Lua Reference manual would perhaps be an improvement, but I'd be willing to accept the current text. The src.md/md.src changes are fine, and I mainly agree with the point about where implementation details should be documented.
I suppose there's a case to be made for splitting it into the non-controversial and controversial bits, merging the former onto kbara-next, and hashing out the rest elsewhere or dropping them.
But the tl;dr version is that while I don't love every detail of this patch, I'm willing to merge it as-is, as well as to accept further patches from anyone changing the text.

lukego:
I still view the branches in the kernel / subsystem model. For me we should have a `documentation` branch, this branch should have an owner, that person should decide what changes are merged and how much review is enough. If we want the documentation done differently then we would send in PRs as input to that person (who would still hold the editorial control).

On the other hand I don't want *every* change to have to pass through the documentation branch so all maintainers need to be basically on the same page about what is expected and what we should ask downstream contributors to provide.

The model we are using now of maintainers volunteering to take changes on a case-by-case basis seems unlikely to scale to me, but I am prepared to be proven wrong.

kbara:
If we want particularly cohesive high-level documentation, I think we need to use an approach like that. For API-level documentation, as you said, not every change should pass through such a mechanism.

eugeneia:
We have `documentation-fixes` which is maintained by me as well and I think it could very well accept PRs that are about high level documentation considerations. 

Regarding this PR: which phrases are preachy? I didn't intend to sound preachy obviously. ;-)

As to the suggestion to point to an external guide: we do this in some way already by saying to use the Lua reference manual as orientation, but for this particular case I find it unpractical to say “derive what we expect from you from this 'book'”. I don't know of any applicable style guide or other existing documentation guide either, and think this PR adds some valuable pointers that imho are not obvious and very useful. E.g. I had to figure this out myself and past-me would have benefited from a brief “general writing guide”.

lukego:
I reflect that I am [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality) on this PR so please take my feedback with a grain of salt. On a meta-level I feel like we will have more of a positive effect on documentation by increasing the rate of contributions (merges) in this area. This can all be changed in follow-on PRs after it lands.

I mark @kbara as the Assignee here based on the discussion above. Seems that @kbara has signaled that she is willing to merge so now it's up to @eugeneia to signal how much more revision/feedback he wants to do before that.

lukego:
(@kbara I can't assign you. Github team invite stuck in your Spam folder?)

lukego:
@eugeneia Generally I am happy to pull documentation improvements from you directly when you decide they have had enough feedback and revision. Like I mentioned above I do want to move towards having subject matter experts (individuals or small groups) who consistently take the lead on upstreaming changes in specific areas (subsystems). I hope that at the moment we are in the process of discovering where upstream people's interests lie.

kbara:
@eugeneia are you willing to remove lines 102-131 as @wingo commented on, or would you prefer me to merge this patch set as-is? I'm willing to merge either.

eugeneia:
@kbara I say merge it as is for now. I am interested in editing the writing to be less preachy but feel like I would not be able to accomplish this right now without more concrete feedback. I do think that without the explanations the section would be of little use. Any concrete suggestions/feedback can be given in form of PRs, issues or further comments which I would then react to.

kbara:
Fair. As you said, the feedback hasn't been particularly actionable, and I think this commit is a net contribution. Merged, and thank you.

PR #754, opened 2016-02-12
[wip] lib.blackbox: New "flight recorder" logging mechanism

lukego:
Comment from the source code:

The "black box" of a Snabb process is like the flight recorder in an airplane: a detailed event log that can be inspected for investigting problems such as crashes or performance degradations that occur in a production environment.

The black box logs entries by fork()ing a child process to produce and log the message. This makes the complete state of the process available for inspection without delaying the parent process (which
can continue to process traffic). This is intended to enable logging detailed information that could be computationally expensive to produce without without impacting the latency of
traffic processing.

For example:

* Print full LuaJIT trace dumps.
* Print a profiler report.
* Scan the heap and summarize all allocated memory.

The high-level problem this would solve is making it possible to do serious performance analysis of any Snabb application while it is running (or after it has terminated). This would replace things like the NFV-specific "benchmark mode" code that enables profiling and trace-dumping only when started with special parameters (`snabbnfv traffic -B`).

Here is the output of the selftest function. Seems that occasional messages don't impact throughput too much. Latency has to be measured though.

```
selftest: blackbox
measuring simple app network throughput with recording intervals
(none) load: time: 1.00s  fps: 76,469,964 fpGbps: 42.211 fpb: 255 bpp: 60   sleep: 0   us
0.500s load: time: 1.00s  fps: 76,532,995 fpGbps: 42.246 fpb: 255 bpp: 60   sleep: 0   us
0.250s load: time: 1.00s  fps: 76,178,983 fpGbps: 42.051 fpb: 255 bpp: 60   sleep: 0   us
0.100s load: time: 1.00s  fps: 75,983,353 fpGbps: 41.943 fpb: 255 bpp: 60   sleep: 0   us
0.050s load: time: 1.00s  fps: 75,648,113 fpGbps: 41.758 fpb: 255 bpp: 60   sleep: 0   us
0.010s load: time: 1.00s  fps: 73,879,869 fpGbps: 40.782 fpb: 255 bpp: 60   sleep: 0   us
0.005s load: time: 1.00s  fps: 70,481,339 fpGbps: 38.906 fpb: 255 bpp: 60   sleep: 0   us
0.001s load: time: 1.00s  fps: 59,536,270 fpGbps: 32.864 fpb: 255 bpp: 60   sleep: 0   us
selftest ok
```

This is a work in progress. TODO:

- [ ] Measure latency impact of `fork()` on parent.
- [ ] Consider file format that is easy to scan/parse.
- [ ] Hook in `jit.dump` and `lib.traceprof` and related tools.

### DIFF
--- a/src/lib/blackbox.lua
+++ b/src/lib/blackbox.lua
@@ -1,0 +1,91 @@
+-- blackbox.lua -- detailed event recorder ("black box flight recorder")
+-- 
+-- ### Overview
+-- 
+-- The "black box" of a Snabb process is like the flight recorder in
+-- an airplane: a detailed event log that can be inspected for
+-- investigting problems such as crashes or performance degradations
+-- that occur in a production environment.
+--
+-- The black box logs entries by fork()ing a child process to produce
+-- and log the message. This makes the complete state of the process
+-- available for inspection without delaying the parent process (which
+-- can continue to process traffic). This is intended to enable
+-- logging detailed information that could be computationally
+-- expensive to produce without without impacting the latency of
+-- traffic processing. For example:
+--
+-- * Print full LuaJIT trace dumps.
+-- * Print a profiler report.
+-- * Scan the heap and summarize all allocated memory.
+--
+-- ### API
+--
+-- record(event_name, func):
+--   Run 'func' in a fork()ed child process and capture its output
+--   (calls to print() and io.write()) in a named black box record.
+
+module(..., package.seeall)
+
+local S = require("syscall")
+
+function record (event_name, func)
+   local ppid = S.getpid()
+   if S.fork() ~= 0 then return end -- parent process
+   local box = ("/var/run/snabb/%d/black.box"):format(ppid)
+   -- Redirect standard output to the blackbox file
+   io.output(assert(io.open(box, 'a')))
+   func()
+   os.exit(0)
+end
+
+function selftest ()
+   print("selftest: blackbox")
+   print("measuring simple app network throughput with recording intervals")
+   local function configure ()
+      local c = config.new()
+      config.app(c, "source", require("apps.basic.basic_apps").Source)
+      config.app(c, "sink",   require("apps.basic.basic_apps").Sink)
+      config.link(c, "source.output->sink.input")
+      engine.configure(config.new())
+      engine.configure(c)
+   end
+   -- Setup timer
+   local interval, deadline
+   local timerhook = function ()
+      if interval then
+         deadline = deadline or engine.now()
+         if engine.now() >= deadline then
+            record("record",
+                   function ()
+                      -- Burn some cycles
+                      for i = 1, 1e7 do end
+                      -- Print a message (to the black box log)
+                      for i = 1, 1000 do io.write("test record\n") end
+                      -- Hang around long enough to be a potential
+                      -- nuisance for copy-on-write memory.
+                      require("ffi").C.usleep(100)
+            end)
+            deadline = deadline + interval
+         end
+      end
+   end
+   timer.activate(timer.new('blackbox', timerhook, 1e6, 'repeating'))
+   -- Warmup
+   configure()
+   engine.main({duration=0.001})
+   -- Run for one second with no recording
+   configure()
+   io.write("(none) ") io.flush()
+   engine.main({duration=1})
+   io.flush()
+   -- Run with recording at different intervals (seconds)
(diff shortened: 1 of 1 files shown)


lukego:
@wingo do you have some code for measuring the latency of the breathe loop that could come upstream for testing here?

lukego:
@eugeneia @wingo @kbara I also adopted an experimental documentation style here: to write real documentation in markdown syntax but include it inline at the top of the source file. The Makefile could extract this for inclusion in a manual. I still think the key to high-quality documentation is to put it somewhere that people will see it every day. Flame on! :-)

aperezdc:
@lukego: I really like the idea of including documentation in the source files :+1: 

If the idea goes forward, it would be amazing to adopt it also for generating the output of `--help` for programs. That way the “command invocation help text” (which, from my POV, is also a kind of “online” documentation) would also be next to the source code that it is describing.

wingo:
@lukego: For breathe() latency I have https://github.com/Igalia/snabbswitch/pull/256.  Will refactor into something mergeable soon, as we need to be able to use it in delivered applications.

Regarding documentation and source files: I really like this style of things.  It is close to the code so it is unlikely to drift out of sync (and it's easier to review for correctness), it is readable as-is as text and it can also be extracted.  I wonder though that the quality of the markdown would slip -- i.e. if reading it on github doesn't highlight rendering faults, then we will tend to make more faults.  Not sure what the solution is there; it's not something that's amenable to automated checking, though automated generation of course is a good idea.

lukego:
Here is an idea: How about if the "black box" is literally a core dump file (or equivalent) produced either on termination of on request (e.g. signal requesting a background fork-and-dump).

This way analysis would be done via inspection of a snapshot of the heap i.e. actually reading the LuaJIT internal data structures. This would be done with the moral equivalent of GDB scripts to extract the state you are interested in e.g. locate the Lua VM via a C global variable and extract all of the bytecode/IR/mcode for the traces and then parse that as binary data for analysis. If ephemeral information is also needed - e.g. information about aborted trace recordings - then that would need to be stored somewhere in the heap e.g. a ring buffer.

The advantages that I see to this approach over explicitly logging messages like trace dumps and profiler reports are:

1. Have all information that exists available for analysis. Less burden of anticipating what information will be needed and logging that explicitly. (Avoid "It crashed in production? Okay, please deploy this new version that will also crash but provide more debug information that we missed the last time.")
2. Simple software interface between runtime and analysis: Just write down the ABI and access everything directly. Compare with e.g. extending the LuaJIT runtime system to export more information, writing Lua code to dump that into an external format, writing a parser for that format. (For example #623 traceprof could be rewritten as an external tool and we could drop the patches I made to export more information about traces from the LuaJIT runtime system.)
3. Less frequent fork-and-dump required. If the complete core dump can be created reliability, even from a process that is crashing hard, then we can save the black box "just in time" rather than as a continuous background process.

btw: I am [exploring](https://twitter.com/lukego/status/701444892296409090) using the [Moose toolkit](http://www.moosetechnology.org/) from the university of Bern to partly automate the analysis I usually do on LuaJIT trace files e.g. cross referencing with profiler information to see which traces are relevant, checking those for red flags like unsunk allocations, looking for problems like lukego/blog#8, etc. Could be useful as a tool for exploring and prototyping ways to inspect the black box. Currently I am parsing the textual trace dumps but I imagine that operating on core files would work too. Early days yet.

Thoughts? cc @wingo @kbara @xrme @mwiget.

kbara:
Small world; I did my bachelor thesis on visualizing Scheme programs in Moose.

lukego:
@kbara Wow! That is a small world indeed :). How did you like it? Can I ask you tech questions? :-)

kbara:
It wasn't bad, though this was back in 2007 and I think it's changed a lot since then. You can ask me questions, but most of what I did was in Scheme, with just a couple of small shunts in Smalltalk to use the metrics I was calculating, so I wasn't ultra-deep into Moose itself (but my supervisor and his grad students were).
Used well, for the right problems (visualizing where complexity is in a system, for instance), Moose is stunning. For crash dump analysis, I honestly don't know. :-)

wingo:
Relatedly, see https://github.com/Igalia/snabbswitch/pull/276.  The kind of telemetry that I am interested in is often real-time in nature, but I guess that's related to the Yang discussion.

lukego:
Closing this PR for now. I will reopen when I have something new to show.

PR #760, opened 2016-02-17
Merge max-next into next

eugeneia:
Contains merges with #742 #743 #747.

Ping @lukego


lukego:
Merged!

I would prefer to have shorter summary lines on the commits so that they don't get elided/truncated when browsing history. See the `...` markers on the commit list above. This is to keep the history easy to browse with limited columns and wanting to include other info e.g. date, author, etc.

PR #762, opened 2016-02-19
Clearly mark all Apache 2.0 licensed code as such

lukego:
This change adds one-line explicit license references for source files that are Apache 2.0 licensed i.e. the vast majority of code in Snabb Switch:

```
Use of this source code is governed by the Apache 2.0 license; see COPYING.
```

This change attempts to preserve all existing license and copyright messages, with the exception of avoiding redundantly referencing Apache 2.0 license more than once in the same file.

The motivation for this change is twofold:

- To make the source code license more obvious to users and contributors.
- To put into practice the agreement made in issue #729.

The change was initially made automatically with a 'sed' script and then I reviewed each file individually.

Notes:
- avx2.c and sse2.c state that they are derived from MIT-licensed code.
- dynasm lua-mode is already clearly marked as public domain.
- checksum.c states that it is derived from BSD licensed code.
- utilities imported into the project are left as-is e.g. StackTracePlus.lua.

I would appreciate if the upstream reviewer would pass through and sanity check this with fresh eyes!

### DIFF
--- a/src/apps/basic/basic_apps.lua
+++ b/src/apps/basic/basic_apps.lua
@@ -1,3 +1,5 @@
+-- Use of this source code is governed by the Apache 2.0 license; see COPYING.
+
 module(...,package.seeall)
 
 local app = require("core.app")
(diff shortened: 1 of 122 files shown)


eugeneia:
Scanned through the lot, looks good to me. Moving existing copyright notices to a central file has to be done in a separate PR.

PR #763, opened 2016-02-21
snabbnfv: Shrink pcap files to reduce repo size

lukego:
Shrinks the disk size of src/ from 12.5MB to 3.7MB by removing redundancy from some test vector files. (Compare with 6.7MB for our "lib/" dependencies including LuaJIT, etc.)

The specific change is that the Snabb NFV test vectors for various packet sizes now each contain a single packet. Previously the files were larger e.g. '9000.pcap' weighed in at 2.2MB for 255 x 9000 byte packets.

I assume that these are inputs to packetblaster and that the packet will be automatically looped and so this will not change system behavior. If this is wrong or confusing then we should perhaps update packetblaster code or documentation.

These files could perhaps be eliminated entirely by using the "synth" option of packetblaster instead but I leave that as future work.

cc @eugeneia does this look okay to you?


lukego:
I noticed this while doing some hacking on my space-constrained Chromebook CB3-111 laptop. Interesting that storage space for source tree is a metric that we overlooked in the beginning of the project.

wingo:
We have about 1.8M, of which a little less than 1M are pcap files, in program/lwaftr.  Some 600K are images for benchmarks.  That might impede a merge, FWIW.  They are not amenable to the "packetblaster synth" treatment, as many of them are error cases and so on.

Issue #758, opened 2016-02-16
Documentation problem: including build artifacts in the repo


lukego:
There is a problem with our current method of documentation generation: that we checkin documentation build artifacts to the repository rather than only source code. This is messy. If you want to change an ascii-art diagram in a README file then we expect you to install Java (!) so that you can regenerate the PNG file using Ditaa. I think it would be much better to include only source in the repository.

How about if we would create a separate repository for the documentation, e.g. `snabbdocs`, and store the documentation there? This could in principle by managed by SnabbBot: it could build the documentation for PRs/merges/releases and check that into the `snabbdocs` repository. Then we would simplify the build dependencies for developers and always have the documentation available in various formats (e.g. markdown, html, pdf, epub).

Hydra (Nix) may also be an easy (easier?) way to build the artifacts than extending SnabbBot but with the difference (downside?) of storing the results on a filesystem of some machine rather than in a Github repo.

cc @eugeneia @wingo @kbara @domenkozar

domenkozar:
Separating docs from source has (a possible) big downside, separating source changes from documentation changes. It's harder to pinpoint the two. It's better to not commit build artifacts in the first place, since that makes the repository quite big.

I'd recommend using a simple Markdown builder like http://www.mkdocs.org/ and https://readthedocs.org/ (supports multiple versions and multiple formats) for hosting the documentation, they're doing an awesome job keeping that site running. We could hook readthedocs into git so it would build the docs, but if that won't be possible due to dependencies we can use Nix+Hydra to publish them. Very SaaSy, but I use this workflow for all my projects and so far it worked well.

eugeneia:
I think that the current documentation handling is extremely simple, sophisticated and scalable. I disagree that having the artifacts in the repository or installing ditaa is a problem. I am intuitively against all suggestions in this thread so far because I expect none are thought through to the end and I expect them to complicate things and/or deprive us of valuable working features.

* SaaS please no. I tried to make mkdocs and readthedocs consume Snabb Switch's markdown files and ended up trying to decipher opaque error messages from their Python/Ruby/Whatever web apps. I would have considered them an option if they managed to build something useful form the Markdown in the Snabb repository without intervention, but even that would probably have been a mistake. I will not email a SaaS support hotline, so I am against SaaS.

[...]

Issue #734, opened 2016-02-02
Create new "snabb" module to define our API


lukego:
I think it is important to create a unified API accessible via the module `snabb`.

This would accomplish a great many things:

1. Define a supported interface with [semantic versioning](http://semver.org/) and consistent user documentation.
1. Define - by exclusion - the *unsupported* interface that represents engine internals.
2. Decouple the naming of modules and apps from the source code layout e.g. `snabb.app.intel10g` vs `app.intel.intel_app` or `snabb.ctable` instead of `snabb.ctable.ctable`.
2. Avoid `require()` by having everything in a ready-available namespace (`snabb.engine`, `snabb.packet`, `snabb.ffi`, `snabb.bit`).
3. Avoid the need to cache modules/functions with `local bit = require("bit")  local lshift = bit.lshift` by representing the `snabb` table with FFI metatables (eagerly bound) rather than ordinary Lua tables (late bound).
4. Could even be possible to use `setfenv()` to avoid having to type the `snabb.` prefix...?

I tried this once before in #465 but didn't manage to fully develop the idea at that time. Since then some things seem clearly, for example it seems that we *do* trust allocation sinking enough to continue returning FFI pointers from core functions like `packet.receive()`, while other things have become murkier, for example this new idea that we could keep the syntax `module.function(x, ..)` instead of `x:method(...)` and still get the [performance benefit of FFI metatable early binding](https://groups.google.com/d/msg/snabb-devel/STgQrTqhDiI/S1l91k58D4UJ) simply by representing the module as an FFI object.

Following up #733 and  cc @wingo @eugeneia 

Is there anybody who wants to take the lead on this? I can contribute strong opinions...

kbara:
If no one has stepped up by March, I'm interested. I'd really like to see this.

lukego:
@kbara it would be awesome if you take the lead on this!

wingo:
I do not like being negative but I am not so sure that this is a good idea :)  If we move the implementation of our code to this module layout, great.  However anything that adds an abstraction between what code I'm using and where it's implemented is less great.  When I see that a module does `require('lib.foo')` I know where to go and look for that code; if it's some synthetic API, not so much.  How we build that API could mitigate that, but I wonder if the extra abstraction would not be a mistake.

lukego:
@wingo The main value I see in this activity is to ask ourselves, "How should Snabb application code really look?"

[...]

PR #764, opened 2016-02-22
[sketch] lib/README.md: Super-rough import of comments

lukego:
Very very roughly import the module comments into README.md to get a feel for how the documentation would be structured if we migrated and edited all of this properly.

If this were done properly then would it be an improvement and a step in the direction that we want to take? cc @eugeneia @wingo @kbara.

Note: The `next` branch currently has very poor consistency in terms of documentation style for `lib/`. The master branch has all of these modules documented in comments whereas the `next` branch has devoted the whole of `lib/README.md` to the `ctable` module. I would like to make this consistent *somehow* before we ship the next release.

Question is: should we do this PR properly or should we document `lib/` modules with comments in sources?

### DIFF
--- a/src/lib/README.md
+++ b/src/lib/README.md
@@ -1,4 +1,202 @@
+## Library routines
+
+### PMU (CPU Performance Monitoring Unit) (lib.pmu)
+
+This module counts and reports on CPU events such as cache misses,
+branch mispredictions, utilization of internal CPU resources such
+as execution units, and so on.
+
+Hundreds of low-level counters are available. The exact list
+depends on CPU model. See pmu_cpu.lua for our definitions.
+
+API:
+
+```
+profile(fn[, event_list, aux]) => value [and print report]
+  Execute 'fn' and print a measurement report for event_list.
+  This is a simple convenience function over the API below.
+
+measure(fn[, event_list]) => result, table {counter->value}
+  Execute 'fn' and return the event counters as a second value.
+  This is a convenience similar to profile().
+
+is_available() => true | false, why
+  Return true if hardware performance counters are available.
+  Otherwise return false with a string briefly explaining why.
+
+setup(event_list)
+  Setup the hardware performance counters to track a given list of
+  events (in addition to the built-in fixed-function counters).
+  
+  Each event is a Lua string pattern. This could be a full event name:
+    'mem_load_uops_retired.l1_hit'
+  or a more general pattern that matches several counters:
+    'mem_load.*l._hit'
+
+  Return the number of overflowed counters that could not be
+  tracked due to hardware constraints. These will be the last
+  counters in the list.
+
+  Example:
+    setup({"uops_issued.any",
+           "uops_retired.all",
+           "br_inst_retired.conditional",
+           "br_misp_retired.all_branches"}) => 0
+
+new_counter_set()
+  Return a "counter_set" object that can be used for accumulating events.
+
+  The counter_set will be valid only until the next call to setup().
+
+switch_to(counter_set)
+  Switch_To to a new set of counters to accumulate events in. Has the
+  side-effect of committing the current accumulators to the
+  previous record.
+
+  If counter_set is nil then do not accumulate events.
+
+to_table(counter_set) => table {eventname = count}
+  Return a table containing the values accumulated in the counter set.
+
+Example:
+  to_table(cs) =>
+    {
+     -- Fixed-function counters
+     instructions                 = 133973703,
+     cycles                       = 663011188,
+     ref-cycles                   = 664029720,
+     -- General purpose counters selected with setup()
+     uops_issued.any              = 106860997,
+     uops_retired.all             = 106844204,
+     br_inst_retired.conditional  =  26702830,
+     br_misp_retired.all_branches =       419
+    }
+
+report(counter_set,  aux)
+  Print a textual report on the values accumulated in a counter set.
+  Optionally include auxiliary application-level counters. The
+  ratio of each event to each auxiliary counter is also reported.
+
+  Example:
+    report(my_counter_set, {packet = 26700000, breath = 208593})
+  prints output approximately like:
+    EVENT                                   TOTAL     /packet     /breath
+    instructions                      133,973,703       5.000     642.000
+    cycles                            663,011,188      24.000    3178.000
+    ref-cycles                        664,029,720      24.000    3183.000
+    uops_issued.any                   106,860,997       4.000     512.000
+    uops_retired.all                  106,844,204       4.000     512.000
+    br_inst_retired.conditional        26,702,830       1.000     128.000
+    br_misp_retired.all_branches              419       0.000       0.000
+    packet                             26,700,000       1.000     128.000
+    breath                                208,593       0.008       1.000
+```
+
+### Checksum calculation (lib.checksum)
+
+```
+This module exposes the interface:
+  checksum.ipsum(pointer, length, initial) => checksum
+
+pointer is a pointer to an array of data to be checksummed. initial
+is an unsigned 16-bit number in host byte order which is used as
+the starting value of the accumulator.  The result is the IP
+checksum over the data in host byte order.
+
+The initial argument can be used to verify a checksum or to
+calculate the checksum in an incremental manner over chunks of
+memory.  The synopsis to check whether the checksum over a block of
+data is equal to a given value is the following
+
+ if ipsum(pointer, length, value) == 0 then
+   -- checksum correct
+ else
+   -- checksum incorrect
+ end
+
+To chain the calculation of checksums over multiple blocks of data
+together to obtain the overall checksum, one needs to pass the
+one's complement of the checksum of one block as initial value to
+the call of ipsum() for the following block, e.g.
+
+ local sum1 = ipsum(data1, length1, 0)
+ local total_sum = ipsum(data2, length2, bit.bnot(sum1))
+
+The actual implementation is chosen based on running CPU.
+```
+
+### Mac addresses (lib.macaddress)
+
+```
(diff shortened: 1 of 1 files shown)


lukego:
First observation: a table of contents is needed to navigate this much content.

wingo:
Regarding lib/README.md, it was requested explicitly by @eugeneia in #722.  I still think lib/README.*module*.md is nicer for browsing on Github, because as you say it is a bit overwhelming to have everything in one vertical scroll, and it seems that GitHub doesn't do tables of contents in markdown files.  Dunno.

lukego:
Could also move each library into a dedicated `lib/foo/` subdirectory with its own README. Then on Github the directory listing would act as a table of contents. Could also hack `require()` to make short names like `require("lib.ctable")` work.

However: I really think we need to nail down which user experience we are trying to optimize foremost. On the one hand if we focus on a stand-alone HTML/PDF/EPUB manual created with `pandoc` then we presumably have a lot of options for structures and showing chapter contents and so on. On the other hand if focus on reading via Github then the file layout is important and we have to make a trade-off between clutter vs latency of clicking around in subdirectory links.

So I think we need to resolve that question over on #758 first i.e. which documentation user experience are we primarily optimizing?

eugeneia:
> Could also move each library into a dedicated lib/foo/ subdirectory with its own README. Then on Github the directory listing would act as a table of contents. Could also hack require() to make short names like require("lib.ctable") work.

I am inclined to go with this and would be willing to refine this PR for that purpose.

> However: I really think we need to nail down which user experience we are trying to optimize foremost. 

I don't think we can get around providing standalone HTML/PDF/EPUB manuals as  the primary target as this is the de-facto standard as far as I can tell. The upside is that this doesn't conflict with organizing the documentation in a GitHub friendly way at all. So I think we can optimize organization for GitHub and render a standalone manual independently without restrictions that arise from this.

wingo:
On Mon 22 Feb 2016 15:10, Max Rottenkolber <notifications@github.com> writes:

>     However: I really think we need to nail down which user experience
>     we are trying to optimize foremost. 
>
> I don't think we can get around providing standalone HTML/PDF/EPUB
> manuals as the primary target as this is the de-facto standard as far as
> I can tell. The upside is that this doesn't conflict with organizing the
> documentation in a GitHub friendly way at all. So I think we can
> optimize organization for GitHub and render a standalone manual
> independently without restrictions that arise from this.

Why not just files in markdown that are meant to be read in github?
Personally I don't see the need for PDF or EPUB at all.  Markdown
documents are easy to improve (once you have a branch with the basic
contents, you can edit online in github to make sure the formatting
looks right; a fairly low-latency cycle) and on github they link really
well to other files, can embed images natively, and ascii art is just as
good as the generated graphs.

I tried hard to avoid giving my opinion but I seem to have failed.  Feel
free to ignore me :)

lukego:
@wingo Fair question. My $0.02 is that the Github UI is too laggy. In practice I am not willing to click around there looking for documentation: I read the sources in my editor instead (like @kbara). I see this as a negative thing because it makes me lose touch with the "real" documentation and not be motivated to improve it.

lukego:
I hope that having a browser tab with the complete Snabb manual open and easy to navigate will fix that situation. (Or if not already open then at least auto-completing as a URL in my browser.)

eugeneia:
> Why not just files in markdown that are meant to be read in github?

While having documentation formatted on GitHub is nice, I feel we can't be limited by GitHub in documentation UI (things like having a ToC etc). Also, Package maintainers for distributions usually want to ship documentation so being able to build a standalone HTML (or maybe even man page one day, I think `pandoc` can do this as well) is useful to them.

lukego:
@capr do you have a suggestion for how we could make `require("foo")` automatically consider `require("foo.foo")` as a fallback? I know that ordinarily the Lua path can be used to make an `init.lua` load but I don't think LuaJIT supports anything like that when loading modules that are compiled into C objects. So we would need to make some kind of extension e.g. wrapper function around `require` or hack the LuaJIT C code or new function in `package.loaders` or ... something?

capr:
@lukego not without any standard-behavior-modifying-magic ;) As you say the `foo/init.lua` convention doesn't work with embedded modules but I guess you could hack the Makefile to generate a temporary `foo.lua` file for each `foo/init.lua` whose contents is `return require'foo.init'` and embed that too. But then all the open files in your editor will be called `init.lua`, ouch.

I know I sound like a broken record, but what's the deal with all the directories anyway? Taxonomic hierarchies (i.e. directories that are not really functional namespaces like `/bin` and `/lib` are in Linux, but cognitive namespaces used solely for classification) only add confusion (eg. `lib` vs `core/lib`) and make you `cd` a lot (or navigate a lot) for no benefit. Modules are APIs and APIs are not hierarchical in nature, the API dependency is a graph. The only legitimate use of directories is for submodules which don't expose an API but serve for breaking down a large implementation, or enable an optional part of the main API, and even then I wouldn't make a directory for less than 20 submodules (`foo_bar.lua` is just fine). That's a long way of saying "don't fight the runtime, work with it".

Also I think the `README.md` convention of github is a terrible idea and promotes deep hierarchies (or cramming unrelated modules into a single doc, take your pick). What's wrong with `foo.md` for each `foo.lua`? Also note that github's flavor of markdown is a small case of vendor lock-in. If you switch to pandoc there might be small incompatibilities that need fixing (no biggie but can be annoying). Alas, pandoc has many cool extensions that won't render on github.

@wingo why did you tried hard to avoid giving your opinion? :)

lukego:
> Modules are APIs and APIs are not hierarchical in nature, the API dependency is a graph. 

The hierarchy is a way of constraining the dependency graph to provide useful structure. Core does not depend on lib, lib does not depend on apps, apps do not depend on programs. The lower layers can be understood completely without reference to the layers above.

This is the same as e.g. a Unix system: the kernel does not depend on libc, libc does not depend on Firefox, etc. This layering seems very typical in my experience i.e. trying to avoid cycles in the dependency graph both between nodes and groups of nodes.

This is practical. For example it allows us to be quite strict in core (refuse duplication, ensure consistency, etc) and a bit more liberal in lib (provide alternative implementations to see which becomes the most popular, introduce new programming styles that may or may not catch on, etc), and downright libertarian in other areas (e.g. allow programs to do whatever they like provided it doesn't mess up the build). Good fences make good neighbours.

wingo:
FWIW, core.lib does import lib.lua.alt_getopt.  Dunno if bug or feature!

lukego:
@wingo I would say that any dependency from core to lib is a bug. The whole `core.lib` module is really a "where should this go, anyway?" miscellaneous dustbin. Refactoring this to be consistent with the layering model would be valuable.

lukego:
@wingo Having said that, there is no big prize for absolute perfect consistency, so it is not an acute problem for a small number of dependencies to break the rules. The main motivation to fix this is the [Fix Broken Windows](http://c2.com/cgi/wiki?FixBrokenWindows) principle.

capr:
The current directory structure doesn't reflect this layering that you're talking about. `lib` and `apps` are both in the root namespace. You have to actually look at the code to see this layering, same as with a flat namespace. From the app dev pov the layering is arbitrary. A dev's question is "show me the spices", not "show me the spice rack". What I think is happening is that those directories are loaded with _policy_ as you explained above, but this policy is not directly visible by users, they only see the directories and wonder what's up with that. So as long as you have to explain the meaning of the layering out-of-band in some documentation, you can have any kind of directory structure or none at all, it's all the same.

capr:
The user wants `require"snabb.spice"` but the layering makes them do `require"snabb.first_row.second_place.spice"`. 

Anyway I think I'm not doing a very good job at explaining this so I'll stop here, but I wanted to at least try. Naming and organizing is complicated business :)

lukego:
@capr So we are talking about two things: the structure of the code - dependency graph - and the presentation to the user - API.

I have described the structure above. This seems sound to me: I don't recognize any objections to this aspect.

I agree that the API is haphazard and that is why I created issue #734. We can work out a better way once @kbara is available to take the lead on that.

Reasonable?

lukego:
@capr Relating the structure vs API idea to Unix too: I see this as like when as a user you call an API function like `clock_gettime()`. Are you calling the kernel? Or the C library? Or the kernel via the C library? You don't really care and the answer may be implementation dependent e.g. the C library may access one time source directly from the CPU and another via the kernel.

lukego:
The filesystem layout is another question. I care much less about that aspect personally.

lukego:
@eugeneia I don't really feel good about releasing `v2016.03` with the documentation style in `lib/` becoming less consistent. I am not sure if we can still fix that. The only solution I see is what both @wingo and @capr have suggested i.e. separate README files in the same directory for each module (`lib/pmu.md`, `lib/ctable.md`, etc). That would avoid the whole mess of moving things into subdirectories and making `require()` still work in the same way. What do you think?

eugeneia:
@lukego I am generally not against `lib/<module>.md` files but I think this leads to a bigger mess and inconsistency than sub-directories and patching require. E.g. without moving files out of sub-directories into `lib/` it would still be inconsistent and I do not think that is what we want. Alternatively we could just move the `README.md` files (e.g. `protocol/README.md -> protocol.md`). The latter I actually see as a viable option, the only downside is that we loose GitHub rendering when browsing the directories.

Is patching `require` off the table? My personal preference would still be to move each library into a sub-directory and patch `require` to load existing files in this order:

```
require("foo") =>
  foo.lua
  foo/foo.lua
```

I do not see right now how this could break things, but making `require` more fancy is obviously a culprit. I generally think we should not rush this decision into 2016.03, is the current situation really that bad?

lukego:
Obsoleted by #790.

PR #765, opened 2016-02-22
lib.bloom_filter: Removed as dead code

lukego:
I believe the Bloom Filter module is not in use. Is that right? If so then we can shrink the size of `lib/` by removing this module (and when somebody needs it in the future they can revert this commit).

This is based on the general operating principle that code in `core/` and `lib/` is a liability rather than an asset i.e. we want to have as little code there as is needed to support current applications. I don't mean any slight to this module which is very well written :).

@alexandergall am I right in thinking this is fine for you and won't impact ALX? (Could be that with `ctable` we can even remove the special purpose mac table some time soon?)

The immediate motivation here is to minimize the size of `lib/` ahead of efforts to make documentation more consistent (which is easier with fewer modules).


alexandergall:
Yes, that was a dead-end (and, yes, replacing my `mac_table` with `ctable` is on the to-do list).

lukego:
Thanks for the quick feedback! Great that we are starting to successfully churn code in `lib` and share code between applications.

wingo:
FWIW @alexandergall  you will probably want to wait until the streaming interface to `lib.ctable` lands before porting; otherwise you may hit the branchy loop problem.  Haven't prepared that yet, probably within a couple weeks though.  FYI :)

wingo:
Merged, thanks.

PR #766, opened 2016-02-22
engine: Set shm path to "app/$name"

lukego:
Set the shared memory path (shm.path) to a private namespace for each app with prefix "app/$name".

This means that apps can create shm objects such as counters and by default these will appear in a local namespace for that app.

Example (for an app update that will be submitted separately):

```
$ sudo ls -1 /var/run/snabb/10773/app/tunnel
drop_bad_cookie
drop_bad_length
drop_bad_local_address
drop_bad_protocol
drop_bad_remote_address
```

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -5,6 +5,7 @@ local lib     = require("core.lib")
 local link    = require("core.link")
 local config  = require("core.config")
 local timer   = require("core.timer")
+local shm     = require("core.shm")
 local counter = require("core.counter")
 local zone    = require("jit.zone")
 local ffi     = require("ffi")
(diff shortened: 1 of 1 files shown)


src/core/app.lua
```diff
@@ -65,6 +66,8 @@ end
 -- Run app:methodname() in protected mode (pcall). If it throws an
 -- error app will be marked as dead and restarted eventually.
 local function with_restart (app, method)
```
eugeneia: I guess we should rename `with_restart` to something like `invoke_app_method`. Nothing critical but this change makes the functions name a misnomer.

eugeneia: @lukego Do you want to revise the name before I merge it or leave it as is?

eugeneia:
Closing in favor of #931.

PR #767, opened 2016-02-22
[wip] Ipv6 tunnel diagnostics

lukego:
Add counters to the `keyed_ipv6_tunnel` app:
- `drop_bad_cookie`
- `drop_bad_length`
- `drop_bad_local_address`
- `drop_bad_protocol`
- `drop_bad_remote_address`

This is to make the reasons for packet drops more transparent.

### DIFF
--- a/src/apps/ipv6/README.md
+++ b/src/apps/ipv6/README.md
@@ -98,3 +98,27 @@ the L2TPv3 header will be overwritten with this value.
 
 *Optional*. Destination MAC as a string. Not required if overwritten by
 an app such as `nd_light`.
+
+### Counters
+
+— Key **drop_bad_length**
+
+Ingress packets dropped due to invalid length (packet too short).
+
+— Key **drop_bad_protocol**
+
+Ingress packets dropped due to unrecognized IPv6 protocol ID.
+
+— Key **drop_bad_cookie**
+
+Ingress packets dropped due to wrong cookie value.
+
+— Key **drop_bad_remote_address**
+
+Ingress packets dropped due to wrong remote IPv6 endpoint address.
+
+— Key **drop_bad_local_address**
+
+Ingress packets dropped due to wrong local IPv6 endpoint address.
+
+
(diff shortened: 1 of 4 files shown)


lukego:
Depends on (includes commits from) #766 to make the engine create the counters locally for each app instance.

eugeneia:
This is super cool obviously. I propose a counterpart to this in `snabbtop`, it should be able to list the counters of a selected app. E.g. so we can do this from a shell:

```
$ snabbtop tunnelXY
drop_bad_cookie         10
drop_bad_length          0
drop_bad_local_address   7
drop_bad_protocol       29
drop_bad_remote_address  3
```

Should be straight forward to implement and I would be willing to tackle this.

lukego:
@eugeneia Cool idea :)

lukego:
Great catch, SnabbBot! Have to think about correctly managing the lifetime of those counter objects to avoid errors and/or memory mapping leaks when recreating app instances.

lukego:
(Added `[wip]` because of bug that needs to be fixed. Possibly overkill since SnabbBot has already marked this change as failing tests.)

lukego:
Thinking about how to manage the lifecycle of counter objects. How about if we leave that to the engine?

For example we could define a new app callback:

```
function myapp:counters ()
   return {'drop_bad_checksum', 'drop_bad_cookie', ...} 
end
```

and then the engine could create these on app init (and delete them on app stop) and make them accessible via:

```
counter.add(self.counter.drop_bad_checksum, 1)
```

Reasonable? I am not sure what a cleaner alternative would be that is not error-prone.

eugeneia:
I would probably use a slot similar to `zone` that can optionally be populated in `new` instead of a dedicated method. E.g.

```
function MyApp:new (arg)
  ...
  local o = {}
  o.counters = {'drop_bad_checksum', 'drop_bad_cookie', ...}
  ...
end
```

Any reason why you chose a method?

PR #768, opened 2016-02-22
[WIP] LISPER program

capr:
The idea of this PR is to start a conversation about what do I need to do in order to make this merge possible. I will hammer this PR until Max is happy with the change list. I can squish the commits if necessary to make this work more intelligible.

TODO(cosmin):

- [ ] re-test lisper with the changes from upstream
- [ ] change current benchmarks from `src/program/lisper/dev-env-perftest` to use the `make benchmarks` environment
- [ ] integrate Max's ipsec code into lisper
- [ ] generate encrypted pcap files and add more benchmarks that include encryption


lukego:
Exciting stuff!

This is the first time we undertake to merge a long-lived application branch onto master.

lukego:
@eugeneia will you be upstream for this? That would be great if so :) Just mark yourself as Assignee.

nnikolaev-virtualopensystems:
A small comment on the git history here. I personally would prefer to see 3 groups of patches (in this order):

1. prepare the landscape for the new feature (i.e. if you need some changes in the core components)
2. implement the new feature
3. adapt the rest of the world to the new feature (add new tests, configurations etc.)

I believe this could be much easier to review now, and to follow the git history in the future.

wingo:
I was once of your opinion, @nnikolaev-virtualopensystems :)  And still my heart goes that way.  But for big apps with a lot of history, I think it can be sufficient to review the changes the branch brings relative to master, understanding that the more "private" files in app/foo and program/foo are more under the control of the "foo" project than Snabb itself.  It would seem to be true that this strategy makes git history less useful, but as long as the merge doesn't bring unwanted changes to non-foo parts of Snabb things should be OK, I think.  Just a thought :)

lukego:
@ nikolay @wingo On reflection that nice structured workflow with topic branches, reviews, etc, is a great benefit of working with upstream and maybe in the future it would make sense for new programs to be part of that workflow from the very beginning. We are still feeling our way around in terms of seeing how to successfully "incubate" new Snabb programs with their own development teams.

lukego:
@capr @eugeneia This PR seems to be stuck in limbo. Let's get it unstuck and merged!

How about quickly putting the branch into [ExtremeNormalForm](http://c2.com/cgi/wiki?ExtremeNormalForm), then merging it upstream, then continuing development with topic branches and pull requests?

Here are some concrete ideas for addressing the "TODO" items above so that they are not merge blockers:

- IPsec support is not finished. Fine. Just delete that code from the branch so that it doesn't block the merge. The code can be reintroduced on a topic branch that can be merged when it is finished.
- More tests and benchmarks are needed. Fine. Just create Issues and/or Pull Requests for these after the code is merged upstream.

Generally the best strategy for new applications seems to be to merge something simple and clean as quickly as possible so that they can benefit from the upstream workflow.

The pitfall I see is to try to do too much before bringing the code upstream e.g. treating every unfinished feature or missing regression test as a blocker. These may be blocking *something* but surely not participation in the upstream community :).

nnikolaev-virtualopensystems:
I agree with @lukego here. Keeping it simple is a good approach here, so one think at a time. But still I prefer to have a cleaner git history :)

lukego:
I notice that @wingo "squashed" the lwaftr history in https://github.com/eugeneia/snabbswitch/commit/8364729b1cc4975e1c03e389de162e7e1bd7f3ed before merging to simplify the history. Could do the same here as the very last thing before merge if we want to shorten the history.

Still: content is king :-)

Looking ahead it seems like new applications would do well to get into the upstream development process earlier e.g. from the very beginning. Even if an application's branch is not actually merging onto master it can still be registered in branches.md and developed with PRs via the main repo.

capr:
I'm trying to identify actionable actions for me:

- extreme normal form -- from my pov. it already is but let's talk about it
- squashing history -- should I do that? I think history is not a very efficient way of understanding the code (it gets more important after the first release)
- removing ipsec support - currently the code is disabled with a simple `if false` until I make it work (. does it help if I remove it?

kbara:
If the ipsec support doesn't work yet, it's probably good to merge something without it. It can be anywhere from a frustrating waste of time to an active security problem to try to use code that doesn't work yet, as a third party who doesn't necessarily realize that at first.

That said, if ipsec is clearly documented as being WIP and is in a program-specific directory, I'm not sure that increasing the integration overhead on you is the right call.

lukego:
@capr thanks for the input. @eugeneia as upstream maintainer can you define the shortest path to merging this code?

lukego:
On reflection the idea of squashing changes seems like quite a can of worms compared with merging.

Merging is simple: branches can bidirectionally merge each other and each time they will pick up any commits that they were missing. If this history seems messy then you need to use more sophisticated query arguments to `git log` e.g. `--first-parent` and so on.

Squashing: How does this really work? Is this a one time thing where the application developer discards the history of their application and then continues working based on the upstream history? Or would the application developer do a dance to maintain their own private history in parallel with the upstream project, akin to `git subtree`?

I don't think I can recommend squashing as an option until I understand it better in this context. I wonder how this is going to be managed with the lwaftr and whether we might end up merging its complete history in the future in order to simplify maintenance.

wingo:
Regarding squashing: in the lwaftr we squashed once, before the initial merge, but will not do so again.  It's merges from here on out.  It is, as you say @lukego, very much the one time in which you throw away history.

wingo:
Also in the lwaftr, any pull requests or other work that were outstanding now need to be rebased onto the new lwaftr branch.  The old branches are dead, long live the new HEAD!

lukego:
@wingo ah thanks for clarifying :). yes that will work and I suppose the consequence is just that if you need to search back into pre-march-2016 history you need to remember to look in another repository for that. not the end of the world. okay, so I am fine with both alternatives again, modulo whatever details I am still failing to understand :)

eugeneia:
@capr In order to merge this I see the following steps:

1. Merge with master to resolve conflicts
2. Merge with `eugeneia:esp2-dev` again (sorry!) to remove obsolete changes to `core/lib*`

I would suggest not squashing any history. I think this might bite us and has not much gain.

*Ideally* I would like all changes outside of `src/program/lisper` to be in individual PRs as @wingo has done it for the lwaftr merge, but this might not be practical anymore in this case.

capr:
@eugeneia Ack

src/apps/ipv6/nd_light.lua
```diff
@@ -77,8 +77,12 @@ local function check_ip_address(ip, desc)
...
 
 function nd_light:new (arg)
-   local conf = arg and config.parse_app_arg(arg) or {}
-   local o = nd_light:superClass().new(self)
+   local arg = arg and config.parse_app_arg(arg) or {}
+   local conf = {}
+	for k,v in pairs(arg) do
+		conf[k] = v
+	end
+	local o = nd_light:superClass().new(self)
```
eugeneia: @capr Not exactly a fan of this code, can you change this to simply copy stuff into `o._config` (preferably without the loop)?

src/core/main.lua
```diff
@@ -33,23 +33,21 @@ function main ()
    zone("startup")
    require "lib.lua.strict"
    initialize()
+   local program
```
eugeneia: @capr I don't think the changes to `core/main` should be in this PR. Are they required for lisper to work?

src/apps/ipv6/nd_light.lua
capr: I removed the tabs and added a comment on why that change is needed: basically if you modify the arg table directly, you won't be able to reuse it and when the app gets reconfigured it crashes or something like that IIRC.

src/core/main.lua
capr: @eugeneia probably not. Thing is I don't remember making those changes myself, maybe they come from a merge. Hm...

eugeneia: @capr I think you can revert them by reverting these commits:

```
commit e5b1a66ed8b73d6256cd5f06fa27ace5f2793931
Merge: 9f74d16 e285507
Author: Cosmin Apreutesei <cosmin.apreutesei@gmail.com>
Date:   Sat Nov 14 13:16:33 2015 +0200

    merge with master

commit 654cc55f03f29fdfa0c20074fb118134c97ae75b
Author: Cosmin Apreutesei <cosmin.apreutesei@gmail.com>
Date:   Sun Oct 25 13:17:41 2015 +0200

    make snabb exe prefer to load modules from filesystem to built-in ones

diff --git a/src/core/main.lua b/src/core/main.lua
index fcde79b..8fdbe3f 100755
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -3,6 +3,7 @@ module(...,package.seeall)
 -- Default to not using any Lua code on the filesystem.
 -- (Can be overridden with -P argument: see below.)
 --package.path = ''
+package.loaders[1], package.loaders[2] = package.loaders[2], package.loaders[1]
 
 local STP = require("lib.lua.StackTracePlus")
 local ffi = require("ffi")
@@ -72,13 +73,13 @@ end
 
 
 -- programname("snabbnfv-1.0") => "snabbnfv"
-function programname (program) 
+function programname (program)
    program = program:gsub("^.*/", "") -- /bin/snabb-1.0 => snabb-1.0
    program = program:gsub("[-.].*$", "") -- snabb-1.0   => snabb
    return program
 end
 -- modulename("nfv-sync-master.2.0") => "program.nfv.nfv_sync_master")
-function modulename (program) 
+function modulename (program)
    program = programname(program)
    return ("program.%s.%s"):format(program, program)
 end
@@ -86,7 +87,7 @@ end
 -- Return all command-line paramters (argv) in an array.
 function parse_command_line ()
    local array = {}
-   for i = 0, C.argc - 1 do 
+   for i = 0, C.argc - 1 do
       table.insert(array, ffi.string(C.argv[i]))
    end
    return array

commit 4c1e8a53c44e8fb3e2c89899e58a0818af13c6bc
Author: Cosmin Apreutesei <cosmin.apreutesei@gmail.com>
Date:   Thu Oct 22 18:14:38 2015 +0300

    lisper (init)

diff --git a/src/core/main.lua b/src/core/main.lua
old mode 100644
new mode 100755
index 9fb08ab..fcde79b
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -2,7 +2,7 @@ module(...,package.seeall)
 
 -- Default to not using any Lua code on the filesystem.
 -- (Can be overridden with -P argument: see below.)
-package.path = ''
+--package.path = ''
 
 local STP = require("lib.lua.StackTracePlus")
 local ffi = require("ffi")
```

capr: I tried to revert that merge but it's given me a lot of conflicts. I wonder, since it's a merge of master, why do I need to revert it since the PR targets master anyway which should have those already. (scratching head)

eugeneia: Maybe just do this (ensure master is up to date):

```
git checkout master src/core/main.lua
git add src/core/main.lua
git commit -m "Revert changes to core.main."
```

eugeneia:
@capr On second thought, @lukego and me are investigating a memory corruption (preumably) somewhere in the ASM code related to the IPsec code. Might be better to strip out the IPsec code from this PR for now so you don't have to wait for us fixing this bug.

lukego:
Ping...

Can we land both IPsec and Lisper upstream now? Or what is blocking?

eugeneia:
@lukego On the IPsec sec I would like some review on #845, but this could as well land seperately. So independent from that, what should be the flow, `ipsec->lisper->max-next` or `ipsec->max-next`+`lisper->max-next`?

eugeneia:
Merged as follows:

```
((ipsec->capr:lisper4)+cleanup)->max-next
```

Issue #761, opened 2016-02-18
CONTRIBUTING.md


eugeneia:
Having a `CONTRIBUTING.md` (this file will be displayed by GitHub when creating a PR for the first time) would help maintainers and contributors by providing an initial “common ground”. This came up a couple of times in previous discussions and I want to create this discussion to figure out what should be communicated in the file. Eventually, when everybody is happy I would open a PR that formulates the results.

I understand that there are plenty of things that could be useful to have written down in a `CONTRIBUTING.md`, but on the other hand I feel like it should be kept concise to reduce reading-overhead. Possible influences for `CONTRIBUTING.md` include:

* https://github.com/SnabbCo/snabbswitch/issues/746 https://github.com/SnabbCo/snabbswitch/issues/725 https://github.com/SnabbCo/snabbswitch/issues/702 - We should probably explain in a few sentences how the merge workflow works roughly, and what is expected of PR authors and what they can expect from maintainers
* https://github.com/SnabbCo/snabbswitch/issues/729 - “Copyright notices belong into `src/COPYRIGHT`; copyright notices in others files are rejected.”
* https://github.com/SnabbCo/snabbswitch/issues/740 - Eventually, we should document coding styles when they emerge and point to that documentation
* Point to the documentation guide

Anything I missed, any new ideas? Feedback from *everybody* is welcome including @kbara, @wingo and @lukego.

kbara:
All of that sounds reasonable to me, at least in light of the way the discussion on copyright notices in source files has gone in this project specifically. We should avoid ever making the coding styles document too heavy-weight.

lukego:
I agree that we want a CONTRIBUTING.md but I think it should be a half-page of helpful information to make contributors understand the process of landing code in Snabb.

Let's please avoid lots of rules and policies that need active policing. My view is that consistency breeds consistency with the code and documentation. If we start sending PRs that improve consistency then this will become a pattern and the results will speak for themselves. A picture says a thousand words, etc.

lukego:
Generally when somebody contributors code to Snabb I think it would be productive to assume that:

1. They have done something valuable that we can benefit from as a community.
2. They are smarter than we are.
3. They have other important things to attend to.

[...]

PR #770, opened 2016-02-22
Add minimal CONTRIBUTING.md

eugeneia:
Adds a `CONTRIBUTING.md` file as discussed in #761. What did I miss, what should I edit out?

Cc @lukego @kbara @wingo

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,20 @@
+Whenever you create a Pull Request (PR), a maintainer will self-assign
+themselves as the *upstream*. The upstream decides if your PR is accepted and
+might require you to amend additional changes before merging. You can expect
+the upstream to communicate clearly if there are any issues preventing your PR
+from being merged, and how they can be fixed. Once merged, the upstream will
+add the `merged` label to your PR.
+
+You can also submit PRs that are not ready to be merged: include the labels
+`[wip]` in the title of PRs that require further work, and `[sketch]` for PRs
+that are not meant to be merged at all.
+
+Please cross reference related issues and PRs extensively by mentioning them in
+descriptions or comments (#XXX), as it helps maintainers to read up on related
+information.
+
+Our [Documentation Guide](https://github.com/SnabbCo/snabbswitch/blob/master/src/doc/documentation-guide.md)
+gives pointers on how to contribute to the project's documentation.
+
+Copyright notices belong into `src/COPYRIGHT`; copyright notices in other files
+will be rejected.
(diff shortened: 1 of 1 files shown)


kbara:
LGTM. There should probably be a way for people submitting PRs to figure out who the maintainers are, perhaps a maintainers.md or similar.

lukego:
Short and sweet!

I would like to tweak the COPYRIGHT section so that we are not encouraging people to register their copyrights but merely telling them how they can do so. For example:

> If you want to record a copyright notice with your contribution then you can optionally do this in the file src/COPYRIGHT. (Copyright notices inside source files will be rejected or removed because that is not our style.)

wingo:
LGTM as well.  (I still am not clear about who is the "upstream" for patches written by @eugeneia, @kbara, or me; I guess in this case with a positive review we should feel free to land them on our own branches?)

kbara:
I've been wondering that as well, though guessing the opposite - that our patches would land on the branches of another maintainer. Both options have quirks. :-)

lukego:
Long term the goal is to bootstrap us into the Linux model where every change has an unambiguous upstream branch based on its subject matter (documentation vs intel drivers vs virtio drivers, etc). The actions we take should be leading us in that direction. But how? :-)

More immediately: if maintainers are getting frustrated that nobody merges their patches then it may be a general process issue with assigning upstreams. In that case it could be counter-productive to create an exception mechanism that only works for maintainers since everybody else will be stuck.

Should we experiment with a different way of assigning upstreams? I would be willing to do this manually based on my own perception of who should review a change and how much they are backlogged if we want to try that.

lukego:
(Or maybe the process today works once we declare that PRs to master should be assigned an upstream to merge them, even when submitted by maintainers.)

lukego:
... blathering a bit ...

Could also define a tag `needs-upstream` that any of us can apply simply to flag when a change risks being reviewed with no clear path forward e.g. immediately after a PR is opened. (This state is perhaps implicit by looking for PRs that are not marked `[sketch]` and have no upstream but maybe explicit is better than implicit, I dunno.)

eugeneia:
> I would like to tweak the COPYRIGHT section so that we are not encouraging people to register their copyrights but merely telling them how they can do so

Good point. Done.

Regarding the question who is upstream for my PRs: I thought it would be either @wingo or @kbara depending on whoever self-assigns. Before we implement a complicated “who-is-upstream” I will propose to rephrase the question to “why should I *not* be upstream?” (e.g. “because I am the PR author”, ”because I don't have time”, “because I know somebody else who should be upstream”, ...). When there is no clear answer to that question we self-assign. If the decision was “wrong” it can be contested at any time.

kbara:
I like that set of heuristics.

kbara:
This looks fine to me. @eugeneia , do you consider it merge-ready?

eugeneia:
@kbara Yes, I am sure we will edit it later on based on feedback from when its prominent but for now I think its a good initial version.

kbara:
Agreed. Merged, thank you.

PR #773, opened 2016-02-23
scripts/lock: Script for locking lab servers

lukego:
Introduce a 'lock' command for synchronizing tests on lab servers.

Users would use `lock` when they are running a command that is likely to conflict with other users e.g. a snabb process that accesses PCI devices or a long running test.

The script can be used for a single command:
```
  lock sudo ./snabb ...
  lock sudo ./mytestscript
```
Or to start a shell that holds the lock (1h idle timeout):
```
  lock
  $
```

The user will see a message like this when the script starts:

```
locking /var/lock/lab.. flock: getting lock took 1.570173 seconds
```

... which could potentially be logged and analyzed to see the pattern of how often commands are blocked and for how long on each server (one day...).

This script would need to be used in a consistent and disciplined way by all users of a given machine to be effective.

### DIFF
--- a/src/scripts/lock
+++ b/src/scripts/lock
@@ -1,15 +1,22 @@
 #!/usr/bin/env bash
 
-if [ "$1" == "--help" -o "$1" == "-h" ]; then
+if [ "$#" == 0 -o "$1" == "--help" -o "$1" == "-h" ]; then
     cat >&2 <<EOF
-Lock a Snabb lab server for mutually-exclusive use.
+Usage: $0 command...
 
-Usage:
-  $0 command...
-    Run a command with the lock held.
-  $0
-    Start an interactive shell with the lock held.
-    Uses TMOUT to set a 1h idle timeout.
+Run a command with the machine-wide Snabb Lab lock held.
+
+Holding this lock entitles the process to use resources such as PCI
+devices and CPU cores that are (informally) reserved for lab tests.
+
+(Please don't access those resources except when using this command to
+hold the lock. That could make your tests interfere with somebody
+else.)
+
+Note: This command sets the TMOUT environment variable so if you start
+an interactive shell such as bash then you will automatically have a
+one-hour idle timeout at the prompt.
+
 EOF
     exit 1
 fi
@@ -18,12 +25,9 @@ lock=/var/lock/lab
 
 echo -n "locking $lock.. ">&2
 (flock --verbose 9 || exit 1
- if [ $# == 0 ]; then
-     echo "starting shell 1h idle timeout"
-     TMOUT=60 ${SHELL} -i
- else
-     "$@"
- fi
+ # Set TMOUT variable in case the command is an interactive Unix shell.
+ # This is a bit magical but hopefully okay for setting an idle timeout.
+ TMOUT=$[60*60] "$@"
 ) 9>/var/lock/lab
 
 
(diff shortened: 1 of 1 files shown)


kbara:
That's an extremely coarse lock - I think a common use case is to only need one or two cards, and not be doing benchmarking.

lukego:
@kbara The best is the enemy of the good. I have been pondering for several years how to make a script like this with finer granularity and never arrived at a solution.

I see this script as potentially boosting the useful capacity of the lab by an order of magnitude. That is because I believe that our servers are actually used less than 10% of the time (while at least one snabb process is running) while they are nominally used nearly 100% of the time (while somebody has their name associated with them).

This is also part of my thinking with acquiring more smaller servers for the lab. If we have 10 machines then taking a machine-wide mutex is not such a big deal compared with e.g. everybody working on chur at the same time.

lukego:
Further thought: We could use `isolcpus` to reserve the majority of the CPU cores and insist that people only use these cores while they are holding the lock. Then background processes (emacs, make, git, nix-env, etc) should not disrupt ongoing test runs.

kbara:
Fair; it seems worth a try.

lukego:
(Can be that this script really belongs in the snabblab/snabblab-nixos repository together with machine init code to create `/var/lock/snabb` with suitable permissions. I could send this script there in a new PR if nobody shoots it down here first.)

kbara:
That does sound like a better place.

eugeneia:
I actually wouldn't mind to have this in `scripts/`. Either way, knowing how long @lukego has been thinking about this I am pleased to think we have found a manageable solution. :-)

In danger of loosing focus and simplicity I have to voice a feature request: Could we add an optional PCI argument? Usually I just want to know if a given PCI address is free, because I want to test functionality and not necessarily performance. So when you want you want to do a CPU intensive  test you could `lock` a whole server (which would fail if I had locked a single PCI adress), but two people could still `lock XYZ` at the same time (which would fail if someone else did `lock` first).

lukego:
@eugeneia I would suggest that we consider this as a future extension if and when it becomes a practical problem that people are waiting too long for their tests to acquire locks. Then we can consider making the locking more sophisticated vs other alternatives like acquiring more hardware.

Goal here is to solve the problem of everybody wanting to work on a separate machine for fear of colliding with each others' tests. I suspect this is costing us most of our effective lab capacity.

lukego:
Note: These days our new servers (lugano model) only have a small number of CPU cores and PCI cards. On these it is not so wasteful to lock a whole machine. This is the environment I have in mind for most people's day-to-day development and testing.

You might say that the new servers have been chosen to make this script easy to write and use. (Or in other words to make the lab resources easier to partition and share.)

The situation is more complex e.g. on a large server with a lot of resources e.g. 24 cores and 20x10G on Grindelwald. However if we would reserve these beasts for testing highly parallel workloads then again it becomes reasonable to use machine-wide locks on them.

@eugeneia I suppose you are thinking in terms of davos and chur that have about 10 different models of network cards installed. These machines could indeed support many people doing independent driver development at the same time with fine-grained locks. Perhaps we will want to improve on this script's workflow for those machines.

For now I would be happy with the modest goal of optimizing utilization of the lugano servers.

eugeneia:
I think its reasonable to incrementally extend this script when the need arises. So should I merge this or is this PR obsoleted by https://github.com/snabblab/snabblab-nixos/issues/20 and can be closed?

lukego:
Closing. I reckon this script belongs in the snabblab repo and hopefully @domenkozar and cherry pick the source from here.

PR #776, opened 2016-02-23
Fixes for 2016.02.1 bugfix release

lukego:
This PR is for @eugeneia and really can be closed after reading.

The intent here is to express my desire to revive the `fixes` branch. This is a branch for collecting fixes that are worth "shipping" on master ahead of the usual monthly release cycle i.e. in parallel with the `next` branch.

Generally I would aim to collect changes on this branch that:

- Have already been accepted e.g. landed on `next`.
- Fix a problem that is really likely to annoy users.
- Generally seem to carry little risk.

The first example I picked is the `#!/usr/bin/env bash` change to make Snabb work better on NixOS machines. I have found myself needing this change several times for new development this month.

My idea is that I would PR this branch to master with the genuine hope that @eugeneia will merge it and treat it as a low-key maintenance release.

On this specific occasion though we are already so close to the next feature release that I don't think we need to really merge this. So please treat this PR more as a request to review the workflow proposed above. (Close at will.)

**Thank you very much** @eugeneia @wingo @kbara for taking over the review and merge of incoming changes! I am really glad to have some bandwidth now for thinking about how to be a better upstream-for-upstreams. Suggestions always welcome :).


eugeneia:
Can do. I suppose this “service” will in practice be limited to one fix release per month, and will occur between the first and third week?

lukego:
I was imaging keeping it unscheduled i.e. we do it when we discover a problem that is likely to cause people pain and we want to fast-track a patch. For example could even be on the day after a release if we discover that we missed a Makefile dependency that will cause a compilation error after `git pull` as nearly happened recently: not much value in keeping people waiting for a fix to such a problem.

lukego:
... but I suppose that could be an exception and maybe it does make sense to try to release generally useful but non-critical fixes around the middle of the cycle.

domenkozar:
Note to myself: this is required for https://github.com/snabblab/snabblab-nixos/issues/13

eugeneia:
@lukego Oh, I actually did not mean to imply to delay bug fix releases but rather that they will occur in the beginning of a release cycle in practice.

Issue #656, opened 2015-11-10
Buffering in the app network: when to hold and when to drop


lukego:
Question: what should be the baseline policy for when an app should hold a packet and when it should drop a packet? This is relevant to the new RawSocket app on #655.

Possibilities:

1. Don't hold packets between breaths. Process or drop but don't wait.
2. Local flow control in apps: don't read from your input queue unless your output queue has capacity.
3. Handle congestion e.g. with an active queue management algorithm.
4. ... other?

lukego:
The idea that makes most immediate sense to me would be for the basic behavior of apps to be (1) and for fancy algorithms like (3) to be implemented as their own apps. Flow control can easily have unintended consequences: better to tune the sizes of the buffers and the amount of work that is accepted in each breath so that packets are dropped when capacity is exceeded or links are out of service.

If we adopted that position we would need to eliminate some flow control from our codebase i.e. cases where an app checks its output capability before reading packets off its input link. Instead the transmit function would detect lack of capacity and drop.

Reasonable people may have a different point of view :).

lukego:
Have been pondering this. My theory is that the simplest and neatest solution is to never apply flow control within the app network. That is: apps will always receive and process all of their available packets and never check whether the output has capacity beforehand. During congestions packets will be dropped at the bottleneck point.

To put this into practice we would need to do a couple of things:

1. Put some "burst" limiting on apps that bring packets into the app network. For example limit each app to introducing 100 packets per breath.
2. Increase the size of our buffers so that they will only overflow due to actual congestion (packet arrival rate exceeding packet processing rate) and not due to burstiness (e.g. `Join` app combining several input links that could reasonably overflow its output link or the next app). In particular we would need to consider increasing the size of `struct link` and NIC transmit descriptor rings to avoid premature overflows.

The benefit of this activity would be more predictable behavior under overload and less code (hallelujah) since apps would always "blindly" receive and transmit and depend on the transmit function to drop packets that are over capacity.

kbara:
I'd definitely like to see this in practice. In an earlier version of the lwaftr, checking whether two output links were full was a non-trivial slow-down, and it didn't actually come with any real benefits.

eugeneia:
I also agree with this rationale. I have come into contact with link drops during #555 and having a clear definition of what it means for a link to drop packets is important to have, imho.

plajjan:
[...]

PR #777, opened 2016-02-23
Measure latency of app.breathe()

wingo:
This commit adds a histogram facility that can record the distribution
of different sampled magnitudes.  In particular, it is useful for
recording distributions of times.  Histograms can be mapped into
/var/run/snabb, where they can be analyzed by a separate process.

This commit also wires up the app.breathe() loop to record latencies for
its breath cycles into a well-known
file (/var/run/snabb/PID/engine/latency), and wires up "snabb top" to do
some basic statistics on this data.

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -236,6 +236,15 @@ function main (options)
       assert(not done, "You can not have both 'duration' and 'done'")
       done = lib.timer(options.duration * 1e9)
    end
+
+   local breathe = breathe
+   if options.measure_latency or options.measure_latency == nil then
+      local histogram = require('lib.histogram')
+      local latency = histogram.create('engine/latency', 1e-6, 1e0)
+      print('hi')
+      breathe = latency:wrap_thunk(breathe, now)
+   end
+
    monotonic_now = C.get_monotonic_time()
    repeat
       breathe()
--- a/src/README.md
+++ b/src/README.md
@@ -198,6 +198,11 @@ following keys are recognized:
    set to boolean values to force or suppress link and app reporting
    individually. By default `engine.main()' will report on links but not
    on apps.
+ * `measure_latency` - By default, the `breathe()` loop is instrumented
+   to record the latency distribution of running the app graph.  This
+   information can be processed by the `snabb top` program.  Passing
+   `measure_latency=false` in the *options* will disable this
+   instrumentation.
  * `no_report` - A boolean value. If `true` no final report will be
    printed.
 
--- a/src/lib/histogram.lua
+++ b/src/lib/histogram.lua
@@ -1,0 +1,252 @@
+-- histogram.lua -- a histogram with logarithmic buckets
+--
+-- API:
+--   histogram.new(min, max) => histogram
+--     Make a new histogram, with buckets covering the range from MIN to MAX.
+--     The range between MIN and MAX will be divided logarithmically.
+--
+--   histogram.create(name, min, max) => histogram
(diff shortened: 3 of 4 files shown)


wingo:
Originally from https://github.com/Igalia/snabbswitch/pull/276.

kbara:
I'd like to nominate @eugeneia to be the reviewer/upstream on this one, if he's willing.

src/core/app.lua
```diff
@@ -236,6 +236,15 @@ function main (options)
       assert(not done, "You can not have both 'duration' and 'done'")
       done = lib.timer(options.duration * 1e9)
    end
+
+   local breathe = breathe
+   if options.measure_latency or options.measure_latency == nil then
+      local histogram = require('lib.histogram')
+      local latency = histogram.create('engine/latency', 1e-6, 1e0)
+      print('hi')
```
eugeneia: I assume this is a debug leftover?

src/core/app.lua
```diff
@@ -236,6 +236,15 @@ function main (options)
       assert(not done, "You can not have both 'duration' and 'done'")
       done = lib.timer(options.duration * 1e9)
    end
+
+   local breathe = breathe
```
eugeneia: Is this not a NOOP?

src/core/app.lua
```diff
@@ -236,6 +236,15 @@ function main (options)
...
+   if options.measure_latency or options.measure_latency == nil then
+      local histogram = require('lib.histogram')
```
eugeneia: From the conservative perspective:

* The `require` should be at the top of the file (I like to use local requires in selftests but I am not sure if we should expand the use of this pattern)
* `core.*` should not require `lib.*`, @wingo pointed out a precedent in https://github.com/SnabbCo/snabbswitch/pull/764#issuecomment-188227602 but I consider it to be a bug
* thus: `lib.histogram` should depend on `core.shm` and be `core.histogram` (and reuse `core.shm.root`, see `core.counter`)

eugeneia:
Super cool feature, love to see `snabb top` extended. Until now I wasn't sure if anybody except me used it. ;-)

@lukego I think the `histogram` module in this PR would be useful in other applications as well and should be a `core.shm` extension (`core.histogram`) similar to `core.counter`. What do you think?

src/core/app.lua
wingo: No :)  It makes a new local variable named `breathe` which is initially bound to the value of the global `breathe` variable.  If we measure latency, then we overwrite our local copy but not the global variable.

src/core/app.lua
wingo: Fine with me, I wasn't sure whether I wanted to propose this code to be "core", but if it seems "core" enough to you it's core enough for me :)  Sure, can re-use core.shm.root too.

src/core/app.lua
wingo: yep, good catch

wingo:
Snabb top has been really useful to us -- for example if we are experiencing packet loss, we need to find where it happens: between which apps, or if not between any apps then on ingress.  It helped us have insight on whether to apply backpressure or not (usually we don't, now, following #656).  It's good stuff :)

I can look at switching to use core.shm.map; good suggestion.  I guess by default core.shm.map should clear the file if a previous PID had it open, though.

src/core/app.lua
eugeneia: Yeah I think it should be in core. :+1: I think to even remember @lukego longing for a histogram shm module once or twice.

PR #755, opened 2016-02-13
[wip] snabbmark: Add preliminary "byteops" benchmark

lukego:
Fun weekend hack...

This branch adds a new command `snabbmark byteops` that measures byte-oriented operations with diverse parameters and produces a comprehensive CSV file. The intention is to systematically measure and compare the performance of operations like memcpy and checksum at different levels of the cache heirarchy, with different alignments, and with different distributions of input sizes. This is in the same spirit as #688 and #744 
.
TLDR: [Full CSV output for 10 runs on lugano-1](https://gist.github.com/anonymous/9e0eb5d65921d882bd8c#file-experiment-log-txt). (45K rows.)

The parameters tested are:

- Function: `memcpy`, `cksum`, `cksumavx2`. (This program seems to "fuzz" a problem in `cksumsse2` that needs to be looked into!)
- Displacement: maximum value for random displacement of source/destination arrays in memory. Uses different values intended to exercise L1/L2/L3/DRAM.
- Distribution of input sizes: 64 bytes, 256 bytes, 1500 bytes, uniform distribution 0..10240, [loguniform](http://www.vosesoftware.com/ModelRiskHelp/index.htm#Distributions/Continuous_distributions/LogUniform_distribution.htm) (proportionally more smaller values) distribution 0..10240 
- Individually, the alignment of src/dst/len values to 1/2/4/8/16/32/48/64 byte boundaries.
This benchmark comprehensively measures the performance of various
byte-oriented operations (copy, checksum) with many different
variations (implementation; distribution of input sizes; alignment of
src/dst/len; etc).

The resulting CSV file includes:

- `nbatch`: Aggregate number of iterations measured using the same parameters.
- `nbytes`: Aggregate bytes for all iterations in the batch.
- `nanos`: Nanoseconds elapsed to process the whole batch.
- `cycles`, `ref_cycles`, `instructions`, `l1-hits`, `l2-hits`, `l3-hits`, `l3-misses`, `branch-misses`: Performance counter report for the batch

Here is some example output:

```
@@ cpu       ;name      ;nbatch    ;nbytes    ;disp      ;lendist   ;lenalign  ;dstalign  ;srcalign  ;nanos     ;cycle     ;refcycle  ;instr     ;l1-hit    ;l2-hit    ;l3-hit    ;l3-miss   ;br-miss   
[...]

Issue #744, opened 2016-02-06
An experimental performance test


lukego:
I did a really basic experiment in the spirit of #688, #741, #692 and following a discussion the Snabb Slack about tuning engine parameters. I am really keen to do performance tuning work based on well-defined repeatable experiments that measure a broad set of applications and workloads. This is a baby step in that direction.

Specifically, this experiment is to see how changing two parameters affects performance as measured by `make benchmarks`. The first parameter is `engine.busywait` with values `true` and `false`. The second parameter is `intel10g.num_descriptors` with values `512`, `1024`, `2048`, `4096`, `16384`, and `32768`. There are 12 distinct combinations of parameters and four benchmarks (basic1, packetblaster-*, snabbnfv-loadgen-dpdk). I run this in a loop for a few hours for a total of 179 tests.

#### Setup

The tests are driven by a shell script that loops through each combination of parameters. Each parameter value is represented by a Git branch that is merged into the test environment, meaning that selecting a parameter value can make an arbitrary code change.

```shell
#!/usr/bin/env bash                                                                                                                   
set -e
n=1
for busy in master busywait; do
  for ndesc in master ndesc/1024 ndesc/2048 ndesc/4096 ndesc/16384 ndesc/32768; do
    echo "run: $n"; n=$[n+1]
    date
    git checkout -B "${busy}+${ndesc}" master
    # grab the '#!/usr/bin/env bash' change for nixos compat                                                                          
    git cherry-pick 1f83b5cb143a41ea8d6bfa71e9315cd9859a59a8 >/dev/null
    git merge -q --no-edit $busy
    git merge -q --no-edit $ndesc
    make >/dev/null
    make benchmarks
  done
done
```

Here is a summary of how the parameter branches look:

```
$ git log --oneline master..busywait
0d04c04 engine: Change default to busywait=true
$ git log --oneline master..ndesc/16384
e95c998 intel10g: Set TX/RX ring size to 16384 packets
```

and here is the script that I used to create the `ndesc/nnn` parameter branches:

```shell
#!/usr/bin/env bash                                                                                                                   
set -e
for ndesc in 1024 2048 4096 8192 16384 32768; do
  git checkout -B ndesc/${ndesc} master
  sed -i -e "s/^num_descriptors = .*/num_descriptors = ${ndesc}/" apps/intel/intel10g.lua
  git add apps/intel/intel10g.lua
  git commit -m "intel10g: Set TX/RX ring size to ${ndesc} packets"
done
[...]

PR #778, opened 2016-02-24
next: Collected changes for the v2016.03 release

lukego:
I am declaring that `next` is now a release candidate for the v2016.03 release. I plan to focus on testing and stability and have it in good shape on March 1st for potential merge by @eugeneia onto master. I leave the `[wip]` tag for the next few days so that we have a chance to test.


lukego:
@eugeneia I want to thoroughly benchmark the `next` branch against the current `master`. The challenge is that I have never been able to get all of the `make benchmarks` running on the NixOS lab servers. The nearest I have come is #744 which runs everything but the iperf tests.

Can you possibly have a look and see if you can get them running on e.g. one of the lugano servers? @domenkozar can help with NixOS questions but I think it is more of a docker/test-env challenge. Probably there is some obvious problem and I simply failed to find the relevant log message.

I wrote up a related issue on snabblab/snabblab-nixos#13.

lukego:
... I suppose that I can run in the fully dockerized environment to avoid this problem. That is probably the best solution for now. I will try it.

eugeneia:
@lukego Outside of docker you will be missing `~/.test_env`, the hacky solution is to copy `/root/.test_env` from the docker image to your `$HOME`. I should talk to @domenkozar  about installing these files for each user by default.

domenkozar:
@eugeneia I'll take a look at this tomorrow

eugeneia:
@domenkozar You would need to add a NixOS package for https://github.com/eugeneia/snabbswitch-docker (see `$(VM_TARBALL)` task in the Makefile) that adds the contents of the created tarball into `/etc/.skel/.test_env/`.

lukego:
Thus spake the wise man:

> A little knowledge is a dangerous thing.

I have taken @sleinen's bait and started learning R for making sense of test results. Bear with me, I am surely doing it all wrong. Here is a first visualization of around 200 test results from `lugano-1`: comparing `master` and `next` by running the DPDK packet forwarding benchmark with a DPDK 1.7 guest and three different packet sizes (64, 128, 256):

r

This seems to be telling us an encouraging message: that 256-byte consistently  hits line rate (~4.4 Mpps), that 128 and 64 byte both exceed this, that variance is low, and that master and next are much the same. If this is indeed the message then this seems like a valuable kind of test to run (and particularly on release candidates).

There is one outlier with next and 64-byte packets: it would be interesting to make that dot into a hyperlink to the relevant log file (which I know can be done in R using SVG output).

I also found the function `aov` (analysis of variance) and that seems to basically work:

```
> summary(aov(dat17.mpps ~ dat17.branch * dat17.packetsize, data=df17))
                               Df Sum Sq Mean Sq   F value Pr(>F)    
dat17.branch                    1   0.00    0.00     3.244 0.0732 .  
dat17.packetsize                2  75.22   37.61 33379.059 <2e-16 ***
dat17.branch:dat17.packetsize   2   0.00    0.00     1.469 0.2328    
Residuals                     195   0.22    0.00                     
---
Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
```

If I am reading this correctly (massive "if") then it is saying that different packet sizes have a major impact on performance (as expected) but different branches do not (as desired i.e. same perf from head and next).

I like the fact that with R this is all scripted and so analysis like this could easily be integrated into the CI.

End braindump. Can be that we look back on this post and chuckle at how I managed to completely misinterpret every aspect of R on my first day trying to use it. But that is then and this is now so I am feeling quite proud for the moment :).

lukego:
(The cool thing about the analysis of variance is that it can tell you both how much effect a variable has and how likely that is to really be caused by random background noise. I am really keen to explore looking for relationships between multiple factors in the context of #755 e.g. how sensitive are memcpy/checksum to alignment and how much does that depend on which level of the cache hierarchy they are hitting.)

lukego:
cc @domenkozar could be that we could have Hydra run these kinds of analysis and make the results available for every PR.

lukego:
Here is a picture of the same test above but with the DPDK 2.1 version of l2fwd:

r

On the one hand it is good for the `v2016.03` release that `master` and `next` have consistent performance. On the other hand we see both lower performance overall (#665) and surprisingly lower results for 128-byte packets compared with 256-byte packets.

I am excited to press ahead with parallel efforts of measurement and optimization. These are both really valuable and mutually beneficial activities. There is a *lot* that can be done in both areas and for more applications. (I am focusing on the NFV application here only because it is the one that has upstream performance tests available: going forward I see this activity as encompassing all programs.)

Onward! One step at a time...

lukego:
@mwiget mentioned seeing lower performance (~10%) when leaving an application idle for a while before starting to process traffic. This is likely some JIT behavior that we need to understand and correct (following @wingo's fine example with #707). The way forward that I see would be to define a test case where the waiting time before processing traffic can be controlled so that we can measure its impact and see which code changes successfully reduce variance.

This is all extremely interesting from my perspective :-)

wingo:
I can believe that.  I feel like sometimes if we start taking too many side traces, LuaJIT doesn't react appropriately, and if you do a jit.flush() things get back to good performance.  Something to look at in the future.

lukego:
Hey wow cool awesome!

I asked R to do an analysis of variance with the branch, the packet size, and the container (DPDK version) and here is what it says:

```
> summary(aov(dat.mpps ~ dat.branch * dat.packetsize * dat.container, data=df))
                                         Df Sum Sq Mean Sq   F value Pr(>F)    
dat.branch                                1   0.01    0.01     1.857  0.174    
dat.packetsize                            2  54.59   27.30  7321.203 <2e-16 ***
dat.container                             1 271.85  271.85 72912.023 <2e-16 ***
dat.branch:dat.packetsize                 2   0.01    0.00     0.734  0.481    
dat.branch:dat.container                  1   0.00    0.00     0.001  0.970    
dat.packetsize:dat.container              2  29.06   14.53  3897.513 <2e-16 ***
dat.branch:dat.packetsize:dat.container   2   0.00    0.00     0.037  0.964    
Residuals                               390   1.45    0.00                     
---
Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
```

and *if* I am interpreting this correctly then it is really providing valuable information:

1. Switching branches (master/next) does not make a difference.
2. Switching packet size does make a difference.
3. Switching containers also makes a difference.
4. The differences caused by the packet size and the container are dependent on each other: packet size impacts each container differently (128-byte is faster with DPDK 1.7 and slower with DPDK 2.1).

This starts to look like *exactly* the thing I was looking for in #755 as a means to search for interactions between data alignment (src/dst/len) and memory hierarchy for different functions. And we already have the data in a CSV file to try this out with!

lukego:
Charging ahead with R and surely making a generous helping of mistakes...

[Post hoc analysis](https://en.wikipedia.org/wiki/Post_hoc_analysis) with [Tukey's test](https://en.wikipedia.org/wiki/Tukey%27s_range_test) is another way to use the ANOVA results. This can be easier to interpret because it tells us how much the average changes between groups of results: the best estimate and the +/- range that can be predicted with 95% confidence.

Here is what Tukey's test says about the effect on Mpps of branch
(`master` vs `next`):

```
> TukeyHSD(aov(mpps ~ branch, data=df), ordered=TRUE)
    95% family-wise confidence level
    factor levels have been ordered

Fit: aov(formula = mpps ~ branch, data = df)

$branch
                   diff        lwr       upr     p adj
master-next 0.008300208 -0.1769718 0.1935722 0.9298626
```

This says that the estimated difference is 0.0083 Mpps (i.e. very small) and we predict with 95% confidence that if we reran the test the effect would be within -0.177 and +0.194 of that value (i.e. not clear that there is a positive or negative difference overall).

Here is the same analysis but with instead comparing the DPDK/L2FWD version in the guest:

```
> TukeyHSD(aov(mpps ~ dpdk, data=df), ordered=TRUE)
    95% family-wise confidence level
    factor levels have been ordered

Fit: aov(formula = mpps ~ dpdk, data = df)

$dpdk
                    diff      lwr     upr p adj
dpdk1.7-dpdk2.1 1.644667 1.554203 1.73513     0
```

This estimates that DPDK 1.7 shows +1.64 Mpps in this dataset and we can say with 95% confidence that the real difference is between +1.55 and +1.74 Mpps.

Cool information!

Just for another angle, maybe pointlessly fancy, we can also look at the effect of branch and DPDK version in combination:

```
> TukeyHSD(aov(mpps ~ branch * dpdk, data=df), ordered=TRUE)
    95% family-wise confidence level
    factor levels have been ordered

Fit: aov(formula = mpps ~ branch * dpdk, data = df)

[... clipped redundant info ...]

$`branch:dpdk`
                                     diff        lwr       upr     p adj
master:dpdk2.1-next:dpdk2.1   0.008071301 -0.1602539 0.1763965 0.9993226
next:dpdk1.7-next:dpdk2.1     1.644434343  1.4748576 1.8140111 0.0000000
master:dpdk1.7-next:dpdk2.1   1.652963458  1.4846382 1.8212887 0.0000000
next:dpdk1.7-master:dpdk2.1   1.636363042  1.4680378 1.8046883 0.0000000
master:dpdk1.7-master:dpdk2.1 1.644892157  1.4778278 1.8119565 0.0000000
master:dpdk1.7-next:dpdk1.7   0.008529115 -0.1597961 0.1768543 0.9992013
```

which makes it look like the effects are independent: same results with same DPDK version, consistent 1.6Mpps boost with DPDK 1.7 vs 2.1.

lukego:
(FYI: My mum tells me that this use and interpretation of Tukey's test is right.)

lukego:
@eugeneia ship'able from my perspective.

PR #780, opened 2016-02-25
lib.equal can compare cdata

wingo:
Cdata values are equal if they have the same size and their bytes are
the same.

### DIFF
--- a/src/core/lib.lua
+++ b/src/core/lib.lua
@@ -20,6 +20,10 @@ function equal (x, y)
          if x[k] == nil then return false end
       end
       return true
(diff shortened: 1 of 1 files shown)


eugeneia:
Cool! Two things come to my mind:

* The documentation of `lib.equal` should probably say what types it supports to make more prominent that this feature exist. E.g. I would not expect it.
* Extending the unit tests of `lib.equal` to test this case would not hurt, even though as is the code looks trivial.

I merged this as is, but if you decide to append anything let me know and I will merge the new commits.

PR #781, opened 2016-02-25
Add packet.shiftright

wingo:
This will be used by the lwAFTR.

### DIFF
--- a/src/core/packet.lua
+++ b/src/core/packet.lua
@@ -72,6 +72,13 @@ function shiftleft (p, bytes)
    p.length = p.length - bytes
 end
 
+-- Move packet data to the right. This leaves length bytes of data
+-- at the beginning of the packet.
+function shiftright (p, bytes)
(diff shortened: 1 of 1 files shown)


src/core/packet.lua
```diff
@@ -72,6 +72,13 @@ function shiftleft (p, bytes)
    p.length = p.length - bytes
 end
 
+-- Move packet data to the right. This leaves length bytes of data
+-- at the beginning of the packet.
+function shiftright (p, bytes)
+   C.memmove(p.data + bytes, p.data, p.length)
```
eugeneia: I think there has to be an `assert` here to check for overflow. See `packet.prepend`.

wingo: Tx for feedback.  I wouldn't want to do this without some deeper perf testing however, and if we know that we're shifting by less than the MTU there shouldn't be a problem; dunno.

Relatedly, shiftleft could underflow p.length.

PR #782, opened 2016-02-25
Add RateLimitedRepeater to basic_apps

wingo:
This app is useful for generating different load levels.  No documentation because the documentation files changed name in next and this is based on master, and the basic_apps documentation doesn't currently specify anything about the API of those applications, so documentation seems to be a moot point currently.

### DIFF
--- a/src/apps/basic/basic_apps.lua
+++ b/src/apps/basic/basic_apps.lua
@@ -5,6 +5,7 @@ local freelist = require("core.freelist")
 local packet = require("core.packet")
 local link = require("core.link")
 local transmit, receive = link.transmit, link.receive
+local clone = packet.clone
 
 
 local ffi = require("ffi")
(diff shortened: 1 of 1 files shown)


eugeneia:
I think this is too fancy for `apps.basic_apps` and belongs under `apps.test.*`. What do you think @lukego?

wingo:
Sure, or `apps.rate_limiter`; dunno.  Or, perhaps we could add a mode to the RateLimiter that doesn't discard packets.

wingo:
Somewhat related to the #656.

eugeneia:
@wingo I like the idea of extending `RateLimiter` and combining it with `Repeater`. It seems like a nice composing solution, but the choice is yours.

mwiget:
@wingo I like the idea as well. While one could chain a RateLimiter after the  Repeater app, it burns avoidable CPU cycles by first creating the packets, then dropping them mid flight. Had to combine RateLimiter also in my crude snabb based lwaftr flow generator in order to have enough CPU cycles to count incoming traffic.

wingo:
Closing this one; #793 is the successor.

PR #784, opened 2016-02-25
[wip] Import lwAFTR into Snabb mainline

wingo:
Depends on:

- [x] https://github.com/SnabbCo/snabbswitch/pull/780
- [x] https://github.com/SnabbCo/snabbswitch/pull/781 (merged into `max-next` but not yet in `next`)
- [x] https://github.com/SnabbCo/snabbswitch/pull/783
- [ ] #793

Still a WIP as it might get rebased/squashed/etc as new commits come in to lwaftr_starfruit and as the dependent PRs are merged.  The `.travis.yml` change definitely needs fixing too; we could keep it for our own purposes but we should switch to SnabbBot and integrate our "end-to-end" tests in a more normal way with the Snabb build/test system.  Cc @kbara @dpino @lukego @eugeneia.


lukego:
@eugeneia do you see any blockers here that we should start looking at in parallel with the work that @wingo has flagged?

eugeneia:
@lukego Nope, as far as I can tell @wingo has kindly opened all changes outside of `apps/lwaftr` and `program/lwaftr` in separate PRs. LGTM.

wingo:
I have rebased this work onto max-next and am going to open a fresh PR so that we only see the changes relative to max-next.

Issue #749, opened 2016-02-10
Support stripping VLAN tags on non-virtualized 82599 NICs


wingo:
When configured with `vmdq=true` and `vlan=1234`, the `intel10g` driver will strip VLANs.  It enables this option by setting the VME bit in RXDCTL:

![screenshot from 2016-02-10 15-34-52](https://cloud.githubusercontent.com/assets/1745903/12950128/ec6be02c-d00b-11e5-849c-2bf5558cba2a.png)

https://github.com/SnabbCo/snabbswitch/blob/master/src/apps/intel/intel10g.lua#L913

However in non-virtualized mode, the VME bit is never set AFAIU and so even if you specify `vlan=1234`, you get a vlan filter but no stripping.  Proposal: we add a `strip_vlan_tag` conf option to the intel10g driver.  Sadly it would have to default to `false` on `vmdq=false` and `true` on `vmdq=true`.  WDYT @lukego?

wingo:
Gaaaah, VLAN *filtering* only works if VMDq is enabled also.  The documentation does not mention this!

wingo:
AFAIU, without VMDq, an 82599 app will accept any incoming MAC address, and will not stamp outgoing packets with the MAC, and won't insert a VLAN tag on outgoing packets either.

wingo:
Apparently app:report() on intel apps also needs vmdq enabled to work: https://github.com/Igalia/snabbswitch/pull/260

lukego:
> Apparently app:report() on intel apps also needs vmdq enabled to work: Igalia#260

Ouch. This is the kind of issue I was alluding to in https://github.com/SnabbCo/snabbswitch/issues/740#issuecomment-182327203. I think the PF/SF/VF factoring of the code is error prone (inheriting methods between objects that have quite different state e.g. available register definitions) and that we should try refactoring in terms of INIT/TX/RX like in the newer I350 driver (see also #687 "I/O 2.0"). See commentary in `intel1g.lua` on master for more on that idea.

The next step that I have planned in that direction is to finish my Intel FM10K (100G) driver make a nice integration between that an Snabb NFV.

PR #788, opened 2016-02-26
Macaddr is not a valid parameter when Intel82599 is SingleFunction

dpino:
MAC address cannot be set if Intel82599 NIC is instantiated as a SingleFunction (no VMDq mode). Method `new_sf(conf)` in `intel10g.app` does not expect a `macaddr` parameter.

On the other hand, there isn't either methods in `M_sf` to set a MAC address AFAIK (the NIC runs in promiscuous mode set at `M_sf:init_receive()`).


wingo:
Related to #749.

For me I am OK merging this as it fixes a bug, but ideally I'd prefer a patch that *adds* support for mac address filtering to the SF interface.  LMK, otherwise I merge this later.

dpino:
@wingo OK, let me try to hack a patch that enables mac address filtering in SF.

dpino:
@wingo The additional patches add support for `promisc` parameter in SF which activates or deactivates promiscuous mode.  When promiscuous mode is off, the NIC filters Rx packets by MAC destination address. That's reflected by printing out the the RXNFGPC (Non-Filtered Good Rx Packets Counter) and RXDGPC (DMA Good Rx Packet Counter). When promiscuous mode is off RXDGPC is zero.

The point is that to properly test non promiscuous mode, it's necessary to craft packets with MAC addresses of the tested NICs.

Another possibility could be to add a method `set_mac` which overrides the actual MAC of the NIC. I think there's a way to restore the MAC from the firmware. Let me know if this more what you had a in mind.

wingo:
Looking good.  Would you mind changing "promisc" to "promiscuous" throughout, and changing set_promiscuous_mode to take a boolean argument instead of having the separate unset function, please.  Thanks :)

lukego:
Background question: Is there an application where it is important to disable promiscuous mode on a single-function NIC?

mwiget:
Is there an API to retrieve the hardware mac of the intel nic via the intel10g app? I couldn't find one. Particularly in single-function app, its desirable to actually use the hardware mac instead of making up one. I know the workaround of loading the ixgbe driver and retrieve the mac via ethtool/ifconfig or read from the filesystem directly, but wonder why there isn't a Snabb function to do so directly.

@lukego : on single-function NIC and disabling promiscuous mode, yes, I do see a use case. It has to do with VLAN filtering. If Snabb is connected to an Ethernet switch port, that port might offer additional VLANs, resulting in Snabb requiring to handle (throw away)  untagged, single or double tagged packets. VMDq helps here.

lukego:
The reason we lack a function for MAC address lookup is historical: we have used VMDq with MAC addresses assigned by QEMU and haven't had a use for the one programmed in EEPROM.

My mind is on the "straight NIC" concept (#801) at the moment so I am viewing every new feature on the `intel10g` driver as a step forward on the one hand (more functionality available) and a step back on the other (further from feature parity between drivers). Still: that is a background consideration since the whole do-it-all-in-software concept is not proven yet.

alexandergall:
A bit late to the party :) @mwiget I've added that code a long time ago: https://github.com/snabbco/snabb/blob/master/src/apps/intel/intel10g.lua#L202

mwiget:
@alexandergall: cool and very useful, thanks!

dpino:
Closing this PR as macaddress is now used in SF mode (commit: 28a3de15)

PR #790, opened 2016-02-29
Restructure documentation in lib/

lukego:
Snabb documentation should be conveniently available as a HTML user manual. This is not the case today. This is an acute problem that we need to fix.

This change is one step in that direction: moving documentation for modules in `lib/` out of source file comments and into per-module README files. There are now files `README.ctable.md`, `README.checksum.md`, and `README.pmu.md` that contain markdown-formatted API descriptions.

This could alternatively have been done by putting all module documentation into a single `lib/README.md` or by creating separate subdirectories like `lib/ctable/README.md`. I resisted because these approaches seem to be motivated by optimizing for Github and I don't think we can afford to be distracted with other mediums until we really have the up-to-date HTML manual always available at users' fingertips.

I am my own upstream on this. I plan to merge onto `next` today unless somebody talks me out of it. I had previously mentioned my concern that we were introducing too much inconsistency into the `lib/` documentation this month and this PR is intended as a correction before release.

cc @eugeneia @wingo @kbara


wingo:
For what it's worth, LGTM

kbara:
I'm ok with it as well.

eugeneia:
@lukego Any reason for the `README.` prefix? I think that is odd, otherwise LGTM.

src/doc/genbook.sh
```diff
@@ -53,7 +53,11 @@ $(cat ../apps/socket/README.md)
 
 # Libraries
 
-$(cat ../lib/README.md)
+$(cat ../lib/README.checksum.md)
```
eugeneia: Oh, the header levels are off in `lib/README*`. Not sure if `pandoc` is smart about this or not.

lukego:
Could remove the `README` prefix. The idea was mostly to keep the docs separate in directory listing and command-line completion. If somebody else says they want that then I will do it before merge :).

PR #793, opened 2016-03-01
Add leaky=false mode to RateLimiter.

wingo:
This adds a configuration argument to RateLimiter, leaky.  If it's true, you have the previous mode of operation where incoming traffic that's above the threshold is dropped.  Otherwise it is left on the incoming ring buffer.

As a functional change, this new parameter defaults to false, meaning that users of RateLimiter should update their configurations.

Also, RateLimiter now provides default values for rate and bucket_capacity.

I have tested this code in the lwaftr test harness (https://github.com/Igalia/snabbswitch/pull/281), and it seems to work correctly.  I have not tested the change to the NFV.  @eugeneia would you mind having a look?  Thanks in advance :)  (This is a follow-on to #782.)

### DIFF
--- a/src/apps/rate_limiter/rate_limiter.lua
+++ b/src/apps/rate_limiter/rate_limiter.lua
@@ -31,6 +31,7 @@ function RateLimiter:new (arg)
    local o =
    {
       rate = conf.rate,
+      leaky = conf.leaky,
       bucket_capacity = conf.bucket_capacity,
       bucket_content = conf.initial_capacity
     }
@@ -43,6 +44,10 @@ function RateLimiter:reset(rate, bucket_capacity, initial_capacity)
    self.rate = rate
    self.bucket_capacity = bucket_capacity
    self.bucket_content = initial_capacity or bucket_capacity
+end
+
+function RateLimiter:set_rate (byte_rate)
+   self.rate = math.max(byte_rate, 0)
 end
 
 -- return statistics snapshot
@@ -70,16 +75,16 @@ function RateLimiter:push ()
    end
 
 
-   while not link.empty(i) and not link.full(o) do
-      local p = link.receive(i)
-      local length = p.length
+   while not link.empty(i) do
+      local length = link.front(i).length
 
       if length <= self.bucket_content then
          self.bucket_content = self.bucket_content - length
-         link.transmit(o, p)
+         link.transmit(o, link.receive(i))
+      elseif self.leaky then
+         packet.free(link.receive(i))
       else
-         -- discard packet
-         packet.free(p)
+         break
       end
    end
 end
@@ -108,7 +113,8 @@ function selftest ()
    -- small value may limit effective rate
 
    local arg = { rate = rate_non_busy_loop,
-                 bucket_capacity = rate_non_busy_loop / 4 }
+                 bucket_capacity = rate_non_busy_loop / 4,
+                 leaky = true }
    config.app(c, "ratelimiter", RateLimiter, arg)
    config.app(c, "sink", basic_apps.Sink)
 
--- a/src/apps/rate_limiter/rate_limiter.lua
+++ b/src/apps/rate_limiter/rate_limiter.lua
@@ -25,8 +25,10 @@ local PACKET_SIZE = 60
 
 function RateLimiter:new (arg)
    local conf = arg and config.parse_app_arg(arg) or {}
-   assert(conf.rate)
-   assert(conf.bucket_capacity)
+   --- By default, limit to 10 Mbps, just to have a default.
+   conf.rate = conf.rate or (10e6 / 8)
+   -- By default, allow for 255 standard packets in the queue.
+   conf.bucket_capacity = conf.bucket_capacity or (255 * 1500)
    conf.initial_capacity = conf.initial_capacity or conf.bucket_capacity
    local o =
    {
--- a/src/apps/rate_limiter/README.md
+++ b/src/apps/rate_limiter/README.md
@@ -7,7 +7,7 @@ the `input` port and transmits conforming packets to the `output` port.
 
 ![RateLimiter](.images/RateLimiter.png)
 
-— Method **RateLimiter:snapshot**
+— Method **RateLimiter:get_stat_snapshot**
 
 Returns throughput statistics in form of a table with the following
 fields:
@@ -15,6 +15,11 @@ fields:
 * `rx` - Number of packets received
 * `tx` - Number of packets transmitted
 * `time` - Current time in nanoseconds
+
+— Method **RateLimiter:set_rate** byte_rate
+
+Configure the rate limiter to refill its empty its bucket at the new
+rate of `byte_rate` bytes per second.
 
 
 ## Configuration
@@ -26,6 +31,14 @@ following keys are defined:
 
 *Required*. Rate in bytes per second to which throughput should be
 limited.
+
+— Key **leaky**
+
+*Optional*.  If true, any incoming traffic that is beyond the configured
+throughput will be dropped.  Otherwise, incoming traffic beyond the
+configured threshold is left on the input link and will be dequeued when
+enough time has passed that the throughput will not be exceeded.
+Defaults to false.
 
 — Key **bucket_capacity**
 
--- a/src/apps/rate_limiter/README.md
+++ b/src/apps/rate_limiter/README.md
@@ -18,8 +18,8 @@ fields:
 
 — Method **RateLimiter:set_rate** byte_rate
 
-Configure the rate limiter to refill its empty its bucket at the new
-rate of `byte_rate` bytes per second.
+Configure the rate limiter to limit throughput to `byte_rate` bytes
+per second.
 
 
 ## Configuration
--- a/src/program/snabbnfv/nfvconfig.lua
+++ b/src/program/snabbnfv/nfvconfig.lua
@@ -48,7 +48,8 @@ function load (file, pciaddr, sockpath)
       if t.tx_police_gbps then
          local TxLimit = name.."_TxLimit"
          local rate = t.tx_police_gbps * 1e9 / 8
-         config.app(c, TxLimit, RateLimiter, {rate = rate, bucket_capacity = rate})
+         config.app(c, TxLimit, RateLimiter,
+                    {rate = rate, bucket_capacity = rate, leaky = true})
          config.link(c, VM_tx.." -> "..TxLimit..".input")
          VM_tx = TxLimit..".output"
       end
@@ -95,7 +96,8 @@ function load (file, pciaddr, sockpath)
       if t.rx_police_gbps then
          local RxLimit = name.."_RxLimit"
          local rate = t.rx_police_gbps * 1e9 / 8
-         config.app(c, RxLimit, RateLimiter, {rate = rate, bucket_capacity = rate})
+         config.app(c, RxLimit, RateLimiter,
+                    {rate = rate, bucket_capacity = rate, leaky = true})
          config.link(c, RxLimit..".output -> "..VM_rx)
          VM_rx = RxLimit..".input"
       end


src/apps/rate_limiter/rate_limiter.lua
```diff
@@ -25,12 +25,15 @@ local PACKET_SIZE = 60
 
 function RateLimiter:new (arg)
    local conf = arg and config.parse_app_arg(arg) or {}
-   assert(conf.rate)
-   assert(conf.bucket_capacity)
+   --- By default, limit to 10 Mbps, just to have a default.
+   conf.rate = conf.rate or (10e6 / 8)
+   -- By default, allow for 255 standard packets in the queue.
+   conf.bucket_capacity = conf.bucket_capacity or (255 * 1500)
```
eugeneia: “Just to have a default” is not an argument for having a default. I don't see a case where a default is useful, and not having an error when accidentally failing to supply `rate` and `bucket_capacity` is dangerous. Please revert this.

wingo: I don't understand what sort of error could occur if the user fails to supply one of these parameters.  For what it's worth, we have been using this bucket capacity to be a fine default in the lwaftr's load tester, across a broad range of throughputs, and there has been no problem.  As for rate, if you use the set_rate() api exclusively, as we do in the load tester via the promise-based transients, it doesn't matter what the initial rate you set is because you will overwrite it from the beginning.

I think the burden is rather on you to assert that either of these is harmful in some way.

wingo: (Sorry, that sounded rude.  Didn't intend that.  My question was, what is the nature of the bug you are worried about when adding these defaults?)

src/apps/rate_limiter/rate_limiter.lua
```diff
@@ -25,12 +25,15 @@ local PACKET_SIZE = 60
...
    conf.initial_capacity = conf.initial_capacity or conf.bucket_capacity
    local o =
    {
       rate = conf.rate,
+      leaky = conf.leaky,
```
eugeneia: I don't like the name “leaky” as it is usually used to describe leaking resources, which is not the case here. I think “drop” is the right name here.

I don't like that the default behavior changes, but on the other hand the fact hat no tests fail make me wonder if `drop=true` is always the wrong choice and if we need it at all. I see two routes here:

1. Show that `drop=false` is consistently superior to `drop=true` for our use cases (hint: the parts of NFV selftest that test `RateLimiter` are currently ridiculously relaxed because the results were... weird). Do not implement `drop` in the first place, instead just change the behavior to `drop=false`.
2. Be conservative and retain the previous default of `drop=true`.

wingo: I think there's no better or worse, both behaviors have their use case.  For processing real traffic, you will want to drop packets.  For processing synthetic traffic from a repeater or a pcap file, you won't.

eugeneia: Actually, looking at `push`:  `leaky` does not change the behavior that much.The only difference is where the packets are dropped (before being pushed on the link or after being popped from the link). I am inclined to think the new behavior is the correct one and that we should drop `leaky` altogether. See also: #656

wingo: Incidentally, boolean configuration values don't default well to "true"; you end up having to interpret a `nil` value to mean the default, when for the user it could be simply that they meant to specify a `nil` value as the boolean value for the parameter (which is a perfectly valid false value of course).  So if we control this via a boolean parameter I think it should be phrased in a way that the default is false :)

src/apps/rate_limiter/rate_limiter.lua
eugeneia: The I don't see why you can't supply the default values in your application. I will not accept this particular change.

wingo: Please provide a reason. Thanks.

wingo: To clarify. If you think my use case is invalid, that is fine, I can go back to use the specialized app. please let me know, thanks.

lukego:
Sorry to be late to the party here...

The big question: Do we rely on flow control (back pressure) in the Snabb app network or not?

If we do then this change makes sense (rate limiting via app composition) but if we don't then #782 makes sense instead (rate limiting as a feature of synthetic packet source apps).

My mental model is that we *don't* want to use back pressure in the app network (#656) and in that case the original PR #782 makes more sense. This PR seems like a good test case for whether there is agreement on that idea or not.

wingo:
Morning :)

I agree that in general we do not want to apply back-pressure.  However for synthetic loads, often you do.  I would agree with the sentiment @lukego that this functionality could belong in a specialized app, but there are a few of these apps already -- the repeater (when joined with the pcap reader), the packet generator in snabb lwaftr, and I think Marcel has one in Snabb VMx.  This functionality should exist in one place.  Whether it's an app or a library is not so important to me, but adding the functionality to RateLimiter does seem to work.

lukego:
@wingo Reasonable. I can accept that making all the upstream apps consistent in their buffering behavior is a separate piece of work (potentially containing some surprises) and until that is done it should not block adding new functionality like this.

I am also skeptical of the defaults. The throttle rate seems naturally like a mandatory parameter to me too. I would also expect the "leaky" behavior to be the default for two reasons: it's more consistent with the plan to avoid back-pressure in apps (i.e. config option is to switch to the unusual behavior) and I also think it is what you normally want in production (where early and deliberate packet drops tend to have better end-to-end behavior than nebulous attempts at link-layer reliability at least in my experience.)

wingo:
OK.  I will change this code to require rate and bucket size even when it's not necessary; a true grumble here but whatever.  (I feel like my use case is not being considered as being as good or as important, and that the existing use case is forcing my code to go through ceremony for no benefit.  The bucket size in particular is a useless knob for synthetic workloads.)

lukego:
FWIW I have no objection to a default bucket size.

I do think that defaults should favor production environments rather than test environments, or else we should have entirely separate apps like in the other PR (which I personally liked better).

I may be overly sensitive because I have spent several years of my life compensating for the side-effects of well-meaning attempts to prevent packet loss on links carrying ethernet traffic, particularly [side-effects](https://www.youtube.com/watch?v=V5cML2lHgvM) introduced by [ARQ](https://en.wikipedia.org/wiki/Automatic_repeat_request) in mobile networks.

(Lots of other people are likely sensitive because of experience with [buffer bloat](https://en.wikipedia.org/wiki/Bufferbloat) too. I am probably over-reacting in this case but I do think one could paraphrase jwz: "A programmer had a problem with packet loss. I know, he said, I will eliminate this by making the layer-1 links reliable. Now he had two problems.")

kbara:
My biases mirror those of @lukego on this, for what it's worth. Buffer bloat reached ridiculous proportions in some systems.

wingo:
Welp, I guess we'll just keep this module local to the lwaftr then; I don't need this on the blocking path to a merge and we can move things around later I guess.

lukego:
@wingo Understandable :).

This discussion has been valuable in terms of coming towards a common understanding of how to handle default parameters and ratification of the "forward-or-drop-but-don't-hold" app behavior.

wingo:
I guess, if you say so ;)  To me the discussion has shown me a few things.  I should never treat a merge as being a no-brainer; that there are many people who can say no, and no one to say yes; that there are many use cases and the ones that are already in git come first; that if one reviewer suggests A and you do it, that another will come later and say that the original thing was better; and so on :)

src/apps/rate_limiter/rate_limiter.lua
eugeneia: I think the use case is perfectly valid and I am completely on board with this PR, just not with the particular change of adding default values to `rate` and `bucket_size` as there are no sensible defaults. You can set default values in your program when instantiating the app without pushing the defaults onto other users of the app. Rejecting the default values does not conflict with your use case as far as I can tell.

lukego:
@wingo Could be that I should have waited with my input for when Max sent the change upstream to me. Then I could have haggled with him about the changes that I want e.g. grumbled about introducing backpressure into the app network and asking him to push back on that in the future.

I chimed in early because I thought you might want the chance to address the feedback earlier since you have previously expressed concern about people changing your code after the initial merge. On the other hand, that is also something that could be addressed with a follow-on PR to next if you wanted to propose undoing that.

Still working all this stuff out. Sorry to meddle.

lukego:
(I am sure that it is a newbie maintainer mistake to over-estimate how interested people will be in your feedback. Forgive me these small missteps :))

eugeneia:
I think there was a misunderstanding and I do not think this PR should be closed.

The issue about buffering/dropping packets is solved to my satisfaction in this PR, and can be reviewed later by a more general initiative about buffering in app networks (I do not like that this PR changes a default for the sake of avoiding a boolean true default value, but this is not a blocker for me).

On the issue of the default rate I stand firmly as explained in the respective sub-thread, but I do not think that this small change conflicts with the PR as a whole.

wingo:
I think @lukego that your input was timely, much better now than later.  To me I see the fundamental conflict as being that Snabb was made with some apps shaping its development, and that extensions proposed during the course of new development might not be as well-suited to the needs of those apps.  In this concrete case, apps which generate synthetic test workloads do not have the same requirements as apps which limit real traffic.  They differ in approaches to backpressure but also as we see in default values (acceptable in one use case, apparently not in the other).

wingo:
Of course it's harder to understand how a new use case could affect existing core code when you can't see how it affects the new code as it's not in git yet.  I think it's a definite best practice to only touch your app/ and program/ directories, and I think we would have avoided a lot of problems in the lwaftr upstreaming if we did that.

lukego:
@wingo Here are the invariants I am ham-fistedly trying to protect:

- Backpressure is not used in the Snabb app network. (That is a legacy misfeature since #656).
- Apps designed for live traffic should not have defaults chosen for test traffic.
- Defaults should only be provided when they are broadly applicable e.g. for the bucket size (who cares) but not for the maximim traffic rate (epsilon chance that the default is correct).

Sorry that I have evidently not done this very well.

lukego:
s/protect/establish/

wingo:
@lukego is backpressure never appropriate for test generators?  Seems silly to include rate-limiting functionality in all test generating apps.  Dunno.

lukego:
> @lukego is backpressure never appropriate for test generators?

That is an important question that can only be answered through discussions like this :).

I would argue that the core simplicity of never using backpressure more than compensates for putting rate limiting into synthetic load generation apps. (End-to-end principle: simplify the core at the expense of the edges.) However, I feel that I need to introduce a PR to pitch that idea in terms of the existing codebase before I push too hard on this with new apps.

wingo:
@lukego ok :)  bear in mind that the future of Snabb includes a RateLimitingPcapReader and a RateLimitingAndRepeatingPcapReader :)

eugeneia:
> bucket size (who cares)

This is a misrepresentation according to the API.

> — Key **bucket_capacity**
> 
> Required. Bucket capacity in bytes. Should be equal or greater than rate. Otherwise the effective rate may be limted.

wingo:
@eugeneia are you sure?  255*1500 suffices to rate-limit to anything between 0 and 10Gb in our test harness.

wingo:
I think rather that a bigger bucket just allows for spikier traffic.  E.g. if you limit to 1Gb and your bucket is 1Gb, you could push 10Gb for 0.1s and 0 for the rest, and you still fit under the limit.

lukego:
@eugeneia Good catch to flag the difference between comments here vs documentation in the rate limiter readme. Certainly the docs should reflect whatever we really think.

Requiring the bucket size as a parameter does seem like a poor user interface to me. My intuition is that, as @wingo says, there will be a sensible default value (either a constant or a function of the rate) and it would be better for the app to know/calculate this than expect the user to do so. The reason it is a required parameter is probably that the person writing the code was not confident of the answer and so punted the issue to the user (reasonable, and certainly better than choosing a bad default, but not optimal).

I would intuitively imagine that around 1 millisecond worth of bucket capacity would be adequate but [citation needed].

lukego:
> bear in mind that the future of Snabb includes a RateLimitingPcapReader and a RateLimitingAndRepeatingPcapReader :)

@wingo :-)

On reflection I see this "leaky vs non-leaky apps" question as like "eager vs lazy functions" in a programming language. Lazy functions do permit some [really lovely](http://worrydream.com/refs/Hughes-WhyFunctionalProgrammingMatters.pdf) formulations of certain problems, but it can also have surprises consequences (e.g. if you try to sum an infinite list), and if you mix both kinds of functions then it's a whole new can of worms about what happens when lazy and eager functions call each other.

So with that analogy it seems like in Snabb we started off without committing to lazy vs eager, then we wrote a bunch of code that is lazy, then we got a vaguely uncomfortable feeling that we should probably pick one way or the other (probably eager), and now on this PR we start to add library code that takes a parameter saying whether to behave eager or lazy.

This seems to open a whole can of worms where users have to understand both evaluation models and the consequences of mixing things together. "How come my router is showing egress packet drops after the latest update?" "Well, your RateLimiter app is now applying backpressure to your Intel10G app, which is causing the NIC to run out of receive buffers, which is causing it to send Ethernet PAUSE frames, which is engaging flow-control on your router port, which is causing the router to drop packets the packets that your rate limiter doesn't want." with addendum: "Be careful not to add a leaky app between the NIC and the rate limiter, like an L2 switch, because then the backpressure will stop there and your router behavior will change again."

Too hard. Better to just pick eager evaluation for its predictability and live with the limitations.

The PR that I am imagining formulating btw is to replace this "lazy" idiom:

```
for i = 1, link.nwritable(l) do ... end
```

with this "eager" alternative:

```
for i = 1, link.burst(l) do ... end
```

and for a simple example `link.burst()` could return a constant e.g. 1/10th of the size of the ring buffer for the link (e.g. bump link capacity up to 1000 elements and burst 100 packets). That way on a breath each app would burst up to 100 packets (by convention), these would all be processed by the app network before the next `pull()`, and no links should overflow because the ring buffers are generously dimensioned.

Taking this a step further we could have a related idiom for any app that generates synthetic load:

```
for i = 1, ratelimiter:clamp(l) do ... end
```

which does not seem so bad. Kind of like how in an eager language every function that returns a list needs to take a length argument saying how many elements you want.

lukego:
... So when `RateLimitedPcapReader` comes along then hopefully we will see a clear pattern and be able to simply add a `rate` parameter to `PcapReader` and every other app that generates packets.

This could also happen asynchronously with follow-on PRs if it starts to bother us that we have too many `RateLimitedFoo` apps in the code.

wingo:
Hi :)  Good morning, I like the musings.  I admit that it can be frustrating (to me) when a PR causes a bunch of musings ;) but I prefer that we share a common understanding of the goals and patterns of the project and this seems an OK way to come to it.

@lukego i think you will find that the concrete approach you are suggesting for rate limiting won't work, because the rate limit is on bytes and not packets.

I think the logical end of your hack-direction is to remove the repeater also, and any other testing-related middlebox.  It could be this is the right thing for production data-planes, but it will make writing test applications much harder -- backpressure is *perfectly fine* in those situations and building app networks is an expressive way to compose a compound network function.  Removing backpressure entirely will leave a hole that should be filled by some pattern :)

lukego:
@wingo So next step for me is to put my code where my mouth is and propose my design in a PR, if I can really formulate it, and see if it gets shot down. Sounds good :).

Musing and bikeshedding are a major potential problem. My mindset here is that the foreground discussion is between the submitter (@wingo) and the upstream (@eugeneia) and everything else is parenthetical background discussion that you can cherry-pick ideas from but otherwise ignore. If this is hard in practice, because the chatter is distracting, then we could declare that general musings are off-topic for incoming PRs and should be raised separately e.g. as their own PRs.

Success here is if we land a steady stream of improvements upstream on master. If musings end up derailing this then they should be curbed somehow e.g. expressing ideas as `[sketch]` PRs rather than comments on other work.

eugeneia:
> Musing and bikeshedding are a major potential problem. My mindset here is that the foreground discussion is between the submitter (@wingo) and the upstream (@eugeneia) and everything else is parenthetical background discussion that you can cherry-pick ideas from but otherwise ignore. If this is hard in practice, because the chatter is distracting, then we could declare that general musings are off-topic for incoming PRs and should be raised separately e.g. as their own PRs.

This! In my opinion this PR should have landed (minus arbitrary defaults, and ideally retaining default behavior). It adds a little feature (back pressure mode in `RateLimiter`) that we don't understand very well but that is proven to be useful and a natural way forward in our orthogonal app ecosystem, and most of all its not intrusive (given it retains the default).

It's perfectly fine to replace this feature with something else once we figured out a general strategy, but right now I think it can't hurt to spell out a requirement (composable RateLimiter) for that strategy in used code. Now that @wingo moved this feature back into a private API it is less visible that it could and should be IMHO.

So I think a good a good etiquette would be to note musings as comments pointing to an issue that will be further discussion of these musings. Also, don't let anyone other than the assignee talk you into closing a PR, what did he review the changes for if they just get closed under his nose? ;-)

PR #795, opened 2016-03-01
Add core.histogram, use it to measure latency of app.breathe()

wingo:
This commit adds a histogram facility that can record the distribution
of different sampled magnitudes.  In particular, it is useful for
recording distributions of times.  Histograms can be mapped into
/var/run/snabb, where they can be analyzed by a separate process.

This commit also wires up the app.breathe() loop to record latencies for
its breath cycles into a well-known
file (/var/run/snabb/PID/engine/latency), and wires up "snabb top" to do
some basic statistics on this data.

A follow-on to #777 rebased onto #794.

### DIFF
--- a/src/core/histogram.lua
+++ b/src/core/histogram.lua
@@ -52,7 +52,7 @@ local log, floor, max, min = math.log, math.floor, math.max, math.min
 local histogram_t = ffi.typeof([[struct {
    double minimum;
    double growth_factor_log;
-   uint64_t count;
+   uint64_t total;
    uint64_t buckets[509];
 }]])
 
@@ -91,7 +91,7 @@ function add(histogram, measurement)
       bucket = max(0, bucket)
       bucket = min(508, bucket)
    end
-   histogram.count = histogram.count + 1
+   histogram.total = histogram.total + 1
    histogram.buckets[bucket] = histogram.buckets[bucket] + 1
 end
 
@@ -123,7 +123,7 @@ function snapshot(a, b)
 end
 
 function clear(histogram)
-   histogram.count = 0
+   histogram.total = 0
    for bucket = 0, 508 do histogram.buckets[bucket] = 0 end
 end
 
@@ -158,8 +158,8 @@ function selftest ()
    h:add(1.5)
    assert(h.buckets[508] == 1)
 
-   assert(h.count == 4)
-   assert(h:snapshot().count == 4)
+   assert(h.total == 4)
+   assert(h:snapshot().total == 4)
    assert(h:snapshot().buckets[508] == 1)
 
    local total = 0
@@ -181,7 +181,7 @@ function selftest ()
    assert(total == 4)
 
    h:clear()
-   assert(h.count == 0)
+   assert(h.total == 0)
    assert(h.buckets[508] == 0)
 
    print("selftest ok")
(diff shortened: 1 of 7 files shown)


wingo:
Ready for review.  Thanks in advance :)

src/core/histogram.lua
```diff
@@ -0,0 +1,189 @@
...
+
+-- Fill a 4096-byte page with buckets.  4096/8 = 512, minus the three
+-- header words means 509 buckets.  The first and last buckets are catch-alls.
+local histogram_t = ffi.typeof([[struct {
+   double minimum;
+   double growth_factor_log;
+   uint64_t count;
+   uint64_t buckets[509];
+}]])
+
```
kbara: I don't like that this has a field called 'count' that refers to the whole type, while the iterator refers to 'count' relative to a particular bucket. Would you be willing to rename the struct field?

kbara:
Why did you choose the order 'index, lo, hi, count' for the iterator, when 'count' (the number of observations in that bucket) is usually the relevant information (or so I'd think)?

src/core/histogram.lua
wingo: Good feedback; certainly!

wingo:
Good q.  The reason is that because the `var_1` variable is threaded through the iteration, it can act as the functional state for the iterator.  If the iterator function doesn't have free variables (which in this case it does, due to "prev"), that means I think that the loop can be better optimized.  But that requires the counter or the iteration state to come first.

I agree that count is more interesting than the index.  Probably just as interesting as lo and hi though.  Given that none of the use cases currently use "index", and "index" can be calculated just by incrementing a variable in the loop, perhaps it's better to do lo, hi, count then and keep more state in the closure; wdyt?  Or count, lo, hi?

kbara:
I'd like count, lo, hi. When I implemented a similar iterator, it didn't have an optional prev parameter - with that parameter, our use cases that actually use the index disappear. (That iterator also had more state in the closure, which seems like a reasonable thing to do here.)

src/core/histogram.lua
kbara: Thanks. It gave me a headache yesterday. ;-)

src/core/histogram.lua
```diff
@@ -0,0 +1,189 @@
...
+local function compute_growth_factor_log(minimum, maximum)
+   assert(minimum > 0)
+   assert(maximum > minimum)
+   -- 507 buckets for precise steps within minimum and maximum, 2 for
+   -- the catch-alls.
+   return log(maximum / minimum) / 507
```
kbara: Any totally compelling reason not to make 509 a named constant? I realize non-4096-byte-pages aren't a huge priority, but I like to avoid exposed magic numbers when feasible. (And things like 507 = 509 - 2 are basic constant folding.)

wingo: I am actually not sure how important it is for this data structure to take precisely a page.  I agree though, will change.

kbara: I'd say not crucial, though it doesn't hurt. Thanks.

kbara:
After those things are addressed, I'm happy with it, though it would be good to hear from @eugeneia about his thoughts on the app/shm/top.lua changes, if any.

wingo:
FWIW the SHM stuff is being reviewed as part of #794.

wingo:
Updated to address feedback.  Thanks for the review!

eugeneia:
LGTM.

PR #797, opened 2016-03-02
[sketch] Prevent typos in option tables

wingo:
Oftentimes we pass around tables of options.  Some of those options are required, some are optional.  Thing is, it becomes hard to add required options over time, as old code might not detect that it needs to change; it's also hard to let old code know when options go away; and most importantly it's easy to make typos.  It's also a bit of irritating effort to provide for optional values which default to false; the idiom `conf.foo = conf.foo or bar` is just too easy to type.

In the lwaftr and in pflua we use helpers to parse options that solve all these problems.  Maybe they're good for snabb itself?  Feedback welcome @lukego @eugeneia :)

I note that @kbara once thought this sort of thing was too much machinery, then it caught a nice bug and she became a believer :)

### DIFF
--- a/src/apps/rate_limiter/rate_limiter.lua
+++ b/src/apps/rate_limiter/rate_limiter.lua
@@ -25,10 +25,12 @@ RateLimiter = {}
 -- Source produces synthetic packets of such size
 local PACKET_SIZE = 60
 
+local required = {'rate', 'bucket_capacity'}
+local optional = {initial_capacity=false}
+
 function RateLimiter:new (arg)
-   local conf = arg and config.parse_app_arg(arg) or {}
-   assert(conf.rate)
-   assert(conf.bucket_capacity)
+   local conf = parameters.parse(config.parse_app_arg(arg or {}),
+                                 required, optional)
    conf.initial_capacity = conf.initial_capacity or conf.bucket_capacity
    local o =
    {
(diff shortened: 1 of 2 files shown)


kbara:
It is too much machinery, it's just the least bad option in these contexts. ;-)

wingo:
Hehe :)  I admit, I really like common lisp's keyword arguments :)

eugeneia:
I would prefer this to be in `core.lib` (and built into `core.config.parse_app_args`). Anything that speaks against that?

Cc @lukego

lukego:
Great! I would really love to have this. I definitely have that uneasy feeling when using table parameters that are not validated for typos etc. Commonly do make mistakes on parameters to `engine.run()` for example.

@eugeneia Good question. To me `core.lib` is a bit of a disorganized mess and I shudder a bit when I read the documentation and see how much stuff we have just thrown in there and thus made part of our API. So I agree that putting this into `core.lib` would be consistent with normal practice but I am not sure how much I want to promote that. Up to you guys.

eugeneia:
@lukego Completely agree that `core.lib` needs a cleanup. As parsing table configurations is a core competency of Snabb (every serious app needs to do it), I would persist on having this in `core.lib` and it being built into `config.parse_app_arg`. The current pattern to parse app configurations is absolutely horrible and this is a chance to end this swiftly. Cleaning up `core.lib` is a separate task.

eugeneia:
I would like to have this, but this PR seems forgotten. I would offer to take over integrating this, given there is demand.

kbara:
I agree; pinging @wingo .

PR #791, opened 2016-02-29
tests require root access

domenkozar:
Otherwise I get `core/shm.lua:111: Unable to create /var/run/snabb: Permission denied`

Issue #740, opened 2016-02-04
Document a coding style for use of object-orientation in Snabb


wingo:
Snabb is a growing project and there is room for many things in it :)  Still, if there are solutions that we agree upon as being particularly snabby, we should document them.

One of these areas is object orientation.  My proposal is to write up a document with the conclusions of this issue and PR it.  However at present it's just a discussion.  Here's a draft:

----

An object in Snabb is data with associated methods.  If you implement a data type in Snabb, we encourage you to set a common Lua metatable on all instances of that type, and to add methods to that metatable exposing the object's functionality.

It's also possible to expose object functionality via functions exported from a module which take an object as an explicit argument.  Doing this or not is up to you.

If a module exports functionality that does not logically operate on an object, such functionality *should* however be exported as a plain function.  Notably, constructors fall into this category.  So for example if you implement a trie:

```lua
local trie = require('lib.trie')

-- The constructor doesn't operate on an object, so it's just a function exported by the module.
local t = trie.new()

-- On the other hand, an "add" method on the trie does operate on an object, so it's a method.
t:add('foo')

-- Maybe there is a trie.add method; if you feel like it.
trie.add(t, 'bar')
```

Thoughts @alexandergall ?

alexandergall:
I agree (and pledge guilty for having abused "class methods" for non-object related functions). I see your point concerning the constructor, too. That could be easily implemented with `lib.lua.class` as well (just a pain to change all the modules that currently use the method style, but that's on me).

wingo:
Yeah probably we should include some implementation notes as well, for example how to provide this api using `lib.lua.class` and so on.  No need for urgent changes, but agreement is a good thing to look for :)

lukego:
Good discussion!

I would like to flag this whole topic as controversial. It is recommending a coding style that is unlike the core modules e.g. packet and link. This needs to be addressed somehow.

The broader discussion could take place under the umbrella of #734 including e.g. should we update the `packet` and `link` modules to follow this example too?

I will advocate for Occam's Razor i.e. the technically simplest and most minimal solution that actually covers our requirements. If that requires procedures, classes, subclasses, class methods, multimethods, predict disaptch, etc, then so be it. But I would presume everything to be unnecessary complexity until proven otherwise.

[...]

PR #798, opened 2016-03-03
Merge max-next/uvilla/1 into next

eugeneia:
Includes: #781 #791 #787 #794 #795

Cc @lukego 

(Is it a good idea to numerate the “merge x into next” PRs? E.g. increment for each merge during a release cycle?)


eugeneia:
Maybe it would be better to give these kind of PRs unique names. E.g. since he next release will be “Uvilla”, something like “Merge max-next/uvilla/1 into next”.

lukego:
Merged, thanks!

I note that sooner or later we need to address some consistency issues in `src/core`:

- Documentation: in source files or in READMEs? Histogram docs are in the source file. Once the manual is online hopefully we can migrate this all into READMEs. (Until then, meh, I hate reading markdown source so I don't complain about having it in the source files instead.)
- OO/metatables: Having `histogram` use one API style and `counter` use another would seem fine to me in `lib` or `apps` but not in `core`. Perhaps in the future we can unify those modules (and perhaps that will be done with some form of OO...). Something to ponder for later.
- Source formatting. The histogram module lacks spaces before function argument lists. This is a trivial detail but it does strike me in the eyes when reading code in `src/core` and I think one day we should do a PR to make this subtree consistent.

Regarding PR names I am not too fussy. I personally think of releases in terms of numbers (`v2016.04`) and not names (`uvilla`) so please don't depend on me remembering which release is called what. I also think it will not always be clear which release I ultimately merge a PR into e.g. if the PR comes late in the release cycle and changes code that I have spent a lot of time testing/benchmarking but most of the time we will probably have a common idea on this.

One way to provide a symbolic name to a PR would be to use a short-lived branch for the PR that ships a ready-to-merge prefix of your integration branch. You could create the branch with `git checkout -b uvilla-1 max-next` and then PR from that. This would also make it possible to continue merging new changes onto `max-next` while you wait for me to merge. (The Linux world does something similar with PRs that say "Please pull branch `foo` at tag `bar`" but I don't think tags and Github PRs would play well, hence the idea to use short-lived branches instead.)

Just ideas.

wingo:
@lukego Seems to me the way in which `histogram` differs from `counter` is that `histogram` offers the OO metatable.  However like `counter` it also provides the functions.  You do not need to use method syntax.  WDYT about that style?  Perhaps these musings should go to #740.

lukego:
@wingo I have separate thoughts on the local level (#740) and the global level (#734).

Locally it all looks fine. The histogram module exports a well-defined API in a style that is familiar from core Lua modules like `string`.

Globally the change does not really fit it with the surrounding code. Core modules like `engine`, `config`, `packet`, `link`, and `counter` all export a function-calling API. I would like these modules to all be consistent, whether than is with a function-calling API, or with an OOP API, or with the dual-mode option.

I would like to pick a canonical style and adopt that for the whole project. Meanwhile I would prefer not to introduce more diverse styles in the subset of our codebase that is already homogenous i.e. `core/`.

PR #802, opened 2016-03-04
Import lightweight 4-over-6 AFTR

wingo:
This PR is to import the lwAFTR into Snabb core.  It is one big ol' commit, on top of max-next, and besides the top-level README.md modification, only adds files to `apps/lwaftr` and `program/lwaftr`.  Please take a look, @eugeneia; thanks!

Cc @kbara @dpino @lukego.


eugeneia:
@lukego I assume this (“lukego merged ...”) was an automated event? Did the branch syncing change? If so, how?

PR #803, opened 2016-03-04
Add lwaftr branch.

wingo:
See #802 for the lwaftr merge.  The current HEAD of the `lwaftr` branch is the same as the current HEAD in that PR.  `lwaftr` will be the development head for new lwaftr development and will not be rebased.  We will merge in from `master` monthly, and will propagate changes back to master through `kbara-next`.

The only big open question to manage for me is how to manage changes to core.  Obviously we will try to avoid them, but let's say we need an urgent fix to something in core and have no time to talk about it with upstream.  What do we do?  Do it in `lwaftr` directly and hope that we'll be able to submit upstream, or revert by the time we go to send to `kbara-next`?  Switch development to a new branch and sort things out later?  To me it's not quite clear.  Suggestions welcome :)

### DIFF
--- a/src/doc/branches.md
+++ b/src/doc/branches.md
@@ -77,3 +77,10 @@ The current state of each branch with respect to master is visible here:
     
     Maintainer: Alexander Gall <gall@switch.ch>
 
+#### lwaftr
+
+    BRANCH: lwaftr git://github.com/Igalia/snabbswitch
+    Lightweight 4-over-6 AFTR application development branch.
+
+    Maintainer: Collectively maintained by lwAFTR application developers.
+    Next hop: kbara-next


lukego:
> let's say we need an urgent fix to something in core and have no time to talk about it with upstream. What do we do?

I would imagine these steps:

1. Create a topic branch for the core change, based on `master`, and PR that to `master`. This makes the change available for discussion / testing / adoption by anybody in the community.
2. Immediately merge that change onto `lwaftr` or any other branches where you need it. This lets you ship the code without waiting for anybody. (Anybody who wants the change quickly can do this.)
3. Asynchronously reach an agreement with upstream about whether the core change is the right thing or whether there is a better solution. This would need to be resolved in order to land the latest `lwaftr` branch on `master`.

This way the code is available to everybody from the beginning, you can ship it without synchronizing with anybody else, and the code lands upstream when everybody is happy with it. Hopefully we will have a regular cadence of merges from `lwaftr` onto `master`, e.g. monthly, and aiming to keep this rhythm will apply gentle pressure on us all to agree on the change in a suitable timeframe.

Reasonable?

Looking ahead, who knows. Could be that the Igalians have multiple lwaftr branches (e.g. latest release vs long-term stable), and @mwiget also has a branch he is maintaining, and open source contributors are sending in changes that need to land on an integration branch before you are ready to merge them into the branches that you ship from, etc. For now having one `lwaftr` branch and aiming to merge that with `master` every month seems like one decent strategy.

The pitfall that I would suggest avoiding is to delay sending the core change upstream e.g. if the first time we see it is within a big `lwaftr` pull request rather than via its own branch and pull request.

Shall I merge this PR directly onto `next`? That seems like a reasonable thing to do with branch registrations to me but I can wait for somebody else to upstream it first if that is preferred.

lukego:
I upstreamed this. Seems reasonable to land branch registrations directly on `next`. (Give a shout if you disagree.)

wingo:
Regarding process, sounds about right to me.  It's a little inconvenient in some ways, because usually you need the lwaftr to test the change, so until now it's been rare that you can make the change directly on master and also test it.  Probably that will change with the merge, as last month's `lwaftr` will usually be in `master` and probably we can test it directly.

I see two social points that we will need to agree on, more or less.  One is that the changes that we make to core, for whatever reason, will always be made with an upstream thought in mind.  Cc @kbara, @dpino.  Otherwise we make them on a side branch that is not meant for merging to `lwaftr`.

The second point is that upstream be willing or interested to take such changes :)  Having an upstream core merge blocking a future `lwaftr` merge is a negative situation, though one we could get into.  I think though that we should generally be OK in this regard, having seen the core changes that the lwaftr merge is bringing in (none, outside the PRs), so probably this is not something to worry about.

I agree with the direct merge to `next` of the branches.md change; thanks!  Onwards and upwards :)

lukego:
Sounds good!

lukego:
You can see `lwaftr` on the Github [branches list](https://github.com/SnabbCo/snabbswitch/branches/all) now. This is automatically mirrored from the branch URL that you registered. This makes it easy to see how far ahead/behind each branch is compared with master. It also means that anybody can `git checkout lwaftr` after cloning the SnabbCo repo.

lukego:
> until now it's been rare that you can make the change directly on master and also test it.

It will really help the upstreaming process to include test coverage with changes to the Snabb core e.g. in `selftest` methods. If we can't be confident that the code works without separately testing it on the `lwaftr` then it will be hard to bring upstream.

wingo:
@lukego Sure.  One of the examples I was thinking of was the `set_rate` / `leaky=false` thing from #793; not sure what our goal should be regarding testing all features of apps.  To me it is reasonable to include those tests as part of the programs using the apps (nfv, lwaftr, etc), though I agree a self-test is better, if more difficult to write :)

PR #805, opened 2016-03-07
Add scripts/snabb_doc.sh

eugeneia:
I added a new cron job “SnabbDoctor” to davos that will render and publish HTML renderings of our documentation.

1. the current release's documentation can be found at http://snabbco.github.io/
2. a tags documentation can be found at `tag/<tag>.html`, e.g. http://snabbco.github.io/tag/v2016.03.html
3. Pull Requests have a new status that points to the documentation for that PR at `sha1/<sha1>.html`

### DIFF
--- a/src/scripts/snabb_doc.sh
+++ b/src/scripts/snabb_doc.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+
+# Snabb Doctor: continuously publish Snabb manual as single HTML file for tags,
+# master and GitHub Pull Requests using GitHub pages. Depends on `jq'
+# (http://stedolan.github.io/jq/).
+
+export SNABBDOCDIR=${SNABBDOCDIR:-"/tmp/snabb_doc"}
+export REPO=${REPO:-"SnabbCo/snabbswitch"}
+export DOCREPO=${DOCREPO:-"snabbco/snabbco.github.io"}
+export DOCURL=${DOCURL:-"https://snabbco.github.io"}
+export JQ=${JQ:-$(which jq)}
+
+function init {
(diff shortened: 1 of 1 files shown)


eugeneia:
Or so I thought, bear with me to figure out why SnabbDoctor failed on this PR.

wingo:
Very nice work!  LGTM; I guess you are the upstream for this though?

lukego:
@eugeneia Great work!

Can you please make a plan with @domenkozar to migrate all of this infrastructure into the nixops-managed snabblab universe? I would like to phase out running infrastructure on single-point-of-failure servers and with out-of-tree scripts e.g. cron jobs. This is fragile and we are lucky to have gotten away with it for as long as we have.

I would then see what solution you guys cook up and adopt that for the little script that mirrors registered branches on the SnabbCo repo :).

domenkozar:
@eugeneia @lukego I can add this cronjob to https://github.com/snabblab/snabblab-nixos/ repository and configure it to run on `eiger`. 

@eugeneia 

What credentials are you using for github?
Can you paste me the cronjob line currently on davos?

Thanks1

lukego:
@domenkozar @eugeneia I believe we have a dependency here on snabblab/snabblab-nixos#13. That issue is closed but I am not sure if we have verified it working.

I also think we need to be conscious of which Github credentials we install on a shared machine like `eiger`. This should not include write access to the Snabb repository for example. We should either lock down the Github permissions to a safe level or host them somewhere more locked-down.

Relatedly: is it a good or a bad idea to make Hydra execute the builds and tests and for SnabbBot to check the results and report to Github? (Why?)

Issue #808, opened 2016-03-08
PMU musings


lukego:
The PMU is a powerful piece of hardware. We are only scratching the surface of how it can be used. This issue is meant to remind us of this fact and perhaps inspire the next step in our usage.

The idea I came up with is simple: to run the `snabbmark basic1` benchmark many times, each time with a different set of PMU events, and each time with `/packet` counts reported.

Script:

```lua
-- Specify groups of PMU events by name pattern.
-- The CPU can track up to 8 events at a time.
-- See lib/pmu_cpu.lua for available counters.
local event_groups = {
   {"^l2_rqsts"},
   {"^l2_trans"},
   {"^dtlb_store"},
   {"^dtlb_load"},
   {"^offcore_requests"},
   {"^itlb_misses"},
   {"^cycle_activity"},
   {"^uops_executed_port"},
   {"^mem_trans_retired[.]load_latency_gt_"},
   {"^mem_uops_retired"},
   {"^mem_load_uops_retired"}
}

local snabbmark = require("program.snabbmark.snabbmark")
local pmu = require("lib.pmu")

local npackets = 10e6

-- Run basic1 benchmark separately for each group of events.
for _, events in ipairs(event_groups) do
   engine.configure(config.new())
   pmu.profile(
      function () snabbmark.basic1(npackets) end,
      events,
      {packet = npackets})
end
```

Output:

```

Processed 10.2 million packets in 0.35 seconds (rate: 29.2 Mpps).
EVENT                                             TOTAL     /packet
cycles                                    1,235,445,846     123.545
ref_cycles                                            0       0.000
instructions                              2,733,223,244     273.322
l2_rqsts.all_code_rd                          1,073,783       0.107
l2_rqsts.all_demand_data_rd                  51,378,621       5.138
l2_rqsts.all_demand_miss                      1,534,342       0.153
l2_rqsts.all_demand_references               97,109,913       9.711
l2_rqsts.all_pf                              21,668,114       2.167
l2_rqsts.all_rfo                             44,657,508       4.466
l2_rqsts.code_rd_hit                            770,365       0.077
l2_rqsts.code_rd_miss                           291,776       0.029
packet                                       10,000,000       1.000
load: time: 0.34s  fps: 58,662,064 fpGbps: 32.381 fpb: 510 bpp: 60   sleep: 0   us

Processed 10.0 million packets in 0.34 seconds (rate: 29.5 Mpps).
EVENT                                             TOTAL     /packet
[...]

PR #809, opened 2016-03-08
snabbmark: Add PMU support to mp-ring benchmark

lukego:
Following on from #808...

The command-line argment -e / --events already existed but now it is actually implemented.

The report is only printed by child process #0 based on the assumption that all children would tell the same story.

Just for a taste here is a little script to run the benchmark with different events and process counts:

```
#!/usr/bin/env bash
for processes in 1 2 3; do
    for events in 'l2_rqsts' 'l2_trans' 'mem_trans_retired[.]load_latency_gt_' 'cycle_activity'; do
        echo "@@@@ processes $processes"
        echo "@@@@ events $events"
        sudo ./snabb snabbmark mp-ring --packets 10e6 --processes "$processes" --events "$events"
    done
done
```

See this [gist](https://gist.github.com/lukego/180e39ec4379c45fb194) for the raw output of the first run.

### DIFF
--- a/src/program/snabbmark/mp.lua
+++ b/src/program/snabbmark/mp.lua
@@ -66,7 +66,6 @@ function mp_ring (args)
    local counters = ffi.cast("uint64_t *",
                              memory.dma_alloc(c.processes*ffi.sizeof("uint64_t")))
    -- Start child processes
-   if c.pmuevents then error("PMU support NYI") end
    local start = C.get_time_ns()
    for i = 0, c.processes-1 do
       if S.fork() == 0 then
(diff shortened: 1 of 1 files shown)


lukego:
cc @xrme @kbara this is hopefully a useful development tool.

lukego:
Last touch for now: I created a [new gist](https://gist.github.com/lukego/062a4c7e4c02cfde92f4) that runs the same test but in a different order, so that the results for 1/2/3 processes are shown together i.e. easier to look at how certain counters changed between the 1 process and 2 process scenarios.

PR #810, opened 2016-03-08
added mention to lisper branch in branches.md

capr:


lukego:
Thanks! Merged to next.

Requests for next time:

- Please write a brief description for the pull request explaining the reason for the change (above it says *No description provided.*)
- Please read the diff to make sure every change is intentional (some unexplained whitespace changes are included on this PR).
- Please don't be annoyed by upstreaming nitpicks :-)

capr:
got it. not annoyed at all, this is useful.

re whitespace -- is there a policy on this? my editor removes them automatically when I save but I see that some people's editors don't do that.

capr:
(in fact my editor normalizes the files in many ways: convert tabs-to-spaces or viceversa, single empty line at eof, no whitespace)

lukego:
I don't see a problem with the specific whitespace change here. I also think it is reasonable to make whitespace cleanups while you are already touching the file anyway. (Git even has a `-w` option to ignore leading/trailing whitespace when looking for diffs.)

I did have to check a couple of things though: this is a `pre` section of a markdown file and those are indicated by consecutive lines that have four leading spaces (seems to render okay) and I am also processing this file with `awk` to mirror the branches (still understands the file okay).

Generally though the important thing is for contributors to know exactly what changes they are sending and to account for them with a well-written commit message. The first thing upstream will do is read the diff and compare it with the commit message and life is easier if everything is consistent. If there are surprises then it looks like the PR has been prepared carelessly.

I would recommend that all contributors read the diff before sending a Pull Request and make sure that every change is intentional and accounted for in the commit message and/or PR text. If something has slipped in by accident, e.g. a debug printout or source reformatting by the editor, then this should either be explained in the commit message or removed with a rebase before sending the PR.

Make sense?

capr:
Makes sense.

PR #813, opened 2016-03-09
[wip] multi-process ring buffers

xrme:
A problem with the current `struct link` and multiple processes is that the `read` and `write` members of the struct are read and modified every single time a packet is queued or dequeued.

http://dl.acm.org/citation.cfm?doid=1345206.1345215 (FastForward for efficient pipeline parallelism: a cache-optimized concurrent lock-free queue ) talks about this, and proposes improvements.

The easy improvement is to make the producer and consumer able to operate more independently by putting some control information into the data.  This is academic-speak for using a null pointer data value to indicate whether there is data available in the ring buffer.  This way, producers only have to look at `write` and consumers only have to look at `read`, and if we arrange for these members to be on separate cache lines, we can avoid the cache coherence transactions that result from modifying both values on every use of the link. That's what I try to implement here.

I haven't yet tried to implement the temporal slipping described in the paper.  Again, this appears to be academic-speak for priming the pipeline so that producers and consumers aren't accessing the same part of the packet data.  In other words, we want to make it so that the producer and consumer are accessing different cache lines.

Another paper (http://www.cse.cuhk.edu.hk/~pclee/www/pubs/ipdps10.pdf, A Lock-Free, Cache-Efficient Multi-Core Synchronization Mechanism for Line-Rate Network Traffic Monitoring) describes MCRingBuffer.  The main idea here is to improve cache locality by the use of local copies of the read and write pointers, and to accumulate a batch of several insertions/removals before updating the shared copy.

### DIFF
--- a/src/core/link.h
+++ b/src/core/link.h
@@ -11,14 +11,10 @@ struct link {
   struct {
     struct counter *txbytes, *rxbytes, *txpackets, *rxpackets, *txdrop;
   } stats;
-  char pad[24];
   // Two cursors:
   //   read:  the next element to be read
   //   write: the next element to be written
-  int read;
-  char read_pad[60];
-  int write;
-  char write_pad[60];
+  int read, write;
   // Index (into the Lua app.active_apps array) of the app that
   // receives from this link.
   int receiving_app;
--- a/src/core/link.h
+++ b/src/core/link.h
@@ -11,10 +11,14 @@ struct link {
   struct {
     struct counter *txbytes, *rxbytes, *txpackets, *rxpackets, *txdrop;
   } stats;
+  char pad[24];
   // Two cursors:
   //   read:  the next element to be read
   //   write: the next element to be written
-  int read, write;
+  int read;
+  char read_pad[60];
+  int write;
+  char write_pad[60];
   // Index (into the Lua app.active_apps array) of the app that
   // receives from this link.
   int receiving_app;
--- a/src/core/link.lua
+++ b/src/core/link.lua
@@ -49,6 +49,27 @@ function receive (r)
    return p
 end
 
+function maybe_rx (r)
+   local p = r.packets[r.read]
+   if p == nil then
+      return nil
+   else
+      r.packets[r.read] = nil
+      r.read = band(r.read + 1, size - 1)
+      return p
+   end
+end
+
+function maybe_tx (r, p)
+   if r.packets[r.write] ~= nil then
+      return nil
+   else
+      r.packets[r.write] = p
+      r.write = band(r.write + 1, size - 1)
+      return p
+   end
+end
+
 function front (r)
    return (r.read ~= r.write) and r.packets[r.read] or nil
 end
--- a/src/program/snabbmark/mp.lua
+++ b/src/program/snabbmark/mp.lua
@@ -47,7 +47,7 @@ function mp_ring (args)
    function opt.r (arg) c.readbytes = tonumber(arg) end
    function opt.w (arg) c.writebytes = tonumber(arg) end
    function opt.h (arg) usage() end
-   local leftover = lib.dogetopt(args, opt, "hn:p:b:e:r:w:", long_opts)
+   local leftover = lib.dogetopt(args, opt, "hm:n:p:b:e:r:w:", long_opts)
    if #leftover > 0 then usage () end
    -- Print summary of configuration
    print("Benchmark configuration:")
@@ -103,6 +103,28 @@ function mp_ring (args)
                -- Sync registers with memory
                core.lib.compiler_barrier()
             end
+	 elseif c.mode == "ff" then
+            local acc = ffi.new("uint8_t[1]")
+	    while counters[i] < c.packets do
+	       local p = link.maybe_rx(input)
+	       if p ~= nil then
+                  -- Read some packet data
+                  for j = 0, c.readbytes do
+                     acc[0] = acc[0] + p.data[j]
+                  end
+                  -- Write some packet data
+                  for j = 0, c.writebytes do
+                     p.data[j] = i
+                  end
+
+		  local sent = nil
+		  repeat
+		     sent = link.maybe_tx(output, p)
+		  until sent ~= nil
+		  counters[i] = counters[i] + 1
+	       end
+	       core.lib.compiler_barrier()
+	    end
          else
             print("mode not recognized: " .. c.mode)
             os.exit(1)
--- a/src/program/snabbmark/mp.lua
+++ b/src/program/snabbmark/mp.lua
@@ -9,11 +9,204 @@ local S = require("syscall")
 local pmu = require("lib.pmu")
 local lib = require("core.lib")
 
+local shm = require("core.shm")
+local band = require("bit").band
+
 -- Ring benchmark:
 -- See how quickly packets are cycled through a ring of processes.
 --
 -- Each process copies packets from its input to its output link. Each
 -- link is populated with an initial "burst" of packets.
+
+local cache_line_size = 64
+
+-- ff (FastForward) link
+
+-- John Giacomoni, Tipp Moseley, and Manish Vachharajani. 2008.
+-- FastForward for efficient pipeline parallelism:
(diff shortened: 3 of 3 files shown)


xrme:
```
lugano-4:src rme$ sudo ./snabb snabbmark mp-ring --mode ff --processes 2
Benchmark configuration:
       burst: 100
  writebytes: 0
   processes: 2
   readbytes: 0
     packets: 100000000
        mode: ff
   pmuevents: false
   7.51 Mpps ring throughput per process
```
Still really bad.  But it may not be a valid result anyway, since using `--processes` of more than 2 doesn't seem to work with mode `ff`.

lukego:
Interesting!

I ran this with PMU enabled. (This should also work on `lugano-4` if you give it a reboot, see snabblab/snabblab-nixos#23.)

```
[luke@lugano-1:~/git/snabbswitch/src]$ make && sudo ./snabb snabbmark mp-ring --mode ff --processes 2 --events mem_load_uops_l3_hit_retired
make: 'snabb' is up to date.
Benchmark configuration:
       burst: 100
  writebytes: 0
   processes: 2
   readbytes: 0
     packets: 100000000
        mode: ff
   pmuevents: mem_load_uops_l3_hit_retired
   8.30 Mpps ring throughput per process
PMU report for child #0:
EVENT                                             TOTAL     /packet
cycles                                   42,160,813,384     421.608
ref_cycles                                            0       0.000
instructions                              3,543,164,335      35.432
mem_load_uops_l3_hit_retired.xsnp_hit         2,788,733       0.028
mem_load_uops_l3_hit_retired.xsnp_hitm       88,033,111       0.880
mem_load_uops_l3_hit_retired.xsnp_miss           82,130       0.001
mem_load_uops_l3_hit_retired.xsnp_none           26,158       0.000
packet                                      100,000,000       1.000
```

Couple of interesting things jump out here compared with https://github.com/SnabbCo/snabbswitch/issues/808#issuecomment-194239854:

1. L3 cache access seems to have decreased but only slightly. Here we see around 0.91 total L3 hits per packet compared with 0.98.
2. Instructions per packet decreased from 60 to 35. This seems like a good thing: it should improve performance once the cache bottlenecks are out of the way. (And make the assembler code in traces easier to read, too.) This is probably from skipping the counters? I have wanted to minimize those for a long time but haven't been able to land that in the past, see #570. (Counters also became a pointer-chasing indirection with the whole `counter_t` business that bothers me quite a lot, too.)

I would have expected a bigger effect from moving the `read` and `write` cursors onto separate cache lines since that seems like classic "false sharing" that would ping-pong the cache line back and forth between cores that need to update it. Given this data it looks like that is not the biggest problem though.

I suppose another "L3 hitm" situation is accessing the pointers in the link ring. One core is writing pointers and another is reading them: second core will have to request them from the first core. Here are a few dubious and overlapping ideas that may be useful for reducing L3 synchronization there:

- Write a whole cache line worth of pointers before bumping the `write` pointer. On x86-64 I suppose this means sending bursts of 8 packets (64 byte cache line / 64 bit pointer = 8 pointers per cache line).
- Write to the link using cache-bypassing stores (`MOVNT`). I am curious about the behavior of these on Haswell: will the effect be to bypass L1/L2 and update L3 (if already populated there)?
- Drop cache lines with `CLFLUSH` once written. Curious about this on Haswell too: can we drop from L1/L2 but keep in L3?

Generally I think a good way forward would be to take all the code on this PR and make it a local part of the `ff` mode of snabbmark. That is, the `ff` would use its own private definition of `struct link` instead of the standard one. This style would allow us to merge a bunch of different implementations into `mp-ring` and continue to tweak and evaluate them over time. Even if we find the optimal design now the answer might change with the next generation of CPU and we may have a new bright idea that we want to try so I think it is important to keep and maintain these benchmarks over the long term.

One more note about `mp-ring`: it could be a problem that the packets are always being accessed in the same order that they were allocated. This could provide an unrealistic consistent "stride" for the memory prefetcher. If we start to see worryingly *high* performance then we may want to mix that up a bit e.g. shuffle the freelist before allocating the packets.

src/core/link.h
```diff
@@ -11,10 +11,14 @@ struct link {
...
   } stats;
+  char pad[24];
   // Two cursors:
   //   read:  the next element to be read
   //   write: the next element to be written
-  int read, write;
+  int read;
+  char read_pad[60];
+  int write;
+  char write_pad[60];
```
lukego: I'm thinking that 56 bytes of padding should be enough here when `int` is 8 bytes. However we should probably be more explicit and declare `read` and `write` as either `int64_t` or `int32_t`. (I have a vague recollection that Mike Pall discourages the use of `uint64_t` with FFI for some reason -- must understand that properly one day.)

Generally LuaJIT does understand declarations like `__aligned__((64))` but I think C only supports that on structs/arrays.

lukego:
One more idea:

There is a potential cache "conflict miss" due to the `struct link` objects being allocated with `shm` that puts everything on a 4096-byte boundary. This seems very likely to be a problem in future benchmarks where one core is accessing a lot of links. One quick solution would be to allocate the links with `memory.dma_alloc()` that happens to allocate shared memory with less extreme alignment (128 bytes, a fairly arbitrary number).

See #566 for where we discovered this conflict-miss problem and made a complicated workaround for counter objects.

lukego:
Yet one more idea:

I wonder whether the CPU "store buffer" introduces meaningful latency on writes i.e. whether the CPU will hold onto the stores for a while before pushing them out to the cache. If so it seems conceivable that one core could keep using a stale value (e.g. of the write cursor position) for a while. This might be avoided if the writer would execute an `SFENCE` instruction to flush the store into cache.

I don't immediately see how this could account for the low performance when taking the PMU reports into account. If cores were continuing to access old data then I would expect to see both high cycles per packet while waiting for the stores to propagate and also high instructions per packet while spinning around in a busy-loop waiting for data to arrive. The PMU shows the former but not the latter.

lukego:
One more reflection is that we should really celebrate the progress on this PR :-). This seems to have wiped 200 cycles per packet off the benchmark. If we do that twice more then we are in good shape :-).

@xrme I mark myself as upstream for this PR. I have one request which is to localize the changes into the benchmark e.g. to add a commit that moves the global changes to `struct link` into a local definition of `struct fflink` for this benchmark. Then I will merge this branch onto the `multiproc` integration branch where we can keep hammering away until we are satisfied and ready to push multiprocess support upstream.

(To me it seems quite luxurious to have integration branches where features can be developed at their own pace and following their own priorities. This fancy distributed Git workflow definitely takes some getting used to be I do enjoy the advantages it offers.)

xrme:
I put back the changes to `struct link` and made everything local to mp.lua.

I added another type of link based on MCRingBuffer (paper cited above).

I think I'll try assembly versions of some rx/tx functions, just so I can be sure of what's going on instruction-wise.  Although fewer instructions are always better (the fastest work is the work that you don't do, after all), we know that the performance challenge here is probably a question of memory/cache efficiency.

lukego:
Thanks! I have merged this onto the `multiproc` branch now. Github has automatically closed this PR now that the changes are merged but you can just reopen it if you want to push more changes from the same branch.

I removed a few lines of dead code in order to proactively minimize the diff between `multiproc` and `master` (particularly the new functions in `link.lua` that we don't need after your new refactoring). You might want to `git fetch origin && git merge origin/multiproc` to pull your changes into this branch.

Ordinarily people are left in peace while they hack on their `[wip]` branches but since the multiprocessing code is interesting to so many people I want to pull it upstream onto `multiproc` in case other people want to contribute. Sorry for the extra fetching and merging that this entails.

lukego:
@xrme Interesting that these variants all have similar performance. I wonder what the bottleneck is? I agree that dropping down to assembler is a good direction so that we will have every detail on the table.

I am not confident of my interpretation of the PMU events yet. I think we need to check the documentation and errata for the events much more carefully. For example some counters are precise (give an accurate per-packet value) while other are sampling (more nebulous to interpret). I have a hard time reconciling the L2 and L3 event counts that I see on some benchmarks and it is possible that they are actually irreconcilable. Probably there is one set of counters that is way more interesting than all the others but I am not certain which it is yet.

PR #815, opened 2016-03-10
[sketch] lib.pmu: Sketch of support for load-latency measurement

lukego:
Counters for measuring the latency of memory loads, like:

    mem_trans_retired.load_latency_gt_4
    mem_trans_retired.load_latency_gt_8
    mem_trans_retired.load_latency_gt_nnn

require the `nnn` cycle count to be programmed into a special MSR. This change does that programming.

The code does not actually work yet due to a hardware dependency on PEBS. See comment in source code.

See also #808.


lukego:
Closing and marking as `[sketch]`. I originally thought this was a simple isolated change and only later realized that it depends on PEBS.

PR #820, opened 2016-03-11
lib.ipsec.esp: IPsec/ESP/AES-128-GCM

eugeneia:
This adds `lib.ipsec.esp`, an implementation of ESP using AES-128 in Galois/Counter-mode. The following features are not yet implemented:

* Auditing is *not* implemented, see the “Auditing” section of RFC 4303 for details: https://tools.ietf.org/html/rfc4303#section-4

* Anti-replay protection for packets within `window_size` on the receiver side is *not* implemented, see `esp_v6_decrypt:check_seq_no`.

* Recovery from synchronisation loss is is *not* implemented, see Appendix 3: “Handling Loss of Synchronization due to Significant Packet Loss” of RFC 4303 for details: https://tools.ietf.org/html/rfc4303#page-42

Furthermore, there is a lot of room for improvement with regards to performance. Currently the library can sustain 1 Gbit/s of traffic (tested on lugano-1).

I would like to see this merged as an initial version that can be further iterated upon, hoping that having this in the tree for all eyes to see will stimulate peer review and uncover further bugs etc.


lukego:
Great to have this in upstream review!

> I would like to see this merged as an initial version that can be further iterated upon, hoping that having this in the tree for all eyes to see will stimulate peer review and uncover further bugs etc.

Counter-proposal: How about if this code lives on this feature branch until it is adopted by a real application e.g. lisper? The upstream flow could be `esp -> lisper -> max-next -> next -> master` for example.

I feel like it would be a nice invariant if all of the upstream code is really in use by at least one application. Application developers are probably in a better position to declare when the code works than passive reviewers. It might surprise people if they find code upstream that has not really been used and tested, too.

eugeneia:
I agree with the reasoning. Should we add a named `esp` branch on upstream then, that—presumably—I will maintain?

lukego:
> I agree with the reasoning. Should we add a named esp branch on upstream then, that—presumably—I will maintain?

Sounds like a good idea. Much like I have done with Multiproc and Mellanox branches.

Going forward it will be interesting to see whether in practice these turn out to be short-lived integration branches that are used to reach "1.0" on a feature or whether they continue to be maintained as the first-hop for new changes to those subsystems.

I do start to think that it will work well to have our collection of fairly general-purposes integration branches close to master (`next`, `max-next`, `wingo-next`, `kbara-next`) and then for more topic-specific branches to spring up downstream (`esp`, `multiproc`, `mellanox`, etc). Then people maintaining `-next` branches could all have a reasonably consistent overall view of the project while people working on more specific stuff downstream could focus more on their own subject matter. So, I am glad that you all pushed back on me when I suggested that we start with topic-specific subsystem branches directly downstream from `next` :).

lukego:
I think having the `esp` branch upstream will help for collaboration too. For example, I am keen to experiment with switching from aligned- to unaligned- instructions in the assembler code and I would prefer to send such a PR to the SnabbCo repo (with target: esp) rather than onto your private fork that is not being watched by the community and SnabbBot and so on.

PR #823, opened 2016-03-14
[sketch] src/Makefile: Flatten Lua module names (no directory)

lukego:
This change puts all Lua module names at the top-level and ignores their directory prefixes. So instead of `core.link` we have `link` and instead of `program.snabbnfv.traffic.traffic` we have `traffic` (which would perhaps be renamed to `snabbnfv_traffic`).

This is based on a suggestion by @capr on #734.

The basic idea here is to make the directory structure a pure implementation detail that is not visible to users. This is motivated by the idea that users don't care about the internal structure and implementors prefer more flexibility about how code is organized.

One way this could simplify the organization of code would be to simply have:

```
src/lwaftr
```

instead of

```
src/program/lwaftr
src/apps/lwaftr
```

i.e. an obvious place to collect all of the code that is specific to one application independent of the content of that code (apps, libraries, command-line programs, etc).

One way it would simplify the API is by hiding the distinction between `core` and `lib` that is perhaps more of an implementation detail anyway.

To make this change actually work we would need to:
- Give every module a unique name.
- Update `require()` lines.
- Invent new rules for finding programs (and perhaps apps).

Some thought would be needed to decide how to do that and how this change relates to the broader topic of #734.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -116,12 +116,12 @@ $(OBJDIR) bin testlog:
 
 $(LUAOBJ): obj/%_lua.o: %.lua Makefile | $(OBJDIR)
 	$(E) "LUA       $@"
-	$(Q) luajit -bg -n $(subst /,.,$*) $< $@
+	$(Q) luajit -bg $< $@
 
 $(PFLUAOBJ): obj/%_lua.o: ../lib/pflua/src/%.lua Makefile
 	$(E) "LUA       $@"
 	$(Q) mkdir -p $(dir $@)
-	$(Q) luajit -bg -n $(subst /,.,$*) $< $@
+	$(Q) luajit -bg $< $@
 
 $(COBJ): obj/%_c.o: %.c $(CHDR) Makefile | $(OBJDIR)
 	$(E) "C         $@"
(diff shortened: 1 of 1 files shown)


lukego:
cc @capr @eugeneia @wingo @kbara. This is intended as a radioactively controversial PR to stimulate discussion :).

wingo:
I like some things out of this but many others are not clear :)  To me you would only want to do this change if we also flattened the directory structure.  But to the extent that directory structure provides value, to me we should reflect that in module names, so that we have just one story there and not two.  WDYT?

I tend to think that core vs lib is a distinction that provides value in the way by selecting some code to set the direction for the project, style-wise, and allow "contrib" code to come in even if it's different in style.  At the same time "core vs lib" has costs too that this would do away with.  Anyway: how would you make this distinction in the absence of hierarchy?

kbara:
I'd rather not see src/lwaftr  - as the number of projects in the ecosystem that ship with Snabb increase, src/ would get incredibly cluttered. I like having things built on Snabb under directories like src/program/lwaftr .

mwiget:
While I agree with @kbara about not cluttering src/, I struggle with having part of a solution under src/apps and src/program. For my snabbvmx work i currently use src/program/snabbvmx and src/apps/nh_fwd, though the latter also contains generator.lua v4v6.lua apps, which is less than ideal. Ideally I'd like to have everything under program/snabbvmx and only move things into src/apps when an app really becomes useful to other programs.

kbara:
I agree that the app/ and program/ distinction is questionable.

eugeneia:
I think the `app/` and `program/` distinction is great because it gives an incentive to write better code. E.g. stuff in `program/` is really just glue code almost by definition, and not the actual core functionality. So when you are writing code outside `program/` you can be a clean-freak, and inside `program/` you can “just make it work”.

I can see why a flat namespace would make sense, but I also agree with @wingo that having “two stories” seems off. On the other hand I really like our directory hierarchy because it structures my work as a reviewer.

Thesis: a framework—even an arbitrary one—will stimulate collaboration, because it provides a shared identity.

PR #822, opened 2016-03-12
doc/core-data-structures.md: Describe packet/link/app

lukego:
This new document explains the three core Snabb data structure: packet, link, and app. For now this is a stand-alone markdown file but it could be incorporated into the introduction in the manual.

This is imported from lukego/blog#11. See also #818.

PR #824, opened 2016-03-14
Documentation: New structure and table of contents

lukego:
This change creates a new table-of-contents for the Snabb manual. This reorganizes the existing material and puts a structure in place for adding new material -- in particular to include the floating markdown files in the repository and to capture the outcome of discussions on Github.

The main changes are:
- New Introduction (making place for #818 and #822)
- Rebalanced and reordered the API and Apps sections
- New chapter for Programs.
- New chapter for the development process.
- New chapter for policies.

See [snabbswitch-ce38aa12.pdf](https://www.dropbox.com/s/5v4ej6hn83kr9vj/snabbswitch-ce38aa12.pdf?dl=0) for the intended output. This will need to be reconciled with the HTML edition separately e.g. if section names are too long or something like that.

Question is: does this represent the content that we want in the manual and the way that we want to organize it? If so then we could set about filling it all in.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,51 +1,110 @@
-# Introduction
+### `packet`: Packet data in memory
+   
+A *packet* is a data structure describing one of the network packets that
+is currently being processed. The packet is used to explicitly manage the
+life cycle of the packet. Packets are explicitly allocated and freed by
+using `packet.allocate` and `packet.free`. When a packet is received
+using `link.receive` its ownership is acquired by the calling app. The
+app must then ensure to either transfer the packet ownership to another
+app by calling `link.transmit` on the packet or free the packet using
+`packet.free`. Apps may only use packets they own, e.g. packets that have
+not been transmitted or freed. The number of allocatable packets is
+limited by the size of the underlying "freelist", e.g. a pool of unused
+packet objects from and to which packets are allocated and freed.
 
-*Snabb Switch* is an extensible, virtualized, Ethernet networking
-toolkit.  With Snabb Switch you can implement networking applications
-using the *Lua language*. Snabb Switch includes all the tools you need to
-quickly realize your network designs and its really fast too!
-Furthermore, Snabb Switch is extensible and encourages you to grow the
-ecosystem to match your requirements.
+— Function **packet.allocate**
 
-![Architecture](.images/Architecture.png)
+Returns a new empty packet. An an error is raised if there are no packets
+left on the freelist.
 
-The Snabb Switch Core forms a runtime environment (*engine*) which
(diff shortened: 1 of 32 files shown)


eugeneia:
What is the “Zen of Snabb”? That's a section I wouldn't expect in a manual, or at least not be able to guess what it is about. Otherwise, I like the structure.

I disagree with use formatting (fixed width fonts) in headers, but that is a detail. More importantly we have to define the module naming scheme that this ToC uses before we use it. E.g. currently there is no `source` anywhere in the code.

lukego:
> What is the “Zen of Snabb”?

I was imagining a short summary of Snabbiness. I was particularly thinking of the summary on [lwaftr slide 3](https://wingolog.org/pub/fosdem-2016-lwaftr-slides.pdf) "Tao of Snabb":

- Simple > Complex
- Small > Large
- Commodity > Proprietary

which seems to put a lot of design decisions in Snabb into context e.g. why our `struct packet` is quite simple and bare whereas other projects like DPDK have adopted all the complex Intel-proprietary offload metadata and so on. Our one is simpler, smaller, and more commodity - and still practical.

Tao of DPDK seems to be `Benchmarks > Everything` and this seems like an important contrast to me.

This is also along the lines of the [Zen of Python](https://www.python.org/dev/peps/pep-0020/) which looks like a joke but is/was actually very useful documentation for newbies to orient themselves with the community.

> More importantly we have to define the module naming scheme that this ToC uses

Good point! Yes, one useful thing about documentation is that it can drive code changes when some things are just too damn ugly to document. The many and varied ways that we have for referring to classes of apps is one such thing. I suspect that we should create a new namespace for apps so that they can have simple string names like `source` that are less dependent on the module that implements them. (And that we should be consistent about how we write names e.g. `lowercase_with_underscores`.)

> I disagree with use formatting (fixed width fonts) in headers, but that is a detail.

Yeah. Just now I am optimizing for the PDF manual and this needs to be reconciled with the HTML edition before merge onto master.

lukego:
@eugeneia Thanks for the fast feedback and thumbs-up on the basic structure btw! I can create an integration branch to collect the contributions that I want to make to documentation this month and then push that upstream to you.

lukego:
The immediate upstream for this change is `pdf-manual`. Further review can be done when that branch is sent upstream to `documentation-fixes`.

kbara:
I definitely like the idea of having a section on Snabbiness, but I might call it "The design principles of Snabb" or "The philosophy of Snabb" or similar. I've been seeing more and more complaints these days about technical projects using religious/philosophical/cultural terms like zen, dao/tao, etc, in contexts that are at most tangentially related.

PR #827, opened 2016-03-15
Optimization documents.

kbara:
Here is some documentation that I think is important for Snabb: how to tune systems and avoid performance pitfalls for deployment.

I'm still getting used to the Snabb next workflow, so merged it to kbara-next, but Luke suggested that I should ask @eugeneia to review it more before it hits next. @eugeneia , patches welcome, and sorry for procedural inconvenience!


README.PERFORMANCE.DEPLOY.md
```diff
@@ -0,0 +1,262 @@
...
+    -chardev socket,id=char0,path=./xe0.socket,server \
+    -netdev type=vhost-user,id=net0,chardev=char0 \
+    -device virtio-net-pci,netdev=net0,mac=02:cf:69:15:0b:00   \
+    -object memory-backend-file,id=mem,size=8000M,mem-path=/hugetlbfs,share=on \
+    -numa node,memdev=mem -mem-prealloc \
+    -realtime mlock=on  \
+    /path/to/img
+```
+
+The allocated memory must match the memory-backend-file size (example shows 8GB). While qemu will fail to boot if there isn't enough hugepages allocated, it is recommended to have some spare and note that the pages are split amongst the NUMA nodes. Check the paragraph on NUMA in this  document.
```
nnikolaev-virtualopensystems: As of the current state of LuaJIT and the way that the guest RAM is mapped to the host process (i.e. SnabbSwitch) when using vhost-user, I am not sure that Snabb will be able to map 8G of RAM. Maybe the example could stay with 512M jus to be on the safe side.

nnikolaev-virtualopensystems: Also it is worth noting here that the memory of the QEMU guest can not be smaller than the HugePage size and to be divisible to that size. Examples:

 - with 2MB huge page size - one can have 512MB guest RAM but can't have 511MB
 - with 1GB huge page size - one can have 1GB or 2GB RAM but can't have neither 512MB nor 1536MB (i.e. 1.5GB).

mwiget: Interesting remark about not handling > 8GB. We've ran many tests with 16G and it works (packets flying). In case this is a limit and we just got lucky, that would be very important to note. Real VNF's tend to demand lots of memory.

lukego:
I am not aware of any reason that LuaJIT would limit the amount of guest memory that can be mapped. Citation needed, Nikolay?

LuaJIT does limit the amount of address space that can be used for garbage-collected Lua objects but that would not apply here.

nnikolaev-virtualopensystems:
@lukego yep - you are right, the limit is about the Lua objects, and here we are accessing everything through directly, so no GC involved. Removing my comment, sory for the confusion.
BTW nice to hear that 16GB VMs were tested and work ;)

mwiget:
On 16G VMs, yes, they seem to work but getting used to in tools like 'htop', because the virtual memory reported tends to be really large for Snabb and highlighted in red, even on servers with ample amount of RAM.

eugeneia:
@mwiget @kbara Huge amount of information about some “dark arts”, great! While skimming over it I got the impression that it is in a WIP state. I would still merge it given its usefulness but I request that the file (not sure if we really need the example PDF in the repository?) be moved to `src/doc/performance-tuning.md`.

Besides fixing the TODOs, in the future I think some sections need to be proofread and clarified, e.g.:

> The default of 512 seems too small, based on load test at IMIX line rate tests against lwaftr, 1024 or 2048 gave equally good results.

This does not really clearly say whats what, IMHO.

Also please use text formatting pervasively, e.g.

> Defined in src/core/app.lua and enabled before calling engine.main() via

vs

> Defined in `src/core/app.lua` and enabled before calling `engine.main()` via

kbara:
I fully agree it's a bit WIP, but I also figure that WIP is much better than totally undocumented - like you said, it's in a realm that's still ```dark arts``` to most of us, where the info available is mainly scattered webpages with less than half of this document's tips. Would you be willing to do the formatting cleanup you mentioned? I'd greatly appreciate it. :-)

kbara:
The 512/1024/2048 numbers are ring buffer sizes. Tuning that to deal with latency variations without getting into the realm of buffer bloat is an ongoing dark art investigation, btw.

mwiget:
Fully agree on WIP. Its basically a structured brain dump. Agree we don't need the PDF in there. Maybe there is a good link to an external web site about it? The graphical view in PDF (or PNG) is very useful.  
On ring buffer size: indeed, tuning is needed, but mostly one should refrain from thinking larger numbers are better. I've seen oscillating effects when setting it too high (> 2048).

kbara:
I've done the file rename suggested by @eugeneia . I haven't removed the PDF, as the diagram is useful and I don't think a PNG is a huge amount better (though it would be a small improvement). I also haven't dealt with the general WIP state of this PR. @eugeneia , concretely, what do you still consider blockers to merging this that need to be resolved?

mwiget:
Providing the lstopo file in png format:

lstopo



Providing the same

kbara:
Thanks, @mwiget . @eugeneia would you like to see the png in the src/doc/ directory, replacing the current PDF?

lukego:
Thanks for PNG. That will likely make this easier to include in the manual alongside the other documentation, eventually.

eugeneia:
@kbara I say swap out the PDF with the PNG version and then I will merge it. Deal?

kbara:
@eugeneia awesome, thanks; I've done that. Deal.

PR #829, opened 2016-03-17
Markdown on demand

lukego:
This branch creates all documentation build products on-demand instead of expecting them to be checked into the repository. This simplifies the source code maintenance.

1. Image files created by `ditaa` are now created by `Makefile` and not stored in the repository.
2. The `.src.md` concept is gone. Now all markdown files are in "source" format, with ascii art, and the Makefile preprocesses them all while building the manual.

The manuals also now build in `obj/doc/` instead of directly in `doc/`.

The benefit that I see here is simplicity for maintaining the sources. Now developers updating documentation don't have to remember to `make markdown` and decide which parts of its output are relevant to check in. I also like the fact that relatively exotic build dependencies like `ditaa` would only be needed for building the manual and not for routine development.

The main downside is that the docs will not show pretty pictures when you browse them on the Github web interface. Bummer. I don't feel we can afford this level of complexity for nice-to-have features of secondary documentation formats though.


src/Makefile
```diff
@@ -184,20 +183,20 @@ obj/jit_vmprof.o: extra/vmprof.c | $(OBJDIR)
 
 book: doc/snabbswitch.pdf doc/snabbswitch.html doc/snabbswitch.epub
 
-doc/snabbswitch.md: markdown Makefile doc/genbook.sh
+obj/doc/snabbswitch.markdown: snabb markdown Makefile doc/genbook.sh
```
eugeneia: @lukego Why does `snabbswitch.markdown` require `snabb` though?

lukego:
@eugeneia Thanks for catching the `snabb` dependency. That was me being lazy about making sure the directories in `obj` existed when they were needed for writing markdown output. Pushed commit 64ee8af with a better solution.

src/Makefile
```diff
@@ -153,7 +153,7 @@ $(JITOBJS): obj/jit_%.o: ../lib/luajit/src/jit/%.lua $(OBJDIR)
 
 $(RMOBJS): obj/%: %
 	$(E) "MARKDOWN  $@"
-	$(Q) scripts/process-markdown $< > $@
+	$(Q) scripts/process-markdown $< $@
```
eugeneia: @lukego This might have broken things? It worked before but now I get:
```
MARKDOWN  obj/program/lwaftr/doc/README.ndp.md
make: *** [obj/program/lwaftr/doc/README.ndp.md] Error 1
```

PR #831, opened 2016-03-17
src/nix/manual.nix: Nix expression to build the manual

lukego:
Add a Nix expression to build the Snabb manual.

You can run it like this:

```
nix-build src/nix/manual.nix
```

and it will build the manual in `./result`.

The main differences compared with simply writing `make book` are:

- Nix installs all of the build dependencies (pandoc, latex, ditaa, fonts, etc).
- Automatically publishes the manual when built via the Hydra CI ([example](https://hydra.snabb.co/build/7)).

Depends on #830 to actually build successfully.

cc @domenkozar could you ack whether this looks like a reasonable starting point for in-tree nix expressions?

### DIFF
--- a/src/nix/manual.nix
+++ b/src/nix/manual.nix
@@ -1,0 +1,34 @@
+# nix expression to build the Snabb manual.
+#
+# Run like this:
+#   nix-build /path/to/manual.nix
+# ... and the files are produced in ./result/
+
+with import <nixpkgs> {};
+
+stdenv.mkDerivation rec {
+  name = "snabb-manual";
+  src = ./../../.;
+
+  buildInputs = [ ditaa pandoc git
+   (texlive.combine {
+      inherit (texlive) scheme-small luatex luatexbase sectsty titlesec cprotect bigfoot titling droid;
+    })
+  ];
+
+  buildPhase = ''
+    # needed for font cache
+    export TEXMFCACHE=`pwd`
+
+    make book -C src
+  '';
+
+  installPhase = ''
+    mkdir -p $out/share/doc
+    cp src/doc/snabbswitch.* $out/share/doc
+    # Give manual to Hydra
+    mkdir -p $out/nix-support
+    echo "doc-pdf manual $out/share/doc/snabbswitch.pdf" \
+      >> $out/nix-support/hydra-build-products;
+  '';
+}
--- a/src/doc/default.nix
+++ b/src/doc/default.nix
@@ -17,7 +17,8 @@ stdenv.mkDerivation rec {
   ];
 
   patchPhase = ''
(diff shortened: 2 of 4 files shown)


src/nix/manual.nix
```diff
@@ -0,0 +1,34 @@
...
+
+stdenv.mkDerivation rec {
+  name = "snabb-manual";
+  src = ./../../.;
+
+  buildInputs = [ ditaa pandoc git
+   (texlive.combine {
+      inherit (texlive) scheme-small luatex luatexbase sectsty titlesec cprotect bigfoot titling droid;
+    })
+  ];
```
domenkozar: I'd rather add here:

```
postPatch = ''
  patchShebangs .
'';

This will make sure all shebangs are converted from any form into full path to bash executable. Then we can keep `/usr/bin/env bash` shebangs that should work on all systems including NixOS.

domenkozar:
One nitpick. I'd also rather place the expression into `src/doc/default`, then it can be just invoked in that directory with `nix-build`.

eugeneia:
I will merge this as soon as @domenkozar acks this.

By the way, I have a nix expression for SnabbBot lying around, should I PR that as well?

lukego:
@domenkozar thanks for feedback, incorporated with latest commits.

> By the way, I have a nix expression for SnabbBot lying around, should I PR that as well?

@eugeneia Yes!

I think we need a subsystem maintainer for reviewing and merging nix expressions. @domenkozar are you up for that? There is a hot-off-the-press description of the process for becoming a subsystem maintainer in #835 so you could also test that and give feedback :).

domenkozar:
@lukego sure! I'd give #835 a go.

lukego:
I did the patchShebangs wrong the first time. See 6a2d652 for the fix.

Generally I have a slightly tricky hacking situation with many documentation-related topic branches that are somewhat interdependent and hard to test in isolation. I am integrating them on `pdf-manual` and plan to send that upstream during this release cycle. So note to upstream maintainers: if there are some bugs in my latest documentation-related scripts then it is not the end of the world for those to slip upstream because we can fix/integrate them properly on `next`.

The main reason I am opening separate PRs is to get feedback and understand how easy/hard it will be to upstream these changes when they are all ready. I am also happy to use `pdf-manual` as the upstream branch for these changes provided that @eugeneia and others are happy with the general direction of these changes and will accept them upstream when ready.

domenkozar:
Done: https://github.com/SnabbCo/snabbswitch/pull/836

kbara:
@lukego What's the status of this PR and its associated changes now?

domenkozar:
This can be closed as it's included in https://github.com/snabbco/snabb/pull/876

PR #835, opened 2016-03-18
doc/git-workflow.md: Rewritten based on new experience

lukego:
I rewrote the git-workflow explanation with the goal of briefly but clearly explaining both how to submit a change to Snabb Switch and also how to be a subsystem maintainer.

This is intended to be a chapter in the manual. I marked `[wip]` because I haven't written the section about how to send your merged changes further upstream yet.

cc @eugeneia @wingo @kbara what do you think of this as a summary of our recent discussions and experiences?


src/doc/git-workflow.md
```diff
@@ -1,86 +1,113 @@
...
 
-### Develop and maintain a new program
+The first step is to create and register your subsystem branch:
 
-Snabb Switch includes programs like `snabbnfv`, `packetblaster`, and
-`snsh`. Here is how you can create a new program and take charge of
-its development.
+1. Pick your technical area of interest. What kind of changes will you be responsible for reviewing and merging? Try to pick an area that is easy to identify, for example "the packetblaster program", "the Intel I350 device driver", or "the Git Workflow chapter of the manual".
+2. Create a branch with a suitable name on your Github fork, for example `packetblaster`, `i350`, or `git-workflow`. This is where you will merge relevant changes.
+3. Describe this branch in the file `src/doc/branches.md` and open a Github Pull Request. This will kick off two discussions: how to clearly identify the changes that you are responsible for, and to which "next-hop" upstream branch you should send the changes that you have accepted by merging.
```
domenkozar: Would be nice to know the branch name to open Github Pull Request to start with. `next`?

eugeneia: We haven't figured that out completely yet. I think in most cases the PRs should be against master, then again some PRs only make sense targeting a specific branch (e.g. PRs related to the IPsec topic branch for instance). I have to update SnabbBot to merge PRs with their target branch instead of a fixed branch (master in our case) for the latter case to work well.

eugeneia:
:+1: Teeny tiny concern: people might expect to find a `drivers` branch and wonder what `<person-next>` branches are about, or even `next`. Imho this should document the whole picture of the current situation, and be updated accordingly when things establish.

I see minor overlap with #761, or at least this should be linked to from `CONTRIBUTING.md`.

src/doc/git-workflow.md
lukego: The `nix` branch should have a well-defined upstream next-hop where @domenkozar sends his Pull Requests. We maintainers should pick one as part of accepting #836.

I propose that `nix` feeds upstream to `kbara-next`. what do you think @kbara and others?

This process is supposed to be explained in point 3 of the text quoted above about updating `branches.md`. How should we make that clearer?

lukego: Note: There are two separate processes both in this document and in actual practice. One is for making an individual contribution, such as a new feature or a bug fix, and one is for sending a subsystem branch upstream, e.g. merging the `nix` branch with its upstream next-hop.

The document is supposed to clearly explain that there are two processes, one for contributors and one for maintainers, and to spell out the details of each. Obviously it is not doing so perfectly :). How could we improve it?

lukego:
@eugeneia Good point. How about if we start maintaining a map of the current branches and then refer to that where appropriate?

Here is a first attempt and using a little creative license in terms of content:

x

Source for ditaa:

```
                                   +--lisper
                   +--max next<----+--documentation<--pdf manual
                   |
       fixes       |
         |         |
master<--+--next<--+--kbara next<--+--nix
                   |               +--mellanox
                   |
                   |
                   +--wingo next<--+--lwaftr
                                   +--multiproc
```

src/doc/git-workflow.md
domenkozar: Maybe there should be a section at beginning explaining the difference between the two so the reader chooses which part to read?

lukego:
Thanks @domenkozar @eugeneia for the feedback. I have the feeling that the first redraft I submitted was not explaining things very clearly. I have pushed another partial rewrite now (half complete, half missing). Are there some parts that are less clearly explained than others now?

I also know that I have a tendency to overuse analogies when trying to elucidate the Git workflow (sorry @wingo...) but I can't look at our branch structure now without thinking of `master` as a river and the subsystem branches as tributary streams, like in this [map](https://en.wikipedia.org/wiki/Rhine#/media/File:Karte_Alpenrhein.png) from the Wikipedia page about the source of the [Rhine](https://en.wikipedia.org/wiki/Rhine) river. Is this worth talking about in the text or would it just be distracting?

src/doc/git-workflow.md
kbara: I'm happy to be the upstream for the nix tributary.

lukego:
I pushed a [new section](https://github.com/lukego/snabbswitch/blob/doc-git-workflow-redux/src/doc/git-workflow.md#becoming-a-maintainer) about how to be an upstream maintainer. I would really like feedback/ack/nack on this from the existing maintainers. cc @eugeneia @wingo @kbara.

kbara:
It looks good, though the next-hop part needs filling in. I think that's currently the fuzziest part of the whole workflow to me.

lukego:
@kbara Thanks for the feedback! I have written and pushed the next-hop section now. I am sure it is missing a lot of information. Can you tell me how it looks and what is still not clear?

Generally I think that having a well-defined branch tree is the key to making upstreaming work smoothly. This is hard now in the early days because we have so few branches and maintainers. Going forward it seems like if new features are first merged onto an appropriate subsystem branch (e.g. `nix`, `mellanox`, `pdf-manual`, etc) and then make their way up the tree (being sanity-checked and conflict-resolved at each step) then it should be easy to coordinate development.

The hope is that this maintenance structure will be scalable i.e. that we can get more useful work done by adding more maintainers. The barrier of entry to becoming a maintainer should be fairly low because the tree above acts as a safety net and mentoring framework. This seems to be how it works in the Linux kernel world.

One aspect I have not touched on in the documentation is having additional upstream communities around another Github fork e.g. the one on `igalia/snabbswitch`. This seems like an important mechanism for long term scalability since the rate of notifications from `snabbco/snabbswitch` is likely to become overwhelming at some point and it will be better to localize specific subsystems to their own repositories. I see a parallel where the Linux Kernel Mailing List was enough in the early days but can't keep up with all the discussions anymore and so these days there are around a hundred [more specific mailing lists](http://vger.kernel.org/vger-lists.html).

However, I feel that it would be premature to promote wide adoption of that model just yet. I tried creating a separate [snabbnfv-goodies](https://github.com/snabbnfv-goodies/snabbswitch) repository for the NFV application and that seems to have been counter-productive compared with working directly on this upstream repository because it reduced visibility of the development without commensurate benefits in that particular case.

lukego:
Ready for upstreaming, please!

src/doc/git-workflow.md
```diff
@@ -1,86 +1,169 @@
-# Snabb Switch Git Workflow
+## Git workflow
 
-This document explains the Git workflows that we use to develop and
-release Snabb Switch.
+How do you engage with the Snabb Switch developer community? The answer depends on what you want to do:
```
wingo: Still uses the name "snabb switch", fwiw, here and below; not a blocker but FYI

kbara:
Is it intentional that this contains both git-workflow.md and git-workflow.src.md?

eugeneia:
I believe this is in need of manual merge conflict resolution.

lukego:
Have merged master and resolved conflicts now. The overall diff seems to reflect only the intended changes against current master.

eugeneia:
For some reason this PR doesn't show up in https://api.github.com/repos/snabbco/snabb/pulls and therefore is invisible for SnabbBot. I will try to close and reopen it just to see what happens.

eugeneia:
Figured it out, see #910

kbara:
Merged, belatedly!

PR #837, opened 2016-03-18
Add src/scripts/snabb_bot.nix

eugeneia:
Adds a Nix expression/module for SnabbBot. Cc @domenkozar

### DIFF
--- a/src/scripts/snabb_bot.nix
+++ b/src/scripts/snabb_bot.nix
@@ -1,0 +1,95 @@
+{ config, pkgs, lib, ... }:
+with lib;
+with pkgs;
+
+let
+  snabb_bot = writeScript "snabb_bot.sh" (readFile (fetchurl {
+     url = "https://raw.githubusercontent.com/SnabbCo/snabbswitch/681129bb42de87eba957d947f6ee7a1d387f6a6d/src/scripts/snabb_bot.sh";
+     sha256 = "ec6c1bb92a1bbe80aa419761fcfcd943ea1413184d9a4441730dae8a542a84da";
+  }));
+in {
+  options = {
+
+    services.snabb_bot = {
+
+      credentials = lib.mkOption {
+        type = types.str;
+        default = "";
+        description = ''
+          GitHub credentials for SnabbBot instance.
+        '';
+        example = ''
+          username:password
+        '';
+      };
+
+      repo = lib.mkOption {
+        type = types.str;
+        default = "SnabbCo/snabbswitch";
+        description = ''
+          Target GitHub repository for SnabbBot instance.
+        '';
+      };
+
+      dir = lib.mkOption {
+        type = types.str;
+        default = "/tmp/snabb_bot";
+        description = ''
+          Target GitHub repository for SnabbBot instance.
+        '';
+      };
+
+      current = lib.mkOption {
+        type = types.str;
+        default = "master";
+        description = ''
+          The branch to merge pull requests with.
+        '';
+      };
+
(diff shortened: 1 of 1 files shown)


src/scripts/snabb_bot.nix
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  config = {
+
+    systemd.services.snabb_bot =
+      { description = "Run SnabbBot";
+        path  = [ busybox bash curl git docker jq ];
+        script =
+          ''
+            export GITHUB_CREDENTIALS=${config.services.snabb_bot.credentials}
```
domenkozar: (nitpick) it would be better to set these also just like `environment.SSL_CERT_FILE`

src/scripts/snabb_bot.nix
```diff
@@ -0,0 +1,95 @@
...
+            export CURRENT=${config.services.snabb_bot.current}
+
+            ${config.services.snabb_bot.environment}
+
+            exec flock -x -n /var/lock/snabb_bot ${snabb_bot}
+          '';
+        environment.SSL_CERT_FILE = config.environment.sessionVariables.SSL_CERT_FILE;
+      };
+
+    services.cron.systemCronJobs =
```
domenkozar: While I prefer to use systemd timers - as they're integrated into systemd/NixOS, I'm fine using a cron here.

domenkozar:
Looks good. I've yet to test it.

Not sure if `src/scripts/` is the best place to keep nixos modules, maybe `src/nixos/`?

lukego:
@domenkozar How come you don't like the idea of putting all nix expressions under `src/nix` from the original #831? Seems a bit funny to sprinkle them around to me, even to have a `src/nixos` directory that only contains nixos modules and not other kinds of nix expressions.

How do other projects organize their in-tree nix expressions? can you link an example or two?

domenkozar:
@lukego the informal convention is:

- to put `default.nix` in directories that are built using that expression (just like Makefiles)
- to put `release.nix` at toplevel, which imports all `default.nix` and builds all parts of the repository
- to put nixos modules upstream *somewhere*, there is really no convention about it. I'd keep them in separate folder to avoid clobber

@lukego so TL;DR #831 was a Nix expression, while this PR is a NixOS module.

lukego:
@domenkozar Thanks for the tips!

src/scripts/snabb_bot.nix
lukego: @domenkozar Can you clarify how systemd timers work e.g. with a link to documentation or an example nixos module? I have never heard of them before and perhaps @eugeneia has not either.

domenkozar: Yes - sorry. Arch has a good quick overview: https://wiki.archlinux.org/index.php/Systemd/Timers

NixOS reference documentation: http://nixos.org/nixos/manual/options.html#opt-systemd.timers
Example usage: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/backup/tarsnap.nix#L318-L322

eugeneia:
@domenkozar Ideally, I would pass this off to you as-is. I am fine with any changes you think appropriate as this (UNIX sysop / Nix) is far from my core expertise.

domenkozar:
@eugeneia Will do, thanks for this expression!

domenkozar:
This is now part of https://github.com/snabblab/snabblab-nixos/blob/master/modules/snabb_bot.nix, thanks @eugeneia

PR #838, opened 2016-03-20
doc/code-of-conduct.md: Add a community code of conduct

lukego:
I propose this short and sweet code of conduct for the Snabb community. This is intended proactively to frame future discussions when problems come up.

The basic suggestions here are to be civil, for upstream to act as a resource for helping people to cooperate efficiently (even when their goals don't immediately align with the master branch), and to keep important discussions on Github so that watching the project here is sufficient to stay "in the loop."

I have found it frustrating in the past to contribute to projects where the upstream community are rude, don't care whether contributors achieve their own goals, or require unproductive rituals that disrespect contributors' time.

Let's not become one of those projects. Instead let's encourage contributors to whack us on the head with this document any time we accidentally step in that direction.

Just to repeat, this is based on past experience with other projects and not related to anything going on in the Snabb community today. I imagine it would be much harder to introduce a change like this in the middle of a heated discussion about project policies and that is why I like the idea of doing it now.

### DIFF
--- a/src/doc/code-of-conduct.md
+++ b/src/doc/code-of-conduct.md
@@ -1,0 +1,7 @@
+### Code of conduct
+
+Follow the [Contributor Covenant](http://contributor-covenant.org/) and report problems to `conduct@snabb.co`.
+
+Help people achieve their goals. Don't stand between people who want to cooperate.
+
+Keep people in the loop. Use Github Issues for important discussions.


lukego:
@eugeneia will you be upstream for this?

what do you think about renaming `documentation-fixes` to simply `documentation` and using that as the next-hop for general documentation changes that are not part of a specific program? (I suppose it would be a bit funny for you to be your own next-hop with `documentation -> max-next` and a simple solution would be for `documentation` to feed directly into `next` instead.)

lukego:
(Could also be that maintaining both `max-next` and `documentation` separately is too much extra work. In that case another alternative would be to "optimize away" the merge to `documentation` and merge directly onto `max-next` instead, which I suppose is roughly the current practice. Even in this case I think it is valuable to have the specific branches well-defined and written up in `branches.md` because it does at least tell us which *person* is responsible for each area even if they may take a short-cut in terms of where the merge things. I feel like we are still feeling around to see how this will play out longer term.)

eugeneia:
`documentation-fixes` is somewhat underused these days. I see value in a documentation branch based on the fact that it will only ever contain documentation changes and never changes to code. So you (`next`) could trust me to ensure that `documentation` never changes code which would potentially ease reviewing the changes.

I think replacing `documentation-fixes` with `documentation` ( a branch for all documentation changes without code changes) is a good move. I think the next hop should be `next`, and I would gladly maintain this branch.

kbara:
That sounds like a good plan to me.

eugeneia:
I would be interested in what others think about the “Contributor Covenant”. I don't think the text is trivial. Are we binding ourselves to follow it literally or is it more of a non-authoritative guideline?

nnikolaev-virtualopensystems:
I guess if "All checks have passed" it should be fine :)

kbara:
It seems like a fairly common sense Code of Conduct - this is a friendly project, and "don't troll or throw insults around" isn't a culture change, happily. If Luke's intention is to actually enforce it if it ever gets violated, and the existing community all buys in, I'm in favour of it.

lukego:
> I would be interested in what others think about the “Contributor Covenant”. I don't think the text is trivial.

Thanks for reading and thinking about it. My take is that the Contributor Covenant rules seem sensible. They pass my basic litmus test of accepting basic Snabb upstream behavior and rejecting basic Linux kernel upstream behavior. This is important to me because I am often explaining that we use the same workflow as the Linux kernel but I don't want to imply that we should conduct ourselves like Linus Torvalds does.

> Are we binding ourselves to follow it literally or is it more of a non-authoritative guideline?

This would be binding and enforced, just like other project policies and related rules that we are bound by such as the [Github Terms of Service](https://help.github.com/articles/github-terms-of-service/).

lukego:
> I think replacing documentation-fixes with documentation ... is a good move

Done on `next`: 06c27c655bf565d83a2234f2617cd6add46587dd.

dpino:
FWIW, I'm aligned with this CoC and I think is positive to have one.

PR #833, opened 2016-03-17
FIX snabbnfv getting-started.md

yu-kasuya:
I tried to use snabbnfv today, Good, It worked.
But in getting-started doc, there were something wrong(old?) command.
So I fixed getting-staretd.md

First, in section Run the Intel 82599 driver selftest, I couldn't use SNABB_TEST_INTEL10G_PCIDEVA/B.
I used SNABB_PCI_INTEL0/1, there worked.

Second, in section Create and launch two VMs and to launch ubuntu2,
Some ids were duprecated ubuntu1's ids, so I fixed some ids in ubuntu2 like char0 -> char1.

PR #840, opened 2016-03-21
Merge max-v2016.04-3 into next

eugeneia:
Includes: #829 #830 #833 #834


lukego:
Merged, thanks!

I dropped the QEMU command line change following comment from @nnikolaev-virtualopensystems. I have not verified but I don't think that part of #833 is needed. Our CI test scripts don't use unique names for the QEMU chardevs.

This seems to me like the upstreaming process working well.

Issue #738, opened 2016-02-04
"Snabb"


lukego:
I propose that we shorten the name of our project to simply *Snabb*.

The longer name *Snabb Switch* comes from the early days of the project when the NFV application (a "hypervisor switch" or "vswitch") was the main focus. However, it has become something of a misnomer now that people are building many new and wonderful other applications. (It's also kind of long to type and abbreviations like *SnS* have not caught on.)

I think it's okay for the name to be slightly ambiguous i.e. for *Snabb* to refer to the core software, the whole software ecosystem including applications, the community, etc. This seems to work fine for "Linux."

What do we think?

#### How

This change would seem to involve a little bit of work:

- Update the source and documentation to `s/Snabb Switch/Snabb/`.
- Rename the Github repo from `snabbswitch` to `snabb`.
- Could alias `snabb-switch` to Snabb NFV if that reduces confusion somehow (or might increase it..)

(The renaming of the default binary from `snabbswitch` to `snabb` was already done long ago.)

#### Related

Couple of related issues:

- Define what the *Snabb* brand stands for i.e. what can users/developers safely assume about an application called *Snabb Foo* (where Foo = packetblaster, nfv, etc). Example: that it is apache licensed, accepts contributions, is engineered to be small/compact, and that the code syncs with the master branch.
- Agree on suitable/unsuitable uses of the project name. (For example I should probably rebrand my own professional activities which for historical reasons are also under the name "Snabb".)

wingo:
Nice!  Before, I was hesitant to use the name "Snabb" because of potential conflicts with SnabbCo; having some clarity will help the project I think.

Regarding *Snabb Foo* and branding, the points you mention make sense to me.  I might relax them to be a bit more aspirational -- e.g. someone could start their project called Snabb Foo with an aspiration to be small/compact, sync with master, etc, but if it's early days you never know how the thing is going to go.  Apache licensing should be a must, though.

Regarding other uses of the word Snabb -- I don't have experience in this domain, but https://wiki.gnome.org/FoundationBoard/Resources/LicensingGuidelines is a not-long document that can hint to what problems might arise.  Have you registered the trademark?  I guess it's a good idea, though who knows :)

[...]

PR #841, opened 2016-03-22
[wip] Rename "Snabb Switch" to "Snabb"

lukego:
I hereby propose that we rename the project from "Snabb Switch" to simply "Snabb" for the next release, following the discussion on #738. The trademark "Snabb" will be available to all users under terms that we will nail down in discussions here on Github. (I will do my own business under another name.)

If we are happy with this idea then I plan to "make it so" using this branch. Specifically, early next week I would update this PR to do the moral equivalent of `s/snabb switch/snabb/` on the entire `next` branch and then we would ship that as `v2016.04`.

Objections? Speak now or forever hold your peace!

### DIFF
--- a/src/README.src.md
+++ b/src/README.src.md
@@ -1,10 +1,10 @@
 # Introduction
 
-*Snabb Switch* is an extensible, virtualized, Ethernet networking
-toolkit.  With Snabb Switch you can implement networking applications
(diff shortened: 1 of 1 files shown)


eugeneia:
I had to point SnabbBot and SnabbDoc to the new repository URLs, apparently the redirects do not work for GitHub's API.

PR #842, opened 2016-03-22
Manage Neighbor Advertisement packet without a Target link-layer address

dpino:
While running the lwAFTR with a conf file were 'next_hop_ipv6_addr' was set, I noticed the address couldn't be resolved into a MAC address. The reason was that although a Neighbor Solicitation was sent, the Neighbor Advertisement that arrived to the lwAFTR didn't include a Target link-layer address. The current code was not managing this case. About this, RFC 4861 says:

"When responding to unicast solicitations, the option can be omitted since the sender of the solicitation has the correct link-layer address"

https://tools.ietf.org/html/rfc4861#page-23

In this case, the source ethernet address of the Neighbor Advertisement packet is used instead.

The pull-request includes an Unit test for a packet of this kind.

### DIFF
--- a/src/apps/lwaftr/lwdebug.lua
+++ b/src/apps/lwaftr/lwdebug.lua
@@ -33,9 +33,38 @@ function print_hex(data, len)
    print(table.concat(gen_hex_bytes(data, len), " "))
 end
 
+-- Formats packet in 'od' format:
+--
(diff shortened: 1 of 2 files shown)


kbara:
Nice catch. While you're solidifying it, mind adding 3-5 retries, 1 second apart, if it doesn't get answers?

PR #843, opened 2016-03-23
program/selftest: Added new "selftest" program

lukego:
This PR adds a program called `selftest` that is intended to help us explore the middle ground between small module-specific unit tests and large application-specific end-to-end tests.

In particular this program would be a nice place to explore test strategies that can be applied to the whole app catalogue e.g. fuzzing, property-based testing, searching for bad case configurations and workloads using optimization techniques like reinforcement learning or genetic programming, and so on.

For now the functionality is modest: roughly equivalent to `snsh -t <module>` but with a command-line interface for manipulating the environment variables used by test cases (which incidentally seem quite ad-hoc and could potentially be revised in the context of this program.)

From the commit message:

> selftest uses built-in test cases to test itself.
> 
> This initial implementation simply makes existing module selftest() functions available and provides a command-line interface for controlling the environment variables provided to test cases.
>
> Future extensions could add whole new testing strategies. The selftest program is intended to act as a framework for adding broadly applicable tests e.g. ones that can test many different kinds of apps (fuzz testers, etc).

From the usage:

```
Usage:
  selftest [options..] module <module-name>
    Execute built-in selftest functions of Snabb.

The common [options] determine the environment variables that are
provided to the test case. These can be inherited from the parent
process (default), reset to a clean slate, and selectively overridden.

  -z,        --zap-env       Unset all test-related environment variables.
  -p ID,     --pci ID        Set variable SNABB_PCI<n> to ID.
                             (Use multiple times to provide <n> = 0, 1, ...)
  -i ID,     --pci-intel ID  Set variable SNABB_PCI_INTEL<n> to ID.
  -s ID,     --pci-solarflare Set variable SNABB_PCI_SOLARFLARE<n> to ID.
  -f FILE,   --pcap-file FILE Set variable SNABB_PCAP to FILE.
```

Example of running the `intel_app` module selftest using only the parameters provided on the command line (nothing inherited from the environment):

```
$ sudo ./snabb selftest -z -i 01:00.0 -i 01:00.1 module apps.intel.intel_app
```

Thought: it would be neat to have a `testing` branch as the upstream for changes like this. Great opportunity for somebody new to become a Snabb upstream maintainer. Any takers?


eugeneia:
> Thought: it would be neat to have a testing branch as the upstream for changes like this. Great opportunity for somebody new to become a Snabb upstream maintainer. Any takers?

I would propose @petebristow, but I don't think he has the time right now. :(

eugeneia:
@lukego In the current state this program seems like feature bloat to me:

* It solves tasks that should probably be solved using `env(1)`
* It introduces a new mapping from short options to environment variables, which are painful enough as they are

I think this should at least replace the existing `./snabb snsh -t`. Right now I can imagine very well in what scenario this is useful, and I think the program has good intentions. On the other hand the “for now” functionality seems unattractive to me.

lukego:
@eugeneia Thanks for the review. Fair point. I will close this PR now. The branch will be available for somebody to merge in the future if they write some more substantial test code and need a place to put it.

PR #846, opened 2016-03-26
doc/trademark: Added tracemark policy

lukego:
This is a simple and permissive trademark license taken directly from
the PostgreSQL one: https://wiki.postgresql.org/wiki/Trademark_Policy.

This would make it clear that everybody is welcome to use the Snabb name and logo for diverse Snabb-related activities such as promoting software, conferences, user groups, support services, books, magazine articles, etc. This means that "Snabb" would refer to the union of all these activities.

Suitable for our project?

### DIFF
--- a/src/doc/trademark.md
+++ b/src/doc/trademark.md
@@ -1,0 +1,34 @@
+## Trademark policy
+
+The Snabb logo is a copyrighted design, and the name "Snabb" is a registered trademark. If you wish to use the name or logo in any way, you must comply with this policy.
+
+1. The name and logo, or derivatives of it may be used to promote Snabb based products or services.
+2. The name and logo, or derivatives of it may be used to promote or serve Snabb or related projects and their communities.
+3. You may not state or otherwise lead people to believe, that you represent the Snabb open source community in any way other than as an individual or corporate contributor to the project.
+
+### Examples
+
+#### Acceptable use
+
+The following are examples of acceptable uses of the logo:
+
+- Inclusion of the name or logo on your website offering Snabb support or consultancy services.
+- Inclusion of the name or logo on your website from which you offer Snabb-related or Snabb-derived software products.
+- Inclusion of a modified version of the logo on a Snabb user forum that you run.
+- Inclusion of the name or logo in your software or documentation to indicate that it runs on or with Snabb.
+- Use of the name or logo on websites, publications or merchandise, modified or unmodified, for Snabb user groups, events, or non-profit organizations which support Snabb.
+- Use of the name or logo to advertise Snabb-related content in a public event.
+- Use of the name or logo in news articles or blogs which reference the Snabb Project.
+
+#### Unacceptable use
+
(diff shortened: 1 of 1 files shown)


plajjan:
I'm a novice on the topic but isn't it relevant where it's a registered trademark? Doubt you could trademark ("varumärkesskydda") "Snabb" in Sweden given it's a common Swedish word :P

Let's say someone wishes to describe their product in Swedish and say it's fast, that would be the word "Snabb", or is there perhaps a distinction being made here through capitalisation? "Snabb" vs "snabb"? Doubt you can prevent Cisco from describing their CSR1000v as fast ;)

Apart from that it's a succinct text which I quite like.

lukego:
@plajjan Good question.

First off, the whole intention here is really to be permissive, not restrictive, and to secure the right for our whole community to use the Snabb brand (name, logo, color scheme, etc). That would include the right for big equipment vendors to ship Snabb-branded products under the terms above too. It would also include the right for people to create Snabb-branded blogs, user groups, etc, on their own initiative.

Then the scope of these rules would actually be really narrow. Implicitly this policy would only cover uses *Snabb* that are protected by trademark rules e.g. that appear to be representing this project. That would include using the name + logo + color scheme for branding networking products but exclude the millions of everyday uses of the word Snabb ("Cisco 1000V: Snabb som tusen!" etc.)

The intention of registering trademarks in a few jurisdictions is also defensively to secure our own right to use the name. I would like to avoid boring trolling situations e.g. where somebody files a trademark and then demands that we pay them royalties to continue using the name ourselves (as happened to Linux). An ounce of prevention is better than a pound of cure.

I am also a relative newbie to trademark issues and just hoping to get the brushstrokes right here and expand the set of people who are comfortable with using the Snabb branding for their activities.

plajjan:
Yeah, it is a rather boring topic and I agree on the intent and I think it's good adding this.

I didn't want to argue to the contrary, it was more of a question. So +1 for merging :)

lukego:
One interesting new case for the Snabb brand is the Youtube channel that I have created, [Luke's Snabb braindumps](https://www.youtube.com/channel/UC-ATWoBhU7ae5w7s5Rc0sSA). This trademark policy says that it is fine for me or anybody else to do these Snabb-branded activities and use the logo and so on.

I originally called it simply "Snabb braindumps." That would also seem to be fine. I changed it to be more distinctive and in keeping with the raw and ad-hoc theme of the videos.

Seems like a reasonable example of how people could use the project name and logo?

If I had called the channel "Linux Braindumps" then I believe I would rather need to contact the [Linux Marks Institute](http://www.linuxfoundation.org/programs/legal/trademark) and buy a sublicense for the name "Linux" for $5000. This has the downside of costing time and money but the upside that I could then presumably be sure that nobody else will use that name and so invest in building up the "Linux Braindumps" brand as my own. I am not sure that this system is worth the fuss... for us it seems to me like being permissive with the Snabb brand provides the best results for the least effort. However, IANAL...

PR #847, opened 2016-03-27
initial import of packetblaster lwaftr

mwiget:
This is a first working prototype of a lwaftr specific IPv4 and IPv6 traffic generator with the following features, for which I'm seeking feedback from the community in general and @kbara, @dpino and @wingo in particular.

- Refactored packetblaster synth and replay into their own directories. 
Send IPv4 and IPv6 traffic concurrently to fully test and report on a LWAFTR "on a stick design"
- Traffic sent either on a PCI port or Linux tap interface
- Simulates any number of B4 clients
- Rate limit generated traffic in MPPS (fractions are fine)
- Report separately and combined on incoming IPv4 and IPv6 traffic in MPPS, Gbps and packet loss
- Report on incoming packet loss by calculating the delta of the embedded packet counter, separately for IPv4 and IPv6 (allowing for separate sender and receiver applications).

See [https://github.com/mwiget/snabbswitch/tree/packetblaster-lw4o6/src/program/packetblaster/lwaftr]() for a detailed description.

Some topics I'm not sure about yet:

- Would it make sense to integrate this tighter with lwaftr, e.g. by importing lwutil.lua or leverage the protocol libraries for udp and datagram?
- Performance seems suboptimal. It max's out at roughly 5 MPPS on receiving. Maybe its just a matter of triggering jit.flush() or there are some low hanging fruits to optimise.


src/apps/test/lwaftr.lua
```diff
@@ -0,0 +1,345 @@
...
+  local src_mac = ethernet:pton(conf.src_mac)
+  local b4_ipv6 = conf.b4_ipv6 and ipv6:pton(conf.b4_ipv6)
+  local b4_ipv4 = conf.b4_ipv4 and ipv4:pton(conf.b4_ipv4)
+  local public_ipv4 = conf.public_ipv4 and ipv4:pton(conf.public_ipv4)
+  local aftr_ipv6 = conf.aftr_ipv6 and ipv6:pton(conf.aftr_ipv6)
+
+  local ipv4_pkt = packet.allocate()
+  local eth_hdr = cast(ether_header_ptr_type, ipv4_pkt.data)
+  eth_hdr.ether_dhost, eth_hdr.ether_shost = dst_mac, src_mac
+  eth_hdr.ether_type = PROTO_IPV4
```
dpino: It seems there's no support of VLAN tags. I would like to have it as it's something we use often.

Since you're defining ethernet_header, one way of implementing support of VLAN tags is to change the definition of the ethernet header if VLAN tags are used. In that case, `ether_header_ptr_type` will point to the definition of a 802.1Q header. See https://github.com/Igalia/snabbswitch/blob/lwaftr_starfruit/src/apps/lwaftr/generator.lua#L103

Same goes for the hardcoded offsets. In case of supporting VLAN tags, offsets should change (14 turns into 16, 34 (20 + 14) turns into 36, etc). As for me I prefer to use constants instead of hardcoded values since it makes easier to understand the meaning of a value (`ipv4_header_size + ethernet_header_size` instead of 34).

src/apps/test/lwaftr.lua
```diff
@@ -0,0 +1,345 @@
...
+
+  local ipv4_hdr = cast(ipv4_header_ptr_type, ipv4_pkt.data + 14)
+  ipv4_hdr.src_ip = public_ipv4
+  ipv4_hdr.ttl = 15
+  ipv4_hdr.ihl_v_tos = C.htons(0x4500) -- v4
+  ipv4_hdr.id = 0
+  ipv4_hdr.frag_off = 0
+
+  local ipv4_udp_hdr, ip4_payload
```
dpino: There's a typo, I think it should be `ipv4_payload` instead of `ip4_payload` (that makes later use of `ipv4_payload` global).

src/apps/test/lwaftr.lua
```diff
@@ -0,0 +1,345 @@
...
+  local ipv4_bytes = self.ipv4_bytes
+  local lost_packets = self.lost_packets
+
+  -- count and trach incoming packets
+  for _=1,link.nreadable(input) do
+    local pkt = receive(input)
+    if cast(uint16_ptr_t, pkt.data + OFFSET_ETHERTYPE)[0] == PROTO_IPV6 then
+      ipv6_bytes = ipv6_bytes + pkt.length
+      ipv6_packets = ipv6_packets + 1
+      payload = cast(payload_ptr_type, pkt.data + 34 + ipv6_header_size + udp_header_size)
```
dpino: First definition of `payload`, should be defined as local.

src/apps/test/lwaftr.lua
```diff
@@ -0,0 +1,345 @@
+module(...,package.seeall)
+
+local lib = require("core.lib")
+local app = require("core.app")
+local packet = require("core.packet")
+local link = require("core.link")
+local ethernet = require("lib.protocol.ethernet")
+local ipv4 = require("lib.protocol.ipv4")
+local udp = require("lib.protocol.udp")
```
dpino: Nit: `udp` is not used.

src/program/packetblaster/lwaftr/lwaftr.lua
```diff
@@ -0,0 +1,201 @@
...
+   end
+
+   local verbose
+   function opt.V (arg)
+      verbose = true
+   end
+
+   function opt.h (arg)
+      print(usage)
+      main.exit(1)
```
dpino: Should return 0 as exit code as the user explicitly requests to print out help.

src/program/packetblaster/lwaftr/lwaftr.lua
```diff
@@ -0,0 +1,201 @@
...
+   end
+
+   local ipv4_only = false
+   function opt.P () ipv4_only = true end
+
+   local ipv6_only = false
+   function opt.E () ipv6_only = true end
+
+   -- TODO how can I use digit options like -4? function opt.4 isn't valid in lua
+   args = lib.dogetopt(args, opt, "VD:hS:s:a:d:b:iI:c:r:PEp:", long_opts)
```
dpino: In case of using invalid characters such as numbers or a dash, which cannot be used as property names directly, should use the alternative hash syntax using brackets. Example:

```lua
> opts = {} opts["4"] = "hi"
> = opts.4
stdin:1: '<eof>' expected near '.4'
> = opts["4"]
hi
```

That said I don't know if `lib.dogetopt` would be able to parse '4' as a short name option.

src/program/packetblaster/lwaftr/lwaftr.lua
```diff
@@ -0,0 +1,201 @@
...
+   end
+
+   if not ipv6_only then
+    print()
+    print(string.format("IPv4: %s:12345 > %s:%d", public_ipv4, b4_ipv4, b4_port))
+    print("      destination IPv4 and Port adjusted per client")
+    print("IPv4 packet sizes: " .. table.concat(sizes,","))
+   end
+
+    print()
```
dpino: Nit: I think there's an extra whitespace.

src/program/packetblaster/replay/replay.lua
```diff
@@ -0,0 +1,88 @@
...
+local intel10g  = require("apps.intel.intel10g")
+local intel_app = require("apps.intel.intel_app")
+local basic_apps = require("apps.basic.basic_apps")
+local main      = require("core.main")
+local PcapReader= require("apps.pcap.pcap").PcapReader
+local LoadGen   = require("apps.intel.loadgen").LoadGen
+local lib = require("core.lib")
+local ffi = require("ffi")
+local C = ffi.C
+
```
dpino: Nit: `intel_app` and `C` are not used.

src/program/packetblaster/replay/replay.lua
```diff
@@ -0,0 +1,88 @@
...
+   local c = config.new()
+   function opt.D (arg) 
+      duration = assert(tonumber(arg), "duration is not a number!")  
+   end
+   function opt.h (arg)
+      print(usage)
+      main.exit(1)
+   end
+
+   args = lib.dogetopt(args, opt, "hD:", long_opts)
```
dpino: `long_opts` defines more options (`src`, `dst`, `sizes`) than short options "hD:" (help, duration). It seems this was already present in master, but since you're moving the code I think is nice to fix it :)

src/program/packetblaster/synth/synth.lua
```diff
@@ -0,0 +1,96 @@
...
+local pci       = require("lib.hardware.pci")
+local intel10g  = require("apps.intel.intel10g")
+local intel_app = require("apps.intel.intel_app")
+local basic_apps = require("apps.basic.basic_apps")
+local main      = require("core.main")
+local Synth     = require("apps.test.synth").Synth
+local LoadGen   = require("apps.intel.loadgen").LoadGen
+local lib = require("core.lib")
+local ffi = require("ffi")
+local C = ffi.C
```
dpino: Nit: `intel_app`, `basic_apps` and `C` are not used.

src/program/packetblaster/synth/synth.lua
```diff
@@ -0,0 +1,96 @@
...
+   function opt.S (arg)
+     sizes = {}
+     for size in string.gmatch(arg, "%d+") do
+       sizes[#sizes+1] = tonumber(size)
+     end
+   end
+
+   args = lib.dogetopt(args, opt, "hD:s:d:S:", long_opts)
+   config.app(c, "source", Synth, { sizes = sizes,
+   src = source,
```
dpino: Nit: wrong alignment.

src/program/packetblaster/synth/synth.lua
```diff
@@ -0,0 +1,96 @@
...
+                 print("Transmissions (last 1 sec):")
+                 engine.report_apps()
+              end
+   local t = timer.new("report", fn, 1e9, 'repeating')
+   timer.activate(t)
+   if duration then engine.main({duration=duration})
+   else             engine.main() end
+end
+
+function is_device_suitable (pcidev, patterns)
```
dpino: I wonder if this function could be moved to a helper file so it can be reused by 'synth' and 'reply'.

dpino:
Looking good. Supporting VLAN tags is the main concern to me.

src/apps/test/lwaftr.lua
mwiget: Many thanks for your detailed feedback. I'll go thru all of them and make the required changes. On VLAN support I'm wondering if I could simply leverage VMDq instead of handling VLANs in software. One drawback is of course the lack of VLAN support on tap interface. On the other hand, my intention with the tap interface was simply a method to make the packets visible via tcpdump if needed.

src/apps/test/lwaftr.lua
mwiget: Indeed. Good catch!

src/apps/test/lwaftr.lua
mwiget: Yep. Same again for the ipv4 handling a few lines down. I wonder if it makes more sense to define payload as local outside the loop or if it doesn't matter.

src/apps/test/lwaftr.lua
mwiget: Indeed.

src/program/packetblaster/lwaftr/lwaftr.lua
mwiget: Agree.

src/program/packetblaster/lwaftr/lwaftr.lua
mwiget: yes indeed. Fixed.

src/program/packetblaster/replay/replay.lua
mwiget: fixed.

src/program/packetblaster/replay/replay.lua
mwiget: Indeed. src/dst are only used in synth.lua. Removed them in replay.lua

src/program/packetblaster/synth/synth.lua
mwiget: fixed.

src/program/packetblaster/synth/synth.lua
mwiget: Yeah, my vim macro to align/autoindent lua has issues with functions across lines ;-)
Fixed it manually.

src/apps/test/lwaftr.lua
mwiget: Hmm. VMDq is probably not a good idea, because it requires to register with a mac address unless a VLAN is given. Just thinking out loud here: maybe VMDq can still be used with VLAN tag, but must be disabled for untagged traffic. Setting MAC address is an issue for the single stick design, which can have two different destination MAC's for v4 and v6 traffic returned from LWAFTR.

mwiget: Went with VMDq set to true when VLAN is given. Its an efficient way to add VLAN support without changing the source code: VLAN handling is done by the Intel NIC. That obviously doesn't work for tap interfaces and I don't have a warning about it just yet.

dpino:
@mwiget thanks for tackling the issues. If VLAN tagging can be managed by VMDq it works for me too.

src/program/packetblaster/lwaftr/lwaftr.lua
```diff
@@ -0,0 +1,212 @@
...
+
+   local ipv4_only = false
+   function opt.P () ipv4_only = true end
+
+   local ipv6_only = false
+   function opt.E () ipv6_only = true end
+
+   local vlan = nil
+   function opt.v (arg) 
+     vlan = assert(tonumber(arg), "duration is not a number!")
```
dpino: Wrong error message (copy-paste error I believe).

mwiget: argh and thanks ;-)

src/program/packetblaster/lwaftr/lwaftr.lua
mwiget: thanks. That info helped a lot: Got it now working using this:

```
   local ipv4_only = false
   function opt.v4 () ipv4_only = true end
   opt["4"] = opt.v4

   local ipv6_only = false
   function opt.v6 () ipv6_only = true end
   opt["6"] = opt.v6
```

src/program/packetblaster/synth/synth.lua
mwiget: @dpino: where would you put it? In a new file in src/program/packetblaster or simply place it in packetblaster.lua (without local) and call it from synth and reply? I kind of like its simplicity (1 print, 1 func call) right where its used right now.

dpino: @mwiget My concern is that the function is now duplicated in synth.lua and replay.lua. Since packetblaster/packetblaster.lua has to exist anyway, it sounds good to me to put it there (instead of creating a new file with a single function) and require it from synth.lua and replay.lua, WDYT?

mwiget: @dpino: agree. I renamed the local reporting fn() into packetblaster.report() and moved also is_device_suitable() into packetblaster.lua. IMHO is_device_suitable() might even be worth moving into lib.hardware.pci, but I'm hesitant to make such a move. Now there is at least no redundant code anymore between synth, replay

mwiget:
I think I fixed all open issues so far, removing the [wip]. Happy to make more changes of course.

dpino:
As for me, no more changes required. LGTM.

src/program/packetblaster/synth/synth.lua
```diff
@@ -0,0 +1,74 @@
...
+     for size in string.gmatch(arg, "%d+") do
+       sizes[#sizes+1] = tonumber(size)
+     end
+   end
+
+   args = lib.dogetopt(args, opt, "hD:s:d:S:", long_opts)
+   config.app(c, "source", Synth, { sizes = sizes,
+     src = source, dst = destination })
+
+   local patterns = args
```
eugeneia: @mwiget I am not a big fan of the code duplication here. Could the shared code below be put into a shared function?

src/program/packetblaster/synth/synth.lua
```diff
@@ -0,0 +1,74 @@
...
+      if packetblaster.is_device_suitable(device, patterns) then
+         nics = nics + 1
+         local name = "nic"..nics
+         config.app(c, name, LoadGen, device.pciaddress)
+         config.link(c, "source."..tostring(nics).."->"..name..".input")
+      end
+   end
+   assert(nics > 0, "<PCI> matches no suitable devices.")
+   engine.busywait = true
+   intel10g.num_descriptors = 32*1024
```
eugeneia: This kind of magic is awkward to have in two places. Does `packetblaster lwaftr` do this too? If not, how does it affect performance?

mwiget: @eugeneia: this is a result of breaking up the single packetblaster.lua command into individual files. Let me look into bringing this logic of building the apps and links into a function of packetblaster.lua and call it from replay and synth. Packetblaster lwaftr doesn't use this because its limited to a single PCI address.

src/program/packetblaster/synth/synth.lua
mwiget: @eugeneia: Hmm, that  was a result of moving synth and replay into their own directories, so they are in line with the added lwaftr. Would it make more sense to leave the original packetblaster.lua as is and simply enhance its logic to also allow commands taken from subdirectories? Kind of  having synth and replay be handled within packetblaster.lua (as before this pull) and only add lwaftr/lwaftr.lua?
Or do you have something different in mind?

src/program/packetblaster/synth/synth.lua
```diff
@@ -0,0 +1,62 @@
...
+       sizes[#sizes+1] = tonumber(size)
+     end
+   end
+
+   args = lib.dogetopt(args, opt, "hD:s:d:S:", long_opts)
+   config.app(c, "source", Synth, { sizes = sizes,
+     src = source, dst = destination })
+
+   local patterns = args
+   local nics = packetblaster.config_loadgen(c, patterns)
```
eugeneia: @mwiget With the risk of being super annoying: the code before and after uses of `config_loadgen` is identical in both cases as far as I can tell. How do you feel about moving that code into `config_loadgen` and rename it to `run_loadgen`?

E.g. instead of

```
   local patterns = args
   local nics = packetblaster.config_loadgen(c, patterns)
   assert(nics > 0, "<PCI> matches no suitable devices.")
   engine.busywait = true
   intel10g.num_descriptors = 32*1024
   engine.configure(c)
   local t = timer.new("report", packetblaster.report, 1e9, 'repeating')
   timer.activate(t)
   if duration then engine.main({duration=duration})
   else             engine.main() end
```

we would have

```
packetblaster.run_loadgen(c, args, duration)
```

mwiget: @eugeneia Ah, now I get it. Did as suggested and it does indeed look cleaner. Made a few more commits to address VLAN support without VMDq and creation of a pcap file plus selftest the pcap file.

plajjan:
@mwiget What is WIP about this one? I'm looking at making additions/modifications to packetblaster and your code seems like a better foundation than what we currently have on master.

mwiget:
@plajjan I wanted to add some more test cases and I have a change on reporting packet loss in % instead of absolute numbers. So nothing major. I should take the WIP away again.

plajjan:
> Would it make sense to integrate this tighter with lwaftr, e.g. by importing lwutil.lua or leverage the protocol libraries for udp and datagram?
@mwiget I think not. It's kind of nice with a second implementation from a testing perspective. One could argue this is just for performance testing and that such validation could be performed by another test app  but.. why? Better that that the test apps share code but we have to make sure we validate the program itself and if we share code with the lwAFTR how can we do that?

I noticed you have 2 space indent. I think Snabb typically uses 3 space indent, no? (Do we have a style guide?)

kbara:
It does indeed usually use a 3 space indent.

dpino:
@plajjan That's correct, the convention is to use 3-spaces indention. With regard to a style guideline I remember reading the codebase follows the Lua Style Guideline http://lua-users.org/wiki/LuaStyleGuide. I remember reading also the rule of thumb is to write code in the same style as the code that it has been written so far. I think those notes were on the "size budget" page in the Wiki, but I cannot find it anymore cc @lukego

mwiget:
@plajjan, @kbara and @dpino: didn't even notice and my .vimrc did technically make the mistake ;-), but I fixed it now.

PR #848, opened 2016-03-29
Add VLAN app to tag or untag packets

plajjan:
Andy Wingo wrote this as part of the lwAFTR but I think it's a useful
component on its own so I'm submitting it for upstreaming.

Maybe this is moot if the lwAFTR is upstreamed but I think it's more
 elegant to have this as another app anyway! :)

### DIFF
--- a/src/apps/vlan/README.md
+++ b/src/apps/vlan/README.md
@@ -1,0 +1,26 @@
+# VLAN
+
+There are two apps, Tagger and Untagger. As the name suggests one is used to
+add a VLAN tag to a packet whereas the other is used to strip one.
+
+## Tagger
+
+Tagger adds a VLAN tag to packets received on the input interface and sends
+them to the output interface.
+
+### Configuration
+
+-  Key **tag**
+
+*Required*. VLAN tag to add or remove from the packet
+
+## Untagger
+
+Untagger remove a VLAN tag from packets received on the input interface and
+sends them to the output interface.
+
+### Configuration
+
+-  Key **tag**
+
+*Required*. VLAN tag to add or remove from the packet
--- a/src/apps/vlan/vlan.lua
+++ b/src/apps/vlan/vlan.lua
@@ -1,0 +1,65 @@
+module(..., package.seeall)
+
+local packet = require("core.packet")
+local bit = require("bit")
+local ffi = require("ffi")
+
+local C = ffi.C
+local receive, transmit = link.receive, link.transmit
+local cast = ffi.cast
+
+Tagger = {}
+Untagger = {}
+
+-- 802.1q
+local dotq_tpid = 0x8100
+local o_ethernet_ethertype = 12
+local uint32_ptr_t = ffi.typeof('uint32_t*')
+
+local function make_vlan_tag(tag)
+   return ffi.C.htonl(bit.bor(bit.lshift(dotq_tpid, 16), tag))
+end
+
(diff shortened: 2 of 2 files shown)


kbara:
The lwaftr has basically been upstreamed - see https://github.com/SnabbCo/snabbswitch/pull/784 for details.

I'm happy to be the upstream for this PR, unless @eugeneia wants to be. @eugeneia , preferences/review/thoughts?

plajjan:
@kbara heh, src/apps/lwaftr/vlan.lua isn't there so it wasn't upstreamed *enough* ;)

Would you agree that it's better to put this as a separate app though? It's a fairly generic component and I think it's much cleaner if it's not inside the directory of a more specific application. I suppose lwaftr should start using this one as well instead of packing its own.

kbara:
It's only in ```next```, so we'll hopefully see it in the April release, which will be quite soon. :-)

I'm quite happy to have the vlan tagger/untagger as separate apps, but I'd like @eugeneia and/or @lukego to chime in and ack/nack it as well.

plajjan:
So I made something rather embarrassing and included some debug print statements that I had used before. Squashed that away :)

nnikolaev-virtualopensystems:
Is there any measurement of the performance impact of the tag/untag functionality?

plajjan:
I can write a selftest for that or at least run some tests.

In general, do we want performance tests as part of selftest() ?

eugeneia:
@plajjan You could add a benchmark to `program/snabbmark/`, this is the place where we are currently collecting various micro benchmarks.

@kbara I would assume that this app is generally useful, so from my point of view it makes sense to have it as an independent app.

@plajjan Since this app is already in `next` in form of `src/apps/lwaftr/vlan.lua` I would like to wait for it to land on master (ETA: early next week), and then have the changes needed to move `src/apps/lwaftr/vlan.lua` to `src/apps/vlan/`` appended to this PR. Does that sound sensible?

plajjan:
@eugeneia sure, that makes sense. Since that is a completely different change from a git perspective I think we better discard this PR and open a new one.

@nnikolaev-virtualopensystems I wrote a quick perftest as a selftest, given @eugeneia feedback I guess it should go into snabbmark instead but anyway, here goes:
```
kll@lingloi320 ~/kod/snabbswitch/src $ make && sudo ./snsh -t apps.vlan.vlan
LUA       obj/apps/vlan/vlan_lua.o
LINK      snabb
BINARY    2,1M snabb
Effective rate - baseline: 65345535.0 Mpps
Effective rate - Tagger  : 40018680.0 Mpps
Effective rate - Untagger: 26598030.0 Mpps
```
This is on my laptop (Intel(R) Core(TM) i7-3687U CPU @ 2.10GHz but turboboost..). The perf drop is quite big but the relative difference seems smaller when used in a more realistic program. Like if pcapreader & repeater is used instead of Source I get:
```
kll@lingloi320 ~/kod/snabbswitch/src $ make && sudo ./snsh -t apps.vlan.vlan
LUA       obj/apps/vlan/vlan_lua.o
LINK      snabb
BINARY    2,1M snabb
Effective rate - baseline: 37955730.0 Mpps
Effective rate - Tagger  : 27742470.0 Mpps
Effective rate - Untagger: 20868690.0 Mpps
```

plajjan:
Anyway, since we'll be using the one from lwaftr I'll close this PR now.

PR #855, opened 2016-04-01
next: Collected changes for the v2016.04 release

lukego:
@eugeneia Here is a release candidate for the v2016.04 release. Sorry to deliver it so late!

I have now renamed the Github repository from `snabbswitch` to `snabb` for consistency with this release (#852). Hopefully that doesn't break anything: Github seems to have the redirection cases covered.

I also made the organization name lowercase (`SnabbCo` -> `snabbco`) just to make the canonical version easier to type. Github organization names are not case sensitive so I don't expect trouble here.

Looks like the snabbdoc CI is failing perhaps due to script/makefile changes. action needed?


src/scripts/snabb_doc.sh
```diff
@@ -75,8 +75,8 @@ function index_out { echo $docdir/index.html;   }
 function build_doc1 {
     ((cd $(repo_path)/src && git checkout $1 && make doc/snabbswitch.html) \
         && mv $(repo_path)/src/doc/snabbswitch.html $2) \
- || ((cd $(repo_path)/src && git checkout $1 && make obj/doc/snabbswitch.html) \
-        && mv $(repo_path)/src/obj/doc/snabbswitch.html $2)
+ || ((cd $(repo_path)/src && git checkout $1 && make obj/doc/snabb.html) \
+        && mv $(repo_path)/src/obj/doc/snabb.html $2)
```
eugeneia: This doesn't work

1. for this PR because the old version of SnabbDoc is run
2. for other PRs because they might be behind `next`

See #856 for a fix targeting `next`. I think merging that PR should make the CI pass.

PR #863, opened 2016-04-08
Add VLAN related apps

plajjan:
This code is based on the work of Andy Wingo as part of the lwAFTR but I have expanded the simple Tagger/Untagger with a Mux program that can mux / demux packets based on VLAN tag. See the README for more info.

I tried to put a few useful functions as functions that can potentially be called from other applications, like if someone want to extract TCI or whatever themselves.

Unfortunately I think there's a HOLB issue in there. It's a tad tricky checking output when there are so many potential outputs. I think this works well enough for most use cases though.


src/apps/vlan/README.md
```diff
@@ -0,0 +1,52 @@
...
+### Configuration
+
+-  Key **tag**
+
+*Required*. VLAN tag to add or remove from the packet
+
+
+## VlanMux
+
+Despite the name, VlanMux can act both as a multiplexer, i.e. receive packes
```
dpino: Typo: s/packes/packets/

src/apps/vlan/vlan.lua
```diff
@@ -0,0 +1,174 @@
...
+local dotq_tpid = 0x8100
+local o_ethernet_ethertype = 12
+local uint32_ptr_t = ffi.typeof('uint32_t*')
+
+
+-- build a VLAN tag consisting of 2 bytes of TPID set to 0x8100 followed by the
+-- TCI field which in turns consists of PCP, DEI and VID (VLAN id). Both PCP
+-- and DEI is always 0
+local function build_tag(vid)
+   return ffi.C.htonl(bit.bor(bit.lshift(dotq_tpid, 16), vid))
```
dpino: I know this code is moved from the lwAFTR but there still some bits that can be improved.

C is cached. In addition, `htonl`, `ntohl`, `htons` and `ntohs` functions are defined at `core/lib.lua`, so actually it's not necessary to access them via the C namespace. The `core/lib.lua` module is always included so it's not needed to require it.

src/apps/vlan/vlan.lua
```diff
@@ -0,0 +1,174 @@
...
+   pkt.length = length + 4
+   C.memmove(payload + 4, payload, length - o_ethernet_ethertype)
+   cast(uint32_ptr_t, payload)[0] = tag
+end
+
+-- extract TCI (2 bytes) from packet, no check is performed to verify that the
+-- packet is carrying a VLAN tag, if it's an untagged frame these bytes will be
+-- Ethernet payload
+function extract_tci(pkt)
+   return ffi.C.ntohs(ffi.cast("uint16_t*", packet.data(pkt) + o_ethernet_ethertype + 2)[0])
```
dpino: `cast`is cached.

There's the functions `rd16, `rd32`, `w16`, `w32` in the lwAFTR codebase (`apps/lwaftr/lwutil.lua`) which do the same (casting of a pointer and derefence) but makes code easier to read. I understand that importing a lwAFTR module maybe understood as if this module was dependent on the lwAFTR. I think ideally these helper functions should be in `core/lib.lua` but that's something to do in another PR IMHO. In conclusion, I would leave as it is for now but I would love to have it changed in the future.

src/apps/vlan/vlan.lua
```diff
@@ -0,0 +1,174 @@
...
+         transmit(output, pkt)
+      end
+   end
+end
+
+
+VlanMux = {}
+function VlanMux:new(conf)
+   local self = setmetatable({}, {__index=VlanMux})
+   self.ethertype_8100 = ffi.cast("uint16_t", 0x0081) -- 0x8100 in network byte order
```
dpino: What about `self.dot1q_tpid = htons(dot1q_tpid)`? The code explains itself (host-to-network-byte-order) and can the comment can be skipped. In addition, I'd keep the same name for the variable (since it was the name that it was given before). Lastly, `conf` is not used (new methods may have an argument, it's not mandatory).

src/apps/vlan/vlan.lua
plajjan: Yes, I am aware. I explicitly did not want to require anything in lwaftr. I agree that these kinds of functions, if generally helpful, should be available in some core lib. I didn't do that here because a) it doesn't belong in same PR and b) I don't know if lwAFTR is the only program using this or if it would be adopted by more - maybe other devs prefer another pattern. Maybe I should have a local rd16 so it's easier to find afterwards if/when we introduce something into core!?

What do you mean with "cast is cached"? I don't know what to do with this information, do you mean I should change something? :)

src/apps/vlan/vlan.lua
plajjan: Yeah, htons is better, will change.

Keep what name?

conf is probably remnant since early versions. The "config" of the app is now ruled entirely by what is connected to its inputs/outputs. Do we have some preference that we always use conf so as to follow a standard pattern kind of?

src/apps/vlan/vlan.lua
plajjan: Okay, see what you mean with cast. NVM :)

src/apps/vlan/vlan.lua
plajjan: The way I interpret this I should be able to do, for example:
```
   self.dot1q_tpid = htons(dot1q_tpid)
```
But:
```apps/vlan/vlan.lua:99: variable 'htons' is not declared```

Comment (dpino) [inline src/apps/vlan/vlan.lua]: Sorry, my fault. `core.lib` is not included by default, it's necessary to import it. This will fixed:

```lua
local htonl, htons, ntohs = lib.htonl, lib.htons, lib.ntohs
```

src/apps/vlan/vlan.lua
dpino: AFAIK there's no pattern. Some apps call its new argument `arg` or `conf`. What is true is that new methods always receive one argument. If the constructor needs more than one argument, a table is passed. Many apps use this pattern:

```
function App:new(arg)
   local conf = config.parse_app_arg(arg)
```

Now, checking what `parse_app_arg` it converts a string with a pattern like "{ property = value,...}" into a Lua table. I think this is legacy code and it's not longer necessary (I'm not aware of any object that passes a string like a Lua table).

plajjan:
@dpino I believe I have addressed all of your concerns now :)

dpino:
@plajjan Cool, thanks. LGTM.

eugeneia:
@dpino @plajjan From my stand point, to merge this I would like to see

1. `apps.lwaftr.vlan` removed
2. Replace all uses of `apps.lwaftr.vlan` with `apps.vlan.vlan`

plajjan:
@eugeneia mmkay, I was thinking that would be a separate PR but if you want to see it in the same I'll see if I can put something together :)

plajjan:
@eugeneia I actually found way fewer references than I thought so I've updated them. PTAL.

dpino:
LGTM.

eugeneia:
Its was rightfully pointed out in #865 that I should probably not be the next hop for this, but @wingo or @kbara instead. Unless there is some argument against merging this I would just go ahead and merge it, and look more closely if I should be the next hop next time, OK? Alternatively you could assign yourselves and take over the wheel.

kbara:
I'm of two minds on that. I'd like @wingo or I to be the next hop on changes within the lwaftr-related directories, but I also like modifications that involve code spilling out from the lwaftr into Snabb as a whole to go through you/someone outside Igalia.

plajjan:
So maybe the best thing here is that we split this. First one PR to add apps.vlan.vlan that @eugeneia can  accept. Then a second one to change lwaftr to use apps.vlan.vlan. @kbara sounds good? Where should I open that second PR? Here? igalia/snabb? That first change need to land on whatever I'm submitting the second PR for...

eugeneia:
@kbara Since @lukego is the next hop for you and @wingo I think its sensible to have one of you guys as the next hop for all lwaftr related changes in the future.

Can I go ahead and merge this anyways?

kbara:
This change set LGTM as is, and I'm happy with it being merged as-is. My comment was with regards to @eugeneia 's musings on who should review such changes. Pragmatically, I think they should be LGTM'd by him and (either Wingo or I), which this now has been, regardless of whether the first or second set is officially the assignee.

PR #866, opened 2016-04-11
match pci addresses with hex a-f in them

petebristow:
This does what is says on the tin, pci addresses are hex and can contain a-f as well as 0-9


src/lib/hardware/pci.lua
```diff
@@ -163,7 +163,7 @@ end
 --
 -- example: qualified("01:00.0") -> "0000:01:00.0"
 function qualified (address)
-   return address:gsub("^%d%d:%d%d[.]%d+$", "0000:%1")
+   return address:gsub("^[0-9a-f][0-9a-f]:[0-9a-f][0-9a-f][.][0-9a-f]+$", "0000:%1")
```
dpino: Good catch! In Lua there's wildcard '%x' which matches hexadecimal digits. See http://www.lua.org/pil/20.2.html.

eugeneia:
@petebristow @dpino Using `%x` will also match upper case letters (whereas `[0-9a-f]` does not), am I right?

petebristow:
No %x does not match upper case letters, or at least it didn't in my testing. 
Linux uses lower case in it's paths, if we accept upper case we should probably flatten it to lower case as part of the process?

eugeneia:
@petebristow Just tested it on Lua and LuaJIT and `%x` managed to match upper case hexadecimal letters for me.

PR #867, opened 2016-04-11
use the CC env var for gcc command in Makefile if available.

petebristow:
Centos 6 ships with an old version of gcc that doesn't support -mavx2. It would be nice for the Makefile to respect CC rather than having to mangle PATH.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -5,6 +5,7 @@ E= @echo
 #E= @:
 
 export TEST_SKIPPED=43
+CC ?= gcc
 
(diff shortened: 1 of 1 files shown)


lukego:
Good catch.

I would really like for us to rewrite the SIMD checksum code with DynASM rather than GCC SIMD intrinsics. This way we could build with older GCC versions and trust that we get the right code (generated by the DynASM in our tree rather than an uncontrolled version of GCC).

One step at a time... DynASM did not support the AVX2 instruction set back when we introduced this code.

PR #869, opened 2016-04-11
Delayed_start, stall snabb so that peer NIC drivers can fully initialize

petebristow:
An app that delays packet forwarding for a configurable period of time.
Useful when feeding pcaps into a physical nic. The delay lets the peer
NIC the link to come up before sending packets so none are dropped.

Waiting for a linkup in the NicDriver:new() delays the whole process
in every situation and is even worse when a port doesn't link as all the 
timers need to expire.

### DIFF
--- a/src/apps/test/delayed_start.lua
+++ b/src/apps/test/delayed_start.lua
@@ -1,0 +1,24 @@
+module(...,package.seeall)
+
+local ffi = require("ffi")
+local C = ffi.C
+Delayed_start = { }
+
+--[[
+An app that delays packet forwarding for a configurable period of time.
+Useful when feeding pcaps into a physical nic. The delay lets the peer
+NIC completely initialize before sending packets so none are dropped.
+]]
+function Delayed_start:new(delay)
+   return setmetatable({ delay = delay }, { __index = Delayed_start})
+end
+
+function Delayed_start:push()
+   if self.delay then
+      C.usleep(self.delay)
+      self.delay = nil
+   end
+   for _ = 1, link.nreadable(self.input.input) do
+      link.transmit(self.output.output, link.receive(self.input.input))
+   end
+end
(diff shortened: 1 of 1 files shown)


eugeneia:
Calling `usleep` will block the whole snabb process as well (timers will not be run for instance). I suggest using [engine.now](http://snabbco.github.io/#engine-core.app) instead:

```
function Delayed_start:new (delay)
   return setmetatable({ start = now()+delay },
                       { __index = Delayed_start})
end

function Delayed_start:push()
   if engine.now() < self.start then return end
   for _ = 1, link.nreadable(self.input.input) do
      link.transmit(self.output.output, link.receive(self.input.input))
   end
end
```

petebristow:
Blocking the whole snabb process with `usleep` was by design. There was an issue floating around a while ago about back pressure was it good / was it bad. I don't remember it reaching a conclusion so it's on my list of things to play with. I wanted to avoid putting back pressure in place with this app or buffering packets in memory, which could go rather wrong if it's being fed by Repeater. 

I'd like to leave it as is if possible, perhaps expanding the comment to make it explicit that it will stall the whole snabb process.

eugeneia:
I don't think blocking is an option as it breaks `engine.main({duration=...})`, but you are right: we need to resolve the back pressure debate! See #656.

As of now anyways, the solution I suggested would work with `Repeater`, as it will only transmit packets when there is room in its output link, e.g. some packets would be buffered in `Delayed_start`.

Cc @lukego

src/apps/test/delayed_start.lua
```diff
@@ -10,14 +10,12 @@ Useful when feeding pcaps into a physical nic. The delay lets the peer
...
+   return setmetatable({ start = engine.now() + delay },
+					  { __index = Delayed_start })
 end
 
 function Delayed_start:push()
-   if self.delay then
-      C.usleep(self.delay)
-      self.delay = nil
-   end
+   if engine.now() < self.start then return end
```
lukego: This looks like a neat solution to me!

petebristow:
@eugeneia Is there anything else you want to see before this can be merged?

eugeneia:
@petebristow I merged it and replaced some tabs with spaces along the way, could you make sure your editor does not emit tabs?

PR #870, opened 2016-04-12
Add bits and bytes methods to Register

petebristow:
Setting the 3rd byte in a register to 45 or bits 0:3 to 5 is currently quite awkward.
This provide 2 helper methods to the Register objects to make life easier.

### DIFF
--- a/src/lib/hardware/README.md
+++ b/src/lib/hardware/README.md
@@ -191,6 +191,19 @@ in read-write mode.
 Clears bits of register according to *bitmask*. Only available on
 registers in read-write mode.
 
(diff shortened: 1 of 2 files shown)


eugeneia:
I was about to merge this when I noticed that this kind of “soft-breaks” the API for RO/RC type registers. E.g. `bits` and `bytes` will not be available on read-only and read-counter registers, even though their reading functionality would be useful. Could you factor out the readers and provide them for RO/RC registers as well?

petebristow:
@eugeneia Is there anything else you want to see before this can be merged?

PR #873, opened 2016-04-18
engine: Add timeline events for the breathe cycle

lukego:
The events are currently always enabled. This needs to be optimized.

Goal for now is simply to have interesting information in the log.

also [experimenting with visualization](https://twitter.com/lukego/status/721985471446917120) of the logs. Very preliminary; no obvious advantage over simple text (yet).


lukego:
Here is a quick screenshot from some tooling I am doing around timeline:

pharoscreenshot 2

This is based on the [Agile Visualization](http://agilevisualization.com/) software stack (Pharo) and I am experimenting with some visualizations. Here is a quick screenshot where each event is drawn as a rectangle, sized based on time (area = cycles), and with each message being assigned a distinct color. (This is the same data as in the table above.)

pharoscreenshot 3

Early days but the timeline does seem able to provide interesting data to go digging in.

lukego:
Here is a summary of the diagram above that shows the sum total cycles for each kind of event in the log and draws edges between events that follow each other:

pharoscreenshot 4

The big square is Virtio-net, the next is Intel10G, etc. On the top-left there are two different events that can precede the start of a breath: the end of the previous breath (top-left) or commit on the counter objects to shared memory (on the right).

This visualization is a bit more like what you see with a traditional profiler i.e. overall proportion of time spent on each activity.

eugeneia:
I really like the visualizations, Pharo seems like a kick-ass tool for this!

PR #875, opened 2016-04-18
test_env.sh: prefer $SNABB_TEST_FIXTURES over ~/.test_env

domenkozar:
This is the last missing piece for https://github.com/snabblab/snabblab-nixos/issues/13

Motivation: on snabblab servers it's too stateful to modify home directories of all users, so we'd like snabb to pick up a specific bash environment variable to discover test fixtures needed for `make test`.

cc @lukego @eugeneia

### DIFF
--- a/src/program/snabbnfv/test_env/test_env.sh
+++ b/src/program/snabbnfv/test_env/test_env.sh
@@ -32,7 +32,11 @@ export qemu_smp=$QUEUES
 export qemu_vectors=$((2*$QUEUES + 1))
 
 export sockets=""
-export assets=$HOME/.test_env
+if [ -z "$SNABB_TEST_FIXTURES" ]; then
+    export assets=$HOME/.test_env
+else
+    export assets=$SNABB_TEST_FIXTURES
+fi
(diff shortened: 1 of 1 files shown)


domenkozar:
On second thought, if `~/.test_env` exists, it should(?) override the `$SNABB_TEST_FIXTURES`. Is that a bit better? Currently one would need to `export SNABB_TEST_FIXTURES=~/.test_env`.

eugeneia:
@domenkozar I had thought of `SNABB_TEST_FIXTURES` as a functionality for edge cases, e.g. automated Nix builds. On regular lab servers you have to make sure that each user has their own `~/.test_env` as it contains persistent VM  images. E.g. when running things in docker, copy on write will cause a fresh VM image to be duplicated on every run. When run bare though, the tests will destructively modify the VM images. So you can't have two test_env instances using the same `SNABB_TEST_FIXTURES` as of now.

domenkozar:
There is one trick to get around that, basically plain COW: https://github.com/domenkozar/snabb-openstack-testing/blob/master/tests.nix#L74

Issue #850, opened 2016-03-31
Maintainers push to snabbco branches directly


lukego:
Github has [extended](https://github.com/blog/2137-protected-branches-improvements) their protected branches feature so that you can say who is allowed to push to which branch. This seems very useful: now we could tell Github who is responsible for each branch (e.g. that @kbara owns `kbara-next`) and then everybody could push to their branches directly.

This would allow us to retire an awkward bit of infrastructure i.e. the shell script that I run to mirror everybody's known branches onto the SnabbCo repo. The main reason that script exists is because when people push directly to SnabbCo it has been easy to make awful mistakes like accidentally making a release by pushing to `master`. (I have done this a couple of times in the past and I didn't want those mistakes to scale up linearly with the number of subsystem maintainers.)

Sound good @eugeneia @kbara @wingo @capr? The specific difference is you would publish your long-lived branches (the ones listed in `branches.md`) by pushing them to the SnabbCo repo rather than your own fork (and Github would now helpfully stop you from pushing to other people's branches). If so then I will setup the permissions and we can make the switch.

capr:
Ok, let us know when it's ready so we can switch the remote.

lukego:
Done!

Now you should please publish your branches by pushing directly to the snabbco repository. These are all "protected branches" which means (a) they don't accept force pushes and (b) each branch has a list of users who are allowed to push to it (I made the assignments that seemed obvious to me, let me know if you have trouble!)

Note that the mirroring script is disabled now so if you push to your own fork then that will not automatically arrive on the snabbco repository.

If you accidentally screw up a public branch and need to force push then please raise an issue in the spirit of #817 and we can work it out together. Github does not seem to have any option to allow force-pushes to protected branches, which is harsh on maintainers (difficult to undo mistakes) but easy on users (won't get their merges messed up by an upstream rebase).

Questions welcome!

cc subsystem maintainers @capr @eugeneia @kbara @wingo @domenkozar

PR #876, opened 2016-04-18
Add Nix expressions to build Snabb

domenkozar:
Hydra is currently using these in [release.nix](https://gist.github.com/83aebe9b7f4b443685ab38405b2b356d.git) and they build successfully at https://hydra.snabb.co/jobset/snabb/master

It's a good idea to put the upstream, because different Snabb versions may need different the Nix expressions. It's also handy to clone Snabb and just run `nix-build release.nix` and you should get the snabb, manual and tests (with logs) for that particular commit (if tests didn't fail).

There are two blockers:

- I have no way to extract Snabb version from the repo. Running a command won't help in Nix. It needs to be statically in a file. Currently no version is added to the derivation name.
- We depend on `git` for manual and also Snabb at build time. This is problematic since hash in Nix is content addressed (`src = ./.;` and `.git` mutates on each commit and thus changes regardless of the commit checkout. See my upstream comment at https://github.com/NixOS/nixpkgs/issues/8567#issuecomment-211362769

PS: There are lots of commits, maybe kbara-next needs to be rebase on master? Or should I just rebase on kbara-next?


domenkozar:
It won't let me force push since it's a protected branch, so can't rebase on top of `kbara-next`. Sorry for the mess :-) cc @lukego @eugeneia

kbara:
@lukego is there anything it'd be helpful for me to do with kbara-next, like merge master into it? I'm still getting used to the upstream Snabb workflow... :-)

lukego:
Hey, cool, there is actually no mess here and everybody is doing the right thing :) we are exercising a new edge in our collaboration graph here :-).

The reason we see many commits here is that downstream branch `nix` has merged a bunch of commits from the `v2016.04` release on `master` but these commits are not yet in `kbara-next`. So Github is neatly showing that the consequence of merging `nix` onto `kbara-next` would be to get all of these commits i.e. the `v2016.04` release in addition to Domen's own change a357165.

The next steps could be either of these things:

1. `kbara-next` merges from `master` and then this PR should automatically update to show only the commits that are still missing i.e. a357165. Then this could be merged.
2. `kbara-next` merges from `nix` which in one fell swoop would bring in the `v2016.04` release and also Domen's change.

So @kbara it is really up to you which you prefer.

In a perfect world we might all merge `master` into our long-lived branches immediately when a new release is made, but this demand-driven merge may well be fine too. The main thing is to transfer all the commits between all the branches and git is good at keeping track of what is missing.

lukego:
Here is one git incantation to summarize the changes more succinctly:

```
$ git log --oneline --first-parent origin/kbara-next..origin/nix
a357165 Add Nix expressions to build Snabb project
03f5958 Merge PR #855 (v2016.04 release) onto master
```

kbara:
Excellent. I've run git merge upstream/master on kbara-next and pushed that; I agree that doing that after each release sounds like a good workflow.

lukego:
@kbara You will need to push your branch directly to the `snabbco` repo now e.g. with `git push snabbco kbara-next` if you have a remote called `snabbco`. The mirroring-from-our-forks hack is gone now - see #850.

kbara:
Interesting. I've done that, it looks to me like it's updated, but the list of commits/etc above doesn't update.

lukego:
Interesting, yep. I see that you have pushed:

```
$ git log -1 --oneline origin/kbara-next
d5d355a Merge remote-tracking branch 'upstream/master' into kbara-next
$ git log --oneline origin/kbara-next..origin/nix
a357165 Add Nix expressions to build Snabb project
```

but I suppose that the Github UI does not automagically refresh in this case. Oh well, probably not a big deal.

kbara:
Yeah; I couldn't find a way to manually do that either. All the more reason to do this right after releases, eh?

kbara:
There is a trailing whitespace after ```+      tar xvzf ${test_env} -C ~/.test_env/ ```; is this something people care about on this project?

kbara:
```
+      export SNABB_PCI0="0000:01:00.0"
+      export SNABB_PCI_INTEL1="0000:01:00.1"
+      export SNABB_PCI_INTEL0="0000:01:00.0"
```
will clearly only work on some machines. Is the intent for this to just be running on the snabblab servers? If not, how are people with other PCI IDs expected to use it - local modifications to this file?

kbara:
Aside from those two nits, it looks good to me, with the caveat that I haven't dug deeply into nix yet. :-)

domenkozar:
You're right. I'll make sure the `SNABB_PCI* exports can be superseeded via bash env. And that whitespace should be gone.

domenkozar:
Pushed the two fixes, but the diff is bigger now:

```
$ git log --oneline origin/kbara-next..origin/nix
6cce8f4 release.nix: whitespace
8982625 release.nix: respect bash env, provide fallback values
e1ff0a3 Swapped in lstopo.png, replacing lstopo.pdf
aa1aac9 Renamed performance tuning file, as per review request
afc0bfb added info on mlock=on
294fc3b braindump on performance parameters
dbe9a9a Initial stub on performance tuning
a357165 Add Nix expressions to build Snabb project
```

domenkozar:
I've removed WIP, the two issues I mentioned can be addressed later.

PR #880, opened 2016-04-18
Merge max-v2016.05-2 into next

eugeneia:
Includes: #867 #866 #863 #875


eugeneia:
[iperf performance regression](https://gist.github.com/SnabbBot/51dcd38f7ab1bd7cb0009736cf4d7e1a) detected but I can not see any reason why so I am having SnabbBot go at it again.

PR #882, opened 2016-04-19
intel10g: Set num_descriptors = 2048 (was 512)

lukego:
Adopt the value for num_descriptors that is recommended in the optimization guidelines:
https://github.com/snabbco/snabb/blob/next/src/doc/performance-tuning.md


lukego:
Interesting that this fails CI. The test that fails is `nfvconfig` that is cycling through very many app network configurations. The failure seems to be a resource leak. I merged #884 to bring in the assertion that the packet freelist is overflowing.

Next step is to work out where the leak is, possibly a case where the `intel_app` is not returning receive descriptors to the freelist and that triggers more easily now that each instance allocates more buffers.

lukego:
Interesting. This is actually not a leak but a resource exhaustion. The `nfvconfig` selftest creates 64 VFs in the same app network and the hard-coded limit on packets in `core.packet` is too limited. The limit is 100,000 packets but with 64 VFs and 2,048 buffers per receive queue we need 131,072 buffers.

One solution would be to increase the maximum number of allocated buffers. The reason I see against this is memory consumption. Each buffer is 10KB and so 100,000 buffers is already 1GB of RAM.

One alternative would be some dynamic sharing. For example for I/O sources to allocate buffers based on their recent activity (is the queue busy or idle?) and system resources (is there a scarcity of packet buffers?). Anybody have a good algorithm for this?

plajjan:
I like KISS (keep it simple stupid). 1GB doesn't sound much. If you have 64 VFs I'm sure you can spare 1GB memory, right?

lukego:
Reasonable idea.

Trouble is that 1GB is not enough if we merge this PR and allocate 2048 buffers per VF. How do you feel about bumping the limit to 10GB?

plajjan:
131072 buffers is ~1.3GB, right?

What kind of limit is this? It's just a limit, right (as opposed to us actually allocating this)? Why do we have it? If you bump it to 10GB and my machine has 8GB, what happens then? PANIC?

lukego:
The allocation is dynamic - pay for what you use - and the limit is currently a hard assertion i.e. PANIC the process on exhaustion. The intention of the limit was to be so high that it only triggers on resource leak bugs but I evidently underestimated our appetite for packet buffers.

You are right that 1.3GB should be enough to make this test pass but with such a strict limit it would be good to have more safety margin on the KISS approach.

lukego:
I suppose also that it is important to allocate resources for the practical worst-case. This seems like a mark in favor of the KISS approach where you always allocate everything you need rather than moving the requirements based on workload (that maybe bites you on the bottom at the worst possible time).

plajjan:
I wasn't necessarily suggesting the limit be set to 1.3GB, just wanted to know what we are actually talking about :)

Catching resource leaks sounds like a good thing but at the same time I'm afraid you will run out of memory before you hit this limit if it's set to 10GB. But we are entering the ops domain, what is typical amount of memory available to host? My gut feeling it doesn't get that much because all apps are running in VMs.

I was about to suggest statically allocating all of it on startup. I remember an issue I had with a linecard that had faulty memory somewhere in its packet buffers. At normal workloads we didn't notice this but then two other paths in the network broke, forcing loads of traffic over this linecard - egress ports were congested, packet buffers filled up, hit the faulty memory region and crashed the linecard. Obviously with the two other paths down this happened at the worst possible time. I don't want something like that happening with my snabb processes, especially if they are supposed to do ddos mitigation...

How hard coded does this need to be? Can it be specified at startup? So like you give it a config file and if that config happens to consume more than the default (1GB?) then snabb simply won't start - you have to raise the limit. This makes it easy for most installations (since we have a reasonable default) and makes it explicit for those that need more.

I'm thinking of tuning postgres. There's a rather reasonable default but if you intend to run heavy queries you need to increase certain values. It's clearly stated that more memory will be consumed.

wingo:
If it's just a soft limit to give feedback to the user when they are forgetting to free a packet, to me it sounds like the intel app should increase the limit appropriately as it instantiates buffers.  If you instantiate 100 VF's with the default ring buffer size, then you are implicitly raising the limit.  The problem that the limit detects AFAIU is that someone forgot to call packet.free, not so much that someone instantiated too many apps.

lukego:
I merged #890 which should make CI pass for this branch. If that change is rejected then I will revert it here and merge in whatever replaces it.

These PRs need an upstream, please!

petebristow:
Just to echo wingo, being to able to raise the limit at run time from inside an app would be very useful.

lukego:
Soft limit for detecting resource leaks -- potentially even slow leaks from unusual code paths -- does sound neat. I don't plan to implement that on this PR though.

My feeling now is that it is reasonable to have a hard-limit on what should be "enough packets for anybody" (now 10GB of packets per process) and also for application test suites to check the application designs against these limitations. This may not be optimal but it does seem fairly practical.

Kudos to @eugeneia for writing a test case for the NFV application that exercises loading a configuration file with the maximum number of network interfaces defined. This has correctly detected that using 4 x more receive descriptors per NIC would cause resource exhaustion with an NFV application configuration that is supposed to be supported (64 receive queues / virtio-net NICs). Great practice to test these boundary-case configurations.

If we had neither this test case nor the hard limit then we may totally lose touch with how much memory the NFV application requires e.g. whether for a large configuration it would try to allocate a 100G+ of packet buffers.

PR #883, opened 2016-04-19
packetblaster: use default number of tx/rx descriptors

lukego:
This depends on the default number of descriptors being sane and for that reason this branch includes a merge of #882.

This seems to work fine - tested line rate with minimum packet size on 10x10G per packetblaster process: https://twitter.com/lukego/status/721693707251105792


eugeneia:
Had to kill of SnabbBot for d616abf because the [nfvconfig selftest got stuck](https://gist.github.com/SnabbBot/b6bd36c1b9351f219e340ec6bf930dd3). Can't really tell why this happened.

PR #893, opened 2016-04-26
Allow opening counters multiple times

teknico:
Sometimes, for instance when running tests, it is useful being able to open the same private counter multiple times in the same program.

The underlying memory is not mapped multiple times: the already existing counter is returned.

### DIFF
--- a/src/core/counter.lua
+++ b/src/core/counter.lua
@@ -42,7 +42,7 @@ local private = {}
 local numbers = {} -- name -> number
 
 function open (name, readonly)
-   if numbers[name] then error("counter already opened: " .. name) end
+   if numbers[name] then return private[numbers[name]] end
    local n = #public+1
    numbers[name] = n
    public[n] = shm.map(name, counter_t, readonly)


wingo:
LGTM, though feedback from @eugeneia / @lukego is welcome of course

wingo:
lgtm!

lukego:
Just an observation that now you can open the counters any number of times but the first close will invalidate all of them. Could be error-prone? Would be helpful to see how this is used.

wingo:
@lukego the use case is in `snabb lwaftr check` -- Nicola needs to check that on runs of the lwaftr that use pcap files as inputs, that the counters change in the ways that we expect.  Since the counters are the public interface, I didn't want to grovel in the app_table.lwaftr state to get them; better to open by name, but they were already opened by the lwaftr...

In practice I don't think anyone is closing counters except via link.free on reconfigure, so it is a moot point practically speaking i think :)

lukego:
Thanks for explaining!

PR #895, opened 2016-04-26
Merge `wingo-next` into `next`

wingo:
This will add PRs #712, #765, and #771, as well as also pulling in changes from the v2016.04.1 bug-fix release that are in master and not yet in next.


lukego:
Merged, thanks!

Sorry about not merging `master` into `next` already :-)

Might be better to use your own Github fork for short-term branches like `wingo-next-06-04-2016` rather than pushing them to `snabbco` just to keep the public [branches list](https://github.com/snabbco/snabb/branches/all) lean. (I have surrendered to branch clutter in my `lukego` fork for now...)

PR #887, opened 2016-04-20
Proposed fixes for 2015.04.1 bugfix release

lukego:
@eugeneia Here is a release candidate for a `v2015.04.1` bugfix release to get #885 and #858 onto master.

PR #874, opened 2016-04-18
Import snabblab docs from the wiki into the manual

domenkozar:
Fixes #864

I've modified it a bit and added minimal `Using the lab` and `Questions` sections.

After this is merged, we should update the wiki to point to the manual.

cc @lukego @eugeneia

PR #896, opened 2016-04-26
Merge max-v2016.05-3 into next

eugeneia:
Includes: #874 #889 #883 #882 #890 #868 #887 #847


eugeneia:
SnabbBot flagged 4a1ab24 for a [SnabbNFV performance regression](https://gist.github.com/SnabbBot/bf300a1bebed9868b264ff5d77a7a70a). I reset the result to get another measurement.

src/core/memory.lua
```diff
@@ -69,15 +69,15 @@ function reserve_new_page ()
...
+      local old = lib.writefile("/proc/sys/kernel/shmmax", tostring(huge_page_size))
       io.write("[memory: Enabling huge pages for shm: ",
                "sysctl kernel.shmmax ", old, " -> ", huge_page_size, "]\n")
    else
-      -- No: try provisioning an additional page
-      local have = tonumber(syscall.sysctl("vm.nr_hugepages"))
+      local have = tonumber(lib.firstline("/proc/sys/vm/nr_hugepages"))
       local want = have + 1
+      lib.writefile("/proc/sys/vm/nr_hugepages", tostring(want))
       syscall.sysctl("vm.nr_hugepages", tostring(want))
```
lukego: Good catch! I was imagining that system call was the preferred kernel interface, shows what I know.

Did you mean to also remove the `syscall.sysctl()` that sets `vm.nr_hugepages`? Seems redundant with the line above now.

lukego:
Merged, thanks!

If the performance regression is real then it will be flagged on #888 and we can resolve it on `next`.

I tried to manually run that benchmark on lugano-1 but the Docker gods are not smiling on me today. @domenkozar is there an easy way to run the NFV tests via `nix-build` atm?

domenkozar:
I just pushed my code for that part yesterday, but tests don't pass yet: https://hydra.snabb.co/eval/475#tabs-still-succeed

src/core/memory.lua
eugeneia: @petebristow ping. Pretty sure it was mistakenly leftover though.

petebristow: Yes that was a mistake. Do I just push the removal change to the original
CR? Or is that something that you'll handle?

eugeneia: Yes please, if the fix is in the original PR we can merge it however we like and at least be sure that you ACKed it.

lukego: I pushed a quick fix in 205375943f0ccc8a5796d50f0fcde488aabbc02e.

lukego: (Rationale: already merged to next so best resolved there either directly or as PR target. New edge in git workflow graoh again...)

eugeneia: @lukego By the way, if you mention the #XXX PR in commit messages these commits will show up in the PR thread, which is nice imho.

PR #897, opened 2016-04-26
Intel1g supporting multiple processes and RSS

petebristow:
Lots to be done, thoughts would be great, the tests are currently the best guide to usage, particularly `test6.sh`. It's fairly opinionated.
1. We shouldn't rely on loopback mode for testing. i350 support has never worked in non loopback mode AFAICT
2. `selftest()` and `selftest.sh` could do with being extended. The contained `selftest.sh` seems adequate for this.
3. A separate process can be in charge of exceptional error counters rather doing them in the fastpath. 
4. Better commentary and datasheet references are required.


plajjan:
Big thumbs up for RSS :)
I'm not qualified to give any real feedback on this although I've skimmed through it - mostly out of curiosity. Did note 2 space indent which is different from the de facto 3 space indent.

petebristow:
From a driver users perspective does it look like the kind of interface you would want, particularly if it was replicated into the 82599 driver?

lukego:
Cool! Pioneering a new design for multiprocess drivers :-)

Personally I find the plumbing rather stylish with using `flock` to become the master, the NIC software-software semaphore to protect global register updates, and automatically provisioning queues that are used. This keeps the interface simple and the code short.

cc @hb9cwp who is surely affected by this driver rewrite and likely has feedback too?

petebristow:
This driver is now in reasonably good shape. An assignee and review would be great.

eugeneia:
@petebristow The top-level comment in intel1g.lua says this supports both Intel I350 and 82599 (awesome?), but it does not implement VMDq as does does the existing Intel10G driver, am I right? If my understanding is correct, maybe rename the module to `intel_rss` to avoid confusion?

A section about this driver in `README.md` would be nice. E.g., describe configuration, modes of operation etc.

src/apps/intel1g/intel1g.lua
```diff
@@ -0,0 +1,684 @@
...
+   local rxdesc_t = ffi.typeof([[
+   struct {
+      uint64_t address;
+      uint16_t length, cksum;
+      uint8_t status, errors;
+      uint16_t vlan;
+   } __attribute__((packed))
+   ]])
+   assert(ffi.sizeof(rxdesc_t),
+   "sizeof(rxdesc_t)= ".. ffi.sizeof(rxdesc_t) .. ", but must be 16 Byte")
```
eugeneia: Should be `assert(ffi.sizeof(rxdesc_t) == 16, ...)`? Not sure how much sense the assertion makes though, in which cases would it not hold true?

petebristow:
This is the first driver that can support a herd of snabb processes. Do you envisage a need for a herd of VMDq processes as well? `intel_rss` probably doesn't make sense if this driver is going to get VMDq support. Obviously `intel1g` is also a poor choice as it supports 10g. `intel_herd` `intel_mp` ?

I've started to manage the herd with something a little like 
https://gist.github.com/petebristow/1231338296c2dfe8c9a28d4939411aee

`README.md` better comments and some tests that exercise the 82599 portion are in the works. 

@lukego Is the Mellanox driver going to take a similar approach to multiple processes?

eugeneia:
@petebristow I am fine with `intel_herd` or `intel_mp`. :-) I just skimmed the code and it looks fantastic (very straight forward). I think with the coming documentation and tests for 82599 this is ready to be merged!

eugeneia:
@petebristow If you merge #942 SnabbBot will run selftest.sh.

src/apps/intel_mp/intel_mp.lua
```diff
@@ -0,0 +1,774 @@
...
+   self.r.EIMC(0xffffffff)
+end
+function Intel:init_rx_q ()
+   if not self.rxq then return end
+   assert((self.rxq >=0) and (self.rxq < self.max_q),
+   "rxqueue must be in 0.." .. self.max_q-1)
+   assert((self.ndesc %128) ==0,
+   "ndesc must be a multiple of 128 (for Rx only)")	-- see 7.1.4.5
+
+   self.rxpackets = ffi.new("struct packet *[?]", self.ndesc)
```
eugeneia: Given that the packet ctype is now public (actually depending on #996), the pattern `ffi.new("struct packet *[?]",...` could be written as

```
desq_t = ffi.typeof("$ *[?]", packet.packet_t)
self.rxpackets = ffi.new(desq_t, self.ndesq)
```

Cc @kbara

kbara: Hm. What do you see as the main advantage of doing it that way?

eugeneia: Seems cleaner to me to not invoke the C name directly, in case the ever definition changes. Also, if `desq_t` was defined module-wide, then the JIT would see one ctype instead of different ctypes for each app instance.

kbara: Ok; the latter is a good enough reason for me. :-)

eugeneia: Never mind my advice above, as we have decided against exposing `packet_t`, see https://github.com/snabbco/snabb/pull/996#issuecomment-243049072.

Issue #789, opened 2016-02-26
core.packet: shiftleft is unsafe


eugeneia:
`packet.shiftleft` as well as `packet.shiftright` (from #781) can underflow and overflow packets.

wingo:
I don't know that this is a bug, fwiw.  Lots of things in Snabb are unsafe in general although they may be safe in particular.  As a user of shiftleft I certainly wouldn't want to propose adding an assertion to it without analyzing perf impact.

kbara:
I think 'safe' variants should be used by default, with unsafe variants with 'unsafe' in the name being available when raw speed is necessary. Call it my Haskell bias. I'm in favor of speed at the expensive of safety only when it's _really_ necessary, never as a default.

lukego:
The real problem in my eyes is lack of an immediate way to measure the impact of the assertions on real applications via end-to-end performance measurements. Removing assertions is only an optimization if it makes something important run faster: otherwise it's a bug.

wingo:
@lukego No argument there; this bug is to *add* assertions, which is fine but requires an analysis of the perf impact IMO.  Just trying to avoid the situation where we could merge in the next release and have to spelunk a perf regression.

eugeneia:
I would personally like the core API and libraries to be memory safe. I see this as one of the major benefits of using Lua instead of C. If that is not possible due to performance requirements then the unsafe API should be explicitly documented as such. So at the very least this is a documentation bug since the documentation of `shiftleft` does not document the dangers:

> — Function **packet.shiftleft** *packet*, *length*
> 
> Truncates *packet* by *length* bytes from the front.

I think this would be a good general rule for the documentation: if its not safe the extent of unsafeness should be documented. A function whose documentation that does not explicitly state unsafeness should be safe.

lukego:
@eugeneia On the one hand that sounds like an entirely reasonable position and I tend to agree. On the other hand contract with the "No hand holding!" statement in the [LuaJIT FFI documentation](http://luajit.org/ext_ffi_semantics.html) that sets a very different tone and is describing the level of abstraction that we are operating at.

kbara:
I don't see this as 'hand holding' so much as having non-stupid defaults that we can override when strictly necessary.

eugeneia:
[...]

Issue #774, opened 2016-02-23
Zen of snabb: ABI > API


lukego:
Just a reflection on the Zen of Snabb (cue from Andy and Kat's talk at FOSDEM):

Application Binary Interfaces (ABIs) are better than Application Programming Interfaces (APIs).

That is, it is often better to concretely describe how something is represented as bits in memory rather than to define it in terms of abstract data types.

The abstract advantage is concreteness. The concrete advantage is interoperability. For example, if core data types like packets, links, freelists, and so on are defined by their memory layout then we can write apps in Lua, C, assembler, Rust, Java, ... and we can even run them on separate hardware like a GPU or Xeon Phi coprocessor or FPGA or ... etc.

This is the same principle that allows us to write native device drivers in Lua: the network card defines its interface as a memory-mapped register layout that is agnostic to the way our software is written.

I have been meaning to braindump this reflection for a while. I take the queue now from #771 which I feel is moving us in the right direction i.e. nailing down the memory representation of core data types.

PR #898, opened 2016-04-26
Truncate and sample

petebristow:
2 very basic apps useful for monitoring rather than forwarding traffic.

### DIFF
--- a/src/apps/basic/basic_apps.lua
+++ b/src/apps/basic/basic_apps.lua
@@ -6,11 +6,40 @@ local app = require("core.app")
 local freelist = require("core.freelist")
 local packet = require("core.packet")
 local link = require("core.link")
+local ffi = require("ffi")
 local transmit, receive = link.transmit, link.receive
 
+--- # `Truncate` app: truncate or zero pad packet to length n
+Truncate = {}
+function Truncate:new(n)
+   return setmetatable({n = n}, {__index=Truncate})
+end
+function Truncate:push()
+   for _ = 1, link.nreadable(self.input.input) do
+      local p = receive(self.input.input)
+      ffi.fill(p.data, math.min(0, self.n - p.length))
+      p.length = self.n
+      transmit(self.output.output,p)
+   end
+end
 
-local ffi = require("ffi")
-local C = ffi.C
+--- # `Sample` app: let through every nth packet
+Sample = {}
+function Sample:new(n)
+   return setmetatable({n = n, seen = 1}, {__index=Sample})
+end
+function Sample:push()
+   for _ = 1, link.nreadable(self.input.input) do
+      local p = receive(self.input.input)
+      if self.n == self.seen then
+         transmit(self.output.output, p)
+         self.seen = 1
+      else
+         self.seen = self.seen + 1
+         packet.free(p)
+      end
+   end
+end
 
 --- # `Source` app: generate synthetic packets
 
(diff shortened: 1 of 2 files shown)


src/apps/basic/basic_apps.lua
```diff
@@ -6,11 +6,40 @@ local app = require("core.app")
...
 
+--- # `Truncate` app: truncate or zero pad packet to length n
+Truncate = {}
+function Truncate:new(n)
+   return setmetatable({n = n}, {__index=Truncate})
+end
+function Truncate:push()
+   for _ = 1, link.nreadable(self.input.input) do
+      local p = receive(self.input.input)
+      ffi.fill(p.data, math.min(0, self.n - p.length))
```
eugeneia: Please don't use the packet fields directly and use the public functions `packet.data` and `packet.length` instead. It's a bit more verbose, but as of now the fields are not really part of the public API.

wingo: Can you point to documentation for this?  To me it is useful to know that a packet is just a data and a length, and part of the core Snabb value proposition.  I haven't seen a PR that goes by that changes this, though I could have missed something :)

eugeneia: http://snabbco.github.io/#packet-core.packet describes the `core.packet` API. It doesn't mention the actual packet structure which is why I am uneasy about using the fields directly instead of going through the API functions.

wingo: Yeah I know what you mean.  Still, Luke has given several presentations with a lovely slide that's just

```c
// The maximum amount of payload in any given packet.
enum { PACKET_PAYLOAD_SIZE = 10*1024 };

// Packet of network data, with associated metadata.
struct packet {
    unsigned char data[PACKET_PAYLOAD_SIZE];
    uint16_t length;           // data payload length
};
```

and so I dunno, it's part of the public messaging.  There are quite a lot of uses out there already fwiw, and many of them introduced since you made the data/length functions in January last year.

lukego: I am in favor of documenting our ABI i.e. the memory layout of these structures.

These basic data structures like `packet` and `link` need to be usable by e.g. assembler code and this does not seem compatible with defining them via an abstract Lua API.

eugeneia: I think documenting the ABI is a separate issue, since that would not define the FFI field names (`data`/`length`) if I understand correctly(?).

One advantage (imho) of the API functions is that they are read-only. I am fine with documenting the field names as well and encouraging their use, or even getting rid of the API functions. Either way, I would like that API to be defined before I recommend its use.

I have encountered direct uses of the fields before as well, but considered them as semi-bugs until now.

petebristow:
A quick and dirty (so take with a pinch of salt) grep of the source code shows ~10 uses of `packet.data()` and ~250 uses of `p.data`. The methods haven't had much impact. 
I'm also in favor of documenting the structure, is there consensus enough to put through a PR to update the [readme](https://github.com/snabbco/snabb/blob/master/src/README.md)?
Should the API functions remain? They seem slightly redundant if the structure is documented.

eugeneia:
I think the best way to find out is to create that PR. I suspect that the PR would be quite involved though as it opens up a can of safety considerations. I also see overlap with: #774 #740 #734 #789

wingo:
I don't see any significant safety consideration, for what it's worth, by using well-known names for fields of a struct with a stable, documented ABI, especially considering that we need to access the packet data directly without any out-of-bounds-checking machinery.  YMMV of course :)

eugeneia:
To avoid stalling this PR I will merge it as is and open a separate PR do discuss the issue at hand.

PR #905, opened 2016-04-29
Add 'start' method to apps

dpino:
I think apps should have a 'start' method. 

The event will be emitted by `ops.start` (in `core/app.lua`). This _start_ method will give a chance to apps to additionally initialize themselves. For instance, right now apps can only access their links or even their own name in `push` and  `pull`methods. The reason is that these properties are setup by `ops.start` but it doesn't emit a _start_ event to their apps. Interestingly, `ops.stop` do emit a stop event to the apps :)

The background for this change is that we want to add counters per app in the lwAFTR and the only chance we have to do that is in their `push` or `pull` methods or doing other hacks that modify the engine https://github.com/Igalia/snabb/pull/299 I think the _start_ method solution is consistent with the current design.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -126,6 +126,14 @@ implemented the app instance is discarded and a new instance is created.
 *Optional*. Print a report of the current app status.
 
 
+— Method **myapp:start**
+
+*Optional*. Starts the app.
+
+At this point links and app name are set. The app has a chance to do
+additional initialization if needed.
+
+
 — Method **myapp:stop**
 
 *Optional*. Stop the app and release associated external resources.
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -178,7 +178,6 @@ function apply_config_actions (actions, conf)
       table.insert(new_app_array, app)
       app_name_to_index[name] = #new_app_array
       app.zone = zone
-      if app.start then app:start() end
    end
    function ops.restart (name)
       ops.stop(name)
@@ -226,9 +225,13 @@ function apply_config_actions (actions, conf)
    for linkspec, r in pairs(link_table) do
       if not new_link_table[linkspec] then link.free(r, linkspec) end
    end
-   -- commit changes
+   -- Commit changes.
    app_table, link_table = new_app_table, new_link_table
    app_array, link_array = new_app_array, new_link_array
+   -- Trigger start event for each app.
+   for _, app in ipairs(app_array) do
+      app:start()
+   end
 end
 
 -- Call this to "run snabb switch".
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -178,6 +178,7 @@ function apply_config_actions (actions, conf)
       table.insert(new_app_array, app)
       app_name_to_index[name] = #new_app_array
       app.zone = zone
+      if app_table[name].start then app_table[name]:start() end
    end
    function ops.restart (name)
       ops.stop(name)
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -178,7 +178,7 @@ function apply_config_actions (actions, conf)
       table.insert(new_app_array, app)
       app_name_to_index[name] = #new_app_array
       app.zone = zone
-      if app_table[name].start then app_table[name]:start() end
+      if app.start then app:start() end
    end
    function ops.restart (name)
       ops.stop(name)
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -230,7 +230,7 @@ function apply_config_actions (actions, conf)
    app_array, link_array = new_app_array, new_link_array
    -- Trigger start event for each app.
    for _, app in ipairs(app_array) do
-      app:start()
+      if app.start then app:start() end
    end
 end
 
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -148,8 +148,8 @@ end
 -- Update the active app network by applying the necessary actions.
 function apply_config_actions (actions, conf)
    -- The purpose of this function is to populate these tables:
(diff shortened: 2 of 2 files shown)


src/core/app.lua
```diff
@@ -178,6 +178,7 @@ function apply_config_actions (actions, conf)
       table.insert(new_app_array, app)
       app_name_to_index[name] = #new_app_array
       app.zone = zone
+      if app_table[name].start then app_table[name]:start() end
```
wingo: I think this could be just `app` instead of `app_table[name]`; am I missing something?

src/README.md
```diff
@@ -126,6 +126,14 @@ implemented the app instance is discarded and a new instance is created.
 *Optional*. Print a report of the current app status.
 
 
+— Method **myapp:start**
+
+*Optional*. Starts the app.
+
+At this point links and app name are set. The app has a chance to do
+additional initialization if needed.
```
wingo: Part of the design or spec should be what additional things you can do in start() compared to the constructor.  In your code, the links aren't set up, so we can't do linky things; dunno.

To expand on @dpino's use case -- we need the app name and we think waiting for the first push() is kinda hacky :)

wingo:
definitely needs some @lukego thoughts here :)

src/README.md
dpino: Right, only the arrays with the links are initialized (`self.input` and `self.output`) but the arrays are empty. Here is a good chance to do anything related with the app name which is the only thing that is available in `start` and it's not in `:new()`. That's the only thing I can think for now.

wingo:
For me if it's only the name of the app, then maybe a start() function is overkill. Too many things it could do, and not having the links set up is a bit irritating.  What if instead we pass some additional table of parameters as a second argument to the constructor?  That table would have {name=..., shmpath=..., ...}, i.e., whatever information the engine decides to make for the app.  Dunno, just an idea.  I don't want to trample on the "arg" to the constructor, as probably someone already has an app with a "name" configuration parameter :P

dpino:
I kind of like the proposed design. Let's say that here is where the app can access any additional attributes that engine adds to the app. If every app would be responsible of passing those attributes as a second argument in the constructor I'm afraid that would end with inconsistent naming of those parameters among different apps.

The links initialization need the apps are placed in the apps array (`new_table_table`) and the `input` and `output` arrays for each app. If there's a need for the apps to do something with the links after they are setup (and before `push` and `pull` starts) perhaps is worth to emit a new event to the apps (`app.linked`?). But so far I've never came across such use case, so I dunno.

wingo:
You say "If every app would be responsible of passing those attributes as a second argument in the constructor I'm afraid that would end with inconsistent naming of those parameters among different apps.".  Not sure what you mean by that; if it's core/app.lua which is responsible for creating and passing the second argument to the constructor, it's a standard interface, and all apps access it in a standard way.

I would go farther and create the links before creating the apps, so that you could already have everything an app needs before constructing it.  No more code, just moving some things around.  YMMV tho :)

dpino:
When I first read your comment I was thinking of every programmer being responsible of creating this second argument:

```lua
config.app("name", class, arg1, { appname = "name", shmpath = "..." })
```

If it's the engine the responsible of creating this second argument and pass it to the constructor, I take my comment aback because the names will be consistent. 

Still it's the programmer responsibility to be aware of this second argument and do something with it. It will go like this:

```lua
function App:new(arg, extra)
   local o = {
      prop1 = arg.prop1,
      appname = extra.appname,
      shmname = extra.shname,
   }
   return setmetatable(o, { __index = App })
end
```

That versus decorating an object with new properties (which the programmer still needs to learn about). I dunno, I see pros and cons in both approaches. Somehow Snabb is already doing this property decoration with the links (`input` and `output`). Before attempting any change I'd like to hear comments from other people, what do they think?

dpino:
I just remembered the use case of the NDP and ARP apps which need to send a solicitation packet to resolve an IP address to a MAC address. Currently these apps need to wait until the `push` event to send that request while they could do it as soon as the links are ready, in a `start` event for instance https://github.com/Igalia/snabb/blob/lwaftr/src/apps/lwaftr/ipv6_apps.lua#L184

eugeneia:
Any specific reason why this was closed? @alexandergall’s bridge apps need a “post setup” method as well:

https://github.com/snabbco/snabb/blob/master/src/apps/bridge/base.lua#L152

dpino:
Sorry, I closed it by mistake.

dpino:
What it was pending to decide in this PR was whether to make links available when `start` is called or not. I'm more lean to this approach as @wingo suggested. The reason is that we have already found use cases in the implementation of the lwAFTR where we need to use the links before `pull`is called. 

For instance, there are cases where we need to send Neighbor Solicitations requests. We wait until the links are available to do that, that means the `pull`method of the app. As a result, the pull method is bloated with logic of sending request packets and processing traffic coming from the links.

https://github.com/Igalia/snabb/blob/lwaftr/src/apps/lwaftr/ipv6_apps.lua#L186

I would rather prefer to put the request packets in the links as soon they are available, although the packets are not going to be processed until the main loop starts.

If everyone agrees in this approach then we need to move the start event after the link setup.

eugeneia:
`apps.bridge.*` would only benefit from a `start` method in which links are available

eugeneia:
I will merge this because we have existing apps (`apps.bridge`) that profit from this, and the need has been made clear in this thread.

@dpino Do you mind if I edit the documentation a bit?

dpino:
@eugeneia No problem, go ahead. Glad to see this PR merged :)

PR #906, opened 2016-04-29
test_env.sh: set initrd and be able to append kernel params

domenkozar:
These two changes are necessary to use NixOS as a guest image for snabbnfv selftest.

- `$SNABB_KERNEL_PARAMS` shouldn't do any harm and it's needed to pass `init=` as it has to point to nix store
- NixOS uses initrd and needs to be passed to qemu

@eugeneia what do you think is best, should we pass initrd only when assets have the file?

### DIFF
--- a/src/program/snabbnfv/test_env/test_env.sh
+++ b/src/program/snabbnfv/test_env/test_env.sh
@@ -104,12 +104,15 @@ function launch_qemu {
     if [ ! -n $QUEUES ]; then
         export mqueues=",queues=$QUEUES"
     fi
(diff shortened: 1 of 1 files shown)


eugeneia:
@domenkozar I guess since as-is the change “breaks” the existing Docker images. Since the initrd is read-only (right?) we could simply use `.test_env/initrd` if it exists?

domenkozar:
@eugeneia I've implemented that approach, ready to be merged.

lukego:
@eugeneia are you upstream for this?

PR #861, opened 2016-04-08
Testing toolbox love

eugeneia:
This PR hijacks #708  and #690 with permission from @petebristow. In short it imports his `Testsink` app as `apps.test.match.Match` and uses this app to simplify some replay-type selftest throughout the code base including the selftests for: `apps.test.match` (see #727), `apps.keyed_ipv6_tunnel.tunnel`, `apps.socket.raw` and `apps.tap.tap`. Other minor changes include:

* Changed documentation of `Source` app to reflect configuration and specify zeroed packet data
* Document `apps.test.synth` and fix it to generate proper Ethernet LLC packets
* Introduce `modest` option for `apps.test.match` for “match at least some”-style tests

PR #907, opened 2016-04-29
Merge kbara-next (2016-april and 2016-march) into next

kbara:
This includes: #884 (no freelist overflow on allocation @lukego ), #861 (testing toolbox love - @eugeneia / @petebristow ), #876 (adding nix expressions for Snabb builds, @domenkozar ), as well as Marcel's performance tuning documentation ( @mwiget ).


lukego:
Merged, thanks!

I was a little concerned about the hard-coded PCI addresses in `release.nix` for running the test suite. The risk is that running the tests on somebody else's machine could hijack an important network interface that happens to have the same address as a testing interface in the lab. However, I interpret `requiredSystemFeatures = [ "performance" ];` to mean that this script can only run on `lugano` servers in practice so it seems safe.

@domenkozar what do you think about renaming `performance` to `lugano` (i.e. be more specific about what hardware we are referring to) and finding a way to remove this from `release.nix` that should be able to run anywhere (not just in our lab)?

PR #908, opened 2016-05-02
Merge wingo-next into v2016.05

wingo:
Add #893 and #903 to `next`.


lukego:
Merged, thanks!

Log shows that SnabbBot reported failure only on the existing performance regression that has sneaked in separately.

PR #909, opened 2016-05-02
Monitor ingress packet drops and jit flush if threshold exceeded

dpino:
This PR adds a method to Intel10g app to fetch ingress packet drops value (QPRDC register). A timer in `core/app.lua` monitors this value periodically (100ms) and runs `jit.flush()` if a certain threshold is reached. There's a grace period of 20 seconds between jit flushes.

The monitor runs by default but it can be disabled on `engine.main(c, { ingress_monitor_drops = false })`

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -43,7 +43,8 @@ function ingress_drop_monitor:sample()
    for i = 1, #app_array do
       local app = app_array[i]
       if app.ingress_packet_drops and not app.dead then
-         sum[0] = sum[0] + app:ingress_packet_drops()
+         local status, value = with_restart(app, app.ingress_packet_drops)
+         if status then sum[0] = sum[0] + value end
       end
    end
 end
@@ -101,13 +102,16 @@ end
 local function with_restart (app, method)
    if use_restart then
       -- Run fn in protected mode using pcall.
-      local status, err = pcall(method, app)
+      local status, result_or_error = pcall(method, app)
 
       -- If pcall caught an error mark app as "dead" (record time and cause
       -- of death).
-      if not status then app.dead = { error = err, time = now() } end
+      if not status then
+         app.dead = { error = result_or_error, time = now() }
+      end
+      return status, result_or_error
    else
-      method(app)
+      return true, method(app)
    end
 end
 
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -99,7 +99,7 @@ end
 
 -- Run app:methodname() in protected mode (pcall). If it throws an
 -- error app will be marked as dead and restarted eventually.
-local function with_restart (app, method)
+function with_restart (app, method)
    if use_restart then
       -- Run fn in protected mode using pcall.
       local status, result_or_error = pcall(method, app)
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -28,37 +28,6 @@ link_table, link_array = {}, {}
 
 configuration = config.new()
 
--- Ingress packet drop monitor.
-ingress_drop_monitor = {
-   threshold = 100000,
-   wait = 20,
-   last_flush = 0,
-   last_value = ffi.new('uint64_t[1]'),
-   current_value = ffi.new('uint64_t[1]')
-}
-
-function ingress_drop_monitor:sample()
-   local sum = self.current_value
-   sum[0] = 0
-   for i = 1, #app_array do
-      local app = app_array[i]
-      if app.ingress_packet_drops and not app.dead then
-         local status, value = with_restart(app, app.ingress_packet_drops)
-         if status then sum[0] = sum[0] + value end
-      end
-   end
-end
-
-function ingress_drop_monitor:jit_flush_if_needed()
-   if self.current_value[0] - self.last_value[0] < self.threshold then return end
-   if now() - self.last_flush < self.wait then return end
-   self.last_flush = now()
-   self.last_value[0] = self.current_value[0]
-   jit.flush()
-   print("jit.flush")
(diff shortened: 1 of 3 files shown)


kbara:
This could be useful if the cause of the drops is side traces, but seems counterproductive for any other cause (such as the machine swapping or being slightly too slow, or other potentially transient timing issues). Perhaps it should be off by default?

wingo:
Will need a documentation update to mention the new parameter; and probably an update to mention "ingress_drop_count" to app writers.  Probably needs a comment as well in the source explaining what it's about.  Draft text, quit verbose:: "When the snabb breathe cycle runs, it usually doesn't drop any packets: packets pulled into the network are fully pushed through, with no residual data left in link buffers.  However if the breathe() function takes too long, it's possible for it to miss incoming packets deposited in ingress ring buffers.  That is usually the source of packet loss in a Snabb program.

There are several things that can cause packet loss: the workload taking too long on average, and needing general optimization; the workload taking too long, but only during some fraction of breaths, for example due to GC or other sources of jitter; or, the workload was JIT-compiled with one incoming traffic pattern, but conditions have changed meaning that the JIT should re-learn the new patterns.  The ingress drop monitor exists to counter this last reason.  If the ingress drop monitor detects that the program is experiencing ingress drop, it will call jit.flush(), to force LuaJIT to re-learn the paths that are taken at run-time.  It will avoid calling jit.flush() too often, in the face of sustained packet loss, by default flushing the JIT only once every 20 seconds."

wingo:
@kbara regarding whether to be on or off by default -- I suspect that if you are already dropping packets at a constant rate, then a once-in-a-while jit.flush() won't matter for your goals -- it won't hurt them anyway, and if the problem is ultimately side traces that should be main traces instead, then it may help.  Dunno.  The idea is that snabb processes should heal themselves if they can.  Of course the overhead of self-doctoring might be too much, as you say.  Curious to see what snabb-bot will say!

wingo:
a failure: https://gist.github.com/SnabbBot/cc90f0b8beb8747648e52d9df85326da#file-log-L294

src/core/app.lua
```diff
@@ -27,6 +27,36 @@ link_table, link_array = {}, {}
...
+      if app.ingress_packet_drops and not app.dead then
+         sum[0] = sum[0] + app:ingress_packet_drops()
+      end
+   end
+end
+
+function ingress_drop_monitor:jit_flush_if_needed()
+   if self.current_value[0] - self.last_value[0] < self.threshold then return end
+   if app.now() - self.last_flush < self.wait then return end
+   self.last_flush = app.now()
```
wingo: this needs to be just `now()`.  i wonder how this worked before :)

src/core/app.lua
```diff
@@ -27,6 +27,36 @@ link_table, link_array = {}, {}
...
+   self.last_value[0] = self.current_value[0]
+   jit.flush()
```
wingo: do we need to `require('jit').flush()` instead?  not sure.

dpino: Just checked `jit` module is included by default:

```lua
$ luajit -e "print(jit.version)"
LuaJIT 2.1.0-beta2
```
but probably it's a good idea to explicitly include it, for the same it's a good idea to do it for the `bit` module.

mwiget:
@kbara hitting Snabb with heavy traffic more often than not required jit.flush() to get to acceptable speed. Thats particularly true when Snabb is launched without any traffic, followed by heavy traffic, which is typical for traffic attracted to Snabb as a result of a routing protocol like BGP (routes announced once Snabb is operational). So I vote for having this enabled at all times. Not having it enabled requires an operator to manually restart Snabb.

wingo:
For me this patch LGTM, and it appears to have no effect on the benchmarks numerically.  I don't know though if the benchmarks drive Snabb to the point that it drops packets though.  Anyway, PTAL @lukego, since this is core stuff an opinion would be welcome.

src/core/app.lua
```diff
@@ -27,6 +28,36 @@ link_table, link_array = {}, {}
...
+   current_value = ffi.new('uint64_t[1]')
+}
+
+function ingress_drop_monitor:sample()
+   local sum = self.current_value
+   sum[0] = 0
+   for i = 1, #app_array do
+      local app = app_array[i]
+      if app.ingress_packet_drops and not app.dead then
+         sum[0] = sum[0] + app:ingress_packet_drops()
```
eugeneia: App methods (except `new`) need to be called in protected mode using `with_restart`, otherwise a single faulty app can bring down the whole process. See how pull/push are called.

src/core/app.lua
```diff
@@ -99,7 +99,7 @@ end
 
 -- Run app:methodname() in protected mode (pcall). If it throws an
 -- error app will be marked as dead and restarted eventually.
-local function with_restart (app, method)
+function with_restart (app, method)
```
eugeneia: I don't understand why it being local broke this PR? Is it because it was lexically defined after it was used? If so I think the proper solution would be to move the function definition up instead of making it accessible. Nitpick, opinions?

dpino: @eugeneia Yep, I made `with_restart` global was it was defined after the ingress drop monitor code. OK, I can move that code after the definition of `with_restart`.

eugeneia:
I have merged this because to me this feature seems critically useful. That being said this is also “new stuff” for me and I would be interested to hear what @lukego thinks about this. From my point of view this PR could be an indicator for a great ability we get thanks to LuaJIT but also could point to some difficulties the JIT nature poses to us.

PR #911, opened 2016-05-03
YANG model for current SnabbNFV configuration

eugeneia:
Initial YANG model for `program.snabbnfv.nfvconfig`. The namespace is imaginary obviously.

Cc @lukego 

nfvconfig


src/program/snabbnfv/snabb-nfvconfig.yang
```diff
@@ -0,0 +1,136 @@
...
+    description "A pcap-filter(7) expression.";
+  }
+
+  typedef gbps {
+    type decimal64 {
+      fraction-digits 4;
+      range "0..max";
+    }
+
+    description "Gigabytes per second.";
```
plajjan: This will be interpreted as Giga**bits** per second and not bytes. Why not use bps ?

```
typedef bps {
  type uint64;
  unit "bps";
  description "bits per second";
}
```

src/program/snabbnfv/snabb-nfvconfig.yang
```diff
@@ -0,0 +1,136 @@
...
+    container tunnel {
+      description "L2TPv3 tunnel configuration.";
+
+      leaf type {
+        mandatory true;
+        type identityref {
+          base "tunnel-type";
+        }
+        description
+          "Tunnel type identifier. Only \"L2TPv3\" is currently allowed.";
```
plajjan: Do you really need to say only L2TPv3 is allowed? That's kind of the point of the model itself - don't denormalise by putting info in multiple places :)

src/program/snabbnfv/snabb-nfvconfig.yang
```diff
@@ -0,0 +1,136 @@
+module snabb-nfvconfig {
+  namespace "http://snabb.co/nfvconfig";
+  prefix snabb-nfvconfig;
+
+  import ietf-yang-types { prefix yang; }
+  import ietf-inet-types { prefix inet; }
+
+  description
+    "This module describes the configuration of SnabbNFV.";
+
```
plajjan: Missing a revision statement (it's mandatory in YANG).

src/program/snabbnfv/snabb-nfvconfig.yang
```diff
@@ -0,0 +1,136 @@
...
+      leaf session {
+        type int32;
+        description
+          "The session field of the L2TPv3 header will be overwritten with this
+        value.";
+      }
+    }
+
+    leaf rx_police_gbps {
```
plajjan: Don't include unit in leaf name.

src/program/snabbnfv/snabb-nfvconfig.yang
eugeneia: Good catch, it should indeed say Giga**bits**.

eugeneia:
This spec reflects the reality of SnabbNFV (literally a one-to-one translation from the API) and I would like to keep it that way. E.g. the (completely legitimate) issues pointed out here (except for my bits/bytes blunder) are really bugs in SnabbNFV imho.

This is a first win for YANG as it showcased the bad configuration of SnabbNFV. I will continue by fixing these issues in SnabbNFV and update the YANG spec accordingly, probably all in this PR since it tracks the bugs nicely.

plajjan:
Ok, understood :)

FYI, YANG is very strict when it comes to making modifications between revisions so for example you can't rename a leaf if you are to follow the rules. We might want to discuss how to deal with versioning (simple way is to put version in the name, like my-model-v1.yang).

eugeneia:
I guess we should mark this PR as WIP and only merge a YANG spec once we are satisfied with it / SnabbNFV has a reasonable configuration API.

PR #912, opened 2016-05-04
quick fix of broken (flooding) bridge

leinilein:
probably broken since 339a80a787cbb0b71209bb29044e52e1cab76a79
nil value in flooding.lua
l24: local src_ports = self._src_ports -- _src_ports was never set
fix is just reactivation of 2 older lines of code

### DIFF
--- a/src/apps/bridge/base.lua
+++ b/src/apps/bridge/base.lua
@@ -123,12 +123,13 @@ function bridge:new (arg)
          end
       end
    end
-
+   
    -- Create list of egress ports for each ingress port, containing
    -- all free ports as well as all ports from different split-horizon
    -- groups
-   local dst_ports = {}
+   local dst_ports, src_ports = {},{}
    for sport, sdesc in ipairs(ports) do
+      table.insert(src_ports, sport)
       dst_ports[sport] = {}
       for dport, ddesc in ipairs(ports) do
          if not (sport == dport or (sdesc.group ~= 0 and
@@ -137,6 +138,7 @@ function bridge:new (arg)
          end
       end
    end
+   o._src_ports = src_ports
    o._groups = groups
    o._ports = ports
    o._dst_ports = dst_ports


eugeneia:
@alexandergall Does this look good to you?

alexandergall:
@eugeneia yes. I guess we need a self-test for the flooding bridge, too.

eugeneia:
@leinilein Would you be willing to add a selftest that tests for this regression as @alexandergall has pointed out? If so I would imagine you could benefit from merging #861 into this branch. If not I would merge this as is.

leinilein:
a selftest makes perfect sense! I will do so but can not promise it before mid of next week

leinilein:
I have implemented a selftest with source and sink app; before pushing it i would like to run the test but did not find any kind of documentation how your selftests really work. Is there any source or most recent example that i missed?

eugeneia:
@leinilein First of all, no need to hurry, take your time! :+1: Regarding selftests: we organize tests as `selftest` functions in modules. E.g. if the module `foo.bar` has a function `selftest` it can be run using any of the following commands:

```
sudo ./snabb snsn -t foo.bar
make foo.bar
make test
```

The `selftest` function should exit with a non-zero status to indicate failure, or `engine.test_skipped_code` to indicate that the test was skipped (for whatever reason). Does that help?

leinilein:
Yes, that was the piece of information i was looking for; unfortunately this is not documented - or at least i did not find this documentation. I will commit probably by the end of this week; thanks for the helping hints @eugeneia.

eugeneia:
I tried this fix and it did not work for me. I used https://github.com/eugeneia/snabb/commit/753fd9f7ea540f45fcbeb8428e990e2d9458f4d2 instead.

eugeneia:
Closing because this was superseded by #968.

PR #913, opened 2016-05-05
Expose packet ctype in documentation, deprecate packet.data and packet.length functions

eugeneia:
Reflecting the discussion from #898, this PR documents the widely used `length` and `data` fields of the packet struct by exposing the packet ctype (ABI?) in the API as well as `packet.max_payload`. The latter is required to safely use the ctype and enables fixing #789.

If we find agreement of this PR I would like to remove all mentions and uses of the then obsolete `packet.data` and `packet.length` API functions before it is merged (instead of deprecating them, I was just lazy here).

Possibly related:  #774 #740 #734


eugeneia:
Cc @petebristow @wingo

eugeneia:
Need upstream for this, @kbara would you be interested?

lukego:
I am in favor. This does bring a related issue to mind that I want to mention.

There seem to be three main plausible definitions of the struct:

```
struct packet_A {
  uint8_t[PACKET_SIZE];
  uint16_t length;
}

struct packet_B {
  uint16_t length;
  uint8_t[PACKET_SIZE];
}

struct packet_C {
  uint16_t length;
  char pad[62];
  uint8_t[PACKET_SIZE];
}
```

which one do we prefer when making this public?

Today we use variant C. Nikolay chose this during "straightline" integration on suggestion from a C struct linting tool. People have suggested it may not be optimal but never really made the case e.g. with both a theory and some relevant data. Maybe now is the time to do so?

lukego:
I will make an attempt to provide a theory and practice for changing the packet struct actually.

### Theory

Here is how they seem to compare to me on paper:

- A: Payload is aligned to start of struct. (This will be 128 bytes in practice due to allocation routine.) Having SIMD-word-size alignment *may* speed up some operations or be exploitable to simplify optimizing assembler code.
- B: First cache line contains all the most important data: length and first 62 bytes of payload (i.e. most or all headers). *May* be a peculiar effect with DMA writing to the same cache line as the length.
- C: Payload is aligned to 64-byte boundary (see above). Length and payload are on adjacent cachelines which *may* help with certain operations e.g. L3 cache fetching cache lines in pairs.

So which is best? I am not sure but it seems plausible to me that B or C is better then A. Even if they were all equivalent there may be one formulation that seems simpler and less surprising to users.

### Practice

I will ask Hydra to create a [performance report](https://hydra.snabb.co/build/3004/download/2/report.html) for each variant and see if there is a detectable difference. Results to follow. Place your bets! :-)

eugeneia:
I bet a drink on C, even though I think its an unfair bet since I don't know which benchmark is used. :P

lukego:
Hydra jobs are submitted! Here are the (currently pending build) report links:

- A: https://hydra.snabb.co/build/3780
- B: https://hydra.snabb.co/build/3658
- C: https://hydra.snabb.co/build/3583

If these don't work out you can look for the latest successful tests at https://hydra.snabb.co/project/lukego-sandbox

lukego:
(Oops! Got some errors on the builds. Our Hydra is still in beta :). Have contacted expert support @domenkozar :-))

lukego:
So! Hydra has completed those benchmark runs and generated the reports linked above. Welcome to take a peek :-)

Couple of things jump out right away:
- basic1 is faster with B and C.
- dpdk64 fails with C.

Other things are harder to interpret:
- iperf scores *may* have a significant difference.
- dpdk scores *may* have a significant difference.

So next steps, which I don't have time for right at this moment, would be to get the dpdk64 benchmark working for C and write some R code to automate the comparison and draw attention to which differences are significant (and which are not) to avoid drawing the wrong conclusions by eyeball.

eugeneia:
@kbara @lukego Can we merge this and decide on the packet structure layout another time?

src/README.md
```diff
@@ -303,35 +303,41 @@ Returns a structure holding ring statistics for the *link*:
...
-left on the freelist.
+— Ctype **packet.packet_t**
 
-— Function **packet.free** *packet*
+```
+struct packet {
+    uint8_t  data[packet.max_payload];
+    uint16_t length;
+};
+```
```
kbara: I'd prefer to document that a packet is guaranteed to have a uint8_t data[pcket.max_payload] field and a uint16_t length field, keeping the actual structure opaque, given the work that's currently being done on potentially redefining it (ie, @wingo 's work on headroom for virtualization).

eugeneia: I think we can just change it when the time comes, no?

kbara: I'd prefer not to be documenting it that explicitly when the odds of it changing so soon are so high.

eugeneia: I get why you are reluctant to documenting an API that may change, but documenting the exact ctype of packets is the whole point of this PR. Besides, we haven't made any promises with regards to API stability. If we document the ctype now, and change it next week, then we will at least have documented the API change.

I don't see a downside to documenting the current state, even if it might change soon. Instead, I see immediate gain by merging this PR now, and not let it be stuck waiting on future changes.

kbara:
I see this PR as deprecating the wrapper methods for the length and data fields, plus making various other documentation improvements, rather than as primarily documenting the exact ctype involved. We're in perfect agreement over everything except whether to include the exact struct definition, as far as I can tell.

src/README.md
kbara: I've merged it, although I disagree with that reasoning. I think documenting the concrete type right now is a mistake, but I agree that the patch is enough of a win overall that it's worth merging despite that.

PR #914, opened 2016-05-06
Set image format as raw

dpino:
Removes the following warning in `qemu0.log`:

```
WARNING: Image format was not specified for '/root/.test_env/qemu-dpdk0.img' 
and probing guessed raw. 
Automatically detecting the format is dangerous for raw images, write operations
on block 0 will be restricted. 
Specify the 'raw' format explicitly to remove the restrictions.
```

This assumes all images used for benchmarking are in raw format.

### DIFF
--- a/src/program/snabbnfv/test_env/test_env.sh
+++ b/src/program/snabbnfv/test_env/test_env.sh
@@ -115,7 +115,7 @@ function launch_qemu {
         -device virtio-net-pci,netdev=net0,mac=$(mac $qemu_n),mq=$qemu_mq,vectors=$qemu_vectors \
         -M pc -smp $qemu_smp -cpu host --enable-kvm \
         -serial telnet:localhost:$3,server,nowait \
-        -drive if=virtio,file=$(qemu_image $5) \
+        -drive if=virtio,format=raw,file=$(qemu_image $5) \
         -nographic" \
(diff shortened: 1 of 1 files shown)


domenkozar:
While I agree this is annoying, it would be better if we started using `qcow2` format as it's more convenient (less space intensive, only takes as much space as actually needed). I think that currently saves ~100MB for copying.

eugeneia:
@domenkozar I agree that raw is a waste of space, on the other hand its simple. I think we should support both in case valuable test images in RAW might be provided to us by third parties?

domenkozar:
Sure, setting `raw` for now is a good step forward to reduce noise in logs :+1:

eugeneia:
Merging this as currently we only really support RAW images.

eugeneia:
@dpino I would prefer to keep merged PRs open and their branches alive until the changes actually hit master (at which point they are closed automatically). My reasoning is that if we find new problems on the way from e.g. max-next to master it is convenient to document/solve these issues in the original PR thread.

dpino:
@eugenia Sorry, my fault. I was batch closing several remote branches on my fork (git push remote :branch) and I think I closed this one by mistake. I agree with keeping remote branches alive until they are merged.

PR #918, opened 2016-05-13
Mellanox ConnectX-4 Driver (WIP)

capr:
Currently the code covers the init procedure of the card up to QUERY_HCA_CAP which fails with "bad block number".

The init procedure so far allocates the command queue and initializes it, and there's some handy methods to send commands with input and output data. The commands are sent synchronously, one at a time (so only the first position of the queue is ever populated) and polled for completion. Later in the init procedure (this is not done!) we'll initialize an event queue and we'll be able to send commands asynchronously and use events to check for completion.

The selftest function does some testing on the bit utilities and tries to initialize the card(s) based on `SNABB_PCI_CONNECTX4[0|1]` env var(s).

The development was done entirely on lugano-4 on 100G cards.

Below is a handy script that will pull the code, reset the cards and run the selftest function. Use it to achieve a fast dev-run cycle if you prefer to edit the code on your home machine (you'll have to make your editor do a commit+push on save).

~~~
cd ~/snabbswitch || exit 1
git pull
cd src
make
export SNABB_PCI0=0000:02:00.0
export SNABB_PCI1=0000:03:00.0
echo 1 | sudo tee "/sys/bus/pci/devices/$SNABB_PCI0/reset"
echo 1 | sudo tee "/sys/bus/pci/devices/$SNABB_PCI1/reset"
sudo -E lock ./snabb snsh -t apps.mellanox.connectx4
~~~

### DIFF
--- a/src/apps/mellanox/README.src.md
+++ b/src/apps/mellanox/README.src.md
@@ -1,0 +1,4 @@
+# Mellanox ConnectX-4 Ethernet Controller App
+
+## MCX4 (apps.mellanox.connectx4)
+
--- a/src/apps/mellanox/connectx4.lua
+++ b/src/apps/mellanox/connectx4.lua
@@ -12,19 +12,36 @@ local index_set = require("lib.index_set")
 local macaddress = require("lib.macaddress")
 local mib = require("lib.ipc.shmem.mib")
 local timer = require("core.timer")
+local bits, bitset = lib.bits, lib.bitset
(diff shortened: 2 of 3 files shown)


lukego:
Great!

Could you please write a paragraph summary of the current state of the code i.e. what you expect to happen when executing the selftest method? This can be the text of the pull request.

Currently the information is much too minimal to understand what this code represents i.e. no pull request text and only a few words in the commit messages. Spending some time and effort on writing these things is a big help to the rest of us!

PR #919, opened 2016-05-13
Update performance-tunning.md

dpino:
The document states the number of descriptors must be a multiple of 128, but there are asserts in the code (lwaftr or firehose) that check whether the number of descriptor is a power of two. I think this is the correct check. OTOH, the fixed paragraph refers to bytes when it should refer to number of packets.

### DIFF
--- a/src/doc/performance-tuning.md
+++ b/src/doc/performance-tuning.md
@@ -14,7 +14,7 @@ require('apps.intel.intel10g').num_descriptors = ring_buffer_size
 config.app(c, "nic", require(device_info.driver).driver, {...})
 ```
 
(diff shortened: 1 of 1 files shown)


dpino:
Backported from https://github.com/Igalia/snabb/pull/322

wingo:
LGTM, as I mentioned in the other PR I think we should work at reducing the number of things we have to document, or at least reduce the number of ways for things to go silently wrong.  For example instead of requiring taskset, we can use sched_setaffinity via a standard --cpu argument.  We can warn if hyperthreading is enabled.  We can warn if we do --cpu to a core that doesn't have NUMA affinity to the PCI devices we use.  We can warn if we are using PCI devices but haven't set up core affinity.  We can warn if IRQ's are being delivered to our core.  The ring buffer settings should be generally fine with the new 1024 packet setting and though we should document the setting, we shouldn't encourage users to change it unless they just want to play around with things.

dpino:
It seems some benchmark tests failed but I think it's not related with the PR.

wingo:
Choo choo, on board for `v2016.06` train departing shortly

PR #923, opened 2016-05-19
Fix malformed error typo in lib.protocol.ipv4

xray7224:
The error for malformed addresses in `lib.protocol.ipv4:pton` had a typo. The `address` variable didn't exist, I believe it should be the `p` parameter that has been passed to the method.


eugeneia:
Looks good to me. Welcome to Snabb @xray7224, and thank you for the fix!

PR #925, opened 2016-05-19
Intel 82599 drivers default to RX/TX statistics counter index 0

### LINKED ISSUE #924: Ingress drop counter for Intel10G not working for vmdq=true
Garrr, there is a limitation in #909 -- it works fine for 82599 cards in single-function mode, but in VF mode it doesn't work unless you pass `rxcounter=0` in the app config.  This is because in VF mode we need to map virtual statistics to one of the global statistics counters, which are scarce relative to the number of possible VF's.

I propose to default `rxcounter` and `txcounter` to 0 if they are not set for `vmdq=true`.  WDYT @eugeneia @mwiget @dpino @lukego ?

wingo:
This PR is based on `next` since it modifies the ingress drop counter code.  Fixes #924.


dpino:
LGTM. This line can be removed too https://github.com/wingo/snabb/blob/2550c50c6c23653affc909fdbbe73a82321a44ed/src/apps/intel/intel10g.lua#L723

wingo:
Thanks for the review, removed the unnecessary line.

PR #927, opened 2016-05-30
Disable ingress_drop_monitor by default

dpino:
As commented in #917, the ingress drop monitor feature needs further testing before turning it on by default. The feature is only run now if requested explicitly with an options parameter.

I made this PR rebased to `next` as #909 is already merged.

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -281,7 +281,7 @@ function main (options)
       breathe = latency:wrap_thunk(breathe, now)
    end
 
-   if options.ingress_drop_monitor or options.ingress_drop_monitor == nil then
+   if options.ingress_drop_monitor then
       local interval = 1e8   -- Every 100 milliseconds.
       local function fn ()
          ingress_drop_monitor:sample()


lukego:
Just an idea: How about if we moved this feature out of the engine and applications (e.g. lwaftr) could run it via some suitable hook such as a `lib.timer` timer?

My reflection is that the main reason I see for embedding this directly in the engine would be to make it universally available but this may be overkill if this is an optional/experimental feature that we disable by default. WDYT?

dpino:
@lukego Sounds good to me. Recently we learned the feature was not working nice with SnabbNFV and that was a big motivation to disable it. Having the feature enabled by default for the lwAFTR but not for SnabbNFV, for instance, sounds like a better approach, or at least I like it more. I'm going to give this a try.

dpino:
Proposed implementation in https://github.com/snabbco/snabb/pull/929

wingo:
FWIW I never understood why the ingress drop counter negatively impacted Snabb NFV, especially on the 550-byte workload we were using to test.  Why would Snabb NFV drop packets on ingress?  Its whole job is to forward them.  I can only think it was misconfigured in some way.

lukego:
One thing to note is that I think the upstream tests for Snabb NFV and LWAFTR are different. On LWAFTR I believe you are gradually increasing the traffic rate until you start to see service impacted. On Snabb NFV we are blasting maximum packet rate and measuring how much of that is successfully forwarded. So in the sense of the automated testing that SnabbBot does the Snabb NFV would always be overloaded because it is blasted at 100% line rate all the time.

This may not be relevant to what you are thinking of here but seems worth keeping in mind for interpreting test results.

mwiget:
Blasting maximum packet rate into a Snabb process is the best case for throughput. Gradually increasing the load, or worse, changing the packets and content, makes it more vulnerable for requiring jit.flush(). At least that where our findings, where we initially "fixed" the throughput on a running stream of packets by simply killing and restarting Snabb. So in that sense, the way Snabb NFV automated test is easier on snabb than what LWAFTR does.

wingo:
I think the warning message in our case should indicate not only that we are flushing the JIT, if that's what we're doing, but also that the user probably wants to take a look at their configuration to verify the CPU affinity, the memory affinity, the PCI affinity, the scheduling, the IRQ delivery, and all that.  In practice for us that's what the "jit.flush" warnings have been meaning for us lately, that the user's setup is borked in some way.

mwiget:
@wingo: my experience with IMIX traffic at line rate however clearly shows the need for jit.flush(), even when all the performance settings are correct. We observed it when changing the binding table in-flight, e.g. from 1M to 2M entries, which btw impacted forwarding by less than 300ms. 
Another use case where we always saw jit.flash() being triggered is with Snabb running idle for some (lets say 30 seconds), then hit with IMIX based line rate traffic, generating an equal amount of return traffic over the same interface. Without a jit.flush(), performance was typically less than 40% of line rate.
I'm surprised no other application has run into this though ...

lukego:
Closing this specific PR as superseded by #929 but noting @mwiget's comments we need to add test cases for more dynamic workloads in the NFV application & likely adopt this mechanism there too. See also #937.

PR #929, opened 2016-06-01
Make ingress drop monitor a timer and enable it only in lwAFTR

dpino:
Moves ingress-drop-monitor out of engine and puts it on its own file as a timer. Applications can use the timer to activate packet-drop monitoring on applications that implement `ingress_packet_drops` method. In this case, ingress-drop-monitor is activated by default for the lwAFTR application.

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -68,7 +68,7 @@ end
 
 -- Run app:methodname() in protected mode (pcall). If it throws an
 -- error app will be marked as dead and restarted eventually.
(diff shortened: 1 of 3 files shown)


lukego:
LGTM! Thanks and sorry about the wait. Merged into `next`.

Likely we can promote this a global always-on feature once we have some more experience. Meanwhile it is interesting to develop a library of timers/hooks that can be installed when desired. Emacs extends quite neatly in that way.

PR #930, opened 2016-06-03
main: fork into worker and supervisor; perform clean up.

eugeneia:
This forks in `main`, leaving a “worker“ and a “supervisor” process. The worker sets a flag that causes it to be killed when the supervisor quits, and resumes as the regular snabb process. The supervisor sets up a `signalfd(2)` to be notified when the worker quits, and cleans up afterwards. See  #509.


src/core/main.lua
```diff
@@ -44,10 +45,30 @@ function main ()
...
+         require(modulename(program)).run(args)
+      else
+         -- Supervisor
+         local exit_signals = "hup, int, quit, term, chld"
+         local signalfd = S.signalfd(exit_signals)
+         S.sigprocmask("block", exit_signals)
+         local signals, err = S.util.signalfd_read(signalfd)
+         assert(signals, tostring(err))
+         if not signals[1].chld then S.kill(worker_pid, "kill") end
+         shutdown(S.getpid())
```
eugeneia: It is a mystery to me why this works. E.g. I had assumed I needed to pass `worker_pid` to `shutdown` since the worker goes on to create the `/var/run` directory based on `S.getpid()`, but apparently when the worker calls `S.getpid()` it returns the parent pid? _confused_

lukego:
I am excited about managing processes directly with Snabb. There are a lot of nice convenience / safety / error-recovery / administrative features we could implement. I also really want to be able to store more information in `/var/run/snabb` directories (e.g. 64MB timeline log files) and start archiving these in a disciplined way (e.g. via a hook or config option to store them in a well-defined location with compression).

Thoughts:

- Major change: the `/var/run/snabb` directories will not exist after the process terminates, right? This is a user-visible change and it seems worth explaining why this makes sense in the PR text.
- Major change: The `snabb gc` command is gone. This also seems worth drawing attention to in the PR text and being clear about why it is obsolete.
- Concern: How can we be confident that `/var/run/snabb` directories won't be left around without cleanup? For example, is this design robust for unexpected-but-inevitable actions like pressing `^C` in the terminal or running `kill -9` on the pid that is using the most CPU?

Generally I would find this change easier to digest with more explanation. Maybe some commentary in the source code explaining the overall intention and how that is achieved with these particular system calls?

eugeneia:
Good points!

> Major change: the /var/run/snabb directories will not exist after the process terminates, right? This is a user-visible change and it seems worth explaining why this makes sense in the PR text.

Yes, this PR makes snabb remove its own runtime directory. This makes sense because as of now the lab servers are littered with stale runtime directories. Ideally there would be a way to pass a `--debug` or `--backup` option to keep the files if that is desired. This would require additional command line option parsing before forking (NYI).

> Major change: The snabb gc command is gone. This also seems worth drawing attention to in the PR text and being clear about why it is obsolete.

`snabb gc` is obsoleted by processes cleaning up after themselves. Also: `snabb gc` never really worked, it failed to remove runtime directories when other processes where reusing PIDs.

>     Concern: How can we be confident that /var/run/snabb directories won't be left around without cleanup? For example, is this design robust for unexpected-but-inevitable actions like pressing ^C in the terminal or running kill -9 on the pid that is using the most CPU?

If you `kill -9` (SIGKILL) the supervisor then you are out of luck (no way to react to that signal): the child will be orphaned. In all other cases this is robust, e.g. sending SIGKILL to the child (the process that will actually use CPU cycles) will lead to clean shutdown of the supervisor, as well as sending SIGTERM or SIGINT to the supervisor or child.

What is currently not covered is handling of SIGCONT to the child, e.g. it is not supported to stop (SIGSTOP) and continue the child. I felt this was not something we want to do, but it might be sensible to have a good default behavior for this case.

eugeneia:
> If you kill -9 (SIGKILL) the supervisor then you are out of luck (no way to react to that signal): the child will be orphaned.

Actually not true, the child will be killed by SIGHUP (`S.prctl("set_pdeathsig", "hup")`), but the runtime directory will not be deleted.

eugeneia:
> What is currently not covered is handling of SIGCONT to the child, e.g. it is not supported to stop (SIGSTOP) and continue the child. I felt this was not something we want to do, but it might be sensible to have a good default behavior for this case.

Also wrong, SIGSTOP/SIGCONT were handled just fine...

teknico:
> Ideally there would be a way to pass a --debug or --backup option to keep the files if that is desired.
> This would require additional command line option parsing before forking (NYI).

An option to keep the data directory around upon termination would be valuable.

eugeneia:
Added support for the worker to be stopped/continued, so that caveat is gone.

> An option to keep the data directory around upon termination would be valuable.

I have made `_G.developer_debug` configurable via environment (`SNABB_DEBUG`) and main no longer deletes the run-time directory when `_G.developer_debug` is true.

teknico:
> I have made _G.developer_debug configurable via environment (SNABB_DEBUG)
> and main no longer deletes the run-time directory when _G.developer_debug is true.

Great, thank you. Are env vars documented anywhere?

eugeneia:
@teknico Good point, currently no. I think this should be documented along with the sub-program mechanism in place of this [outdated section](https://snabbco.github.io/sha1/127d21fa6e6dccc0cd00ceb9b11dd24ba838b723.html#main).

lukego:
Is it safe to say that we would benefit from having well-defined configuration options and convenient methods of supplying them e.g. config files and command line?

and further that such config items would best be described with a YANG schema for the Snabb engine and parameters like busy-wait, developer-debug, etc?

kbara:
I'd like to retain ```snabb gc```. This cleanup model isn't robust in the face of situations like losing power or the supervisor process getting hit with a kill -9, OOM killer, etc.

eugeneia:
@kbara `snabb gc` doesn't work (never did) because it fails when a (new, possibly non-snabb) process with a matching pid exists. It could maybe find out the name of that process and disregard it if it is not “snabb” but for now `rm -r /var/run/snabb` is more reliable.

kbara:
Snabb gc wasn't ideal for those reasons, true. After talking to my team, I'm ok with removing it. Thanks.

PR #933, opened 2016-06-08
next: Changes queued for the v2015.06 release

lukego:
Sorry about the wait!


eugeneia:
@lukego Can you merge #926 as well, please?

lukego:
@eugeneia Pulled!

This is the first instance where we have pulled in the complete history of an application-specific fork (`program/lisper`) rather than squashing it. This is something we have talked about being open to in the past but not actually done.

Can be that we will start bumping the limits of default settings for Git log browsing and need to start thinking about what information we are looking for and then how to query it (e.g. using options like `--merges` and `--first-parent` etc).

eugeneia:
@lukego Agree, I only “trust” this history because I reviewed every changed file myself (and reverted a handful of minor unrelated changes I deemed out of scope), the history as shown per GitHub is opaque to me.

PR #935, opened 2016-06-08
virtio/net_device: Fix bug with mergeable rx buffers

lukego:
Snabb could over-report the size of jumbo packets delivered to the VM under certain circumstances.

This was detected when using jumbo frames with VMs that were bootstrapped with nix. Could not be reproduced with Ubuntu guests.

I resisted the urge to refactor the virtio device code more because this will be faster and safer once we complete some related work with Hydra for testing many software combinations.

### DIFF
--- a/src/lib/virtio/net_device.lua
+++ b/src/lib/virtio/net_device.lua
@@ -306,7 +306,16 @@ function VirtioNetDevice:tx_buffer_add_mrg_rxbuf(tx_p, addr, len)
       self.tx.finished = true
    end
 
-   return to_copy
+   -- XXX The "adjust" is needed to counter-balance an adjustment made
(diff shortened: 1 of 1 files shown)


wingo:
Interesting!  Note that https://github.com/Igalia/snabb/pull/354/commits/df8e83c6771a7b892324922f25d5123a69f651d9 implies that this code is running in all cases, whether or not MRG_RXBUF is negotiated, if QEMU first negotitated to have MRG_RXBUF on.

lukego:
I fast-tracked this change onto `next` to support @domenkozar's CI efforts i.e. to try and run the NFV test suite using VM images that were bootstrapped via Nix expressions. (The change is already landed on `max-next`.)

PR #942, opened 2016-06-13
intel1g: canonicalize env var name, pass it through dock.sh

eugeneia:
This makes `dock.sh` pass through the required environment variable for the intel1g selftest, enabling SnabbBot to run it.


src/apps/intel/intel1g.lua
```diff
@@ -35,7 +35,7 @@
 -- Note: section and page numbers in the comments below refer to the i210 data sheet
 
 -- run selftest() on APU2's second/middle NIC:
---  sudo SNABB_SELFTEST_INTEL1G_0="0000:02:00.0" ./snabb snsh -t apps.intel.intel1g
+--  sudo SNABB_PCI_INTEL1G0="0000:02:00.0" ./snabb snsh -t apps.intel.intel1g
```
kbara: I really don't approve of hardcoding PCI addresses, but I don't consider that a blocker for this patch, as it was also previously hardcoded.

PR #934, opened 2016-06-08
vhost_user: Fix & comment "feature cache"

lukego:
Added a long description of the "feature cache" and how it supports reconnecting to a running QEMU.

Added a restriction so that the cache is only considered for the first feature negotiation performed by a vhost-user app instance. This is to avoid using the cached features when renegotiating with a guest e.g. when the guest has switched to a new driver.

Not yet tested! Should fix #932. cc @wingo

PR #945, opened 2016-06-16
Merge max-next-v2016.07-1 into next

eugeneia:
Includes: #889 #866 #904 #905 #934 #935 #936


src/README.md
```diff
@@ -99,6 +99,18 @@ Tables of named input and output links.  These tables are initialized by
...
+— Field **myapp.appname**
+
+Name of the app. *Read-only*.
+
+
+— Method **myapp:start**
+
+*Optional*. Start the app. At this the `input` and `output` link tables are
+populated, and `appname` is set. The app can perform additional initialization
+if required.
```
lukego: The details of this callback are surprising to me.

Based on the name `start` I would expect the callback to be run once when an app instance is first created and before any traffic is processed. Looking at the code it seems like `start()` is actually always called, on every app instances, each time you run `engine.configure()` (even if your new config is identical to the old one).

Could you please update either the name, the description, or the implementation to make the behavior & appropriate uses more obvious?

eugeneia: @lukego @dpino What do you think about the name `configured`, e.g. the callback to be called after an app was configured?

dpino:
@eugeneia @lukego I agree with Luke's point of view, the snippet of code is run every time `configure` is called so better call the event _configure_ instead of _start_. However that's not I originally intended. Originally the trigger of the event was from `ops:start()`, but when doing it from there the links are not set up at that point, so that makes _start_ less useful. So I moved to after the links are setup, but that changes the semantic of the event as trigger points out.

I'm OK with calling it `configure`, but as apps can implement _stop_ I think having a _start_ counterpart makes sense. If going for this solution, the start loop should only be called if ops.start was executed.

eugeneia:
@dpino `new` is the counterpart to `stop`.

eugeneia:
@dpino I would say `new` is the counterpart to `stop`. I think `configure` makes sense, as this callback lets apps react to the way they were configured.

dpino:
@eugeneia Taking a look at Alexander's patch I think it makes sense to call it `configure` too, as he originally called that method `post_config`. Agree with the name change.

lukego:
Thanks for the name change!

I would still like to address these issues before merge:

- Exactly when is `configure` called and what are the intended uses? Seems to me like the implementation is a hook for every time `engine.configure()` is called independent of the content of the configuration changes and that seems potentially excessive to me.
- Do we really want to have both `configure` and `reconfig` callbacks? This is the situation on the submitted branch. I think we should either merge these together or otherwise clearly explain when to use one vs the other. (If we keep both then we should also be consistent about abbreviating their names or not i.e. config vs configure.)

lukego:
I would like to please withdraw my objection to the implementation of `configure`.

I checked how the `relink()` callback had been implemented (added in commit 4f166b3a62daa0809061edd5fbe7b2953743d5ad and removed in commit c4a78856731cdb4bbc062f410cd46042e7ee32fa) and it is exactly the same. Could be seen as premature optimization to make a more complex implementation that runs the callback less frequently e.g. skipping for apps whose own links and configurations have not been touched.

So my only request now is to please reconcile `configure` with `reconfig` either by merging them together somehow or making their names/purposes more clearly distinct :).

dpino:
I think I can explain best what I need if I describe the use cases I'm trying to fix.

I would like to give apps a chance to complete their initialization once they have been configured (with their links setup too). `app:reconfig` is only called if an app already exists in the `app_array` so it cannot be used.

One example that could take advantage of this new event is the NDP app in the lwAFTR.

https://github.com/Igalia/snabb/blob/lwaftr/src/apps/lwaftr/ipv6_apps.lua#L186

There's a case where the NDP app needs to send Neighbour Solicitation packets until a parameter is resolved (`self.dst_eth`). At this moment, we are implementing this logic in the `push` method, because only at that point links are setup. However, we could start sending NS packets as soon as the app and the links are ready.

It could be argued that the links won't be processed until `push` is called in any case, but putting this logic in two different methods would make the code cleaner and the `push` method less bloated IMHO.

Another use case is creating counters by app name. An app name is added to an app on `start`, but since there's no `start` hook apps can react to, the app has to wait until `push` to create its counters and it has to do it only once. Something like:

```lua
function App:push ()
    if not self.added_counters then
       self.counter = counter.open(self.appname.."/counter")
       self.added_counters = true
   end
   local input, output = assert(self.input.input), assert(self.output.output)
   while not link.empty(input) do
      ...
   end
end
```

I think a method where apps could further initialize themselves (in `new` appname or links do not exist), seems necessary.

Solutions I can think of:

* We could discard this new `configure` method and call `reconfig` on `start` too, but that would make the `reconfig` and `start`states less clear. In addition to that, the first time `reconfig` is called, links wouldn't exist.
* I go back to my original request of adding a `start` event (I agree having a `configure` and `reconfig` callbacks would be confusing). I wanted to defer the call of the event until the links were created, but I didn't do it right. Should be something like:

```lua
function apply_config_actions (actions, conf)
   ...
   local start_called = false
   function ops.start (name)
      ...
      start_called = true
   end
   ...
   if start_called then
      for _, app in ipairs(app_array) do
         if app.start then app:start() end
      end
   end
end
```

lukego:
@diego Thanks for providing that context. That helps a lot.

How about implementing this all locally in the `push` method like this?

```lua
function NDP:push ()
   self:maybe_send_ns()
   ... process packets ...
end

-- Send rate-limited neighbor solicitation message.
function NDP:maybe_send_ns ()
   self.next_ns_time = self.next_ns_time or engine.now() -- first message is immediate
   if self.next_ns_time <= engine.now() then
      self:send_ns()
      self.next_ns_time = engine.now() + self.ns_interval
   end
end
```

This seems neat, predictable, and non-bloated to me. The `send_ns()` function will be called at regular intervals and take the appropriate (non-)action based on the current links and configuration. How about it?

Note that `lib.timer` should not be used from apps at all and this seems to be the root of our difficulty here. Those timers are persistent, like crontab entries, and can easily drift out of sync with the app's own lifetime and configuration. You don't want zombie timers accumulating and running in the background and referencing obsolete objects via closures. I suspect that the root of our whole difficulty here is trying to match the lifecycles of apps and `lib.timer` timers which don't align very well. (Maybe we should rename `lib.timer` to `lib.cron` to make its appropriate usage more obvious?)

WDYT?

Incidentally I would quite like to have a `tick()` timer in apps that runs at a fixed frequency e.g. 1 millisecond. This could be used for taking care of administrivia such as checking whether it is time to send a neighbor solicitation message. This is something we could add if it would simplify or speed up existing apps in practice.

dpino:
@lukego Thanks for the detailed explanation, I like this new design more. In fact, before I needed to add a `timer.deactivate` method which I can remove now. I was not aware that apps should not depend on timers to implement their logic.

eugeneia:
@lukego Last commit renames and documents the callback as per #946.

lukego:
Merged. Thanks and sorry about the wait!

PR #950, opened 2016-06-20
Asynchronous Snabb

eugeneia:
This implements @lukego’s asynchronous design from #656 (faithfully, I hope).

- Link capacity is increased to 1000
- A new constant `engine.pull_npackets` is introduced with is a floored 10th of the link capacity (e.g. 100)
- `link.full` and `link.nwriteable` functions are removed
- All usage of the removed functions is removed/replaced with use of `engine.pull_npackets`

This means that all I/O apps pull at most `engine.pull_npackets` per breath. All other apps will forward all packets they process (usually all packets in their input links) and put them on output links (where they might be dropped) each breath. This yields the advantage that packets will be dropped where the bottleneck actually is. It also simplifies the Snabb API.

If we can find agreement on this I will happily remove the `[sketch]` tag. :-) 

Cc @lukego @plajjan @kbara @petebristow


plajjan:
`link.full()` still seems to have a purpose for a QoS app. Should we just not use `link.full()` from normal applications but let it remain for special apps that need to detect congestion on egress NIC? Or am I missing something? (just back from vacation ;)

eugeneia:
I reverted the removal of `link.full` and `link.nwritable`, given their potential use in QoS apps. Also fixed some bugs. I am removing the `[sketch]` and pushing for this to land! :-)

eugeneia:
@kbara Do you think this could land in the September release?

kbara:
Thanks for the ping on this; I'll review it now.

kbara:
Nice work, I've PR'd it to next.

PR #931, opened 2016-06-07
App-local statistics counters

eugeneia:
This branch implements per-app statistics counters and uses this feature to implement statistics for Intel_app (per pciaddress actually) and vhost_user.

The basic mechanism is that each app now has a `shmpath` (automatically set based on its name) to which `core.shm.path` is set to when its methods are executed. `core.counter` is updated to use `core.shm.resolve`, so counters opened in `app:new` are created under the app's `shmpath`. Another convention in this PR is that apps store their counters in their `counters` field.

[doc/statistics.md](https://github.com/eugeneia/snabb/blob/statistics-superset/src/doc/statistics.md#statistics-counters-in-snabb) describes the schema used for the counters defined by this PR. It deviates from related standards in the way it counts packets. E.g. it defines  broadcast packet counts ⊆ multicast packet counts ⊆ total packet counts. I think this makes the most sense because it allows app authors to choose how fine grained their stats are. I also don't think we should buy into the [ifType](https://www.iana.org/assignments/ianaiftype-mib/ianaiftype-mib) madness.

The SNMP code formerly in `intel10g.lisp` was extracted to `lib.ipc.shmem.iftable_mib` using the schema / API described above.

Anything in this PR is open for discussion change (obviously), but I would like to get it merged and then develop these features incrementally to get things rolling. E.g. @alexandergall has pointed out that we need more data types than counters, etc. My rationale is that with this PR merged we can start recording important statistics in other apps.

App counters (as well as link counters) can be read using `snabb top -c`, e.g.:

```
bash-4.3# program/snabbnfv/selftest.sh > nfv.log &
bash-4.3# ./snabb top -c A_Virtio
 dtime                          1,465,310,436                 
 rxbytes                        18,186,581,184                
 rxdrop                         0                             
 rxmcast                        9                             
 rxpackets                      12,012,296                    
 txbytes                        85,150,962                    
 txmcast                        0                             
 txpackets                      990,127                       
 type                           4,097                         
bash-4.3# ./snabb top -c 'A_NIC.tx -> A_Virtio.rx'
 dtime                          1,465,310,436                 
 rxbytes                        174,939,590                   
 rxpackets                      2,032,431                     
 txbytes                        174,939,590                   
 txdrop                         0                             
 txpackets                      2,032,431                     
bash-4.3# ./snabb top -c 0000:01:00.0
 dtime                          1,465,310,436                 
 macaddr                        0                             
 mtu                            9,014                         
 promisc                        1                             
 rxbcast                        0                             
[...]

PR #953, opened 2016-06-24
Various statistics counters for NFV apps

eugeneia:
Depends on: #931 #949 

This adds statistics/diagnostics counters for various NFV apps.

- L2TPv3 app counts `rxerrors`  broken down into why
- ESP tunnel counts `rxerrors` (not insanely useful, can most of the time be inferred from link statistics)
- nd_light has `status` (initially link down, link up once neighbor established), `rxerrors` (erroneous packets from south dropped, boken down into why), `txerrors` (erroneous packets from north), and `txdrop` (packets from north dropped while link down)
- pcap_filter counts `rxerrors` and `sessions_established`
- rate_limiter counts `txdrop` (again, not insanely useful)

If the recurring boilerplate code wasn't obvious until now, this PR shows it pretty prominently. My next step will be to abstract away the boilerplate I added in my recent PRs... :-)


kbara:
Both dependencies have been merged into next, but not via the same upstream branch. Any thoughts on the most reasonable way to proceed on this patch set short of waiting for an extra release, @lukego ?

eugeneia:
@kbara You could merge `next` into `kbara-next`, then merge this PR.

kbara:
Good idea; I've done so.

PR #954, opened 2016-06-28
Intel1g fixes

alexandergall:
Commits     acc3595,    1024ea1 and 93d1c53 should be no-brainers. I'm not sure why the author has used local functions to manipulate registers instead of the standard utilities of `lib/hardware/register.lua`.  I have stuck to that method but believe that this should eventually be changed to have more uniformity in drivers. Also, this driver should probably be included in the `apps/intel/intel_app` framework, I guess. 

Commit 05565fe adds MTU-handling similar to what's already in `intel10g`, but this time I have enforced egress drops of oversized frames. Drop counters are only exposed through SNMP via 1ccd4d7. Note that the functionality relies on the assumption that all VLAN-handling by the NIC is disabled. This might require some discussion (also in the context of `intel10g` and future drivers).

Commit 1ccd4d7 implements "old-style" SNMP support like in the current `intel10g` driver. This will be removed once #931 (or something similar) is on master, I suppose. I wouldn't mind if this commit is excluded from this PR (i.e. I would maintain it on the `l2vpn` branch until it becomes obsolete unless other people want to use SNMP). Note that I reverted to the utilities from `lib/hardware/register.lua` for the statistics registers to make use of the accumulator functionality.

### DIFF
--- a/src/apps/intel/intel1g.lua
+++ b/src/apps/intel/intel1g.lua
@@ -50,6 +50,8 @@ local bits, bitset = lib.bits, lib.bitset
 local compiler_barrier = lib.compiler_barrier
 local tophysical = core.memory.virtual_to_physical
 local register = require("lib.hardware.register")
+local mib = require("lib.ipc.shmem.mib")
+local timer = require("core.timer")
 
 Intel1g = {}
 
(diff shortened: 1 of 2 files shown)


eugeneia:
@petebristow Since this PR overlaps a lot with #897 could you pull any changes that are not covered by #897 (e.g. applicable to your HEAD), *except* 1ccd4d7.

@alexandergall I don't want to accept 1ccd4d7 because I just extracted that code from intel10g in #931 and there is no need to duplicate it, besides I don't want SNMP / ifMIB / RFC7223 specific code in drivers. Once #897 lands I will merge it with my statistics branch and `lib.ipc.shmem.init_snmp` from #931 will be able to attach to intel1g through the standard statistics interface.

Issue #951, opened 2016-06-22
NFV appears to allocate


wingo:
When I do a perf -p PID for a snabb-nfv PID and then look at it with `perf report`, I see:

```
   6.09%  snabb    snabb              [.] lj_BC_TGETS
   5.62%  snabb    perf-546.map       [.] 0x000000000fa7face
   5.06%  snabb    perf-546.map       [.] 0x000000000fa7eaf5
   4.20%  snabb    snabb              [.] lj_meta_tget
   2.94%  snabb    snabb              [.] lj_cdata_index
   2.69%  snabb    snabb              [.] lj_alloc_free
   2.42%  snabb    snabb              [.] lj_BC_JLOOP
   2.33%  snabb    snabb              [.] lj_BC_CALL
   2.02%  snabb    snabb              [.] lj_tab_getstr
   2.00%  snabb    snabb              [.] lj_tab_get
   1.92%  snabb    snabb              [.] lj_ctype_getfieldq
   1.86%  snabb    snabb              [.] lj_BC_RET1
   1.84%  snabb    snabb              [.] gc_sweep
   1.69%  snabb    snabb              [.] lj_cconv_ct_ct
   1.65%  snabb    snabb              [.] lj_cf_ffi_meta___index
   1.64%  snabb    snabb              [.] lj_carith_op
   1.57%  snabb    libc-2.21.so       [.] __memmove_avx_unaligned
   1.29%  snabb    snabb              [.] lj_BC_FUNCC
   1.27%  snabb    snabb              [.] lj_alloc_malloc
   1.26%  snabb    snabb              [.] lj_cont_ra
   1.08%  snabb    snabb              [.] lj_BC_RET0
   1.05%  snabb    snabb              [.] lj_BC_UGET
   1.04%  snabb    snabb              [.] lj_tab_setstr
   1.04%  snabb    snabb              [.] cksum_generic
   0.93%  snabb    snabb              [.] lj_BC_JFUNCF
   0.90%  snabb    snabb              [.] release_unused_segments
   0.89%  snabb    snabb              [.] propagatemark
   0.88%  snabb    snabb              [.] lj_vm_exit_interp
   0.88%  snabb    snabb              [.] lj_vm_call_dispatch_f
   0.88%  snabb    snabb              [.] lj_meta_lookup
   0.86%  snabb    snabb              [.] lj_ctype_intern
   0.84%  snabb    snabb              [.] lj_BC_TGETV
[...]

PR #956, opened 2016-06-29
next: Changes queued for the v2016.07 release

lukego:
Changes queued up for v2016.07.

Fixes needed:

- [x] Selftest in `core.timer` is printing tens of thousands of debug messages. Have to account for why this has started happening and likely remove some debug printouts (seems excessive to print a message on every timer invocation even when developer_debug is enabled... could be a use case for timeline instead soon).
- [x] NFV selftest is failing CI. Could be related to migration of davos from Ubuntu to NixOS that is ongoing. @eugeneia?
- [ ] [Performance regression](https://hydra.snabb.co/build/81169/download/2/report.html) detected by Hydra tests developed over on #951. This benchmark is automatically executing as the [issue-951](https://hydra.snabb.co/jobset/domenkozar-sandbox/issue-951) jobset. Results can be found by clicking an Evaluation (i.e. test run) and locating the `benchmark-report` job.


src/core/main.lua
```diff
@@ -28,8 +29,8 @@ ffi.cdef[[
       extern char** argv;
 ]]
 
-_G.developer_debug = false
-debug_on_error = false
+_G.developer_debug = lib.getenv("SNABB_DEBUG") and true
+debug_on_error = _G.developer_debug
```
eugeneia: @lukego Ok where do I start: the breaking change is that this sets `_G.developer_debug` to `nil` instead of `false`. Then in `core.timer`, `debug` is set to `_G.developer_debug`. The next thing I can't explain:

- When `_G.developer_debug` is `true` or `false` the value of `timer.debug` will be `true` or `false` respectively
- However when `_G.developer_debug` is `nil` the value of `timer.debug` is the following table:

```
Snabb> for k,v in pairs(timer.debug) do print(k,v) end
traceback	function: builtin#141
setlocal	function: builtin#133
getupvalue	function: builtin#134
setupvalue	function: builtin#135
upvalueid	function: builtin#136
getlocal	function: builtin#132
getregistry	function: builtin#126
getinfo	function: builtin#131
sethook	function: builtin#138
setmetatable	function: builtin#128
upvaluejoin	function: builtin#137
gethook	function: builtin#139
debug	function: builtin#140
getmetatable	function: builtin#127
setfenv	function: builtin#130
getfenv	function: builtin#129
```

eugeneia: This patch would fix it, but I am honestly interested in why this happens.

```diff
diff --git a/src/core/main.lua b/src/core/main.lua
index f736e25..343e2f9 100644
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -29,7 +29,7 @@ ffi.cdef[[
       extern char** argv;
 ]]
 
-_G.developer_debug = lib.getenv("SNABB_DEBUG") and true
+_G.developer_debug = (lib.getenv("SNABB_DEBUG") and true) or false
 debug_on_error = _G.developer_debug
 
 function main ()
```

lukego:
@eugeneia Hah! I had just done a `git bisect` and tracked the problem to the same commit and also written a message saying I don't understand then problem. Now I see your note above and suddenly it makes sense :).

The issue is the [semantics of nil values in Lua tables](http://lua-users.org/wiki/StoringNilsInTables). Lua does not permit `nil` as a table value and the effect of `foo.bar=nil` is to remove the key `bar` from table `foo`.

So if we set `debug = nil` in the `timer` module and then reference the key `debug` it will see that the key is missing and delegate the lookup to the global environment (via metatable) where it will resolve as `_G.debug` which is the Lua debug module.

I will push a fix along the lines of your patch above.

lukego:
Hydra has found a [performance regression](https://hydra.snabb.co/build/81169/download/2/report.html) from `master` to `next` that needs to be resolved. Go Hydra! I updated the status comment on the PR text.

Looks like it should be bisectable because one branch is consistently outperforming the other on lugano servers. @domenkozar How might I execute this test as part of a manual `git bisect`? I mean, is there an easy way to interactively (on `lugano-1`) run the test case based on a locally checked out Snabb and eyeball the result to see if it is good or bad? I am thinking a `nix-build ...` with the right args is all I need?

domenkozar:
@lukego with `git bisect` that would indeed be the easiest. I'll look into https://github.com/snabblab/snabblab-nixos/issues/35 over the weekend, which should be what we need here.

lukego:
Are you sure we need snabblab/snabblab-nixos#35? Seems like I could run the test locally on `lugano-1` with a plain `nix-build` if I would choose the right nix expression and arguments from in the `snabblab-nixos` repo?

domenkozar:
@lukego sure. Use inputs from https://hydra.snabb.co/build/81169/#tabs-buildinputs and use `--arg snabbBsrc /home/luke/snabb/`

lukego:
Trying that. Relatedly: is there also a convenient way to run one benchmark by name e.g. `l2fwd_snabb=master_dpdk=1-8-0_qemu=2-4-1-num-1`? Seems a bit round-about to use the Hydra jobset as an entrypoint from the command-line (?)

lukego:
I tried this to test the master branch:

```
[luke@lugano-1:~/git/snabblab-nixos]$ nix-build jobsets/snabb-matrix-packet-ABC.nix --arg snabbAname '"bisect"' --arg snabbAsrc ~/git/snabb-bisect --arg numTimesRunBenchmark 1 --arg benchmarkNames '[ "dpdk" ]' --arg reportName '"report-by-snabb"'
```

but it gave me an error:

```
setting SOURCE_DATE_EPOCH to timestamp 1467385365 of file snabb-bisect/lib/pflua/tools/pflua-asm
warning: file snabb-bisect/lib/pflua/tools/pflua-asm may be generated; SOURCE_DATE_EPOCH may be non-deterministic
patching sources
patching file src/program/snabbnfv/test_env/test_env.sh
Reversed (or previously applied) patch detected!  Assume -R? [n] 
Apply anyway? [n] 
Skipping patch.
1 out of 1 hunk ignored -- saving rejects to file src/program/snabbnfv/test_env/test_env.sh.rej
configuring
grep: : No such file or directory
grep: : No such file or directory
no configure script, doing nothing
building
ln: failed to create symbolic link 'src/snabb': File exists
builder for ‘/nix/store/d6z1a94vnqxr3xblxv8krrp3z6w1cwsp-l2fwd_snabb=bisect_dpdk=16-04_qemu=2-4-1-num-1.drv’ failed with exit code 1
cannot build derivation ‘/nix/store/awll8igr3kbf04nivfqj9gyhg9pkld99-csv-builder.sh.drv’: 1 dependencies couldn't be built
cannot build derivation ‘/nix/store/xqm6g4qw42kbmcq7wckjhglmsijnh337-csv-builder.sh.drv’: 1 dependencies couldn't be built
cannot build derivation ‘/nix/store/w758adxjpygkp2i7mwvpavbw64m0y01s-snabb-report-csv.drv’: 1 dependencies couldn't be built
error: build of ‘/nix/store/w758adxjpygkp2i7mwvpavbw64m0y01s-snabb-report-csv.drv’ failed
...

Do you know what is wrong?

**Comment by lukego** [comment]
Awesome! I was able to use `nix-build` to run the benchmark locally and then isolate the problem using `git bisect` and related commands. Great to be able to address all versions of all software dependencies with one command-line!

Here is the command that worked for running one test against local checkouts:

```
$ sudo lock nix-build jobsets/snabb-matrix-packet-ABC.nix \
                      -A benchmarks.l2fwd_snabb=bisect_dpdk=16-04_qemu=2-4-1-num-1 \
                      --arg numTimesRunBenchmark 1 \
                      --arg snabbAname '"bisect"' \
                      --arg nixpkgs ~/git/nixpkgs
                      --arg snabbAsrc ~/git/snabb-bisect \
                      --arg benchmarkNames '[ "dpdk" ]' \
                      --arg reportName '"report-by-snabb"' \
```

which runs the nfv/dpdk benchmark and prints output like this:

```
Processed 100.0 million packets in 36.74 seconds (6400001562 bytes; 1.39 Gbps)
Made 1,338,361 breaths: 74.72 packets per breath; 27.45us per breath
Rate(Mpps):     2.722
```

The [Hydra report](https://hydra.snabb.co/build/81169/download/2/report.html) shows that "good" is around 2.7 Mpps and "bad" is around 2.5 Mpps so for `git bisect` I used 2.6 Mpps as the cutoff point.

### Diagnosis

The ~10% performance regression comes with the new statistics counters (#931) that landed on `kbara-next` with commit 9906acaaf69147e3d1869d4075d8609bdc8c7536 and on `next` with commit 00706efafc2bde086ae013d3db4b347cb5914b7f. So @eugeneia we need to find the slow part, optimize that, and land a fix on `next`.

Great to have performance regression tests, let's do more of this!!

lukego:
I cherry-picked the virtio-net fix df8e83c6771a7b892324922f25d5123a69f651d9 from lwaftr repo for renegotiating virtio-net options with guests that switch drivers while running.

eugeneia:
@lukego I have tracked down the regression to the statistics counters in vhost_user. I pushed e3fcbbf to #931 which reverts a premature optimization I did, and this removes half of the slowdown. That leaves us with a 2.5% regression. No looking into what can be done about the remaining slowdown.

eugeneia:
Profiling tells me that 10% of time is spent in `counter.add`, and does not highlight vhost_user in particular. Theory: more counters → more work. Possible solutions I can think of:

- optimize `counter.add`
- update vhost_user to invoke `counter.add` less, e.g. aggregate updates like intel10g does

eugeneia:
Alternatively we could omit counting successfully processed rx/tx packets in vhost_user, and rely on the existing link statistics to reflect these.

lukego:
@eugeneia Looking at a [new Hydra run](https://hydra.snabb.co/build/85492/download/2/report.html) it looks like the same performance regression is there on both `next` and your `statistics-superset` branch. If you push new code to your branch it should automatically rerun. Hopefully you can find results by clicking around the [snabb-new-tests](https://hydra.snabb.co/project/snabb-new-tests) project on Hydra. Currently we have two relevant CI jobs, `regression-min-2016.07` running a quick 5-iteration benchmark and `regression-2016.07` running a thorough 30-iteration benchmark.

@domenkozar I would like to generalize these CI jobs. Instead of specifically testing for a performance regression on the 2016.07 release they could be continuously showing the comparative performance of a collection of branches: `master`, `next`, `lukego/optimize`, `eugeneia/optimize`, `wingo/optimize`, `kbara/optimize`, etc. That way everybody can always interact with the CI simply by pushing to their optimization branch and waiting to see the result. Just a matter of renaming the jobsets and locking in the branch names?

Meta: I reckon that all "out of band" information we are discussing should be brought "in band" somehow. For example, if profiler information is required for interpreting results then the CI tests should include runs that are profiled and publish the reports. This way we are all looking at the same information and can see exactly how it was produced.

lukego:
@domenkozar This is really cool testing to have up and running, btw! I am especially impressed that Hydra is smart enough to reuse all previous results that are still valid. So if one branch is updated then its benchmarks will be rerun but the results for the unmodified branches will be reused. This seems extremely efficient and means we can include a lot of branches in the benchmark campaigns.

domenkozar:
@lukego I've renamed the CI jobs to `next-regression-benchmarks` and `next-regression-benchmarks-small`: https://hydra.snabb.co/jobset/snabb-new-tests

Once `/optimize` branches exist we can also change those :) 

By "profiler information" do you mean we should write those tests and include them? I'm +1 on that, let's add more regressions tests.

domenkozar:
Looks like this errored out due to https://github.com/snabblab/snabblab-nixos/issues/52

PR #964, opened 2016-07-08
Fixes for next/2016.07

eugeneia:
- This should fix the performance regression reported in #956 (only did a quick comparison, we should wait for hydra to confirm) by removing statistics callbacks from virtio code and reverting an attempted optimization that backfired
- Also includes #963


lukego:
Hydra report here: https://hydra.snabb.co/build/136747/download/2/report.html

Great that we are putting our new benchmarking infrastructure into practice! The report clearly shows a subtle performance regression that is very easy to miss in interactive benchmarking due to subtle variation between runs.

I am having a read of some JIT dumps to understand why counter bumping is influencing the benchmark scores so much. Seems like it should be a very cheap operation in principle.

PR #966, opened 2016-07-14
Several Snabb-bot improvements

dpino:
* Remove temporary dirs before creating them.
* Allow Snabb-bot to work with GitHub private repositories.
* Use OAuth CLIENT_ID and CLIENT_SECRET parameters to increase number of requests per hour.


src/scripts/snabb_bot.sh
```diff
@@ -26,13 +26,18 @@ function init {
     fi
     export logdir="$SNABBBOTDIR/log"
     export tmpdir="$SNABBBOTDIR/tmp"
+    rm -Rf "$SNABBBOTDIR" "$logdir" "$tmpdir"
```
eugeneia: `$logdir` is actually supposed to persist (it caches results, otherwise everything is recomputed all the time).

PR #971, opened 2016-07-19
pci_bind: A utility to bind/unbind PCI cards

kbara:
Everyone seems to have their own shell scripts to bind/unbind PCI cards. This patch adds a tool to consolidate the key options into Snabb. It's more Snabb-like to not need to have Snabb *and* an external shell script to use Snabb.

```
  pci_bind [-ubir] <PCI-ADDRESS>
  pci_bind [-ah]

  -a, --all
                             Restore all PCI addresses to kernel management
  -b, --bind <PCI-ADDRESS>
                             Bind the PCI address to the kernel
  -h, --help
                             Print usage information
  -i, --info <PCI-ADDRESS>
                             Print out the associated Linux interface name
  -r, --remove <PCI-ADDRESS>
                             Remove the specified PCI address
  -u, --unbind <PCI-ADDRESS>
                             Unbind the PCI address, so Snabb can use it

```

See the README for more details.

PR #974, opened 2016-07-25
Merge max-next-v2016.08-1 into 'next'

eugeneia:
Includes: #969 #971


lukego:
Can we move `virtual_ether_mux` into e.g. `program/snabbnfv/traffic`? If it wants to live in lib/ then I think it needs documentation.

eugeneia:
@lukego Good point, done.

lukego:
@eugeneia Looks like the renaming of `packetblaster_bench.sh` to `dpdk_bench.sh` is breaking the Hydra test script. The [lukego/next-tributary jobs](https://hydra.snabb.co/jobset/lukego/next-tributary#tabs-jobs) shows the DPDK benchmarks failing on max-next with an [error finding the script](https://hydra.snabb.co/build/168677/download/1/log.txt). Tricky... the new name is better, but renaming successfully requires cross-repo changes i.e. `snabblab-nixos` test expression needs to be updated to work with both the new and the old name.

On balance I think it makes sense to revert the change of the script name. This keeps the interface compatible for external test scripts (esp. Hydra's nix exp) and should mean that the `next-tributary` jobset will successfully run the benchmarks to help me avoid accidentally merging a performance regression onto `next`.

Could you please push a revert of that name change and check that the `lukego/next-tributary` successfully benchmarks `max-next` & then resubmit the PR? (Sorry for the inconvenience. Intention is to establish a workflow where Hydra successfully benchmarks the *-next branches before I merge them to `next`.)

eugeneia:
@lukego I have added a symlink instead, lets see if that works.

lukego:
@eugeneia That did the trick, thanks!

Going forward could I please pull directly from `max-next` rather than creating the dedicated branches like `max-next-v2016.08-1`? This is just because I am now doing extra testing of `max-next` via my Hydra jobset and that helps me to merge with confidence. The consequence would be that when you have a PR open from `max-next` you would need to only push commits that you consider ready for merge (since they will be automatically added to the PR).

I actually pulled `max-next` to get these changes now. I realize now that this brought along #975 before you PR'd that. I hope that is okay with you.

Closing now because the relevant commits have already landed.

PR #977, opened 2016-07-31
Anti-replay (RFC4303 Appendix A)

vanfstd:
This implements and adds tests for IPsec anti-replay as per RFC4303 Appendix A, except for the resynchronization (App. A3) which will get its own PR.
Note that the test cases produce "auditable events" which end up on stdout.

Feedback appreciated.


vanfstd:
Committer email was wrong, please disregard.

Issue #976, opened 2016-07-29
NFV test matrix preview


lukego:
Check out this new [NFV test matrix report](https://hydra.snabb.co/build/176830/download/2/report.html)!

This is a fully automated workflow where Hydra detects changes on branches under test, executes many end-to-end benchmarks (>10,000) with VMs in different setups, and produces a report that shows benchmark results both in broad overview and slice-and-diced by different factors. The tests run on a 10-machine cluster and take around one day to complete. *Awesome work, @domenkozar!!!*

We should be able to hook this up to upstream branches including `master` and `next` once we land #969 that allows Snabb NFV to run without a physical hardware NIC. Just for now it is connected to a couple of branches that do include this PR.

### Observations

This is a preview in the sense that we have not been able to thoroughly sanity-check the results yet and some of them are different than we have seen in the past. So take all of this with a grain of salt.

On the **Overall** graph we can see that the two Snabb branches being tested, `matrix` and `matrix-next`, look practically identical.

On the **iperf** graph we can see results clustered around three places: 0 Gbps (failure), 6 Gbps, and 10-15 Gbps. On the **iperf configuration** graph we can see the explanation: the *filter* tests are failing (likely our packet filter is blocking iperf); the *ipsec* tests are delivering ~6 Gbps; and the baseline and L2TPv3 tests are delivering ~10-15 Gbps.

On the **l2fwd** graph we can see that most results are failing. Question is, why? I am not sure. The **l2fwd success and failure** stats may point to QEMU version and packet size being important factors? (These tests are much more reliable with SnabbBot so what is the difference here? Can be related to the guest kernel.)

### Onward

So! Now we can automatically see whether changes are good for Snabb NFV. We want to merge changes that make the humps higher (more consistency) and that move the area under the curves to the right (faster average). Then we also want harsher test cases that reveal problems by spreading the curves out and moving them to the left :).

If this works well in practice then we could gradually add coverage for all the other Snabb applications too.

### Feedback?

Questions, suggestions, comments?

kbara:
Looks awesome overall. The main thing that worries me is the lack of any indication of expected error/noise - when I see two overlapping graphs that diverge a bit, I have no idea how likely it is to be significant or random.

lukego:
[...]

PR #981, opened 2016-08-01
next: Changes queued for the v2015.08 release

lukego:
Here is some good stuff for v2016.08.

@eugeneia The one blocker I see is to resolve the counter-related performance regression that we introduced in v2016.07. Could be done either with optimization or with the minimum sufficient disabling of functionality. (Having the minimal difference that removes the problem would also make it easier for people to help with optimization because then there is a very clear A/B comparison to make.)

@eugeneia Sorry but I removed your branch from the Hydra jobset `snabb-new-tests/releases-lugano-small-20x`. I want to keep this as a stable configuration that tests releases only. I think that for development you can define a new jobset that tests e.g. `v2016.06` (good) and `v2016.07` (bad) and `max-dev-branch` (work in progress) and it should show you what you need to see.

Then once you push to `max-next` the changes will be benchmarked on the `lukego/next-tributary` jobset. (This may even reuse the results of your run if nix detects from the sha256 hashes that the code on max-next has already been tested separately.)


lukego:
Cool! Now we have merged the performance regression fix (#983) and we can see what Hydra reckons about this as a release candidate. Could be that the full test matrix (#976) is not available until tomorrow.

lukego:
@eugeneia This release is good to go from my perspective. I updated the [releases benchmark](https://hydra.snabb.co/job/snabb-new-tests/releases-lugano-small-20x/benchmark-reports.report-by-snabb/latest/download-by-type/file/HTML) to include `next` and it shows that the regression we saw in v2016.07 is gone.

The full test matrix is harder to interpret... work in progress... I have some fixes to push that should have a big impact there (either for 2016.09 or 2016.08.1) by resolving most of the error cases.

PR #982, opened 2016-08-01
Merge Max's "Extended support for managing shared memory #972"

kbara:


eugeneia:
As I just commented in #972 I don't think this is ready yet.

kbara:
Re-opened as per Eugenia's comment on #972 about a false alarm.

eugeneia:
Thanks, again sorry for the confusion!

PR #983, opened 2016-08-01
Revert vhost_user statistics

eugeneia:
Reverts statistics for vhost_user app to avoid the performance regression introduced by it.

See https://hydra.snabb.co/build/178439/download/2/report.html

Depends on #982

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -104,6 +104,13 @@ the engine for use in processing and are *read-only*.
 Name of the app. *Read-only*.
 
 
+— Field **myapp.shm**
+
+Can be set to a specification for `core.shm.create_frame` during `new`. When
+set, this field will be initialized to a frame of shared memory objects by the
+engine.
+
+
 — Method **myapp:link**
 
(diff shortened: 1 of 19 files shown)


lukego:
@eugeneia @kbara Interesting workflow issue :-)

This PR depends on #982 but that has not landed on `next` yet and I am not the upstream (@kbara is).

So there are a few different ways that this change could land on next:

1. I could merge this onto next. This way #982 is coming along for the ride and landing before @kbara includes it in a PR.
2. @eugeneia could rebase this PR to remove the dependency. However, this is likely wasted work if #982 is ready to land anyway.
3. We could make @kbara the upstream for this PR and then she can combine it with #982 on `kbara-next` and PR the combination to `next`.

How about we go with option 3 i.e. that @kbara becomes upstream for this PR and we humbly request a fast-track on it? (I already changed Assignee.)

lukego:
Oh, sorry, my bad!

I am the upstream for #982 and @kbara has already asked me to merge the dependent changes.

Sorry about the noise :).

PR #990, opened 2016-08-10
Add method 'exists' to core/shm

dpino:
Trying to open a shared memory object which has not been mapped yet results in an error. For instance:

```
core/shm.lua:37: shm open error (26039/shm/selftest/obj):No such file or directory
```

I think it would be useful to have an `exists` method which could be used to check whether an object exists before trying to open it.


eugeneia:
I have previously done this by using `shm.children` and probing the resulting table. Alternatively we could use `pcall` to catch the error, but that would be less granular. Obviously convenient, merged (I will edit the documentation a bit to describe the return type). :+1:

PR #999, opened 2016-08-23
Packet.lua: add clone_to_memory

kbara:
It's useful to be able to clone packets to arbitrary memory, not just freshly allocated packets. In the lwaftr, we need this facility for cloning packets to within ctables that embed, rather than point to, packets.

@eugeneia

### DIFF
--- a/src/core/packet.lua
+++ b/src/core/packet.lua
@@ -71,12 +71,17 @@ function new_packet ()
    return p
 end
 
+-- Create an exact copy of srcp at memory pointed to by dstp.
+function clone_to_memory(dstp, srcp)
+   ffi.copy(dstp, srcp, srcp.length)
+   dstp.length = srcp.length
+   return dstp
+end
+
 -- Create an exact copy of a packet.
 function clone (p)
    local p2 = allocate()
-   ffi.copy(p2, p, p.length)
-   p2.length = p.length
-   return p2
+   return clone_to_memory(p2, p)
 end
 
 -- Append data to the end of a packet.


eugeneia:
How about calling the function `to_pointer` (as a counterpart of `packet.from_pointer`)?

kbara:
I'd  be perfectly happy with that; it sounds like a useful improvement.

eugeneia:
On second look `clone_to_memory` does something different than I had thought: it expects the destination to point to a `struct packet` (it sets `dstp.length`), so its not really the inverse of `from_pointer`. Wouldn't it make more sense to add an optional destination arguments to `clone` and `new_packet`?

```
-- Create a new empty packet.
function new_packet (ptr)
   local p = ffi.cast(packet_ptr_t, ptr or memory.dma_alloc(packet_size))
   p.length = 0
   return p
end

-- Create an exact copy of a packet.
function clone (p, p2)
   if not p2 then p2 = allocate() end
   ffi.copy(p2, p, p.length)
   p2.length = p.length
   return p2
end
```

kbara:
The point, as I originally said, is to have a fully valid struct packet at the end - this originated (albeit in a more complex form) from the needs of the lwaftr's memory-bounded fragment reassembly code.

I don't particularly want to modify new_packet. 

I want to be able to clone a packet to a memory address, and have packet.lua take care of all the details that depend on how struct packet is currently defined (in the lwaftr, we use a more complex struct). I don't particularly like adding an extra parameter to clone, because at that point the arguments are src, (optional dst) rather than dst, src as is found in most of the codebase, which is asking for stupid bugs.

eugeneia:
Closing as per #998.

PR #1000, opened 2016-08-26
Handling Loss of Synchronization due to Significant Packet Loss (RFC4303 Appendix A3)

vanfstd:
This implements and adds tests for RFC4303 Appendix A3, re-synchronization of the anti-replay mechanism in the event of massive packet loss.

### DIFF
--- a/src/lib/ipsec/esp.lua
+++ b/src/lib/ipsec/esp.lua
@@ -169,16 +169,17 @@ function esp_v6_decrypt:decapsulate (p)
       -- Assuming b) means 'in time', since systematic loss could stall resync indefinitely.
       self.decap_fail = self.decap_fail + 1
       if self.decap_fail >= ESP_RESYNC_THRESHOLD then
-         local seq_high_tmp = seq_high
+         local resync_start
          if seq_high >= 0 then -- We failed to decrypt in-place, undo the damage to recover the original ctext
             gcm:encrypt(ctext_start, iv_start, seq_low, seq_high, ctext_start, ctext_length, ffi.new("uint8_t[?]", gcm.AUTH_SIZE))
+            resync_start = seq_high + 1
          else
-            seq_high_tmp = self:seq_high() -- use the last seq_high we saw if it looked replayed
+            resync_start = self:seq_high() + 1 -- use the last seq_high we saw if it looked replayed
          end
          self.decap_fail = 0 -- avoid immediate re-triggering if resync fails
-         seq_high_tmp = self:resync(p, seq_low, seq_high_tmp + 1, ESP_RESYNC_ATTEMPTS)
-         if seq_high_tmp >= 0 then
-            seq_high = seq_high_tmp
+         local seq_high_resynced = self:resync(p, seq_low, resync_start, ESP_RESYNC_ATTEMPTS)
+         if seq_high_resynced >= 0 then
+            seq_high = seq_high_resynced
             -- resynced! the data has been decrypted in the process so we're ready to go
          else
             return false
--- a/src/lib/ipsec/esp.lua
+++ b/src/lib/ipsec/esp.lua
@@ -48,6 +48,8 @@ local PAYLOAD_OFFSET = ETHERNET_SIZE + IPV6_SIZE
 local ESP_NH = 50 -- https://tools.ietf.org/html/rfc4303#section-2
 local ESP_SIZE = esp:sizeof()
 local ESP_TAIL_SIZE = esp_tail:sizeof()
+local ESP_RESYNC_THRESHOLD = 4 -- Resync after this many consecutive failed decaps
+local ESP_RESYNC_ATTEMPTS = 6 -- Try this many increments of seq_high
 
 function esp_v6_new (conf)
    assert(conf.mode == "aes-128-gcm", "Only supports aes-128-gcm.")
@@ -102,7 +104,7 @@ function esp_v6_encrypt:encapsulate (p)
    self.esp_tail:pad_length(pad_length)
    self:next_seq_no()
    local ptext_length = payload_length + pad_length + ESP_TAIL_SIZE
-   gcm:encrypt(payload, self.seq, self.seq, payload, ptext_length)
+   gcm:encrypt(payload, self.seq, self.seq:low(), self.seq:high(), payload, ptext_length, payload + ptext_length)
    local iv = payload + ESP_SIZE
    local ctext = iv + gcm.IV_SIZE
    C.memmove(ctext, payload, ptext_length + gcm.AUTH_SIZE)
@@ -127,6 +129,7 @@ function esp_v6_decrypt:new (conf)
    o.window_size = conf.window_size or 128
    assert(o.window_size % 8 == 0, "window_size must be a multiple of 8.")
    o.window = ffi.new(window_t, o.window_size / 8)
+   o.decap_fail = 0
    return setmetatable(o, {__index=esp_v6_decrypt})
 end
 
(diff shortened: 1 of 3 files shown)


src/lib/ipsec/esp.lua
```diff
@@ -168,10 +162,59 @@ function esp_v6_decrypt:decapsulate (p)
                    "flow_id=" .. tostring(self.ip:flow_label()) .. ", " ..
                    "reason='" .. reason .. "'";
       logger:log("Rejecting packet ("..info..")")
-      return false
+      -- RFC4303 is somewhat unclear on a) whether or not packets that look replayed are eligible to trigger
+      -- recyns, and b) what 'consecutive' means in "number of consecutive packets that fail authentication":
+      -- Consecutive in time? Consecutive in their seq_low's?
+      -- Assuming a) should be 'yes' since the to-be-resynced packets might have seq_low's that just happen
+      -- to fall inside the replay window (seq_lo-wise) and would look replayed, yet aren't.
+      -- Assuming b) means 'in time', since systematic loss could stall resync indefinitely.
```
eugeneia: Do we know how other implementations interpret the RFC? If not, could you compare with a mature implementation? OpenBSD?

src/lib/ipsec/esp.lua
```diff
@@ -168,10 +162,59 @@ function esp_v6_decrypt:decapsulate (p)
...
+         end
+         self.decap_fail = 0 -- avoid immediate re-triggering if resync fails
+         seq_high_tmp = self:resync(p, seq_low, seq_high_tmp + 1, ESP_RESYNC_ATTEMPTS)
+         if seq_high_tmp >= 0 then
+            seq_high = seq_high_tmp
+            proceed = true -- resynced! the data has been decrypted in the process so we're ready to go
+         end
+      end
    end
+   if not proceed then return false end
```
eugeneia: Looks to me as if we could `return false` in the failure branch if resync failed and avoid the `proceed` variable altogether?

src/lib/ipsec/esp.lua
```diff
@@ -48,6 +48,8 @@ local PAYLOAD_OFFSET = ETHERNET_SIZE + IPV6_SIZE
 local ESP_NH = 50 -- https://tools.ietf.org/html/rfc4303#section-2
 local ESP_SIZE = esp:sizeof()
 local ESP_TAIL_SIZE = esp_tail:sizeof()
+local ESP_RESYNC_THRESHOLD = 4 -- Resync after this many consecutive failed decaps
+local ESP_RESYNC_ATTEMPTS = 6 -- Try this many increments of seq_high
```
eugeneia: These should be configurable in `new` like spi etc. What could be sensible defaults?

src/lib/ipsec/esp.lua
vanfstd: We could drop `proceed` at the cost of two `else return false`, yes.  Good catch.

src/lib/ipsec/esp.lua
```diff
@@ -168,10 +161,61 @@ function esp_v6_decrypt:decapsulate (p)
...
-      return false
+      -- RFC4303 is somewhat unclear on a) whether or not packets that look replayed are eligible to trigger
+      -- recyns, and b) what 'consecutive' means in "number of consecutive packets that fail authentication":
+      -- Consecutive in time? Consecutive in their seq_low's?
+      -- Assuming a) should be 'yes' since the to-be-resynced packets might have seq_low's that just happen
+      -- to fall inside the replay window (seq_lo-wise) and would look replayed, yet aren't.
+      -- Assuming b) means 'in time', since systematic loss could stall resync indefinitely.
+      self.decap_fail = self.decap_fail + 1
+      if self.decap_fail >= ESP_RESYNC_THRESHOLD then
+         local seq_high_tmp = seq_high
```
vanfstd: Rename to `resync_start`?

src/lib/ipsec/esp.lua
```diff
@@ -168,10 +161,61 @@ function esp_v6_decrypt:decapsulate (p)
...
+   local gcm = self.aes_128_gcm
+   local payload = p.data + PAYLOAD_OFFSET
+   local iv_start = payload + ESP_SIZE
+   local ctext_start = payload + self.CTEXT_OFFSET
+   local ctext_length = p.length - self.PLAIN_OVERHEAD
+   for i = 1, n do
+      if gcm:decrypt(ctext_start, seq_low, seq_high, iv_start, ctext_start, ctext_length) then
+         return seq_high
+      else
+         C.memmove(p.data, p_orig.data, p_orig.length)
```
vanfstd: clone/free instead?

src/lib/ipsec/esp.lua
vanfstd: Some considerations: a) Anyone who can inject or forge packets might be able to trigger an attempt to resynchronize by sending packets with the right (unencrypted) header fields or replaying very old packets.  Every `SP_RESYNC_THRESHOLD`th such packet will trigger resync (as opposed to *every* such packet after the `SP_RESYNC_THRESHOLD`th due to [Line 180](https://github.com/vanfstd/snabb/blob/ef2b4027d3baaabf337e8fb521d3bdd76f994b15/src/lib/ipsec/esp.lua#L180))

Next, the actual cost of a resync operation is roughly `ESP_RESYNC_ATTEMPTS` times the cost of decrypting a packet payload. That's not much in absolute terms, but compared to the overhead of regular packets, it is.  However, the larger the value of `ESP_RESYNC_ATTEMPTS`, the larger a packet loss we can deal with before losing the ability to resynchronize. The number of lost packets we can tolerate is 2^32 times `ESP_RESYNC_ATTEMPTS`.

Based on that, I'd tend to default to a smallish value for `ESP_RESYNC_ATTEMPTS`. Even a value as low as 5 would result in being able to resynchronize after well over a year of network outage, if the sender keeps up a packet rate of 1000 p/s. We could bump it to 6 or even 8 to feel extra sure about it.

Contrary, the value of `ESP_RESYNC_THRESHOLD` should probably default to something substantially bigger. Maybe 1024.  If we are in a situation where we'd resync, we're already past 2^32 lost packets, so it doesn't matter if we require some more to fail before spending resources on resync, making it less attractive to try and abuse resync for DoS purposes in the process.

src/lib/ipsec/esp.lua
```diff
@@ -168,18 +161,72 @@ function esp_v6_decrypt:decapsulate (p)
...
+            gcm:encrypt(ctext_start, iv_start, seq_low, seq_high, ctext_start, ctext_length, ffi.new("uint8_t[?]", gcm.AUTH_SIZE))
+            resync_start = seq_high + 1
+         else
+            resync_start = self:seq_high() + 1 -- use the last seq_high we saw if it looked replayed
+         end
+         self.decap_fail = 0 -- avoid immediate re-triggering if resync fails
+         local seq_high_resynced = self:resync(p, seq_low, resync_start, self.resync_attempts)
+         if seq_high_resynced >= 0 then
+            seq_high = seq_high_resynced
+            -- resynced! the data has been decrypted in the process so we're ready to go
```
eugeneia: I just wondered: what happens to the state stored in the anti-replay window on resync? Doesn't it need to be updated/cleared to reflect the advance in `seq_high`?

vanfstd: That is being taken care of by the call to track_seq_no a few lines below this. Effectively it calls [advance_window](https://github.com/vanfstd/snabb/blob/ef2b4027d3baaabf337e8fb521d3bdd76f994b15/src/lib/ipsec/track_seq_no.c#L44) with the resynced-on sequence number, resulting in `diff` in that function to be large so the entire window is cleared.
Might still make sense to add a test for this, though.

eugeneia:
We decided to merge this with some refactoring, and additional bug fixes (RFC4303 never gets old):

- [Hydra report](https://hydra.snabb.co/build/599710/download/2/report.html#density-plot) (`resync-reorg` is what was merged)
- [Diff of resync-reorg](https://github.com/vanfstd/snabb/compare/resync-pr...eugeneia:resync-reorg)

PR #1005, opened 2016-08-31
Getter/setter for Intel10g's ring_buffer_size

dpino:
Intel10g's ring buffer has to comply several restrictions (power of two and not larger than 32*1024). Currently these constraints are implemented independently in several apps (firehose, lwaftr). The PR adds a single entry point for getting/setting intel10g's ring buffer.

### DIFF
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -23,7 +23,18 @@ local timer = require("core.timer")
 local bits, bitset = lib.bits, lib.bitset
 local band, bor, lshift = bit.band, bit.bor, bit.lshift
 
-num_descriptors = 1024
+local num_descriptors = 1024
+function ring_buffer_size (arg)
+   if not arg then return num_descriptors end
+   local ring_size = assert(tonumber(arg), "bad ring size: " .. arg)
+   if ring_size > 32*1024 then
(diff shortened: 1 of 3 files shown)


src/apps/intel/intel10g.lua
```diff
@@ -23,7 +23,18 @@ local timer = require("core.timer")
 local bits, bitset = lib.bits, lib.bitset
 local band, bor, lshift = bit.band, bit.bor, bit.lshift
 
-num_descriptors = 1024
+local num_descriptors = 1024
+function ring_buffer_size (arg)
```
eugeneia: This looks like its a global setting, shouldn't this be per PCI device / app instance? Maybe even a configuration option of the driver app?

dpino: I'm not sure. The reason I think this should be per module and not per instance is that `num_descriptors` is used to set RDLEN and TDLEN registers in the intel10g driver. In SF that should be OK, but what would happen in VF mode? VF inherits `set_receive_descriptors` and `set_transmit_descriptors` from SF. If two VF instances are started with different `num_descriptors` sizes the latter would override RDLEN and TDLEN registers in the card. Wouldn't that be a problem?

PR #1008, opened 2016-09-02
Remove apps.vpn

alexandergall:
Remove dead code, obsoleted by the l2vpn program on the l2vpn branch.


eugeneia:
I would prefer to merge this once the replacement lands upstream. Might not make a difference, but I don't know who is using `apps.vpn`?

alexandergall:
The new VPN code will remain on the `l2vpn` topic branch (in that sense, the replacement is already there and has been so for a long time). You can phase out the old `vpws` code whenever you like.

PR #1012, opened 2016-09-05
manual: support NixOS 16.03 and 16.09 texlive package sets

domenkozar:
Otherwise manual doesn't build on NixOS 16.09 (beta to be released today).

### DIFF
--- a/src/doc/default.nix
+++ b/src/doc/default.nix
@@ -13,8 +13,8 @@ stdenv.mkDerivation rec {
 
   buildInputs = [ ditaa pandoc git
    (texlive.combine {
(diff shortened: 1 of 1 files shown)


domenkozar:
cc @kbara

eugeneia:
I think this change might need to be replicated in the snabb_doc module in snabblab-nixos, am I right?

See: https://github.com/snabblab/snabblab-nixos/blob/master/modules/snabb_doc.nix#L80

domenkozar:
Yeah. We could deduplicate the list, but still you'd have to bump snabb source to get an update.

PR #1014, opened 2016-09-07
avoid allocating a packet on every breath

petebristow:
Further to @alexandergall comments in slack, now tap doesn't allocate/free a packet on every breath.

### DIFF
--- a/src/apps/tap/tap.lua
+++ b/src/apps/tap/tap.lua
@@ -31,6 +31,7 @@ function Tap:new (name)
    end
    return setmetatable({sock = sock,
                         name = name,
+                        pkt = packet.allocate(),
                         shm = { rxbytes   = {counter},
                                 rxpackets = {counter},
                                 rxmcast   = {counter},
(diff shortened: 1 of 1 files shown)


src/apps/tap/tap.lua
```diff
@@ -68,7 +67,9 @@ function Tap:pull ()
       if ethernet:is_bcast(p.data) then
          counter.add(self.shm.rxbcast)
       end
+      p = packet.allocate()
    end
+   self.pkt = p
```
eugeneia: I think we can omit these two lines if we use `link.transmit(l, packet.clone(p))`?

eugeneia: On the other hand that would double the copying done, never mind.

petebristow: Doesn't that result in an extra copy for every packet?
I guess worrying about extra copies is probably mute given that it's a Tap interface, but it seems a slight dubious habit to get into.

eugeneia: Agree! Will merge this as is.

petebristow:
@eugeneia If snabb bot agrees can you merge this update. The code above breaks under load, packets with length 0 start flying around. I'm not really sure why, but the new code lets luajit hoist the self.pkt rather than doing it manually.

eugeneia:
Do you think the [failure](https://gist.github.com/SnabbBot/e9efa5b9f53402aa11d2511601f24d7d#file-log-L1644) in #1016 is related?

Would be interesting to know why 82eecef fixes things, if it does.

petebristow:
Yes that was the failure mode. The socket write is being asked to write a packet of length 0, which causes an `EINVAL`. I haven't been able to work out why 82eecef fixes things, but I haven't been able to replicate the issue. Which isn't very satisfying.

 The original patch would reliably break after a few hundred thousand packets, and sometimes break under `selftest`. 82eecef has run several hundred million packets and not had any issues.

alexandergall:
I believe, in your first version, you did not store the packet in the object's `pkt` slot when the loop was left via the `return`. Adding `self.pkt = p` before the `return` would have fix it as well, I suppose (or use `goto` to break out of the loop instead of `return`).

petebristow:
Thanks, a sleep deprived blonde moment.

PR #1019, opened 2016-09-08
config.app: parse app args

eugeneia:
This PR addresses two issues:
- awkward `lib.parse_app_arg` dance in the constructor of many apps
- proper app arg validation and errors in case of missing mandatory arguments and undefined arguments

To fix this a `parse` function (adapted from #797) is added to `core.lib`, and `config.app` will use this function to parse app args when `appclass.config` is set.

Example:

```
MyApp = {
   config = {
      a = {required=true},
      b = {default=42}
   }
}

local c = config.new()
config.app(c, "valid", MyApp, {a="foo"}) => MyApp:new({a="foo", b=42})
config.app(c, "invalid1", MyApp) => error
config.app(c, "invalid2", MyApp, {a="foo", c="bar"}) => error
```

This supersedes #797, and provides a tightly integrated version thereof.

Depends on: #1008


eugeneia:
I have added a change that sets the zone during `config.app` and `ops.reconfig`, and makes the handler in `core.main` print the current zone. This makes it more obvious where errors come from.

eugeneia:
Cc @kbara @wingo

src/apps/pcap/pcap.lua
```diff
@@ -34,11 +34,8 @@ end
 
 PcapWriter = {}
 
-function PcapWriter:new (filename, arg)
-   local conf = arg and config.parse_app_arg(arg) or {}
-   local mode = conf.overwrite and "w+" or "w"
-   local file, errno = io.open(filename, mode)
-   if errno then error(errno) end
+function PcapWriter:new (filename)
+   local file = assert(io.open(filename, "w"))
```
dpino: The modifier flag was introduced to avoid Snabb returning an error when the pcap file already exists. I initially set it to "w+" always but others preferred to introduce a flag. The way the flag was added doesn't match how other apps work (all of them expect one arg, which is either a string or a table), I didn't know at that time.

Is there any way to keep this behaviour?

dpino: @eugenia That would work for me :+1:

Issue #987, opened 2016-08-04
The Snabb side of a NETCONF-managed system


wingo:
Greetings, Snabbfolk.  As you know we don't really have a story yet for NETCONF, something that a lot of operators would like.  We discussed this back in January on #696, but since then there have been a few developments on all sides.  Below is what we in Igalia see as a good way forward; it's the result of talking with operators and NETCONF experts, seeing how we can provide what they need while also remaining responsible for data plane performance. 

# YANG data modelling for Snabb

## Big picture

```
                                +--------------------+
                                | on-disk persistent |
                                | configuration      |
                                +--------------------+
                                          /\
       +---------------+ counters +-------\/------+          +---------------+
       | dataplane     | shm tree | snabb config  |          | sysrepod      |
       |               |~~~~~~~~~~>               |   get    +----------+    |
       |               |          |               |   set    | snabb    |    |
       |          ring | commands |               <--listen--> config   |    |
IPv4<->= NIC0   buffer O<----------               |  delete  | endpoint |    |
       |               |          |               |   add    +----------+    |
IPv6<->= NIC1          ---------->O               |          |               |
       +---------------+   ack    +---------------+          +---------------+
```

The different boxes here represent processes.

The NETCONF agent, `sysrepod` in this case, communicates with the Snabb data plane by running the `snabb config` program.  `snabb config` can read "state data" from the data plane via SHM counters, just as now.  Configuration queries are served by reading the state of the data plane from disk, without actually talking to the data plane.  The data plane doesn't touch disk of course.  `snabb config` can also send commands to the data plane to modify the data plane's configuration.  When the data plane ACKs the change, the `snabb config` program updates a persistent configuration store on disk so that the next `snabb config` invocation can know the state of the data plane without asking.  Access to `snabb config` is serialized, so that the disk state is in sync with the in-memory state of the data plane.  The protocol between `snabb config` and the data plane is private to the particular network function in question.

`snabb config` will speak a data model that is based on YANG, for minimum impedance mismatch between NETCONF agents and Snabb applications.

Initially, and perhaps for an indefinite period of time, there will be ad-hoc implementation for each YANG model supported by `snabb config`.  In this design document we'll use the specific example of the lwAFTR, but I think these considerations apply to all Snabb network functions.

For the lwAFTR we plan to implement include support for the YANG module for lwAFTR systems in general and also support for a custom YANG module including all of the additional functionality of
[...]

PR #1021, opened 2016-09-19
Engine support for multiprocess orchestration

lukego:
This is a work-in-progress effort to support multiprocess app networks.

The basic idea:
- Add `engine.start()` to enable automatically running the current engine configuration in the background. (Like an asynchronous version of `engine.main()`).
- Extend `engine.configure()` to optionally take a table of configurations that will each execute with a dedicated process and CPU core. Example: `engine.configure({worker1 = conf1, worker2 = conf2, ...})`.
- Make sure `engine.configure()` delta-updates work across all processes. That is, the parent process will diff the configuration to start/stop worker processes as needed, and then each worker process will do a local diff to update its own app network (the same logic that already exists today).
- Implement a sharing and error handling policy. For example all DMA memory is globally available between workers and failure of any process terminates everything.

I would imagine this to be a new API alongside the synchronous/single-process `engine.main()`. Hopefully this would be suitable for all applications to use in the future. In the short term we may need to invent some new communication mechanisms (e.g. shm conventions) to support use cases where application code is inspecting the app network directly (won't be possible when the apps are running in a child process).

The immediate use case here is to update the NFV application to have a single 100G NIC shared between many worker processes that are each serving a distinct set of VMs. So the NFV application would take many port configuration files (one per process/core) but only one NIC PCI address.

Thoughts? cc @wingo @eugeneia @kbara @plajjan @alexandergall @dpino @petebristow and well everybody else too...


lukego:
Just thinking aloud...

Child processes will need an efficient way to check for an app network update and then a simple way to load it. The check could be done by polling a serial number in a shm object like `//config/serial` that the parent bumps after each call to `engine.configure()`. The configuration could be loaded by reading the overall configuration from a text file `//config/configuration.$serial` and picking the relevant bit.

This may involve (re)introducing some shm path syntax for accessing objects from the parent process. The children may also need to be able to access each others' DMA memory e.g. to make it easy for one app to allocate a DMA ring for a NIC and then another app to attach to the queue by using that memory.

kbara:
My first knee-jerk response is "I'll be glad to have a standardized sharing and error handling policy." :-)

eugeneia:
The configuration serial is already stored in [engine/configs](https://github.com/snabbco/snabb/blob/master/src/core/app.lua#L120). The parent PID could be stored in `parent-pid` so that the parent can be accessed using `/{parent-pid}/foo`.

lukego:
I have pushed the next step now: a `snabb worker ...` program.

This is intended as an internal interface that a parent Snabb process can use to spawn a child process that will simply execute an app network on a dedicated core & react to configuration updates. (On balance I think that fork() without exec() is too messy and risks inheriting a lot of state that we don't want the child processes to have.)

The processes cooperate using a shared shm folder called `group/`. This is created by the parent process and then aliased into each child process (as a symlink).

The shm objects that I anticipate having in the `group` folder are:

```
group/myworker/config.lua       # current configuration for myworker
group/myworker/configs.counter  # incremented by parent on new config
group/dma/7e42ca00              # DMA HugeTLBs shared by all processes
group/pci/01:00.0/...           # Objects made available by driver
```

Here is how this would work in NFV with 100G (ConnectX):

- Master process initializes the NIC, sets up switching, allocates all descriptor rings, polls counters (non-realtime activities).
- Worker processes are each run a separate app network (serving different VMs) provided by the master process (based on the configuration file).
- Worker process drivers attach to the NIC via descriptor rings (allocated in shared DMA memory & with addresses provided as shm objects).
- DMA memory would stay allocated until the whole process tree shuts down (cleaned up by the supervisor of the master process).
- Supervisor would also automatically disable PCIe bus mastering (DMA) for all PCI devices. This is a safeguard against the descriptor ring memory being reused while still referenced by a NIC.

Whaddayareckon?

Next steps...

- Add a SIGSEGV handler to automatically map DMA memory from other processes (have prototyped this before);
- Add the engine API for running multiple app networks in subprocesses.

lukego:
(Could be overkill to make the worker into `snabb worker` program. I'll try factoring it simply into a `core.worker` module.)

lukego:
I pushed a refactoring where a simple `core.worker` module implements both the parent process and the child process code. This is currently sketch quality (compiles but won't run). Currently creating the worker with `fork()` but will switch this to `fork()+exec()` with some simple entry point (maybe snsh).

I have a small obstacle to overcome and I'm curious for @eugeneia and @wingo point of view because this is a precursor of sorts to the YANG work in #987.

The problem is: How do I serialize a `config` object to pass it from the parent process to the child process?

Today we already have functions like `lib.store_conf()` and `lib.load_conf()` that can serialize Lua objects to text files (pretty-print them and then parse them). The catch is that a `config` object includes Lua objects that cannot be directly serialized, particularly app classes which are typically tables of functions and closures.

So how should we serialize a `config` object? One idea would be to define a canonical string name for each app and then to use this to indirectly refer to the class object.

Could also be that there are related challenges e.g. places where we use app configuration tables that include non-serializable objects though I cannot immediately think of any.

Just now I have punted on the problem by calling non-existent `config.save()` and `config.load()` functions.

src/core/worker.lua
```diff
@@ -0,0 +1,110 @@
...
+local children = {}
+
+local function child (name)
+   return children[name] or error("no such child: " .. name)
+end
+
+-- Start a worker process with affinity to a specific CPU core.
+-- The child will execute an app network when provided with configure().
+function start (name, core)
+   local pid = S.fork()
```
eugeneia: Pedantic: should be  `assert(S.fork())`

eugeneia:
@lukego Hah, sorry for removing `alias` like a couple weeks ago. ;-)

I think the YANG effort will require a way to serialize configs as well, so it seems sensible to expect to have `config.save` and `config.load`.

I like the code in this PR very much! :+1:

lukego:
I have pushed some more code. The overall intention is for a group of Snabb processes (the main one + the workers) to be able to automatically share DMA memory between themselves, and to implement clean shutdown semantics where DMA is disabled for all PCI devices before any DMA memory is freed.

Here are the specific invariants that this code aims to create:

- Each HugeTLB allocated by a running group of Snabb processes is accessible to each process via the shm path `group/dma/xxx.dma`. (`xxx` is the physical address of the memory.)
- DMA memory is never freed while any PCI device has bus mastering (DMA) enabled by a process in the group.
- DMA memory is always freed during shutdown of a Snabb process group i.e. backing files on hugetlbfs do not outlive the processes.

The next step is to make DMA memory pointers globally valid for all Snabb processes in a group. This will be achieved with a SIGSEGV handler that detects access to unmapped DMA memory and automatically maps it in (if the address belongs to DMA memory allocated by a process in the group). I will use this immediately in the Mellanox driver so that the main process can create descriptor rings and then worker processes can easily access them.

lukego:
I pushed the SIGSEGV handler now. This adds another invariant:

- DMA memory allocated by any process is automatically usable by every other process in the same group.

This means that any address returned by `memory.dma_alloc()` can be shared between processes. You have to be a bit careful about this, e.g. if you pass a `packet` from one process to another then it may end up on the wrong freelist when you are done with it, and you would need to take care of this e.g. as in #601.

The immediate benefit is that the Mellanox driver will be able to allocate all of the descriptor rings in the parent process and then make the addresses available to worker processes via shm objects.

lukego:
Overall I feel like this branch represents a pretty reasonable approach to multiprocessing. The tricky bits we need to review are potential race conditions in the interactions between processes e.g. are configurations created atomically, are are notifications made after all data is available, are all combinations of signals and termination orders handled equivalently, etc.

I suppose we also need a clear description of this new "process group with workers" concept in the manual. This concept as evolved a little during implementation.

lukego:
I pushed an update so that the worker processes use `execve(2)` to create an entirely new process image. This way the worker processes are all normal Snabb instances that have been initialized from scratch. This is intended to avoid a whole class of errors from inheriting state with fork().

Added an initial selftest function too.

The main thing lacking now is a working way to serialize a configuration object so that the parent can actually provide configurations to the children.

Overall I feel like the design here is quite reasonable and the code is short, but the exact formulation has quite some rough edges. For example I have somewhat awkwardly broken the shm abstraction in order to operate directly on the underlying directories (e.g. to create symlinks to HugeTLB files that live in a directory that is not addressable by shm paths).

eugeneia:
So this is the simplest way I could think of to “serialize” app classes: https://github.com/lukego/snabb/compare/multiprocess-engine...eugeneia:multiprocess-engine

It extends `core.config` to accept app classes as string identifiers in the form `"<module>[/<var>]"` and updates `core.app` accordingly. The main caveat is that the programmer has to ensure he only uses string identifiers for classes when using `core.worker`. We could just stop supporting app classes passed as Lua objects, but that seems harsh.

Another issue is that this change conflicts with #1019, which uses `class.config` to reject invalid app arguments during `config.app`(!). I chose to parse app arguments in config because I felt it was a good idea to fail fast, but this could be done in `engine.configure` as well.

If I come up with something better I will let it be know. :-)

wingo:
Neat PR :)

Regarding configuration.  I think different network functions are going to have different needs.  In particular in the lwAFTR, I don't think a generic serialization facility is going to work for the sorts of configuration updates that we expect.  For the lwAFTR we need to compile the configuration to an efficient binary format that can be just slurped into memory at startup.  #987 is where we would like to go, I think.

I would like to see other Snabb programs migrate towards declarative YANG configurations.  In that way we have a model for what "configuration change" means that besides reloading an entire configuration also encompasses add/remove/set operations on container elements, we have a protocol for communicating changes between processes (#987) that won't cause the data plane to lose packets, and we can verify configurations both against a YANG schema and with program-specific checks.

I will be working on this full-time for the next few weeks fwiw.  I look forward to seeing what patterns end up working for you for the receive-side steering and virtual 100G cases you are looking at here.  I do want to push back a little though on any generic attempt to codify what configuration means in a Snabb context without considering the lwAFTR's use cases.  Thanks :)

wingo:
What operational CPU/NUMA affinity concerns does a multi-process Snabb have?  My working idea is that only the data planes should be bound to the data-plane CPUs, and that the kernel should even be configured to have nohz on those cores.  Does that match your impression?

wingo:
I would imagine for a PCI device with affinity for NUMA node 1, that the first CPU on that node should handle any auxiliary processes, as that CPU is the one that will be handling interrupts for that socket.  So the process should have NUMA affinity but unspecified or specified-to-first-CPU-on-node CPU affinity.

lukego:
@wingo I would be happy to replace this with a YANG-oriented mechanism. Would you like to take the lead on the multiprocess support and just cherry-pick anything you want from this branch?

Here are the ideas on this branch that I think are significant:

1. The notion of a _group_ of related processes. This provides a scope for sharing certain resources e.g. PCI devices and DMA memory. Here it is simply implemented as a symlink between shm folders.
2. Orderly shutdown of the group. Here we have a shutdown step that disables DMA on all PCI devices before allowing any DMA memory to be freed to the kernel.
3. Having DMA memory mappings shared between all processes in the group (via the SIGSEGV handler.)

I do like the idea of having automatic assignment of suitable CPU cores and e.g. reserving the first core of each socket for the kernel. I would ideally prefer to know that worker processes always have affinity to one specific core rather than being exposed to kernel scheduling, but I don't have a scheme worked out for that and maybe it is not practical.

lukego:
(Oh you are talking about the CPU affinity of the management process. Yes, I agree it would be fine to let the kernel schedule that somewhere out-of-the-way, and as you say possibly worth explicitly setting affinity to a core that we are _not_ planning to use for worker processes.)

wingo:
I would be happy to take more responsibility around here though I worry that I might not fully page in your needs.  In fact, maybe the manager process can handle configuration-related concerns?  Probably best to do configuration in a sub-process for isolation reasons, but I don't know.

As far as the NFV (e.g.) and YANG go -- we have a branch in which we use YANG data models to configure the lwAFTR, I think the first step would be to get that into shape, publish, and see what you think.  It will take a week before we can start doing "snabb config"-like things I think.  I wil retake the configuration part of the conversation then.

lukego:
The process model here is:

- Manager process is a normal Snabb process, running the engine, but is understood to be reserved for non-realtime administrative tasks such as handling configuration updates.
- Worker process is also a normal Snabb process but has CPU affinity and will expect the manager process to take care of anything that would require an unacceptable interruption of traffic.
- Each Snabb process - manager and workers - is monitored by a tiny supervisor process that performs minimal cleanup actions on termination e.g. safely turning off DMA and cleaning up shm folders.

How does that match your model?

wingo:
Interesting, for some reason I thought the manager would also be the supervisor.  Cool.  In that case that works for me.  What is the mapping of the various processes to /run/snabb/PID ? Do they share one directory?  cc @xray7224

lukego:
The new constraint here is that for the Mellanox driver I would really like to be able to run the NIC initialization & update logic in the manager process. The Mellanox firmware interface is really designed around the expectation that a centralized stateful thing - traditionally the kernel - will be responsible for provisioning queues etc. I _could_ make this state available to all of the worker processes so that they can each perform provisioning actions, but at the moment this is not my preferred implementation strategy.

lukego:
@wingo The supervisor process is actually already introduced in #930 and I am pigging backing on that here. Originally the only thing that it did was to remove the shm directory on shutdown but I found it an appealing place to implement clean DMA shutdown semantics too.

Just now on this branch every process has its own `/var/run/snabb/$pid` folder like always, but underneath this we have a `group/` folder that is shared between all processes (in the workers it is a symlink to the manager.) So from the outside looking in `/var/run/snabb` we see a flat list of all Snabb processes on the machine but internally they can use `group/` to share objects with only their siblings.

eugeneia:
> Originally the only thing that it did was to remove the shm directory on shutdown but I found it an appealing place to implement clean DMA shutdown semantics too.

Can’t resist telling history: it was about DMA shutdown semantics (#509) 400 PRs earlier, for which I failed to find a solution back then. Then came #930 and now we are full-circle :+1:

eugeneia:
Wait, now who is the lead on this? @wingo? If so I will un-assign myself.

wingo:
As this is an RFC/WIP does it need to be assigned to anyone, @eugeneia ?  I didn't think of myslf as lead fwiw :)

lukego:
@wingo I would like to extend the NFV application to support multiprocess support (shared 100G NIC) in the short term (this month). Do you think it is better to base that on the multiprocess support on this branch (which we will then need to reconcile with the YANG extension) or would you like to supply an alternative (aligned with YANG from the beginning)?

wingo:
Good question... I am also on the hook for deliverables at the end of this month and so we will be racing together :)

I think the text approach you mention here will not work for the lwaftr, even in the case where you can't make an incremental update and instead you have to just reload the config -- it will be too slow.  For that reason we are compiling configurations to a binary format (Igalia#517) so that the data plane can load the configuration faster.  I hope to be done with that bit by tomorrow.

However although that configuration-compilation approach is generic and can work for any YANG model -- see design here: https://github.com/Igalia/snabb/blob/8c8dc727b0ded04be1899412c766bebd9530345d/src/lib/yang/README.md -- it does mean a different configuration syntax for your users and probably a different form of the configuration internally, relative to how any specific app (in your case the NFV but this also affects the lwaftr) is currently configured.

For that reason there's a bit of change cost and I think you are the right person to decide when/if you will be able to pay it.  Again, I expect to have this config facility shippable by the end of the week.  What does your timeline look like in that regard?

lukego:
@wingo This discussion is helpful. I think I see how to make this multiprocess branch simpler and more orthogonal to your YANG work. I reckon that we can avoid the conflict and both continue hacking on our merry ways and then sync up without a collision :).

The problem I see is this maldesigned API function:

```lua
-- Configure a worker process with a new app network.
worker.configure(workername, config)
```

I proposed that the parent process will pass app network configurations to the worker processes. This created an immediate problem: How do we transport the `config` object between processes when it has no serialized representation? Then it also creates the critical path dependency: Should we model the app network with YANG and make it serializable?

I think a simpler solution would be for the worker process API to be agnostic to configuration. You would instead start a worker process and provide it with a startup entry point (Lua code) that can do anything. The worker would then be responsible for obtaining its configuration in some program-specific way. For example on Snabb NFV each worker could independently parse the configuration file and cherry-pick the parts that are relevant (parent is responsible for starting children but then they configure themselves.) For lwAFTR the configuration could be provided via the high-performance "back-channel." Over time we will probably discover common requirements but we can factor that as library code.

So - thanks for the input. If that suits everybody else I will move forward with this branch for managing child processes without committing to any one parent-child configuration interface (and upstreaming via @eugeneia)?

wingo:
@lukego yes your analysis is exactly how I see it.  We are doing things that sound similar on the surface but that doesn't mean that they are actually similar in essence or that they should share implementation.  There could be points that we need to coordinate on but I don't think configuration right now is one of them; configuration-as-app-network-specification works for you right now and not for me and that's OK :)

eugeneia:
:+1: for the generalized solution!

@lukego I will happily be upstream.

lukego:
Groovalicious. I will hack in that direction & remove the RFC/WIP marker when ready.

lukego:
@eugeneia @wingo So! I would love to reach a good solution for how to specify which core a worker process should run on. (We should nail this down at least on the Lua worker API level even if different programs choose to use the API in different ways.)

Just now on this branch the API requires the parent process to nominate the exact core number for the worker when it starts:

```lua
-- Start a worker process with affinity to a specific CPU core.
-- The child will execute an app network when provided with configure().
function start (name, core)
```

The simplest way to incorporate this into the NFV application could be to specify which core each virtual port should be handled by. Example config file with added `core` parameter:

```lua
-- NFV ports configuration: two ports on core #1 and one port on core #2.
{ { id="vm1", mac="00:00:00:00:00:01", core=1 },
  { id="vm2", mac="00:00:00:00:00:02", core=1 },
  { id="vm3", mac="00:00:00:00:00:03", core=2 } }
```

This would seem to work okay but it is quite low-level. Is it worth inventing a new notation, e.g. to give cores symbolic names (`core="traffic0"`), to assign actual core numbers heuristically, and to allow the user to supply some optional hints/constraints on core selection (`cores = { traffic0 = "numa=0", traffic1 = "core=1-6", ... }`).

lukego:
I chatted with @wingo about core assignment on [Slack](http://snabb.slack.com/) and this is what sounded best at the moment:

- The user specifies exact core numbers (as in the example above.)
- Snabb assigns cores using `lib.numa()` (#1038) to make sure it is done right.

This is a simple and direct approach where the person creating the config file is responsible for deciding which exact cores to assign. The alternative would be to use a "do what I mean" heuristic to automatically select a core but this seems a bit error prone (might need to consider isolcpus, hyperthreads, other applications on the same machine, etc to avoid making mistakes.)

@eugeneia what do you reckon about  that scheme?

eugeneia:
I agree with the decision, its unrealistic to reliably tell how the host is configured, and what’s already running on it. AFAIK there is no way to tell if a core is free? Anyhow, this doesn’t shut any doors, we can add smart middle ware later on.

src/core/worker.lua
```diff
@@ -0,0 +1,145 @@
...
+   if pid == 0 then
+      -- Lock affinity for this child
+      S.sched_setaffinity(0, {core})
+      local env = { "SNABB_PROGRAM_LUACODE=require('core.worker').init()",
+                    "SNABB_WORKER_NAME="..name,
+                    "SNABB_WORKER_PARENT="..S.getppid() }
+      -- /proc/$$/exe is a link to the same Snabb executable that we are running
+      S.execve(("/proc/%d/exe"):format(S.getpid()), {}, env)
```
wingo: This gives me the shivers -- what if someone upgraded this snabb in place?  What if it was moved or made clean?  In any case we need an assert and anerror message so we don't fall off and keep running the parent.

wingo:
I like this code. In the short term I'd like to use this code, and I would like be able to run my own code to configure the app network in the worker (and the main process of course).  For now I only need a single worker.  What's your timeline here?  I'll use an app pair to communicate between the main process and the worker, so until it lands I can still make progress by running the manager-side app from an unrelated process.

Open question: does C-c on the manager reliably stop workers?  I don't see any call to stop() in your patch.

lukego:
@wingo I would quite like to land this code on your branch before next/master if that suits you. I like the idea of feature branches propagating quickly to users and the next/master branches acting as the "trailing edge" that aggregates everything in the background without time pressure. I realize that this is a bit complicated e.g. with the same branch potentially having multiple upstreams/PRs. Have to see how it works in practice.

So it sounds like what you need is:

1. Reliable mechanism for making sure that all processes run the same software.
2. More flexible entry-point to the worker process e.g. Lua expression instead of app network.
3. Well-defined process shutdown semantics w/ test.

Is that about right?

#### shutdown

Thinking about shutdown... I am a fan of the Erlang style where you make cleanup code as minimal as possible and also isolate it from the rest of the code. So I would like shutdown of the parent to reliably stop the children but I would prefer not to do this with explicit stop() calls in the parent because that could fail in scenarios to heap corruption, SIGKILL, OOM, etc. The [best mechanism](http://stackoverflow.com/questions/284325/how-to-make-child-process-die-after-parent-exits) that I see now would be for the child to run something like `prctl(PR_SET_DEATHSIG, SIGTERM)` to automatically receive a signal from the kernel when the parent terminates, and for this signal to be unhandled so that the kernel shuts down the worker automatically. what do you think?

#### in place updates

Thinking about making the process forking robust to in-place updates. The optimal would be if the kernel would provide a way to access the exact binary that a process is running (instead of a symlink). Is there a way? I haven't found one.

Failing that I have two backup ideas.

One would be that during startup we `fork()` a persistent "template" process that serves requests to fork new children with the same software. This way we preserve the executable in the running heap of that process.

Another variation would be for the main process to _copy_ the snabb binary that it starts from into the shm group folder `/var/run/snabb/$pid/group/snabb` where it is accessible to every process in the group.

These all sound messy. Great if there is a better solution? Should grep some codebases for execve() to see what other people are doing.

eugeneia:
> Thinking about shutdown... I am a fan of the Erlang style where you make cleanup code as minimal as possible and also isolate it from the rest of the code. So I would like shutdown of the parent to reliably stop the children but I would prefer not to do this with explicit stop() calls in the parent because that could fail in scenarios to heap corruption, SIGKILL, OOM, etc. The best mechanism that I see now would be for the child to run something like prctl(PR_SET_DEATHSIG, SIGTERM) to automatically receive a signal from the kernel when the parent terminates, and for this signal to be unhandled so that the kernel shuts down the worker automatically. what do you think?

The worker already does this in the relation to the supervisor:
https://github.com/snabbco/snabb/blob/master/src/core/main.lua#L176

lukego:
@wingo This is potentially ready for you to use now.

Commit 1ce4ca8 introduces a simpler interface `worker.start(name, luacode)` where `luacode` is evaluated as a string and expected to take care of core affinity, configuration acquisition, etc.

Commit 5393afa makes the child processes terminate when the parent does.

Have to still formulate a nice solution for ensuring that every process runs the exact same software version. This will ideally also work for applications like `snabb config`, `snabb top`, etc, when they want to communicate with a process using its own native software version.

WDYT?

wingo:
Nice, thanks!  I think @kbara is looking into this in the short term; I will help if needed.  FYI also I now have the leader/follower thing implemented from #1069 and am just fixing a couple of lwaftr bugs.  Things are looking good!

lukego:
@kbara Could you please gist the output of this command?

```
sudo strace -f ./snabb snsh -t core.memory
```
 
Can you also confirm that other branches are working on your dev laptop? (My own Linux laptop lacks HugeTLB support and doesn't run Snabb.)

kbara:
I do most of my dev work for Snabb on this laptop, and I've confirmed that current upstream (snabbco) master's tests pass on it.

Gist: https://gist.github.com/kbara/eaa8da0f9113296185746866afdf4268

lukego:
Thanks for the info. Could you please gist the output of that command when running/failing on this branch? (The gist above seems to be from a successful from of a different branch.)

lukego:
(I have tried to reproduce this problem on a couple of lab servers but for me `make test` is running fine. I have also tried purging state like `/var/run/snabb`. I am hoping `strace` will provide a hint about what is different on your machine.)

kbara:
It seems to have gone away almost entirely. I can _sometimes_ reproduce the symptom in the unix socket test when merged onto the lwaftr branch, though. (It failed, then I ran it with strace and it succeeded, and now I'm struggling to get it to fail again.)

lukego:
@kbara The line failing above is [memory.lua:151](https://github.com/lukego/snabb/blob/514784ada3a8d1c485abc986053f893e44cfb57b/src/core/memory.lua#L151). So we have created a file on hugetlbfs, we have made it hugepage-sized, but the call to map it failed. alas, we do not have the error message.

Seems likely to be some obscure kernel-related issue. Could be an edge case e.g. how it behaves when it is unable to provision a hugetlb page? I will dig a little. Could be useful if you check `dmesg` for a relevant message. also, what kernel are you running?

kbara:
Nothing even slightly related in dmesg. I'm on 4.4.0-45-generic.

lukego:
I have reproduced it now. Fix + explanation coming.

kbara:
Several hundred test runs later, that's looking a lot like a fix, thank you. What's the basic idea behind retrying these allocations, by the way?

lukego:
Good question. This is a DWIM feature to make Snabb easier to get started with. (Could be that it should be disabled for programs that are more concerned about correctness than convenient in installation.)

Just now you can install and run Snabb on a new machine very easily:

```
git clone https://github.com/snabbco/snabb
make -C snabb
snabb/src/snabb packetblaster -S 60 01:00.0
```

and if your machine was not already perfectly prepared then you may see some messages like this:

```
[mounting /var/run/snabb/hugetlbfs]
[memory: Provisioned a huge page: sysctl vm.nr_hugepages 0 -> 1]
[memory: Provisioned a huge page: sysctl vm.nr_hugepages 1 -> 2]
[memory: Provisioned a huge page: sysctl vm.nr_hugepages 2 -> 3]
[memory: Provisioned a huge page: sysctl vm.nr_hugepages 3 -> 4]
[memory: Provisioned a huge page: sysctl vm.nr_hugepages 4 -> 5]
```

This is Snabb attempting to DWIM instead of exiting with error messages like:

```
Error: Failed to provision a huge page.

Please take the following actions:

1. Reserve huge pages in your grub.conf:
      hugepages=nnn
   (where nnn is the max number of pages you will need for Snabb.)
2. Create directory /mnt/hugetlbfs
3. Add this entry to /etc/fstab:
      hugetlbfs       /mnt/hugetlbfs  hugetlbfs       defaults        0 0
4. Reboot.
```

which feels like creating a lot of schlep work for the user.

(You could also say that it is a workaround for the Linux kernel lacking a suitable mechanism for dynamically allocating a hugetlb page, at least that I am aware of.)

lukego:
(Could add a `DWIM` configuration option to the engine to control this behavior. Once the YANG support lands perhaps we can model the engine config that way?)

lukego:
@kbara Please pull this branch again to get commit 7adc708 which fixes a file descriptor double-close bug in `memory.lua`. This can have unpredictable effects - was seen to asynchronously close vhost-user sockets in the NFV application. See also justincormack/ljsyscall#205.

lukego:
This branch depends on #1077. Please merge @kbara. I did not pull here because it is based on a different ancestor (newer version of master) and would make this PR much more noisy to pull that too.

lukego:
I pushed API documentation for the `core.worker` multiprocess API: https://github.com/snabbco/snabb/blob/82f335573a743fd608baa5ee966d3c5ef35183ab/src/README.md#multiprocess-operation-coreworker.

src/core/worker.lua
```diff
@@ -0,0 +1,132 @@
...
+   local name = assert(lib.getenv("SNABB_WORKER_NAME"))
+   local parent = assert(lib.getenv("SNABB_WORKER_PARENT"))
+   print(("Starting worker %s for parent %d"):format(name, parent))
+
+   -- Create "group" alias to the shared group folder in the parent process
+   shm.alias("group", "/"..parent.."/group")
+
+   -- Run the engine with continuous configuration updates
+   local current_config
+   local child_path = "group/config/..name"
```
wingo: @petebristow notes that this looks like the wrong thing (and if it isn't it needs a comment I think): https://github.com/snabbco/snabb/pull/1133#pullrequestreview-38643931

WDYT @lukego?

wingo:
You will need Igalia#809 at least if this is to work with intel_mp.  Do you want to make a separate merge branch that is based here, then merges in master, then pr's back to master?

wingo:
Probably you want the worker.lua part of 3e3281998c7c1f3636b1b379be890bc1d3aeb011 as well.

wingo:
You might also need 876addc17287ed7c8fc133e3e24452aa3373f429

src/core/worker.lua
lukego: ahem! That whole `init()` function is actually dead code. Removed in 876addc.

That's from an older version where `core.worker` imposed specific behavior on the worker process i.e. running an app network provided by the parent. These days the worker simply links itself into the process tree and runs the provided Lua code.

PR #1023, opened 2016-09-23
Cast VLAN tags as int32

dpino:
There are certain VLAN tags that are not correctly untagged, for instance VLAN: 128.

The problem is that VLAN tags are read as uint32. This value is compared to a tag, which is obtained as the result of a bitwise operation. As LuaJIT BitOp module works in the signed int32 range, certain tags return a negative number while being compared to an uint32, which is always a positive number.

The solution is to read VLAN tags as int32 or to read them as uint32 and to apply the `bit.tobit()` operation. `tobit()` guarantees successful comparisons between the result of bitwise operations in LuaJIT (for instance when comparing a bitwise operation to a constant). Certainly I think reading the tags as int32 is simpler.

Here is a small snippet that renders the issue:

```lua
local ffi = require("ffi")
local bit = require("bit")

local C = ffi.C
local htonl = bit.bswap

local dot1q_tpid = 0x8100
local function build_tag(vid)
   return htonl(bit.bor(bit.lshift(dot1q_tpid, 16), vid))
end

local data = ffi.new("uint8_t[?]", 4)
data[0] = 0x81
data[1] = 0
data[2] = 0
data[3] = 128

print("Target tag: "..build_tag(128))
print("Cast to uint32_t: "..ffi.cast("uint32_t*", data)[0])
print("Cast to int32_t: "..bit.tobit(ffi.cast("int32_t*", data)[0]))
print("Cast to uint32_t (tobit): "..bit.tobit(ffi.cast("uint32_t*", data)[0]))
```

Output:

```
Target tag: -2147483519
Cast to uint32_t: 2147483777
Cast to int32_t: -2147483519
Cast to uint32_t (tobit): -2147483519
```

### DIFF
--- a/src/apps/vlan/vlan.lua
+++ b/src/apps/vlan/vlan.lua
@@ -17,7 +17,7 @@ Untagger = {}
 -- 802.1q
 local dot1q_tpid = 0x8100
 local o_ethernet_ethertype = 12
-local uint32_ptr_t = ffi.typeof('uint32_t*')
+local int32_ptr_t = ffi.typeof('int32_t*')
 
 
 -- build a VLAN tag consisting of 2 bytes of TPID set to 0x8100 followed by the
@@ -41,7 +41,7 @@ function push_tag(pkt, tag)
    local length = pkt.length
    pkt.length = length + 4
    C.memmove(payload + 4, payload, length - o_ethernet_ethertype)
-   cast(uint32_ptr_t, payload)[0] = tag
+   cast(int32_ptr_t, payload)[0] = tag
 end
 
 -- extract TCI (2 bytes) from packet, no check is performed to verify that the
@@ -85,7 +85,7 @@ function Untagger:push ()
    for _=1,link.nreadable(input) do
       local pkt = receive(input)
       local payload = pkt.data + o_ethernet_ethertype
-      if cast(uint32_ptr_t, payload)[0] ~= tag then
+      if cast(int32_ptr_t, payload)[0] ~= tag then
          -- Incorrect VLAN tag; drop.
          packet.free(pkt)
       else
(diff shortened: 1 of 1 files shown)


dpino:
I think this regression happened when the implementation of `htonl` and `htons` changed in commit ca86c9403a32a7c800c63143324613b1ee476ba5, from using C API to `core/lib.lua` (which relies on LuaJIT's `bitwise` function). C's htonl and htons return an unsigned 32-bit, so the original code was correct.

The cast should have changed in that commit too. Interestingly it was me who asked Kristian Larsson to use `htonl` and `htons` from `core/lib.lua`, so I take all the blame here :)

kbara:
I think this is entirely the wrong approach. Vlan tags are a 12-bit positive integer, and should be represented as such. Messing around with negative 32 bit int comparisons is introducing an unnecessary epicycle into the code.

dpino:
@kbara Definitely dealing with 12-bit numbers is harder than to deal with 16-bit or 32-bit numbers. Are you suggesting to check only the TCI part of a VLAN tag? That's one possible approach, but that's not what the original code did. Since I am OK with the original implementation I changed as little code as possible to fix the issue.

kbara:
The root cause, as you said, is that we're using a broken htonl/htons; the C type signature specifies 'unsigned' very clearly.

dpino:
Using Snabb's core/lib `htons` and `htonl` is preferred than `C.htonl` or `C.htons` because it saves making a C call, saving a lot of CPU cycles. As a rule of thumb, Snabb prefers executing Lua code than making C calls, that's my understanding. In addition, other parts of Snabb's codebase use the core/lib.lua functions too. I think all code than needs to use htonl and htons should use the functions from core/lib.

When we wrote down this app we were not aware of the core/lib functions. In my opinion, the problem is not to use the functions from lib, it's to not have updated our code accordingly.

alexandergall:
Can you elaborate why you believe your fix is simpler than changing just the check to
```
if cast(bit.tobit(uint32_ptr_t, payload)[0]) ~= bit.tobit(tag) then
```
The `tobit` on the rhs is redundant, but it seems safer to make this symmetric in case a future change would make `tag` to no longer be the result of a BitOp.

dpino:
I agree that check is safer if the implementation of how tag value was obtained changes.I made my comment assuming we know tag is a value obtained as a result of a LuaJIT bitwise operation (signed int). In that case casting a uint32 to a signed int can be saved if directly casting the original value as signed int.

eugeneia:
@dpino @kbara What’s the status on this PR in light of #1024 ? I am with @kbara that using a signed type is probably unnecessary, but the test case introduced here looks to be still valid?

dpino:
@eugeneia @kbara I agree #1024 is the right fix for this case. Agree also the test case can become handy to prevent future regressions. I can rework the path and add only a test case.

eugeneia:
I can cherry pick the test from this PR; no problem. Merging the test then.

PR #1024, opened 2016-09-23
Make ntohl/htonl unsigned, in line with C and expected semantics

kbara:
I believe ntohl should be unsigned and have the same semantics as C.ntohl. Switching between signed and unsigned definitions of it breaks code. Using a non-standard definition of such a fundamental function is an unnecessary trap.

This is labelled a sketch because it breaks the tests for 4 apps:
```
ERROR     testlog/lib.protocol.datagram
ERROR     testlog/apps.ipv6.nd_light
ERROR     testlog/apps.vpn.vpws
ERROR     testlog/apps.socket.raw
```

If @eugeneia and @lukego are conceptually happy with this change, I will update those 4 apps to not be broken and remove the [sketch] marker.

### DIFF
--- a/src/core/lib.lua
+++ b/src/core/lib.lua
@@ -10,6 +10,8 @@ require("core.clib_h")
 local bit = require("bit")
 local band, bor, bnot, lshift, rshift, bswap =
    bit.band, bit.bor, bit.bnot, bit.lshift, bit.rshift, bit.bswap
+local tonumber = tonumber -- Yes, this makes a performance difference.
+local cast = ffi.cast
 
 -- Returns true if x and y are structurally similar (isomorphic).
 function equal (x, y)
@@ -373,7 +375,8 @@ if ffi.abi("be") then
    function htonl(b) return b end
    function htons(b) return b end
 else
-   function htonl(b) return bswap(b) end
+  -- htonl is unsigned, matching the C version and expectations.
+   function htonl(b) return tonumber(cast('uint32_t', bswap(b))) end
    function htons(b) return rshift(bswap(b), 16) end
 end
 ntohl = htonl
--- a/src/README.md
+++ b/src/README.md
@@ -871,14 +871,14 @@ Returns a copy of *array*. *Array* must not be a "sparse array".
 — Function **lib.htons** *n*
 
 Host to network byte order conversion functions for 32 and 16 bit
-integers *n* respectively.
+integers *n* respectively. Unsigned.
 
 — Function **lib.ntohl** *n*
 
 — Function **lib.ntohs** *n*
 
 Network to host byte order conversion functions for 32 and 16 bit
-integers *n* respectively.
+integers *n* respectively. Unsigned.
 
 
 


alexandergall:
Are you sure those are the only things that break? it could be that other cases simply don't excercise the case where the difference matters (i.e. use numbers < `0x80000000`), couldn't it?

kbara:
Nope, more things could indeed break, but those are the ones that have test cases which show it. The whole Snabb codebase only has 30 ntohl and 50 htonl calls, though, which can be manually audited in a reasonable amount of time. (There's also no guarantee that we don't currently have things broken by the signed definition - I see you've seen https://github.com/snabbco/snabb/pull/1023 ).

It's a bit of a pain to change this to be unsigned now, but that pain will only get worse with time. I don't want the signed semantics to hound the project 10 years from now.

alexandergall:
Unfortunately, I don't really remember any of this, but doesn't `tonumber` involve a conversion to a double that you'd like to avoid in many (most?) cases? That would be something to consider, I guess. This point would be moot if LuaJIT optimized the conversion away when it would matter, which may well be the case :)

One could argue that the same issue exists for anything that involves BitOps and to make this clear in the description of the endian conversion API, which would have to document how to use it safely.

eugeneia:
Geat catch! :+1:

kbara:
I believe it's optimized away, but would indeed have to verify that.

There are several legitimate ways to define bitops, and the signed version used in LuaJIT is the only reasonable choice given constraints described at http://bitop.luajit.org/ . On the other hand, I think ntohl/htonl should always be unsigned.

eugeneia:
I agree that ntohl/htonl should be plain bit reverses or no-ops.

kbara:
I see the signed nature of LuaJIT bitops as an implementation detail that shouldn't change ntohl/htonl's external behaviour.

src/core/lib.lua
```diff
@@ -373,7 +373,8 @@ if ffi.abi("be") then
    function htonl(b) return b end
    function htons(b) return b end
 else
-   function htonl(b) return bswap(b) end
+  -- htonl is unsigned, like the C version, to avoid traps...
+   function htonl(b) return tonumber(ffi.cast(uint32_t, bswap(b))) end
```
dpino: Should be ```lua ffi.cast("uint32_t", bswap(b))```

kbara: Excellent catch. I'm so used to dealing with the results of whichevertype_t = ffi.typeof(....) that I entirely overlooked that, but it does indeed make the tests pass locally. Uses of it will still need to be manually checked, though, since this is a breaking semantic change.

kbara:
This change has non-zero overhead - but none of the difference appears to be floating-point related, and the original version is already more expensive than one might expect.

```
% cat signed.lua 
ffi = require("ffi") -- not needed, but here to ease comparability.
bit = require("bit")

function ntohl(x) return bit.bswap(x) end

for i = 1,100 do
   print(ntohl(0xffffffffff))
end
```

```
ffi = require("ffi")
bit = require("bit")

function ntohl(x) return tonumber(ffi.cast('uint32_t', bit.bswap(x))) end

for i=1,100 do
   print(ntohl(0xffffffffff))
end
```

I compared the machine code from each one.
```
luajit -jv -jdump=mbi,dump.signed signed.lua
luajit -jv -jdump=mbi,dump.unsigned unsigned.lua
```

They're very similar; the instructions other than mov/cmp/jnz are identical in both files - and mov/cmp/jnz are not floating point operations.
```
% grep 0b dump.signed| grep -v mov | grep -v cmp | grep -v jnz        
0bcafe81  cvttsd2si ebx, [rdx]
0bcaff90  xorps xmm7, xmm7
0bcaff93  cvtsi2sd xmm7, ebx
0bcaffa1  sub eax, edx
0bcaffa6  jb 0x0bca0014 ->1
0bcaffe1  add edx, +0x38
0bcafff4  jmp 0x0041f809
0bcafe4a  addsd xmm7, xmm5
0bcafe4e  ucomisd xmm6, xmm7
0bcafe52  jb 0x0bca0014 ->1
0bcafe61  jmp 0x0bcafe6d
```

Oddly, xorps is for single-precision floating point point numbers, which I didn't expect from reading http://bitop.luajit.org/semantics.html (which mandates doubles, since 24 bits of precision isn't enough - so, without digging into this, I'll assume that's not related to the bswap/bitops) - but addsd, cvtsi2sd, and ucomisd  work on doubles. Notably, both versions use ```cvttsd2si``` to convert a double to an int.

But there are significantly more mov/cmp/jnz instructions in the unsigned dump. The signed dump has 74 machine instructions, the unsigned dump has 97. (Nowhere near all of those are ntohl costs; trace2 is purely loop-related overhead, for instance.) The unsigned version contains extra IR:

```
0016 >  p32 HREFK  0015  "tonumber" @8
0017 >  fun HLOAD  0016
0018 >  p32 HREFK  0015  "ffi" @35
0019 >  tab HLOAD  0018
0020    int FLOAD  0019  tab.hmask
0021 >  int EQ     0020  +31 
0022    p32 FLOAD  0019  tab.node
0023 >  p32 HREFK  0022  "cast" @6
0024 >  fun HLOAD  0023
...
0033 >  fun EQ     0024  ffi.cast
0034 }  cdt CNEWI  +10   -1  
0035 >  fun EQ     0017  tonumber
```
Notably, nothing uses 0034, so the CNEWI is dead rather than alarming.

I'd be inclined to say that the performance of this version is acceptable though suboptimal if the performance of the previous version was, unless hydra says differently.

kbara:
I've isolated the calls to ntohl, by changing the loop in the test scripts to

```
for i = 1,100 do
   collectgarbage()
   a = ntohl(0xffffffff)
   collectgarbage()
   print(a)
end
```
(Why collectgarbage()? It's the first never-compiled function in the base library list on http://wiki.luajit.org/NYI ). 

This results in 3 traces, and the second one is purely related to ntohl (including function lookups/etc), plus setting up the second collectgarbage.

Looking more at the IR, the local dance helps, as always:
```
local bswap = bit.bswap
local cast = ffi.cast
```

The dumps (generated in the same way as the previous post, with byte code, IR, and machine code, three traces in each) shrink nicely with the locals dance:

```
% wc -l dump*
  214 dump.lsigned -- with the locals dance
  247 dump.signed

  240 dump.lunsigned -- with the locals dance
  290 dump.unsigned
```

Adding ```local tonumber = tonumber``` helps further:
```
% wc -l dump.l2unsigned
222 dump.l2unsigned
```

222, the number of dump lines with everything declared local,  is barely larger than the 214 of the localized signed version (and better than the 247 of the version currently in snabb!).

If I isolate the machine code of trace 2, both the signed and unsigned versions are now 50 instructions long - the extra IR is optimized away by this point. They visually look identical, aside from inconsequential details like exact addresses.

I expect with the ```local``` change, this will be a mild performance boost over current upstream.

kbara:
Correction: current upstream already localizes bswap, so I expect performance will be exactly the same.

kbara:
A few details of the above analysis are subtly wrong due to the example code having the optimizer constant-fold away the bswap, as shown by the lack of BSWAP in the IR. The overall point about the generated assembly being identical with and without unsigned semantics continues to hold, though.

The code I'm using to avoid consonant folding and maximize isolation of the trace is this:

```
ffi = require("ffi")
bit = require("bit")
io = require("io")
math = require("math")

local cast = ffi.cast
local bswap = bit.bswap
local tonumber = tonumber
local write = io.write
local random = math.random

function ntohl(x) return tonumber(cast('uint32_t', bswap(x))) end

for i=1,100 do
   local base_num = math.random(0, 0xffffffff)
   collectgarbage()
   a = ntohl(base_num)
   collectgarbage()
   write(a)
end
```

kbara:
To be even more precise, it's not absolutely identical, once constant folding is eliminated - there's the use of a 64-bit register vs only its lower 32 bits in one instruction, and some reordering that I think is trivial. But it's the same number of basically the same operations.

```
Signed:
0bcafdf6  movd r15d, xmm7
0bcafdfb  bswap r15d
0bcafdfe  mov rdi, 0xfffffffb41b1c310
0bcafe08  cmp rdi, [rbp+0x5d8]
0bcafe0f  jnz 0x0bca0010        ->0
0bcafe15  cmp dword [rbx+0x10], +0x00
0bcafe19  jnz 0x0bca0010        ->0
0bcafe1f  xorps xmm7, xmm7 ; zero out xmm7
0bcafe22  cvtsi2sd xmm7, r15d  ; Note the use of r15d here; it's just the lower 32 bits
0bcafe27  movsd [rbp+0x5d0], xmm7
```

```
Unsigned:
0bcafd9d  movd r15d, xmm7
0bcafda2  bswap r15d
0bcafda5  xorps xmm7, xmm7 ; zero out xmm7
0bcafda8  cvtsi2sd xmm7, r15 ; note the lack of a d - this is all 32 bits
0bcafdad  mov rdi, 0xfffffffb41568e60
0bcafdb7  cmp rdi, [rbp+0x5d8]
0bcafdbe  jnz 0x0bca0010        ->0
0bcafdc4  cmp dword [rbx+0x10], +0x00
0bcafdc8  jnz 0x0bca0010        ->0
0bcafdce  movsd [rbp+0x5d0], xmm7
```

How to see what it's doing in practice:

1) Get a dump:
```luajit -jdump=mbi,dump.unsigned -Ohotloop=1 unsigned.lua```

2) Use GDB:
```gdb --args luajit -Ohotloop=1 u.lua ```

In gdb, run the following (based on the assembly for the unsigned version, above), interpreting # to be a comment (although it isn't gdb syntax):
```
(gdb) start # necessary for hbreak to work on x86_64
(gdb) hbreak *0bcafd9d
```
Set up gdb to show some useful info - r15, r15d, and xmm7 determined from the asm snippet for unsigned at the start of this comment:
```
(gdb) display /i $pc
1: x/i $pc
=> 0x403e90 <main>:     push   %r12
(gdb) display $r15
2: $r15 = 0
(gdb) display $r15d
3: $r15d = 0
(gdb) display $xmm7
4: $xmm7 = {v4_float = {0, 0, 0, 0}, v2_double = {0, 0}, v16_int8 = {
    0 <repeats 16 times>}, v8_int16 = {0, 0, 0, 0, 0, 0, 0, 0}, v4_int32 = {0, 0, 0, 0}, 
  v2_int64 = {0, 0}, uint128 = 0}
```

Continue running it:
```
(gdb) c
Breakpoint 2, 0x000000000bcafd9d in ?? ()
1: x/i $pc
=> 0xbcafd9d:   movd   %xmm7,%r15d
2: $r15 = 1073777184
3: $r15d = 1073777184
4: $xmm7 = {v4_float = {-6.61220694, 184, 0, 0}, v2_double = {6755402676148019, 0}, 
  v16_int8 = {51, -105, -45, -64, 0, 0, 56, 67, 0, 0, 0, 0, 0, 0, 0, 0}, v8_int16 = {
    -26829, -16173, 0, 17208, 0, 0, 0, 0}, v4_int32 = {-1059875021, 1127743488, 0, 0}, 
  v2_int64 = {4843621402472060723, 0}, uint128 = 4843621402472060723}
```

Changing values is easy (the following has no real impact, because $r15 is about to be set when the next instruction runs)
```
(gdb) set $r15d = 0xffffffff
(gdb) print $r15
$1 = 4294967295
(gdb) print $r15d
$2 = -1
```

Go forward one assembly instruction:
```
(gdb) ni
0x000000000bcafda2 in ?? ()
1: x/i $pc
=> 0xbcafda2:   bswap  %r15d
2: $r15 = 3235092275
3: $r15d = -1059875021
4: $xmm7 = {v4_float = {-6.61220694, 184, 0, 0}, v2_double = {6755402676148019, 0}, 
  v16_int8 = {51, -105, -45, -64, 0, 0, 56, 67, 0, 0, 0, 0, 0, 0, 0, 0}, v8_int16 = {
    -26829, -16173, 0, 17208, 0, 0, 0, 0}, v4_int32 = {-1059875021, 1127743488, 0, 0}, 
  v2_int64 = {4843621402472060723, 0}, uint128 = 4843621402472060723}
```
Recognize $r15d as the first value in v4_int32.

...
Typing return in gdb on an empty line repeats the last command (ni). The cvtsi2sd, the one opcode that meaningfully diverges, is about to be called:

```
(gdb) 
0x000000000bcafda8 in ?? ()
1: x/i $pc
=> 0xbcafda8:   cvtsi2sd %r15,%xmm7
2: $r15 = 865588160
3: $r15d = 865588160
4: $xmm7 = {v4_float = {0, 0, 0, 0}, v2_double = {0, 0}, v16_int8 = {
    0 <repeats 16 times>}, v8_int16 = {0, 0, 0, 0, 0, 0, 0, 0}, v4_int32 = {0, 0, 0, 0}, 
  v2_int64 = {0, 0}, uint128 = 0}
(gdb) 
0x000000000bcafdad in ?? ()
1: x/i $pc
=> 0xbcafdad:   movabs $0xfffffffb40008168,%rdi
2: $r15 = 865588160
3: $r15d = 865588160
4: $xmm7 = {v4_float = {-3.68934881e+19, 25.2245655, 0, 0}, v2_double = {865588160, 0}, 
  v16_int8 = {0, 0, 0, -32, -23, -53, -55, 65, 0, 0, 0, 0, 0, 0, 0, 0}, v8_int16 = {0, 
    -8192, -13335, 16841, 0, 0, 0, 0}, v4_int32 = {-536870912, 1103743977, 0, 0}, 
  v2_int64 = {4740544288130072576, 0}, uint128 = 4740544288130072576}
```

After calling it, you can see 865588160 in v2_double.

**Comment by kbara** [comment]
Audit of ntohl/htonl signedness safety change in the Snabb source tree.

Summary: most code using the signed version appears to believe it's unsigned (every single backing type with defined signedness in a struct in unsigned). Code where the signedness matters uses one of two other implementations(!), both unsigned. All in-tree uses seem to be fine; ```git blame``` on uses of ntohl in code without significant in-tree use suggest that @alexandergall, @eugenia, and @capr might want to make sure no out-of-tree code they have breaks.

----

The analysis.

- I'm excluding documentation or 'local dance' lines, as well as .c or .h files and the defintions (rather than uses) of ntohl/htonl.

* apps/lwaftr/binding_table.lua:      return map:lookup(ffi.C.ntohl(ipv4_as_uint), port)
  * This was sidestepping the core/lib.lua version, perhaps because of the signedness issue.
* apps/lwaftr/fragmentv6.lua:   ```return ntohl(rd32(pkt.data + frag_id_start))```
  * This is safe; it's the return value of a function (get_frag_id) which in turn is only compared to other results of that function.

* apps/lwaftr/lwaftr.lua:   return ntohl(rd32(get_ipv4_src_address_ptr(ptr)))
* apps/lwaftr/lwaftr.lua:   return ntohl(rd32(get_ipv4_dst_address_ptr(ptr)))
  * These two lines are used to get src/dst IPv4 addresses and are used in two ways
     * To generate ICMP (this goes through apps/lwaftr/lwaftr.lua, and then into a ```uint8_t[4]``` header field in lib/protocol/ipv4.lua, and is fine)
     * to 'enqueue' encapsulation to IPv6 within the lwAFTR; these are put in softwire_key_t's ```uint32_t ipv4``` field before they can cause trouble.
* apps/test/lwaftr.lua:      local ipdst = C.ntohl(rd32(ipv4_hdr.dst_ip))
  * Another C.ntohl....
* core/lib.lua:ntohl = htonl
* core/lib.lua:     [32] = { ntoh = ntohl, hton = htonl }
  * These two just imply that htonl also needs to be audited
* lib/protocol/esp.lua:      return(ntohl(h.spi))
* lib/protocol/esp.lua:      return(ntohl(h.seq_no))
  * Both of these clearly intend an unsigned ntohl; the backing header contains ```uint32_t spi; uint32_t seq_no;```. The only in-tree user is lib/ipsec/esp.lua, which in turn only seems to be setting unsigned backing fields. Heads up to @eugeneia in case there are any out of tree uses he knows of.
* lib/protocol/gre.lua:      return ntohl(self:header().key)
  * This clearly intends to have a signed ntohl, as the is a uint32_t in the data structure. Not used elsewhere in the snabb tree. @alexandergall should probably get a heads-up about this one, as the person who seems most likely to need to check whether it breaks any out-of-tree code.
* lib/protocol/keyed_ipv6_tunnel.lua:      return ntohl(h.session_id)
  * Once again, the backing data type is a ```uint32_t```. No substantial in-tree uses. Another heads up to @alexandergall.
* lib/protocol/tcp.lua:   return ntohl(h.seq)
* lib/protocol/tcp.lua:   return ntohl(h.ack)
  * Backed by ```uint32_t    seq; uint32_t    ack;```; all in-tree uses are in this file are fine with either signedness.
* program/lisper/lisper.lua:   return ntohl(p.spi)
* program/lisper/lisper.lua:   local sid = ntohl(p.session_id)
* program/lisper/lisper.lua:      ntohl(p.session_id),
  * The first two key into local tables and should be ok with either signedness, the last is in a print. Heads up to @capr just in case.

----

* apps/keyed_ipv6_tunnel/tunnel.lua:      psession[0] = lib.htonl(conf.local_session)
  *  ```psession_id_ctype = ffi.typeof("uint32_t*")``` - It's instantly cast to unsigned.
* apps/lwaftr/conf_parser.lua:   return ffi.C.htonl(ffi.cast('uint32_t*', addr)[0])
  * Yet another use of ffi.C's signed version 
* apps/lwaftr/fragmentv6.lua:   wr32(base + 4,  htonl(frag_id))
  * It's writing a raw bit pattern, signedness doesn't matter.
* apps/lwaftr/icmp.lua:local htons, htonl = lwutil.htons, lwutil.htonl
* apps/lwaftr/lwaftr.lua:local htons, htonl = lwutil.htons, lwutil.htonl
* apps/lwaftr/lwheader.lua:local htons, htonl = lwutil.htons, lwutil.htonl
  * Yes, these three use a *third* definition of htonl, explicitly unsigned...
* apps/test/lwaftr.lua:      ipdst = C.htonl(ipdst + self.b4_ipv4_offset)
  * Yet another use of ffi.C's signed version 
* apps/vlan/vlan.lua:   return htonl(bit.bor(bit.lshift(dot1q_tpid, 16), vid))
  * In the end, the use boils down to ```cast(uint32_ptr_t, payload)[0] = tag``` - with an explicit cast to unsigned.
* lib/ipsec/aes_128_gcm.lua:   h.padding = htonl(0x1)
  * High bit isn't set, signedness doesn't matter.
* lib/ipsec/aes_128_gcm.lua:   h.spi = htonl(spi)
* lib/ipsec/aes_128_gcm.lua:   h.seq_no[0] = htonl(seq_h)
* lib/ipsec/aes_128_gcm.lua:   h.seq_no[1] = htonl(seq_l)
  * The backing types are unsigned, so these three are implicitly cast.
* lib/protocol/esp.lua:      h.spi = htonl(spi)
* lib/protocol/esp.lua:      h.seq_no = htonl(seq_no)
  * The backing types are unsigned, so these two are implicitly cast.
* lib/protocol/gre.lua:      self:header().key = htonl(key)
  * The backing types is unsigned, so this is implicitly cast.
* lib/protocol/keyed_ipv6_tunnel.lua:      h.session_id = htonl(id)
  * The backing types is unsigned, so this is implicitly cast.
* lib/protocol/tcp.lua:      h.seq = htonl(seq)
* lib/protocol/tcp.lua:      h.ack = htonl(ack)
  * The backing types are unsigned, so these two are implicitly cast.

**Comment by kbara** [comment]
@eugeneia I believe this is ready to be merged. It has ~no performance impact (same number of assembly instructions, effectively identical beyond the use of a 64 bit vs 32 bit subset of a 64 bit register in a single one of them, slightly reordered), and does not appear to break any in-tree code. (In tree code that actually cares about signedness seems to use other implementations, mainly the ffi.C one).

**Comment by alexandergall** [comment]
Just out of curiosity: would
```
function htonl(b) return band(0xFFFFFFFF, bswap(b)) end
```
have made any difference (unless I'm overlooking a semantic difference, in which case please ignore)?

**Comment by kbara** [comment]
There is a semantic difference between this patch and ```function htonl(b) return band(0xFFFFFFFF, bswap(b)) end```, which has almost the same broken semantics as the version without the band.

That would still return a signed int, rather than an unsigned one, because luaJIT bitops are all signed by definition. IE, we would still have had #1023 and awkwardness with dealing with a signed ntohl. https://github.com/Igalia/snabb/issues/453 would still have been a bug, for the same underlying reason - the expression would have been 32 bit rather than sign-extended 64, but we'd still be comparing a signed 32 bit int to an unsigned 32 bit int and having the comparison fail.
In short:
```
> a = 0xf0000000
> print(bit.band(0xffffffff, a) == a)
false
```
As we're using a LuaJIT with IEEE doubles as the backing numeric type, an unsigned 32 bit int will never be equal to a negative signed one.

Demonstration that the backing type is double:
```
> z = 18014398509481985
> print(z)
1.8014398509482e+16
> print(z == 18014398509481984)
true
```

Secondarily, it also might not do exactly what you expect if you give ctypes to an ntohl implemented purely with bitops:
```
print(ffi.sizeof(bit.band(0xffffffff, ffi.cast("uint32_t", 1))))
8
```

PR #1025, opened 2016-09-23
standardise link/counter names

eugeneia:
As discussed in #871 this PR the shorthands “rx” and “tx” are removed from link and counter (SHM) names and replaced with “input” and “output”. This change affects many modules and the changes to each module should be reviewed by their authors/maintainers.

The most notable change that the translation is reversed for link counters, e.g. rx is now output and tx is input, e.g.:
- input = packets enqueued on the link (was: tx = packets transmitted onto the link)
- output = packets dequeued from the link (was: rx = packets received off the link)

Depends on: #1019


lukego:
Great initiative! Have only given it a quick run-through but no problems jump out at me.

eugeneia:
See further discussion in #1030

PR #1026, opened 2016-09-26
Merge max-next into next

eugeneia:
Includes: #1024 #1023


src/apps/vlan/vlan.lua
```diff
@@ -147,6 +147,29 @@ function VlanMux:transmit(o, pkt)
...
+      01 10 08 00 9c d4 07 c0 00 01 bc fa e3 57 00 00
+      00 00 f3 44 01 00 00 00 00 00 10 11 12 13 14 15
+      16 17 18 19 1a 1b 1c 1d 1e 1f 20 21 22 23 24 25
+      26 27 28 29 2a 2b 2c 2d 2e 2f 30 31 32 33 34 35
+      36 37
+   ]], 82))
+   local payload = pkt.data + o_ethernet_ethertype
+   local vid = 0
+   for i=0,15 do
+      for j=0,255 do
```
kbara: Pedantically, only 12 of those 16 bits are strictly the tag, but it doesn't really matter for this test case.

PR #1028, opened 2016-10-04
next: Changes queued for the v2015.10 release

lukego:


lukego:
@eugeneia Release at will!

PR #1030, opened 2016-10-12
Merge PR #1025 (Max's tx/rx->input/output)

kbara:


lukego:
Hm. The code looks very innocent but in the Hydra job this really [looks like](https://hydra.snabb.co/build/636545/download/2/report.html#l2fwd) it introduces a performance regression for the l2fwd benchmark. @eugeneia any ideas?

src/program/snabbmark/snabbmark.lua
```diff
@@ -45,18 +45,18 @@ function basic1 (npackets)
...
-   config.link(c, "Tee.tx1 -> Sink.rx1")
-   config.link(c, "Tee.tx2 -> Sink.rx2")
+   config.link(c, "Source.output -> Tee.input")
+   config.link(c, "Tee.output1 -> Sink.input1")
+   config.link(c, "Tee.output2 -> Sink.input2")
    engine.configure(c)
    local start = C.get_monotonic_time()
    timer.activate(timer.new("null", function () end, 1e6, 'repeating'))
-   while link.stats(engine.app_table.Source.output.tx).txpackets < npackets do
+   while link.stats(engine.app_table.Source.output.output).output_packets < npackets do
```
eugeneia: So here is a potential mistake in translation that could affect benchmark results (although not of the l2fwd benchmark). I changed “packets transmitted onto the link” to ”packets read off the link” because I felt it was correct in this instance, and further instances below. E.g. I think its “cheating” to stop when the Source has emitted N packets, as opposed to when these N packets have actually been processed.

Opinions?

eugeneia:
I am not sure how to read the results. I started a [new run](https://hydra.snabb.co/jobset/max/link-name-regression#tabs-evaluations) for 51c135f51c19dad9295b3e100048ab53e8be8144 and get this:

box

dens

```
> summary(aov(score ~ snabb, read.csv("~/tmp/bench.csv")))
               Df Sum Sq Mean Sq F value   Pr(>F)    
snabb           1     33   33.07   44.96 2.08e-11 ***
Residuals   13998  10296    0.74                     
---
Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
2800 observations deleted due to missingness
```

```
> model.tables(aov(score ~ snabb, read.csv("~/tmp/bench.csv")), type="means")
Tables of means
Grand mean
         
6.501847 

 snabb 
snabb
link-name-regression               master 
               6.550                6.453 
```

eugeneia:
Looks like the issue is not with absolute performance but the distribution of results. So where does the variance come from? :O

src/core/link.h
```diff
@@ -9,7 +9,10 @@ struct link {
   //   http://en.wikipedia.org/wiki/Circular_buffer
   struct packet *packets[LINK_RING_SIZE];
   struct {
-    struct counter *dtime, *txbytes, *rxbytes, *txpackets, *rxpackets, *txdrop;
+    struct counter *dtime,
+                   *input_bytes, *output_bytes,
+                   *input_packets, *output_packets,
+                   *input_drop;
```
lukego: Is the naming of `input` vs `output` clear to everybody else?

Previously we had `link.transmit()` incrementing the `txpackets` which seems like a direct match. Now we have `link..transmit()` incrementing the `input_packets` which seems more ambiguous to me. (Does the sender input packets _onto_ a link, or does the receiver input packets _from_ a link?)

If this is just me then it doesn't matter but otherwise we could consider e.g. `enqueue` and `dequeue` as perhaps being harder to misinterpret.

lukego: (Or even `added_bytes` and `removed_bytes` or something.)

lukego:
Good testing. Yes, it does look like there is a voodoo problem here. I have no idea what it could be. Least crazy guess is that the new names might be creating a hash collision in a Lua table somewhere. (That _could_ indirectly cause JIT behavior changes because hashtable lookups are specialized on the size of the hashtable and any change will require a side-trace. If this is the case then the solution may actually be to get #1037 working so that we don't have to worry about such crazy stuff. But maybe there is a more innocent explanation.)

The best idea that I have for debugging, which is not a great one, would be to split the change up into multiple smaller commits (e.g. four parts) and have Hydra test them all in parallel. Then when you see which commit first introduces the problem you can split that one up and test the parts. (In the spirit of `git bisect`.)

lukego:
@domenkozar Just checking: Could there also be any change on the lab servers in between the test runs that Max has made? Just thinking that e.g. if we have upgraded NixOS then we might see a difference if one branch was tested before and one after.

lukego:
(These results are sufficiently suspicious that it could be worth forcing them to rerun from scratch somehow.)

domenkozar:
@lukego the only reason I can think of is monitoring https://github.com/snabblab/snabblab-nixos/issues/89

I'd test the hypothesis by commiting a dummy change to the branch causing problems. I'd turn off monitoring before that. Sounds good?

lukego:
I think we should start with the more general question: Has anything changed in the lab that influences test results? Then if that is affirmative we could look for the more specific cause e.g. munin monitoring.

Simple way do that could be for @eugeneia to add two more branches, each with a trivial change added to the branches tested before, and then we see whether the test results are the same as we see now (i.e. determined by snabb code) or whether they are different (e.g. determined by date on which teh tests were executed).

domenkozar:
Yeah let's go with that. I'll deploy meanwhile a monitoring script that uses `lock` just be sure it doesn't clobber benchmarks.

eugeneia:
Pollution of test results was also my first suspect. Here is the evaluation the with the dummies added: https://hydra.snabb.co/eval/4011

Let’s see what that gives.

lukego:
NB: If the test results come back clean I will merge this branch. The idea about renaming link counters could be taken in a separate PR targeting `next` if we decide we care about that. (I need to check out the new Github "reviews" feature, it could be neat if that provided a way to distinguish between comments in passing vs specific requests. Anybody done much with it?)

lukego:
@eugeneia Could you drop the priority ("scheduling shares") from 1000 down to e.g. 50? This jobset is really big (~15K tests) and seems to be blocking various much smaller jobs.

@domenkozar One day I need to ask you more about the Hydra "declarative jobsets." Seems like our jobset definitions would be much easier to maintain as Nix code in a Git repo where we could use common parameters and library functions to keep them in harmony. Just now I am sure we have a haphazard mix of priorities, nixpkgs verisons, snabb branch selections, etc in our jobsets. (Hopefully we can still use the webui for ad-hoc jobs too... I would be curious to know more about the feature.)

domenkozar:
@lukego it's quite easy, see https://hydra.mayflower.de/project/hydra-pr-jobs

It creates `.jobset` jobset, which fetches a git repository containing jobsets, for example https://github.com/mayflower/hydra-prs

Then if new configuration appears, a new jobset is created or modified.

lukego:
@domenkozar Oh that looks really wonderful. Great that it plays nicely with the imperatively managed jobsets and that each project can have a separate jobset repo.

lukego:
(Oh and I am able to use the Hydra command "Bump to front of queue" to fast-track the small jobs now as a once-off. Cool.)

domenkozar:
Yes :) The main issue is getting Hydra built with declarative jobsets support, since Nix and Hydra master are currently in need of fixing :) But Mayflower have a branch from which they run their hydra, worth taking a look.

eugeneia:
The regression is in fact real: https://hydra.snabb.co/build/689209/download/2/report.html

kbara:
Looks like it is. Interesting.

By the way, what's 'id', and why is 'score' on y axis when it's usually on the x axis in the other graphs?

eugeneia:
`id` is the identifier for a given test run. This jobset tested two benchmarks (basic l2fwd with 64 and 265 byte packets), on four different versions, with 100 iterations. Each benchmark/version combination is run 100 times, hence the id. I don’t think the id axis in meaningful for the box plot. As to the choice of axis, I don’t know.

lukego:
@eugeneia Interesting! I would be curious to know if you can see the issue on the `basic1` benchmark or only the more complex ones?

eugeneia:
Here is a report with the basic benchmark included:https://hydra.snabb.co/build/737970/download/2/report.html

lukego:
Interesting :). Challenging not only to account for the difference but even to describe it.

How about splitting up this change (e.g. rebase on a temp branch) into many smaller commits, for example to update app separately, and then have Hydra compare them all in parallel? This could be a divide-and-conquer strategy to isolate the issue(s).

What do you think about scaling back the Hydra job a bit too? Just now it's running 7000 tests per branch with high scheduler priority (1000). I think you'd probably be fine with 1/10th the tests at 1/10th the priority.

eugeneia:
I split up the changes into smaller chunks like so, note that every commit follows the previous one, e.g. they are linear.

```
commit 07d8d2bfafc63c72391b69af71477058b394667d
Author: Max Rottenkolber <max@mr.gy>
Date:   Mon Oct 24 19:04:45 2016 +0200

    lwaftr

commit 3cde879cc82bd4f1832b803dcff7bb54043dc736
Author: Max Rottenkolber <max@mr.gy>
Date:   Mon Oct 24 19:04:37 2016 +0200

    lisper

commit a41c9278c514781570cd6d23db552afd941843e5
Author: Max Rottenkolber <max@mr.gy>
Date:   Mon Oct 24 19:04:26 2016 +0200

    standardize stats counters.

commit 4af450e9566ea93b0192a9cca1f15f88ef107be5
Author: Max Rottenkolber <max@mr.gy>
Date:   Mon Oct 24 19:03:01 2016 +0200

    standardize device ports.

commit 73f4effd31a3b69460bc6f452164829dcc61fd92
Author: Max Rottenkolber <max@mr.gy>
Date:   Mon Oct 24 20:44:29 2016 +0200

    standard link counter names
```

Here are the results (slightly hard to read because they are so crowded): https://hydra.snabb.co/build/767819/download/2/report.html

eugeneia:
It indeed looks to me as if there is some hash collision thing going on, or at least, I find it peculiar that `link-counter-names` seems to be the only branch not affecting performance (its a metatype). Maybe this tells us something about regular LuaJIT tables that we have so far ignored, and points us towards metatypes?

Then there is the `stats-counter-names` sweet-spot which seems to improve things over master. This commit mostly entails longer/different keys for statistics counters.

`port-names` drops the ball, OK, maybe the port keys are now *too* regular and collide somewhere. We had a mix of rx/tx input/output, now its lots of input/output.

`lisper` and `lwaftr` only change the respective programs, I am inclined to rule them out as factors. Here is a report without them:

https://hydra.snabb.co/build/764551/download/2/report.html

eugeneia:
The issue with ruling out `lisper` and `lwaftr` is that linearly speaking, without them this branch is indicated as a significant net plus in l2fwd. The big dip in `port-names` is gone after amending `stats-counter-names`, and reintroduced by the seemingly unrelated changes.

lukego:
First thought is that there seems to be an easy to reproduce test case here. The `basic1` benchmark usually scores 30 for one branch and 32 for another. This test probably only takes some tens of seconds to run on a local machine.

lukego:
So the commit that changes behavior is eugeneia/snabb@4af450e? I have taken a quick look at profiling this but I don't immediately see the difference (not so surprising because it is relatively small). The best path forward that I see is recursive divide-and-conquer i.e. keep on finding the smallest change that causes a behavior change and splitting that up until we only have a few lines left.

lukego:
Digression: Just imagine how cool it would be if our built-in profiling support was so good that we could solve mysterious performance problems like this one offline and after the fact :). Then instead of running new tests all the time we could zoom in closer on the results that we already have.

Here is a babystep in that direction: I ran `snabbmark basic1` with timeline logging support and then converted the timeline to CSV. This way we can use R to analyze the performance of many different engine breaths all sampled from the same process.

First simple example is a density plot showing how long (cycles) the push and pull calls to each app instance take. This is for `basic1` that generates packets with a `Source`, duplicates them with a `Tee`, then discards them with a `Sink`.

appdensity

Just a proof of concept, please take with a grain of salt. Could potentially use this to make comparisons between branches and find very specific things e.g. if one particular app slows down or becomes less consistent.

kbara:
What's the plan with the rx-related changes? I'd like to be able to feed other things to next via kbara-next. :-)

kbara:
Closing, as per discussion with @eugeneia

lukego:
(Just observing that Github seems to have successfully kept the above discussion intact even though we reset the branch to remove these commits. Good to know.)

wingo:
I wonder if this naming change regressed because previously the ordering of the app_array and link_array could depend on how things hashed, and thus the expected number of visits through that array could vary depending on how names changed.  Maybe after #1066 the name change is not a regression!

eugeneia:
Thanks for that insight/theory, I will retest this together with #1066 to check.

eugeneia:
Great news! @wingo’s theory seems to manifest: https://hydra.snabb.co/build/858603/download/2/report.html#density-plot

lukego:
Just fantastic :).

PR #1034, opened 2016-10-14
Virtio uses shiftleft/shiftright to tag packets

wingo:
This builds on #1032 and uses shiftleft/shiftright in the virtio-net driver to tag and untag packets.  It also switches guest snabb applications to use direct descriptors.

### DIFF
--- a/src/apps/ipv6/nd_light.lua
+++ b/src/apps/ipv6/nd_light.lua
@@ -123,7 +123,6 @@ function nd_light:new (conf)
    -- Prepare packet for solicitation of next hop
    local nh = { nsent = 0 }
    local dgram = datagram:new()
-   nh.packet = dgram:packet()
    local sol_node_mcast = ipv6:solicited_node_mcast(conf.next_hop)
    local ipv6 = ipv6:new({ next_header = 58, -- ICMP6
          hop_limit = 255,
@@ -148,6 +147,7 @@ function nd_light:new (conf)
    dgram:push(ethernet:new({ src = conf.local_mac,
                              dst = ethernet:ipv6_mcast(sol_node_mcast),
                              type = 0x86dd }))
+   nh.packet = dgram:packet()
    dgram:free()
 
    -- Timer for retransmits of neighbor solicitations
@@ -176,7 +176,6 @@ function nd_light:new (conf)
    -- Prepare packet for solicited neighbor advertisement
    local sna = {}
    dgram = datagram:new()
-   sna.packet = dgram:packet()
    -- Leave dst address unspecified.  It will be set to the source of
    -- the incoming solicitation
    ipv6 = ipv6:new({ next_header = 58, -- ICMP6
@@ -197,6 +196,8 @@ function nd_light:new (conf)
    -- Leave dst address unspecified.
    dgram:push(ethernet:new({ src = conf.local_mac,
                              type = 0x86dd }))
+   sna.packet = dgram:packet()
+
    -- Parse the headers we want to modify later on from our template
    -- packet.
    dgram = dgram:new(sna.packet, ethernet)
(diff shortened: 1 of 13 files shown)


wingo:
stats will be here: https://hydra.snabb.co/build/659075

wingo:
@lukego wdyt about this one?  Only affects Snabb guests, not DPDK or other guests.  There is no corresponding change for the host because the choice of descriptor layout is up to the guest, not the host.  The PR is hard to read though because it includes earlier patches from #1032; perhaps just look at the last two commits.

Nicola Larosa and @domenkozar are working on getting lwaftr tests into Hydra but I think we don't yet have them wired up.  Still, obvious improvement I think, simplifying the code and giving us twice the descriptor count.

wingo:
ping :)

lukego:
I suggest workflow of @eugeneia or @kbara being upstream for this and then sending it onwards to me for rubber-stamping. I have no strong feelings on this module - wasn't involved in the development - and have no access to test cases for it either.

kbara:
It looks like a solid win on iperf results. On the three 'l2fwd by configuration' graphs, it looks like it may be slightly worse; I can't tell eyeballing it if that is significant.

Interestingly, 2.1.0 noind is still bimodal.

kbara:
I'd be willing to upstream it, if @lukego is happy with an Igalian upstreaming for an Igalian.

wingo:
That linked perf analysis includes the result of #1032.  Let me see if I find a new one.

lukego:
Oh, or. perhaps I have misunderstood this change. Is this changing the device side of virtio-net? I will look more closely now...

wingo:
This is changing the driver side of virtio-net.

kbara:
It changes  src/lib/virtio/virtq_driver.lua

lukego:
OK. In that case I prefer not to be upstream as per ^^^ and I don't think this code has any coverage in existing Hydra tests (which currently always runs DPDK inside the guest and never Snabb).

@kbara Sure but `virtq_device` is the code being exercised by CI i.e. the device (hypervisor/snabbnfv) side.

wingo:
@kbara there is no up-to-date perf thing.  however it will not show on the graphs because #736 hasn't landed; we don't have any automated tests with snabb guests AFAIU, neither integration nor benchmarking :(

lukego:
(The ideal would be to add a Hydra case that runs Snabb inside the VM in addition to DPDK. This should be fairly straightforward FWIW.)

lukego:
(i.e. it would be an extension of the existing `l2fwd` benchmark that we run on Hydra but it would run with a different VM image i.e. one that runs the Snabb-based l2fwd instead of the DPDK-based one.)

wingo:
I verrrrry much understand not wanting to land something without CI, but I think the right thing to do here is just merge this so we can merge the lwaftr.  Merging it does not make the situation worse and this is the code that we are shipping and we are the only users of this code, and in parallel @dpino  has #736 open and Domen and Nicola are working on getting hydra+lwaftr+qemu+... working too, so things are on the right track IMO.  @kbara wdyt?

kbara:
Looks like @lukego has thumbed up it. I'm ok with that workflow, with the caveat that I think we *really* need to do our best to have test coverage of it in the next ~2 weeks, before the next release is cut.

kbara:
I've reviewed it. LGTM, @wingo . As per out of band discussion, please merge this to wingo-next, since you'll need it there for the lwaftr merge and there's no point putting it in two maintainer-next branches. :)

eugeneia:
Slightly off-topic...

> we are the only users of this code

There have been two instances of students asking for help with virtio_net on snabb-devel (latest instance: https://groups.google.com/forum/#!topic/snabb-devel/8eLfuiKksjY ). Sadly, I couldn’t help them because I have never used virtio_net and neither have a way to test it. Since there seems to heightened interest in this, I think it would be very valuable to make virtio_net a first class Snabb citizen like vhost_user, and integration with the l2fwd benchmark would go a long way!

PR #1037, opened 2016-10-16
Self-optimizing engine [sketch]

lukego:
This is a highly experimental engine extension to automatically detect JIT code that is suboptimal and force it to recompile. The high-level goal is to keep the Snabb process optimized for its current workload at all times without needing to globally `jit.flush()` and without waiting for an operational problem (e.g. packet drop) to take action.

This is a feature I have been hashing out over on the LuaJIT repo:
- Trace profiling feature (LuaJIT/LuaJIT#224).
- Side-trace efficiency issue (LuaJIT/LuaJIT#218).

This implementation will print a message any time it recompiles a trace. It also prints an overall summary of JIT code behavior every one second, like this (example from `snabbmark basic1` behavior):

```
[trace profile:  93.9% loop;   6.1% head;   0.0% foreign]
[trace profile:  93.5% loop;   6.5% head;   0.0% foreign]
[trace profile:  94.2% loop;   5.8% head;   0.0% foreign]
```

Generally the idea is that if code spends a lot of time in the non-looping ("head") parts of traces then it's probably not well optimized. This could be due to JIT heuristics making poor choices, which this change should automatically remedy, or programming styles that don't work well with the JIT, which we can better understand as the next problem.

The Snabb change is commit aacd8ff and the rest is the LuaJIT trace profiling extension.

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -14,6 +14,8 @@ local zone      = require("jit.zone")
 local jit       = require("jit")
 local ffi       = require("ffi")
 local C         = ffi.C
+local jitutil   = require("jit.util")
+local traceprofile = require("jit.traceprofile")
 require("core.packet_h")
 
 -- Packet per pull
@@ -268,6 +270,7 @@ function main (options)
       breathe = latency:wrap_thunk(breathe, now)
    end
 
+   traceprofile.start(1)
    monotonic_now = C.get_monotonic_time()
    repeat
       breathe()
@@ -275,6 +278,7 @@ function main (options)
       if not busywait then pace_breathing() end
    until done and done()
    counter.commit()
+   traceprofile.stop()
    if not options.no_report then report(options.report) end
 end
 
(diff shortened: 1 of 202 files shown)


src/core/app.lua
```diff
@@ -444,25 +442,46 @@ function report_apps ()
...
-         if nonloop > loop*2 then
-            print("flushing "..tr)
-            -- More than twice as much time in the non-loop? Suspicious.
-            jit.flush(tr)
+         local nonloop, loop, ffi = traceprofile.tracestats(tr)
+         if (loop+nonloop > 100) and (nonloop > loop*2) then
+            -- Suspicious trace: has used a significant amount of CPU
+            -- and less than a third of its time spent in the loop.
+            print(("[flushing trace %d: low loop intensity (%d%%)]"):format(
+                  tr, 100*loop/(loop+nonloop)))
```
eugeneia: I think there is a `jit.flush(tr)` missing here.

lukego: Thanks! I had missed that & could well be the reason that my early tests are showing slow traces being detected but not corrected :-).

lukego:
This fine-grained trace-at-a-time approach currently seems less promising than the more coarse-grained approach in #1039 based on my experiments to date. More thoughts at https://github.com/LuaJIT/LuaJIT/issues/218#issuecomment-254448416.

wingo:
Did you mean to tag #1309, @lukego ?  AFAIU the change there is orthogonal; you pretty much always want to know if the system is dropping on ingress.

PR #1038, opened 2016-10-17
Numa upstream

wingo:
This PR is a distilled port of https://github.com/Igalia/snabb/pull/341 to Snabb master, adding a couple of interfaces to ljsyscall, pulling in a getcpu() bugfix from @kbara, and then adding a Snabb module that lets you do `numa.bind_to_cpu(CPU)` in Snabb.  That will bind the current thread to that CPU, bind the memory for that thread to that CPU's NUMA node, and migrate all memory to that NUMA node.  It's ideal for hooking up to a `--cpu` command-line argument, with no `taskset` / `numactl` shenanigans, and with the additional benefit of migrating pages.  (That's what we do in the lwaftr; this commit is from Igalia/lwaftr and something we need to land upstream before merging upstream.)

The ljsyscall change is going upstream here: https://github.com/justincormack/ljsyscall/pull/193.  Embarassingly I can't yet run docker on this guixsd machine so I haven't been able to add/fix the tests upstream yet :P  However this patch does run in production, so there's that.

### DIFF
--- a/lib/ljsyscall/syscall/linux/c.lua
+++ b/lib/ljsyscall/syscall/linux/c.lua
@@ -347,6 +347,13 @@ function C.sched_setparam(pid, param)
   return syscall(sys.sched_setparam, int(pid), void(param))
 end
 
+function C.get_mempolicy(mode, mask, maxnode, addr, flags)
+  return syscall(sys.get_mempolicy, void(mode), void(mask), ulong(maxnode), ulong(addr), ulong(flags))
+end
+function C.set_mempolicy(mode, mask, maxnode)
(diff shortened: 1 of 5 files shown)


kbara:
LGTM. I'm assuming this should also go into wingo-next; if you'd rather it go into kbara-next, give me a shout.

wingo:
Thanks @kbara, will merge to `wingo-next`.  Just the NFV thing left to do before merging.

eugeneia:
LGTM! `lib.numa` should probably have a README though.

wingo:
Tx for review, already in next.  Readme on the way in #1048.

PR #1039, opened 2016-10-17
Avoid NFV reconfigure when nothing has changed

wingo:
[Note: this PR originally contained two unrelated changes and was edited to contain just one.  What follows is the original PR note.  See https://github.com/snabbco/snabb/pull/1039#issuecomment-254790578 for the update.]

Two changes in this PR from downstream `lwaftr` branch: one to add the "warning" ingress drop monitor to the NFV.  This will simply detect drops and print out a warning to the console when many drops are detected.  The second avoids calling `reconfigure` when the configuration hasn't actually changed.  The former helps us track down which part of a system is slow, and the latter is a minor perf improvement. I'll push to my `optimize` branch to see what hydra thinks.

### DIFF
--- a/src/program/snabbnfv/traffic/traffic.lua
+++ b/src/program/snabbnfv/traffic/traffic.lua
@@ -83,19 +83,22 @@ function long_usage () return usage end
 function traffic (pciaddr, confpath, sockpath)
    engine.log = true
    local mtime = 0
-   if C.stat_mtime(confpath) == 0 then
-      print(("WARNING: File '%s' does not exist."):format(confpath))
+   local needs_reconfigure = true
+   function check_for_reconfigure()
+      needs_reconfigure = C.stat_mtime(confpath) ~= mtime
    end
+   timer.activate(timer.new("reconf", check_for_reconfigure, 1e9, 'repeating'))
+   -- Flush logs every second.
+   timer.activate(timer.new("flush", io.flush, 1e9, 'repeating'))
    while true do
-      local mtime2 = C.stat_mtime(confpath)
-      if mtime2 ~= mtime then
-         print("Loading " .. confpath)
-         engine.configure(nfvconfig.load(confpath, pciaddr, sockpath))
-         mtime = mtime2
+      needs_reconfigure = false
+      print("Loading " .. confpath)
+      mtime = C.stat_mtime(confpath)
+      if mtime == 0 then
+         print(("WARNING: File '%s' does not exist."):format(confpath))
       end
-      engine.main({duration=1, no_report=true})
-      -- Flush buffered log messages every 1s
-      io.flush()
+      engine.configure(nfvconfig.load(confpath, pciaddr, sockpath))
+      engine.main({done=function() return needs_reconfigure end})
    end
 end
 


wingo:
hydra: https://hydra.snabb.co/build/685340

lukego:
Great! I am becoming very interested in these JIT flushing heuristics lately :).

wingo:
No idea why there are new hydra failures here; will look into it I guess

wingo:
(For that matter, what is going on with the standard tests?)

wingo:
Bah, I think there are ingress drop monitor patches missing from this PR.  Will fix.

wingo:
Apologies for the confusion; I had missed a commit that gave the ingress drop monitor the ability to have different actions (https://github.com/Igalia/snabb/pull/361).  Rebased to include that patch before anything else.

wingo:
perfies will be https://hydra.snabb.co/build/704184

lukego:
I would like to be upstream on this please :) I am very interested in this topic at the moment.

src/lib/timers/ingress_drop_monitor.lua
```diff
@@ -5,45 +5,59 @@ module(...,package.seeall)
...
       if app.ingress_packet_drops and not app.dead then
-         local status, value = with_restart(app, app.ingress_packet_drops)
-         if status then sum[0] = sum[0] + value end
+         sum[0] = sum[0] + app:ingress_packet_drops()
       end
    end
 end
 
-function ingress_drop_monitor:jit_flush_if_needed ()
+local tips_url = "https://github.com/Igalia/snabb/blob/lwaftr/src/program/lwaftr/doc/README.performance.md"
```
lukego: Could we do this in a way that is more general? This code is being merged onto other branches and called from other programs now.

lukego:
Have you considered making this also react to drops on Virtio interfaces? (Hard? Bad idea?)

src/lib/timers/ingress_drop_monitor.lua
wingo: Yeah, sorry that this crept in; good catch :)  Does making this point to a non-lwaftr document in snabbco/snabb work for you?  If so I will make sure that such a document exists and use its URL.

wingo:
Regarding virtio, it's hard.  Right now it's impossible to drop packets on a host -> guest interface, since the packet allocation is controlled by the guest.  Even now we don't fully drain the last link in the app graph that feeds VhostUser:pull().

In practice to know whether it's the NFV or the guest which is being slow, you look at that last link.  If the NFV is dropping packets on ingress but the last link is full, that means the guest is too slow.  Otherwise if the NFV is dropping packets but that last link regularly empties then the guest is fine and the problem is in the host.  If the host isn't dropping packets on ingress then all is fine.

What about this do you think should change?  I would think we could drain the ->VhostUser link in VhostUser:push or so, perhaps emitting a warning.

lukego:
Please allow me to zoom out for a moment, not to suggest changing this patch, but rather to map out the context for this whole ingress-drops business mostly for the sake of future directions.

Here is how I have it all in my head:

The problem we want to solve is to detect when a Snabb process is overloaded and then to run a callback. The callback could take corrective action (JIT flush) or provide diagnostics (log message).

Definition: Snabb is _overloaded_ when the average packet arrival rate exceeds the average packet processing rate.

This means that during overload a backlog will grow on one or more ingress interfaces, and once the backlog exceeds the queue size we will start to drop packets. Conversely, during "underload" we are processing packets faster than they are arriving, on average, and this means that any backlog on ingress interfaces will be shrinking and each interface will regularly be drained empty during a `pull()`.

So how do we detect overload? There are a bunch of conditions and we could use one or more of them as proxies:

1. Packets are being dropped on ingress. Pro: Closely related to overload that exceeds tolerable bounds. Con: Has to be implemented specially for each I/O driver.
2. I/O interface runs out of packet buffers for receive (requires a complete refill on `pull()`). Pro/Con: Similar to (1) but perhaps easier/harder to implement for certain I/O drivers.
3. I/O interface consistently outputs the maximum burst size on `pull()` i.e. ~100 packets per breath (`engine.pull_npackets`) meaning that the packet arrival rate is >= the packet processing rate (because otherwise the queue would be shrinking towards zero). Pro: Should work for any I/O driver with an ingress queue size >= ~100. Con: ?

So putting this into the context of detecting ingress overload on a Virtio-net interface if we don't immediately have a "dropped packets counter" for (1) then we could use (2) by detecting if every descriptor on the receive vring is occupied with a packet (no room for more) or (3) by checking if the number of packets pulled by the virtio app is consistently `engine.pull_npackets`.

How much of this makes sense and what am I missing?

src/lib/timers/ingress_drop_monitor.lua
lukego: Yes that sounds fine to me.

wingo:
This PR originally had two unrelated things: a change to the way the NFV loop runs, and a way to use the ingress drop monitor to warn when the NFV is slow.  Those changes were unrelated.  I have now force-pushed the branch to only make the former change and will do something else for the latter one.  I will push to `wingo/optimize` so we can get numbers/CI on this one and will ping when results are in.

wingo:
Hydra results will be https://hydra.snabb.co/build/719109

wingo:
Hydra results up.  The relevant comparison is between wingo-optimize and master, as wingo-optimize is just one commit beyond master.

wingo:
Well, I dunno.  Hydra results do not show a regression and also do not show a clear progression.  WDYT @lukego?  I could go either way on this, though I lean towards landing it.

lukego:
Yes, I like the new formulation, it's stylish. Merge it onto wingo-next?

wingo:
Will do, thanks :)

PR #1042, opened 2016-10-18
apps.test.synth: send correct amount of packets per breath.

eugeneia:
The `Synth` up would send a multiple of `engine.pull_npackets` if it had more than one `size`, now it sends `pull_npackets + (pull_npackets % #sizes)` per breath.


eugeneia:
I will be my own upstream on this minor change, if that’s OK.

PR #1044, opened 2016-10-19
Update perf docs

wingo:
This PR merges the lwAFTR performance documentation into the main performance-tuning.md document, preferring the wording from the lwAFTR doc.  Thoughts?

### DIFF
--- a/src/doc/performance-tuning.md
+++ b/src/doc/performance-tuning.md
@@ -6,7 +6,7 @@ system-oriented tips and move towards application configuration.
 
 ## BIOS settings
 
-### Prioritize performance over efficiency
+### Prioritize performance over energy efficiency
 
 Go into your BIOS and verify that you or your hardware vendor have not
 enabled aggressive power-saving modes that could downclock your
@@ -123,10 +123,11 @@ $ cat /proc/interrupts
 ### Disable IOMMU
 
 The oldest CPU supported by Snabb is the Sandy Bridge series, and this
-series happens to have a bug in their IOMMU support that cripples
-snabb performance.  As a result, we never test Snabb data planes with
-the IOMMU on, so in production it should also be disabled via a kernel
-boot parameter:
+series happens to have a bug in their IOMMU support that can manifest
+itself either as total packet loss or severe performance penalties.
+As a result of this bug, we never test Snabb data planes with the
+IOMMU on even on later CPU models, so in production you should also
+disable the IOMMU via a kernel boot parameter:
 
 ```
 # Ubuntu
@@ -330,9 +331,14 @@ So, a bigger ring buffer insulates packet processing from breath
 jitter.  You want your ring buffer to be big enough to not drop
 packets due to jitter during normal operation, but not bigger than
 that.  In our testing we usually use the default ring buffer size.  In
-your operations you might want to increase this up to 2048 entries.
-We have not found that bigger ring buffer sizes are beneficial, but it
-depends very much on the environment.
+your operations you might want to increase this up to 1024 or 2048
+entries.  We have not found that bigger ring buffer sizes are
+beneficial, but it depends very much on the environment.
+
+Remember that buffers are fundamentally mechanisms for smoothing over
+variations in latency.  The real problem is the latency variation --
+focus your effort on getting that right and you won't have to fiddle
+much with ring buffer sizes.
 
 ## Virtualization
 
--- a/src/doc/performance-tuning.md
+++ b/src/doc/performance-tuning.md
@@ -21,7 +21,7 @@ disable energy efficiency settings or profiles, typically named "Max
 performance", "Energy Efficiency" and "Custom".  Select "Max
 performance" for latency sensitive Snabb use.
 
-### Hyperthreads
+### Disable hyperthreads
 
 Hyperthreads are a way of maximizing resource utilization on a CPU core,
 driven by the observation that a CPU is often waiting on memory or some
@@ -32,7 +32,7 @@ want.  We do not want another thread competing for compute and cache
 resources on our CPU and increasing our latency.  For best results and
 lowest latency, disable hyperthreading via the BIOS settings.
 
-### Tune Turbo Boost
+### Consider disabling Turbo Boost
 
 Intel Turbo Boost Technology allows processor cores to run faster than
 the rated operating frequency if they're operating below power,
@@ -55,7 +55,7 @@ for CPUFREQ in /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor; do
 done
 ```
 
-### Isolate CPUs
+### Reserve CPUs for data plane usage
 
 When running a Snabb data plane, we don't want interference from the
 Linux kernel.  In normal operation, a Snabb data plane won't even make
@@ -85,7 +85,7 @@ For a standard Ubuntu or other system, edit `/etc/default/grub` to add the appro
 GRUB_CMDLINE_LINUX_DEFAULT="... isolcpus=1-5,7-11 ..."
 ```
 
-### Avoid interrupts
+### Avoid interrupts on data-plane CPUs
 
 Normally Linux will handle hardware interrupts on the first core on a
 socket.  In our case above, that would be cores 0 and 6.  That works
@@ -445,8 +445,8 @@ pid 1's current affinity list: 0-15
 
 #### Avoid fragmentation
 
-Fragment reassembly in particular is a costly operation.  Make sure
-that MTUs are set such that fragmentation is rare.
+Fragment reassembly is a costly operation.  Make sure that network
+MTUs are set such that fragmentation is rare.
 
 #### Ingress and egress filtering
 
@@ -497,7 +497,7 @@ caches.  The actions of one processor causing the TLBs to be flushed
 on other processors is what is called a TLB shootdown.
 
 Occasional packet loss at ingress has been observed while TLB
-shootdowns happened on the system. There are per cpu counters of TLB
+shootdowns happened on the system.  There are per-CPU counters of TLB
 shootdowns available in the output of 'cat /proc/interrupts':
 
 ```
@@ -505,8 +505,5 @@ $ cat /proc/interrupts |grep TLB
  TLB:       2012       2141       1778       1942       2357       2076       2041       2129       2209       1960        486          0          0          0          0          0          0        145          0          0   TLB shootdowns
 ```
 
-They must remain 0 for the CPU serving Snabb (pinned via taskset or
-numactl). One possible source of such TLB shootdowns can be the use of
-SHM between processes on different nodes or pinning Snabb to a list of
-CPUs instead a single one (TODO: needs confirmation. Have seen
-periodic TLB shootdowns before optimization based on this doc);
+They should remain 0 for the CPU serving Snabb (pinned via taskset or
+numactl).
--- a/src/doc/performance-tuning.md
+++ b/src/doc/performance-tuning.md
@@ -1,154 +1,184 @@
 # Maximizing deployment performance
 
 For maximum performance, several hardware, operating system and Snabb
-parameters need to be tuned. Note that this document is only on tuning
-for deployment performance, not on how to write performant Snabb code.
+parameters need to be tuned.  We start with some generic
+system-oriented tips and move towards application configuration.
 
-## Snabb
+## BIOS settings
 
-### ring-buffer/num_descriptors
+### Prioritize performance over efficiency
 
-Defined in src/apps/intel10g.lua and adjustable before (!) the NIC
-gets initialized:
+Go into your BIOS and verify that you or your hardware vendor have not
+enabled aggressive power-saving modes that could downclock your
+processors.  A CPU in a power-saving mode typically takes some time to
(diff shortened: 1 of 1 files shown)


src/doc/performance-tuning.md
```diff
@@ -257,6 +439,71 @@ root@c819e0f106c4:/# taskset -cp $$
...
 
+## Application-specific performance notes
+
+### lwAFTR
+
+#### Avoid fragmentation
+
+Fragment reassembly is a costly operation.  Make sure that network
+MTUs are set such that fragmentation is rare.
+
```
kbara: It might be worth mentioning tuning the size of the reassembly buffers?

src/doc/performance-tuning.md
```diff
@@ -257,6 +439,71 @@ root@c819e0f106c4:/# taskset -cp $$
...
+#### Ingress and egress filtering
 
+Simply enabling ingress and/or egress filtering has a cost.  Enabling
+all filters adds 4 apps to the Snabb graph, and there is a cost for
+every additional Snabb app.  In our tests, while a normal run can do
+10 Gbps over two interfaces in full duplex, enabling filters drops
+that to 8.2 Gbps before dropping packets.  However we have not found
+that the performance depends much on the size of the filter, as the
+filter's branches are well-biased.
```
kbara: It would be useful to have Hydra re-testing this overhead routinely - @tekniko.

kbara: (Notably, I wonder whether it really kills performance that badly on on-a-stick mode)

src/doc/performance-tuning.md
```diff
@@ -110,95 +205,184 @@ Node 1 HugePages_Total:    32
...
+switch or the lwAFTR.
+
+So, a bigger ring buffer insulates packet processing from breath
+jitter.  You want your ring buffer to be big enough to not drop
+packets due to jitter during normal operation, but not bigger than
+that.  In our testing we usually use the default ring buffer size.  In
+your operations you might want to increase this up to 2048 entries.
+We have not found that bigger ring buffer sizes are beneficial, but it
+depends very much on the environment.
+
```
kbara: I've heard contradictory things on the benefits of 512 vs 1024 - the previous version strongly stumped for 1024. It might be worth explicitly preserving the information that some people have found 1024 to be a win compared to 512.

src/doc/performance-tuning.md
```diff
@@ -1,91 +1,185 @@
...
+Hyperthreads are a way of maximizing resource utilization on a CPU core,
+driven by the observation that a CPU is often waiting on memory or some
+external event, and might as well be doing something else while it's
+waiting.  In such a situation, it can be advantageous to run a second
+thread on that CPU.  However for Snabb that's exactly what we don't
+want.  We do not want another thread competing for compute and cache
+resources on our CPU and increasing our latency.  For best results and
+lowest latency, disable hyperthreading via the BIOS settings.
+
```
kbara: Might be worth mentioning that having hyperthreading enabled can cause packet loss.

src/doc/performance-tuning.md
```diff
@@ -1,91 +1,185 @@
...
+### Consider disabling Turbo Boost
+
+Intel Turbo Boost Technology allows processor cores to run faster than
+the rated operating frequency if they're operating below power,
+current, and temperature specification limits.  The conservative thing
+to do is to turn off Turbo Boost when doing performance tests of Snabb
+data planes, because that way you reduce variance due to CPU clock
+speed changing.
+
```
kbara: I seem to recall Turbo Boost backfiring slightly when you switch back from it, as well. Does your recollection match mine?

src/doc/performance-tuning.md
```diff
@@ -257,6 +439,71 @@ root@c819e0f106c4:/# taskset -cp $$
...
+```
+engine.busywait = true
+engine.main(...)
+```
+
+If true then the engine will poll for new data in a tight loop (100%
+CPU) instead of sleeping according to the Hz setting. This will reduce
+overall packet latency and increase throughput at the cost of
+utilizing the CPU hosting Snabb at 100%.
+
```
kbara: Want to comment on what we've empirically observed with enabling/disabling busywait?

src/doc/performance-tuning.md
```diff
@@ -1,91 +1,185 @@
...
-### Turbo Mode
-Intel Turbo Boost Technology allows processor cores to run faster than the rated operating frequency if they're operating below power, current, and temperature specification limits. (TODO: impact not yet analyzed on Snabb, nor if it is controlled by the performance profile). 
-## Linux Kernel
 ### Disable IOMMU
-Sandybridge CPUs have a known issue on its IOTBL huge page support, impacting small packet performance.  Newer CPUs don't have this issue.
-(TODO, only found this info here: [http://dpdk.org/ml/archives/dev/2014-October/007411.html]()
-(TODO: pass through mode: [https://lwn.net/Articles/329174/]())
 
-Add IOMMU=pt (pass through) to the kernel:
+The oldest CPU supported by Snabb is the Sandy Bridge series, and this
```
kbara: I think it's worth documenting that on older supported processors, having IOMMU on can (IIRC) lead to 30%-100% packet loss - that's pretty dramatic, and was hard-won non-obvious knowledge that it would be sad to have people have to troubleshoot from scratch again.

src/doc/performance-tuning.md
```diff
@@ -110,95 +205,184 @@ Node 1 HugePages_Total:    32
...
-#Should irqbalance be enabled?
-ENABLED="0"
-...
-```
+If the Snabb network function that you are using does not have a
+`--cpu` option, file a bug.  Binding Snabb to a CPU and NUMA node can
+also be done using the `numactl --membind` and `taskset -c` commands,
+but we recommend the `--cpu` argument as it is easiest and it also
+runs various consistency checks, such as checking that PCI devices
+that you use are local to your NUMA node.
```
kbara: Thank you for writing --cpu!

src/doc/performance-tuning.md
wingo: I have the feeling that when people were tweaking ring buffer sizes in the hopes of dropping fewer packets, they were grasping at straws... if you are fine with it I prefer this tone.  If we think there should be a new default we should do more science and change the default, not recommend that users choose a specific different default.  As it is users are welcome to do what experimenting they like.

src/doc/performance-tuning.md
```diff
@@ -110,95 +205,184 @@ Node 1 HugePages_Total:    32
...
 lstopo --of pdf > lstopo.pdf
 ```
 
-Example from a Lenovo RD650 (Intel(R) Xeon(R) CPU E5-2650 v3): 
+Example from a Lenovo RD650 (Intel(R) Xeon(R) CPU E5-2650 v3):
 ![lstopo.png](lstopo.png)
 
-### PCI Card and Snabb on same NUMA node
-Memory shared between Snabb (by means of huge page mapping) and the NIC must share the same node. This is achieved via 'numactl', once the correct node is identified for a given NIC port/PCI address. 
-To find the node for a given PCI address, use cpulistaffinity combined with numactl (TODO: is there a more direct way??):
```
kbara: Is it worth documenting cpulistaffinity+numactl, in addition to the grovel around in /sys approach?

src/doc/performance-tuning.md
```diff
@@ -110,95 +205,184 @@ Node 1 HugePages_Total:    32
...
+
+### QEMU & Vhost-User
+
+[Vhost-User](http://www.virtualopensystems.com/en/solutions/guides/snabbswitch-qemu/)
+is used to connect Snabb with a high performance virtual interface
+attached to a QEMU-based virtual machine.  This requires hugepages
+made available to Qemu:
 
 ```
-$ lspci|grep 10-
```
kbara: The new doc doesn't mention lspci at all; it seems to me like it perhaps should.

src/doc/performance-tuning.md
wingo: I think it's implicit here.  I am interpreting "might be nice" comments to mean "think about it but nothing required" and I have thought and here I choose not to act :)  Please do lmk if I misinterpret this.

src/doc/performance-tuning.md
wingo: What do you mean? :)  This does not ring any bells.

src/doc/performance-tuning.md
```diff
@@ -1,91 +1,185 @@
...
+Under NixOS, edit `/etc/nixos/configuration.nix` to add this parameter:
 
 ```
-engine.busywait = true
-engine.main(...)
+boot.kernelParams = [ "isolcpus=1-5,7-11" ];
 ```
-If true then the engine will poll for new data in a tight loop (100% CPU) instead of sleeping according to the Hz setting. This will reduce overall packet latency and increase throughput at the cost of utilizing the CPU hosting Snabb at 100%. 
 
-### Monitor ifInDiscards
```
wingo: Yes.  IMO people should use the ingress drop monitor or the ingress drop counter, mentioned later.

kbara:
Tons of nits, but this looks awesome overall. Thank you.

src/doc/performance-tuning.md
wingo: I don't remember enough to say something here.  Do you? :)

src/doc/performance-tuning.md
wingo: Not sure how this squares with this point -- if the oldest supported CPU is Sandy Bridge, it's not just a perf drop, sometimes iommu on means you get no packets at all iirc.  Best to always say "don't have iommu on"!

src/doc/performance-tuning.md
kbara: I suspect it's also grasping at straws, but I'm a little reluctant to not document that it was done and what was tentatively found, even if just in passing - ie, a line like "While we don't have robust evidence that changing the size of the ring buffer is a win, some users have reported better results with 1024 than 512. YMMV."

src/doc/performance-tuning.md
kbara: https://github.com/Igalia/snabb/blob/lwaftr/src/program/lwaftr/doc/README.configuration.md
lwaftr.conf example includes:
```
max_fragments_per_reassembly_packet = 1,
max_ipv6_reassembly_packets = 10,
max_ipv4_reassembly_packets = 10,
```

The counters docs, in turn, show how to use lwaftr query -l to get the actual memory use for the ctable buffers.

I'm ok with it not getting documented yet, but it is in our main lwaftr branch.

src/doc/performance-tuning.md
wingo: To me that is more in basic configuration and not performance tuning.  Not sure.  You have any reason that it should?

src/doc/performance-tuning.md
kbara: You're not misinterpreting. I just lazily think that being able to ^F for 'packet loss' should be able to find the most common red flags. :-)

src/doc/performance-tuning.md
kbara: I have a very fuzzy memory of switching back and forth between turbo boost and non-turbo boost mode having some overhead/latency, but perhaps I'm confounding it with hyperthreading switching costs.

src/doc/performance-tuning.md
kbara: Hmm. Probably just "it was in the previous doc" bias. README.running.md and README.troubleshooting.md mention it, and as you said it's basic, so I think that's good enough.

src/doc/performance-tuning.md
kbara: Okdokie.

src/doc/performance-tuning.md
kbara: I think it's worth explicitly saying "having IOMMU on can cause severe (up to 100%) packet loss on some older supported processors." It's highly non-obvious, and the current documentation doesn't reflect just how severe the problems can be.

src/doc/performance-tuning.md
kbara: Nope, hence fishing to see if you did with that wording. :-)

src/doc/performance-tuning.md
kbara: Ok.

src/doc/performance-tuning.md
wingo: As this is a PR for master which doesn't have these knobs yet afaiu, I'll leave these out for now.  But yes we should doc them!

wingo:
Updated, ptal @kbara.  Thanks :)

src/doc/performance-tuning.md
kbara: Having thought a bit more, I'd really like this to explicitly say that we have observed hyperthreading causing packet loss.

kbara:
LGTM aside from not explicitly saying that hyperthreading causes packet loss (I think the "best results and lower latency" wording is too vague, at least by itself.

wingo:
Tx for review, added mention of packet loss for hyperthreading.

wingo:
I will merge to `wingo-next` and send to luke for merging today :)

kbara:
Awesome. :-)

PR #1046, opened 2016-10-19
Merge wingo-next (#1038, #1039, #1044)

wingo:


lukego:
Merged, thanks!

I do think some of the commit messages are a bit terse :-) "Tweaks" could be "Tweaks to performance-tuning.md."

I note that `numa.lua` has no description of its API but I tend to see this as a symptom of a systemic documentation problem (is anybody _really_ using the markdown docs instead of reading the source files? Github analytics suggests not more than one person per day.)

wingo:
Thanks for the merge, and I will be less terse in the future :)

kbara:
I use markdown docs a bit (ie, the lwaftr ones on counters, troubleshooting, etc), but not as an API reference. I tend to view them locally in vim.

lukego:
@kbara Do you think I should have required a README.md markdown document for `numa.lua` to merge?

lukego:
@kbara (well, moot question perhaps, since the `numa` module issue is about markdown API documentation and you are saying that you don't tend to refer to that.)

lukego:
@kbara (anyway, we can pick up the API discussion again in the future, and maybe the conclusion will be to keep doing what we are doing, or maybe we will come up with a new idea. Just now my problem is that I feel too bad about making markdown API documentation a hard requirement on merging changes because I'm not confident that people are really using it.)

kbara:
Frankly, I have no opinion on this in general, was just trying to throw in a data point. I'd be reluctant to lobby for Andy to write more documentation in the absence of stronger opinions.

lukego:
I would have been comfortable asking for API documentation as comments in the source file, which are actually very valuable to me in practice, except that I don't feel that I have a mandate for this based on past discussions.

wingo:
I am happy to write docs fwiw, can do that in a followup

PR #1047, opened 2016-10-20
Merge lwaftr to next

wingo:
This PR merges in the last 5 months or so of lwaftr development to upstream Snabb.  It also adds Snabb vMX, the system that joins up a Snabb dataplane to an external virtualized control/management plane, also adding "on-a-stick" operation for the lwaftr in particular but in a way that can apply to any Snabb dataplane.

With a few minor exceptions, this diff includes only changes to `src/apps/lwaftr`, `src/programs/snabbvmx`, and `src/programs/lwaftr`.  Review is very welcome on any of it, but I expect that you will want to focus on the non-lwaftr, non-vmx bits.


wingo:
Hydra run here: https://hydra.snabb.co/build/737038

src/apps/lwaftr/podhashmap.lua
```diff
@@ -75,6 +70,32 @@ function PodHashMap.new(key_type, value_type, hash_fn)
    return phm
 end
 
+-- FIXME: There should be a library to help allocate anonymous
+-- hugepages, not this code.
```
kbara: Still true. :-)
(Not a blocker, no action required for this PR).

src/core/timer.lua
```diff
@@ -70,6 +70,7 @@ end
 
 function selftest ()
    print("selftest: timer")
+
```
kbara: This seems like a spurious new line. (Not a blocker)

eugeneia:
There is a new “nh_fwd”  app that seems to be very lwaftr centric, or at least the README is vague to me. Why does it not belong into `apps/lwaftr`?

wingo:
The nh_fwd (next-hop forward) app is part of snabb vmx.  It sits outside (around, really) a snabb data plane and forwards control traffic to a virtualized control plane, such as an Ubuntu image or a Juniper vMX switch. That way it learns the next hop and can stamp on the L2 address of where the packets should go next.  It is not lwaftr-specific.

src/lib/timers/ingress_drop_monitor.lua
```diff
@@ -2,48 +2,78 @@ module(...,package.seeall)
...
+      action = args.action or 'flush',
+      last_flush = 0,
+      last_value = ffi.new('uint64_t[1]'),
+      current_value = ffi.new('uint64_t[1]')
+   }
+   if args.counter then
+      if not args.counter:match(".counter$") then
+         args.counter = args.counter..".counter"
+      end
+      if not shm.exists("/"..S.getpid().."/"..args.counter) then
```
eugeneia: I think this should be `shm.exists(args.counter)`. SHM paths are no longer resolved based on a changing `shm.path`.

kbara:
I haven't spotted anything else that @eugeneia hasn't.

kbara:
The three new commits look ok to me.

wingo:
Added a couple patches to `lwaftr` that address Max's feedback.  @kbara how does it look to you?

kbara:
LGTM, thanks for all your work on this.

dpino:
Originally `nh_fwd` was its own app but then it was moved to apps/lwaftr. It's true that it could be used for other apps that need to forward traffic to a service (in this case the lwaftr) and to a VM, but until there's a real need to reuse it we decided to move it to apps/lwaftr. What @eugeneia spotted is he remaining README that we forgot to remove.

wingo:
So I made the rookie mistake of just making this a PR from lwaftr instead of branching off lwaftr and PR'ing from that.  I will merge https://github.com/snabbco/snabb/pull/1047/commits/3d33fb6d834222306b32fef8eee39a59de69516e to wingo-next and pr that to next.

wingo:
@dpino thanks for the clarification on that, let's fix the readme issue in a followup

wingo:
Closing this PR as it's already merged and I made the rookie mistake of PRing this from a live branch which continues to get updates and thus false-alarm notifications.  Thank you all!

PR #1051, opened 2016-10-25
Reset IPv4 checksum before recalculating it

dpino:
Checksum method in `lib/protocol/ipv4.lua` always recalculates an IPv4 header checksum. The checksum field must be set to zero before recalculation.

There was a checksum unit test function in the module but it couldn't detect this error because the sample packet had its checksum field equals to zero.


eugeneia:
Good catch. I am pretty sure that the SnabbBot performance regression indicated here is spurious, so I will merge it.

Issue #1050, opened 2016-10-21
snabb.co redirect to Github


lukego:
I have now transitioned the Snabb domain (http://snabb.co and http://www.snabb.co/) to redirect to our Github project page.  (Could take a little while to propagate.)

This is simply following through on the plan discussed over at https://github.com/snabbco/snabb/issues/738#issuecomment-180727157 i.e. for the Snabb name/trademark to refer to this open source project.

Previously the `snabb.co` website was my own professional  shopfront but I will recreate that later under a different brand.

I suppose the next step is to provide `@snabb.co` email aliases to any maintainers who would like one.

wingo:
Neat.  Thanks for being an amazing project technical lead and steward; you have made a really lovely virtual place to work :-)

lukego:
Thanks Andy :). 'Tis a pleasure to be hacking with you all!

eugeneia:
Closing because the issue is resolved.

PR #1052, opened 2016-10-25
README.md: link to documentation

eugeneia:
For me this is a no-brainer, we have fairly detailed documentation but no obvious public reference towards it. This puts a reference in a place where people looking for it will likely find it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,11 @@ join the [snabb-devel mailing
 list](https://groups.google.com/forum/#!forum/snabb-devel) and read
 on.
 
+## Documentation
+
+- [API Reference](http://snabbco.github.io/)
+- [Contributor Hints](https://github.com/snabbco/snabb/blob/master/CONTRIBUTING.md#hints-for-contributors)
+
 ## How does it work?
 
 Snabb is written using these main techniques:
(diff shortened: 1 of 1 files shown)


eugeneia:
Related: https://github.com/snabbco/snabb/issues/1050

lukego:
Merged, thanks!

I would love to get the usage rate on documentation up. My idea to create a PDF manual was a total failure.

Just now I still don't find the HTML docs very appealing. Comes up really compressed in the left-hand side of the screen. However maybe that is just my bad taste. Screenshot in case it looks different on my computer:

screenshot - 102816 - 11 16 16

eugeneia:
@lukego I edited that screenshot to show my rationale for this layout:
snabbd

The white space is going to be somewhere, dividing it is wasting it. Every other way to deal with this is worse. *Mic drop* ;-)

lukego:
@eugeneia Thanks for the screenshot! This helps me to understand the motivation for the layout. Having everything squeezed on the left-hand side lets you put it behind an editor on the right. I can see why you like that.

This does not work for me because I have different window-management habits. I probably need to build a separate interface to the documentation that works for me. Previously I tried that with PDF but I was not happy with the results. Ongoing process...

PR #1053, opened 2016-10-25
[wip] Add JIT optimization tool: lib.specialize(fn)

lukego:
This branch contains a work-in-progress "design pattern" for making LuaJIT perform really aggressive optimizations under certain circumstances. This is based on the same technique that was recently used to optimize vhost-user (#1001).

The high-level intention is to make the `pull()` and `push()` methods of each app _instance_ be JITed individually and to treat the app configuration parameters as if they were constants. JITing instances separately means that each one will have dedicated machine code and they will not interfere with each other (for example no collisions between instances like the ctype diversity in #612). Treating configuration parameters as constants is essentially "inlining" the configuration values as if they were written as constants in the source file i.e. allowing the optimizer to take advantage of their exact values. (The methods are re-JITed when the app configuration changes.)

The JIT is able to take this even further and to generate machine code that reference singleton objects (such as the app `self` object) using absolute addresses so it is not even necessary to assign a register to hold the address.

Likely some experience will be needed in order to use this feature effectively. Can also be that certain of these benefits are already available in practice without using this special feature.

Notably this will only be effective for apps that start _JIT traces_ in their pull or push methods, which is loosely any app that includes a loop (`for`, `while`, or `repeat`) directly in its pull or push method. If we wanted this to also affect traces started in subroutines then we would need to take a slightly different approach. Could be that we can formulate this in a better way that is easier to understand. Caveat emptor.

This is intended as part of an overall effort to better understand the internal workings of LuaJIT in order to eventually explain them concisely and potentially also to identify tricks for generating code better than what is possible with an ahead-of-time static compiler.

See background discussion at LuaJIT/LuaJIT#208.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -909,6 +909,19 @@ lib.parse({foo=42, bar=43}, {foo={required=true}, bar={}, baz={default=44}})
   => {foo=42, bar=43, baz=44}
 ```
 
+— Function **lib.specialize** *function*
+
+Returns a clone of the given function with a shallow copy of its
+function definition. The clone function has a separate bytecode
+definition and any JIT traces that start in the clone will be
+aggressively specialized by LuaJIT for the environment of the clone.
+
+This can lead to especially efficient machine code when:
+- The function being specialized contains a loop (`for` or `while` or `repeat`) directly in its source code (not in a subroutine because the bytecode cloning is shallow).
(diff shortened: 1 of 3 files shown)


eugeneia:
This seems to trigger a [bug visible in apps.vlan](https://gist.github.com/SnabbBot/21cbe691ef0b782d5ac699dfcbd95f22#file-log-L553). Some edge case related to the shallow copying of functions?

PR #1055, opened 2016-10-28
Merge max-next into next

eugeneia:
Includes:
- SnabbNFV/IPsec fixes #1054
- YANG schema for SnabbNFV #911 
- documentation fixes

I think its a good idea to merge the YANG schema because it is versioned in such a way that nothing prevents us from iterating over it in any direction. E.g. it reflects the current state of SnabbNFV, the schema’s name is versioned (“snabb-nfvconfig-v1.yang”), when we break backwards compatibility we can just create v2.


wingo:
For the YANG schema, isn't the internal versioning sufficient?  SGTM though, we can work in parallel :)

Also as regards the relationship between YANG schemas and data configuration, you might be interested by https://github.com/Igalia/snabb/issues/519, https://github.com/Igalia/snabb/pull/517, and https://github.com/Igalia/snabb/pull/525.

eugeneia:
@wingo @plajjan [pointed out](https://github.com/snabbco/snabb/pull/911#issuecomment-216917284) that it depends on the kind of change, quote:

> FYI, YANG is very strict when it comes to making modifications between revisions so for example you can't rename a leaf if you are to follow the rules. We might want to discuss how to deal with versioning (simple way is to put version in the name, like my-model-v1.yang).

Basically the versioned name makes sure we can “reuse” the “snabb-nfvconfig” even if there are fundamental changes at the price of minimal ugliness.

eugeneia:
I appended #1051 because I thinks its obviously not positive.

lukego:
I am concerned that renaming the traffic policing parameters in the NFV config file  )(commit 16fba64a08bb0992959d4006fa46dad30106cefb) will break any existing configuration files outside the git  tree (actually to make them silently ignore the policing parameters.)

Is it important to make incompatible changes to the NFV config file here?

eugeneia:
Good catch! I don’t think its strictly necessary, I considered this more of a style issue (avoid units in leaf names).

I see multiple options to avoid the silent breakage:

- make nfvconfig accept the old keys as a deprecated fall back
- back out of the change
- …?

lukego:
@eugeneia I reckon old names as a deprecated fallback is good for now :+1:.

Issue #1033, opened 2016-10-14
Introduce named programs to Snabb


xray7224:
As part of the work to support a NETCONF-managed system through YANG modelling (issue #987) we need to be able to assign names to each Snabb program running on a system. This name will be used by the user in the `snabb config` command to specify which program they wish to get/set/etc. on.

The name should be unique string on the system which can be enumerated along with their corresponding PIDs. This then allows for sending messages between processes using their names.

My proposal for implementing this would be to have each program specify its own unique name via a new function in `core.app`. Setting the name would be optional, this is because most programs, at least in the beginning likely won't use this feature until they have a need for it. The function would look something like this:

```lua
function run (params)
    ...
    engine.set_name("myprogram-1")
end
```

The reason for using setters and getters are because upon setting the name, three things will occur, these are:

1. Sets a "name" attribute on the configuration of the running app which can be retrieved later by a `get_name` function.

2. Validates that the name is unique on the system

3. Creates a directory with the PID in `/var/run/snabb/` and creates a symlink in `/var/run/snabb/by-name/<program name>`

These three things ensure the names remain unique on the system using validation and allows for enumeration of programs by their names as mentioned above. The third job of creating the directory and symlink is done at this stage largely for validation however, the setup done in this step will be used by the Shared Memory functionality in `core.shm`.

The NETCONF-managed system will use, as laid out in #987, `core.shm` to create a ring buffer to pass messages between the config app and the program being configured. The user will specify the app whilst using `snabb config` which will then find the shared memory and communicate a set/add/delete/etc. to the program.

wingo:
Looks good in general, note though that the PID directory gets created already.  I.e. anonymous snabb processes have those dirs already.  Maybe it should be "claim_name()" or something, as it is an exclusive resource, and claiming a name could fail.  Dunno :)

lukego:
Just curious: How come the PR for this issue is being reviewed on the Igalia repo (Igalia/snabb#518) instead of here on the upstream snabbco repo? This is fine of course but I see a risk that feedback from people here will come late and potentially create more work for you.

lukego:
(Sorry for not providing some feedback on the idea already btw. Just been busy. I will try to provide some useful input when the code is PR'd to this repo.)

lukego:
[...]

PR #1058, opened 2016-11-02
Introduction of named programs

xray7224:
Hello,

This PR introduces named programs, as per the issue #1033. I plan to merge this via igalia/snabb but posting PR as I'd love more feedback from snabb devs.

Thanks!

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -128,84 +128,46 @@ function configure (new_config)
    counter.add(configs)
 end
 
--- Claims a name for a program so it can be identified by name by other processes.
+-- Claims a name for a program so it's identified by name by other processes.
 --
--- The name given to the function must be unique, if a name has been used before by
--- an active process the function will error displaying an appropriate error message.
--- Any other process can enumerate programs by their name and PID using enumerate_named_programs.
+-- The name given to the function must be unique; if a name has been used before
+-- by an active process the function will error displaying an appropriate error
+-- message. The program can only claim one name, successive calls will produce
+-- an error.
 function claim_name(name)
-   function name_from_path(dir)
-      -- Extracts last file/folder name from path
-      -- e.g. "/var/run/snabb/hello" -> "hello"
-      local ds, de = string.find(dir, "/[^/]*$")
-      return string.sub(dir, ds+1, de)
-   end
-
    configuration[name] = name
    local namedir = "by-name/" .. name
-   local namedirqualified = named_program_root .. "/" .. name
+   local namedir_fq = named_program_root .. "/" .. name
    local piddir = shm.root .. "/" .. S.getpid()
-   local backlinkdir = piddir.."/name-backref"
+   local backlinkdir = piddir.."/name"
 
    -- Verify that the by-name directory exists.
    shm.mkdir(namedir)
 
    -- Verify that we've not already claimed a name
-   if S.lstat(backlinkdir) then
-      -- Read name of program for error message.
-      local fnamedir = S.readlink(backlinkdir)
-      local fname = name_from_path(fnamedir)
-      error("Program already claimed name '"..fname.."', cannot claim name '"..name.."'")
-   end
-
-   -- Check if a symlink to a named program with the same name exists, if not, unlink it.
-   if S.lstat(namedirqualified) then
-      -- It exists, lets check what the PID is.
-      local piddir = S.readlink(namedirqualified)
-      local pid = tonumber(name_from_path(piddir))
-      if S.kill(pid, 0) then
-         -- It's a running process, we should display an error message.
-         error("Name has been claimed by a currently running process: "..namedirqualified)
-      else
-         -- It's dead, we should be able to remove the symlink and relink.
-         S.unlink(namedirqualified)
-      end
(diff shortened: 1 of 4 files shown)


lukego:
Just a note that I have checked for a potential conflict with "process groups" over on #1021 and I do not think there will be a problem.

The idea of "process groups" is that Snabb processes will form a tree with parent/child relationships. The implementation does not change directory layout of `/var/run/snabb` and should be compatible. Every process (parents, children, grandchildren, etc) has its pid listed in `/var/run/snabb` and so any process should be able to call `claim_name()`.

I suppose that as application developers we would need to decide which processes should be named e.g. only the parent or also the children. Seems like we can decide this later.

I would welcome feedback on the process groups idea btw. Could be that it is ill-conceived or ill-described?

eugeneia:
@xray7224 I am curious, is there a specific reason why you did not reorganize the `<pid>` directories into `by-pid/` after all?

xray7224:
@eugeneia I decided it was probably fine as is. One of the main reasons for deciding this would be having to change `SNABB_SHM_ROOT` which is user configurable to not be the directory with the PID files in. Currently my code uses that as a base and creates "by-name" inside of it however, changing it to be `by-pid` leaves me with two choices:

1. Change what the `shm.root` (`SNABB_SHM_ROOT`) actually points to (a directory with both `by-pid` and `by-name`).

2. Use it as is but step back one directory from the root which seems like encroaching outside of the space set aside by the user.

If i were to do this I think I'd definitely go for option one and perhaps this change should be made WDYT?

lukego:
Does this PR make `snabb top` work with process names? That would seem like a really nice feature. I am not using `snabb top` myself and the reason is perceived difficulty of identifying the right process (solved very neatly with names).

wingo:
Probably we should put off making a change to the PID structure of /run/snabb until some time later -- people have different versions of tools and I think this PR lets old "snabb top" keep on working for a while.

eugeneia:
@wingo I don’t think we can or should support a use case where an out-of-tree snabb top is used.

@xray7224 I don’t know what could possibly break so its easy for me to advocate a clean directory structure, but from my experience these hacks will just bite us later. I also think option 1 is reasonable.

Also :+1: to what @lukego said. This is a long standing issue, and I am not innocent in this regard. :-)

wingo:
@eugeneia the use case wasn't external `snabb top`, it was rather a `snabb top` from some previous version or some other mainline-headed but potentially precisely mainline branch

eugeneia:
@wingo That’s what I was referring to, too. Since top is built into the snabb executable, a matching version should always be available and used I think.

lukego:
I am also concerned about the potential for mismatched Snabb versions to communicate via private interfaces that are incompatible. There could be many Snabb executables installed on a machine, $PATH could be different when a process is spawned via systemd vs managed via shell, software can be upgraded on disk while the previous version is still executing, etc. So I worry about this happening accidentally even if the policy would be that you should always use the same version.

I am not sure of an ideal solution. One possibility could actually be for processes like `snabb top` and `snabb config` to try and respawn themselves using the same executable as the target process. If there is a Snabb process running with pid 123 then the kernel will provide `/proc/123/exe` as a symlink to the executable that this process started from. I am using this over on #1021 to try and spawn worker processes using the same software as the parent. The weakness I see is that a symlink is based on name rather than inode and so it seems like you will still get a mismatch of the executable has been replaced on disk.

Could be possible to make a more fool-proof mechanism if the main Snabb process would immediately fork() a process that serves as a template for request processing i.e. could serve requests to fork new children...

eugeneia:
@lukego In that case, how crazy would using a hard link be? I.e. `/var/run/snabb/123/snabb.exe`

lukego:
@eugeneia Hard link to the executable would be nice but the implementation challenge is that unix hardlinks cannot span filesystems. So we cannot have a hardlink on the ramdisk (`/var/run`) pointing to the disk (e.g. `/usr`).

Could consider copying the whole executable instead? In the context of #1021 this could go into the `shm/group/` folder so we would only have one copy per process group. This would also make the `core.worker` process-spawning API robust to the snabb binary on disk being replaced while running.

wingo:
Sorry I have to insist about the version thing.  There may be many versions of snabb on a system and you don't want "snabb top" to start doing weird things depending on which version/branch/etc of snabb top you run.  I do not want these bugs from my users!

eugeneia:
@xray7224 Sorry for the long silence. I think this change should also update `snabb top` so that it can attach to named instances (see https://github.com/snabbco/snabb/pull/1058#issuecomment-258377887). I would like to merge it once that issue is addressed.

The remaining issues can be fixed later on:

- implement `snabb top` so that it always uses the snabb executable of the instance it attaches to
- possibly clean up directory structure

eugeneia:
Ehh... I had forgotten that this is upstreamed via igalia/snabb, nevermind my last comment.

eugeneia:
This did land upstream and can be closed, no? ❓ 🕵️‍

xray7224:
Yeah it's landed upstream, thanks for the reminder. I'll close it.

PR #1059, opened 2016-11-03
Add data type mapping between FFI and Lua values

wingo:
From the README:

  Ever been annoyed that you can't create a hash table where the keys are
  FFI values, like raw IPv4 addresses, but the values are Lua objects?
  Well of course you can key a normal Lua table by any Lua value, but the
  key is looked up by identity and not by value, which is rarely what you
  want.  `foo[lib.protocol.ipv4:pton('1.2.3.4')]` will not be the same as
  `foo[lib.protocol.ipv4:pton('1.2.3.4')]`, as the `pton` call produces a
  fresh value every time.  What you usually want with FFI-keyed tables is
  to be able to look up the entry by value, not by identity.

  Well never fear, *cltable* is here.  A cltable is a data type that
  associates FFI keys with any old Lua value.  When you look up a key in a
  cltable, the key is matched by-value.

### DIFF
--- a/src/lib/cltable.lua
+++ b/src/lib/cltable.lua
@@ -3,8 +3,6 @@ module(..., package.seeall)
 local ffi = require("ffi")
 local ctable = require("lib.ctable")
 
-CLTable = {}
-
 function build(keys, values)
    return setmetatable({ keys = keys, values = values },
                        {__index=get, __newindex=set})
--- a/src/lib/cltable.lua
+++ b/src/lib/cltable.lua
@@ -26,10 +26,7 @@ function set(cltable, key, value)
    local entry = cltable.keys:lookup_ptr(key)
    if entry then
       cltable.values[entry.value] = value
-      if value ~= nil then
-         cltable.keys:remove_ptr(entry)
-         -- FIXME: Leaking the slot in the values array.
-      end
+      if value == nil then cltable.keys:remove_ptr(entry) end
    elseif value ~= nil then
       table.insert(cltable.values, value)
       cltable.keys:add(key, #cltable.values)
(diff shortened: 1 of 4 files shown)


wingo:
seems i forgot to add https://github.com/Igalia/snabb/commit/580d8c7ee7b118c1bda7da9b3d941f27fbf9ee2b to this branch

wingo:
@eugeneia would you perhaps mind having a look at this one?  No rush though, I just wanted to get it on board one of the upstream trains :)

src/lib/cltable.lua
```diff
@@ -0,0 +1,66 @@
...
+   return cltable.values[entry.value]
+end
+
+function set(cltable, key, value)
+   local entry = cltable.keys:lookup_ptr(key)
+   if entry then
+      cltable.values[entry.value] = value
+      if value ~= nil then
+         cltable.keys:remove_ptr(entry)
+         -- FIXME: Leaking the slot in the values array.
```
eugeneia: This!

eugeneia: Wait, shouldn’t this be `value == nil`, as in “if value is nil, delete the entry and set the corresponding key in the backing table to nil, too”.

src/lib/cltable.lua
```diff
@@ -0,0 +1,66 @@
...
+function pairs(cltable)
+   local ctable_next, ctable_max, ctable_entry = cltable.keys:iterate()
+   return function()
+      ctable_entry = ctable_next(ctable_max, ctable_entry)
+      if not ctable_entry then return end
+      return ctable_entry.key, cltable.values[ctable_entry.value]
+   end
+end
+
+function selftest()
```
eugeneia: As per the above I would suggest to test deletion here.

src/lib/ctable.lua
```diff
@@ -91,8 +91,9 @@ end
 -- FIXME: For now the value_type option is required, but in the future
 -- we should allow for a nil value type to create a set instead of a
 -- map.
-local required_params = set('key_type', 'value_type', 'hash_fn')
+local required_params = set('key_type', 'value_type')
```
eugeneia: In theory you could use the new `core.lib.parse` function for this.

wingo:
Fixed nit, good catch :)

Also it turns out the slot is not leaked.  The next slot will be allocated at `#values`, which may be any index which precedes a nil and in practice seems to be the first one.  So the normal case of adding and removing shouldn't grow the values array.  Additionally if it did grow and get holey, that would be LuaJIT's problem to switch to a sparse array representation and not a packed array.

Re ctable and parameters it is a good idea for later :)

PR #1060, opened 2016-11-03
next: Changes queued for the v2016.11 release

lukego:
Let's see what Hydra has to say...


lukego:
Ready when you are @eugeneia!

PR #1061, opened 2016-11-07
Lwaftr: changes needed for Nix benchmarks

domenkozar:
I can run the benchmark without the chroot using these patches.

I've also merged `igalia/lwaftr` since some of the changes by @tekcnico are needed from there for benchmarking to work.

Note: OpenStack support is optional in order to only copy 25MB of Snabb executable to the VM instead of 500MB using OpenStack, which MySQL pulls in.

cc @teknico @eugeneia @wingo


domenkozar:
I'll cherry-pick only the changes needed from `igalia/lwaftr`, let me rebase.

domenkozar:
I'll get this upstream via `igalia/lwaftr`.

PR #1062, opened 2016-11-07
Merge `lwaftr` to next

wingo:
This PR mainly includes fixes to `snabbvmx`, getting it to the point where it works reliably under Hydra.  Thanks very much to @dpino, @teknico, and @domenkozar for this work..  It also includes #1033 from @xray7224.


wingo:
I think this PR is pretty low-risk; the only tricky change is the "named programs" one.  For me it could hit master for v2016.11, or we could wait until v2016.12, either way.  WDYT @eugeneia and @lukego ?

domenkozar:
`ERROR     testlog/program.snabbvmx.tests.selftest.sh`

lukego:
Let's wait for v2016.12. I have already advertised `next` as ready for v2016.11 release in https://github.com/snabbco/snabb/pull/1060#issuecomment-258388939 so I prefer to only merge fixes to release-blockers at the moment.

wingo:
Makes sense to me.  I didn't realize that non-core code could be risky but it turns out it can be if the test suites are flaky or don't run on the main CI servers.

wingo:
Closing this PR, as it keeps getting new commits as we merge to `lwaftr`.  I'll instead merge via `wingo-next`.

PR #1066, opened 2016-11-10
Topologically sort app graph

wingo:
This commit sorts the app graph topologically instead of running it to a
fixed point on each breath.  This eliminates quadratic breath time
complexity as a function of graph size.

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -250,6 +250,50 @@ function apply_config_actions (actions, conf)
    for _, app in ipairs(app_array) do
       if app.link then app:link() end
    end
+   compute_breathe_order ()
+end
+
+-- Sort the NODES that are reachable from ENTRIES topologically
+-- according to SUCCESSORS via reverse-post-order numbering.  This
+-- implementation is recursive; we should change it to be iterative
+-- instead.
+function tsort (nodes, entries, successors)
+   local visited = {}
+   local post_order = {}
+   local function visit(node)
+      visited[node] = true
+      for _,succ in ipairs(successors[node]) do
+         if not visited[succ] then visit(succ) end
+      end
+      table.insert(post_order, node)
+   end
+   for _,node in ipairs(entries) do
+      if not visited[node] then visit(node) end
+   end
+   local ret = {}
+   while #post_order > 0 do
+      table.insert(ret, table.remove(post_order))
+   end
+   return ret
+end
+
+breathe_pull_order = {}
+breathe_push_order = {}
+
+function compute_breathe_order ()
+   breathe_pull_order = {}
+   local inputs = {}
+   for i, app in ipairs(app_array) do
+      if app.pull then table.insert(breathe_pull_order, app) end
+      for _,link in pairs(app.input) do inputs[link] = app  end
+   end
+   local successors = {}
+   for i, app in ipairs(app_array) do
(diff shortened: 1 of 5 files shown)


lukego:
Cool idea! The engine code today is super naive and I am sure that this sort of optimization will be very beneficial for larger app networks.

I suspect that this optimization depends on some specific invariants about the app network. Is that right? If so then what are they?

I suppose that if each `push()` method is called only once per breath then this could affect the behavior of an app network that contains a cycle. Contrived example would be an app network that loops a packet through a decapsulation app as many times as it takes to remove every encapsulation:

```
source -> decap -> sink
          ^   |
          +---+
```

On the master implementation this would fully process each packet every breath while on the optimized branch it would only process one decap step per breath, right? (This is not necessarily a problem but I would like to understand.)

btw this would probably make the "cycles per app callback" density graph over at https://github.com/snabbco/snabb/pull/1030#issuecomment-256485005 look nicer. Just now you can see a bimodal distribution on the calls to the `Sink` app: sometimes it has work to do (~4000 cycles/call) and other times it has no packets (~650 cycles/call).

(NB: Those cycle counts include relatively high timeline logging overhead because this is an experimental branch that reads a bunch of PMU counters on each log message. RDPMC has reciprocal throughput 37 cycles according to Agner i.e. reading 7 PMU counters has latency 259 cycles.)

wingo:
Indeed, multi-app cycles will cause packets to stay in links after a breathe() finishes.  However as long as the network is capable of processing the input traffic rate, this will not lead to packet drops (at least as long as we don't artificially drain/drop non-empty links after each breath) as the next breath will process those packets.  I don't think there is a significant difference in this case relative to what happens now with the fixpoint loop though.

src/core/app.lua
```diff
@@ -229,27 +216,79 @@ function apply_config_actions (actions, conf)
...
+      if not visited[node] then visit(node) end
+   end
+   local ret = {}
+   while #post_order > 0 do
+      table.insert(ret, table.remove(post_order))
+   end
+   return ret
+end
+
+breathe_pull_order = {}
```
eugeneia: Naive question: why is the pull order relevant?

wingo: Well... in my patch it isn't.  This is more of a set; you want pulls to come first.  Probably the ideal is to run all pulls needed by each chain; if there are parallel disjoint graphs you should run pull-1 then all pushes in that chain, then pull-2 then all pushes in that chain, and so on.  Right now the patch does the simple thing though.  It could be that pull-1 and pull-2 are both consumed by push-N though, so it's a reasonable first step.

wingo:
this error appears to be a timeout.

kbara:
@wingo I'm happy to put this on kbara-next as soon as the tests pass.

wingo:
What is going on with the test suite?  I do not understand the NFV failure.

wingo:
Hydra for this branch: https://hydra.snabb.co/build/842422/download/2/report.html
https://hydra.snabb.co/build/843455/download/2/report.html

eugeneia:
Probably a spurious failure, davos is suffering from those... I have reset the cache for this PR so it should be reevaluated soon.

kbara:
I've opened https://github.com/snabbco/snabb/pull/1070 to upstream it

PR #1067, opened 2016-11-10
Optimize reconfiguration

wingo:
This PR refactors (re)configuration so that compute_config_actions
results in an ordered list of actions, O(n) in the size of the change
rather than of the graph, such that apply_config_actions can just mindlessly
apply them.

apply_config_actions is now a mutating procedure which mutates the app
table, the link table, and the current configuration to reflect the
result of applying the config diff in compute_config_actions.

This PR also includes #1066.

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -250,6 +250,50 @@ function apply_config_actions (actions, conf)
    for _, app in ipairs(app_array) do
       if app.link then app:link() end
    end
+   compute_breathe_order ()
+end
+
+-- Sort the NODES that are reachable from ENTRIES topologically
+-- according to SUCCESSORS via reverse-post-order numbering.  This
+-- implementation is recursive; we should change it to be iterative
+-- instead.
+function tsort (nodes, entries, successors)
+   local visited = {}
+   local post_order = {}
+   local function visit(node)
+      visited[node] = true
+      for _,succ in ipairs(successors[node]) do
+         if not visited[succ] then visit(succ) end
+      end
+      table.insert(post_order, node)
+   end
+   for _,node in ipairs(entries) do
+      if not visited[node] then visit(node) end
+   end
+   local ret = {}
+   while #post_order > 0 do
+      table.insert(ret, table.remove(post_order))
+   end
+   return ret
+end
+
+breathe_pull_order = {}
+breathe_push_order = {}
+
+function compute_breathe_order ()
+   breathe_pull_order = {}
+   local inputs = {}
+   for i, app in ipairs(app_array) do
+      if app.pull then table.insert(breathe_pull_order, app) end
+      for _,link in pairs(app.input) do inputs[link] = app  end
+   end
+   local successors = {}
+   for i, app in ipairs(app_array) do
+      local succs = {}
+      for _,link in pairs(app.output) do table.insert(succs, inputs[link]) end
+      successors[app] = succs
+   end
+   breathe_push_order = tsort(app_array, breathe_pull_order, successors)
 end
 
 -- Call this to "run snabb switch".
@@ -310,10 +354,8 @@ function breathe ()
    -- Restart: restart dead apps
    restart_dead_apps()
    -- Inhale: pull work into the app network
-   for i = 1, #app_array do
-      local app = app_array[i]
---      if app.pull then
---         zone(app.zone) app:pull() zone()
+   for i = 1, #breathe_pull_order do
+      local app = breathe_pull_order[i]
       if app.pull and not app.dead then
          zone(app.zone)
          with_restart(app, app.pull)
@@ -321,25 +363,14 @@ function breathe ()
       end
    end
    -- Exhale: push work out through the app network
-   local firstloop = true
-   repeat
-      local progress = false
-      -- For each link that has new data, run the receiving app
-      for i = 1, #link_array do
-         local link = link_array[i]
-         if firstloop or link.has_new_data then
-            link.has_new_data = false
-            local receiver = app_array[link.receiving_app]
-            if receiver.push and not receiver.dead then
-               zone(receiver.zone)
-               with_restart(receiver, receiver.push)
-               zone()
-               progress = true
-            end
-         end
+   for i = 1, #breathe_push_order do
+      local app = breathe_push_order[i]
+      if app.push and not app.dead then
+         zone(app.zone)
+         with_restart(app, app.push)
+         zone()
       end
-      firstloop = false
-   until not progress  -- Stop after no link had new data
+   end
    counter.add(breaths)
    -- Commit counters at a reasonable frequency
    if counter.read(breaths) % 100 == 0 then counter.commit() end
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -95,19 +95,19 @@ function restart_dead_apps ()
 
    for name, app in pairs(app_table) do
       if app.dead and (now() - app.dead.time) >= restart_delay then
-         local function add_action(action, ...)
-            table.insert(actions, { action, { ... } })
-         end
          io.stderr:write(("Restarting %s (died at %f: %s)\n")
                          :format(name, app.dead.time, app.dead.error))
-         add_action('stop_app', name)
-         add_action('start_app', name,
-                    configuration.apps[name].class,
-                    configuration.apps[name].arg)
+         local info = configuration.apps[name]
+         table.insert(actions, {'stop_app', {name}})
+         table.insert(actions, {'start_app', {name, info.class, info.arg}})
          for linkspec in pairs(configuration.links) do
             local fa, fl, ta, tl = config.parse_link(linkspec)
-            if fa == name then add_action('link_output', fa, fl, linkspec) end
-            if ta == name then add_action('link_input', ta, tl, linkspec) end
+            if fa == name then
+               table.insert(actions, {'link_output', {fa, fl, linkspec}})
+            end
+            if ta == name then
+               table.insert(actions, {'link_input', {ta, tl, linkspec}})
+            end
          end
       end
    end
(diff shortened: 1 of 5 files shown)


wingo:
For motivation for this PR see https://github.com/Igalia/snabb/issues/562.  Specifically I want to be able to compute a set of changes in the main process and then ship that set of changes to the worker to be applied.

src/core/app.lua
```diff
@@ -119,66 +122,134 @@ end
 -- new app network by making the changes needed.
 function configure (new_config)
    local actions = compute_config_actions(configuration, new_config)
-   apply_config_actions(actions, new_config)
-   configuration = new_config
```
wingo: I am not sure that what you are working on is a good idea.  it does not seem to be a strategy that will survive a multiprocess world.

eugeneia: You’re right, I was overly complicating things

wingo:
Seems I am neglecting to call the link methods, and that's making the bridge unhappy.

src/core/app.lua
eugeneia: @wingo Still, why is it preferable to mutate `configuration` (in many surgical steps) instead of going with the pattern proposed by the `core.config` API to just swap it out?

src/core/app.lua
```diff
@@ -119,66 +122,138 @@ end
...
+      table.insert(actions, { action, { ... } })
+   end
+
+   -- First determine the links that are going away and remove them.
+   for linkspec in pairs(old.links) do
+      if not new.links[linkspec] then
+         local fa, fl, ta, tl = config.parse_link(linkspec)
+         if old.apps[fa] then add_action('unlink_output', fa, fl) end
+         if old.apps[ta] then add_action('unlink_input', ta, tl) end
+         if not new[linkspec] then add_action('free_link', linkspec) end
```
eugeneia: this looks fishy, `new[linkspec]` makes no sense, even if its a typo for `new.links[linkspec]`.

eugeneia: Should probably be
```
if new.apps[fa] then add_action('unlink_output', fa, fl) end
if new.apps[ta] then add_action('unlink_input', ta, tl) end
add_action('free_link', linkspec)
```

wingo: indeed!

src/core/app.lua
wingo: A very good question!

I guess I would first note that what we are doing is a fundamentally place-oriented operation underneath -- we are mutating the graph of instantiated apps and the links between them, not creating an entirely new graph.  When we're done we can't use the old idea of what the apps and links are because it is no longer valid.

You are right though that don't have to mutate the description of the app graph that is passed in to `app.configure` or `app.compute_config_actions`.  Even though we are mucking with links and such, we could avoid mutating the configuration.  However consider: in a multiprocess configuration as in https://github.com/Igalia/snabb/blob/yang/src/apps/config/README.md, the follower would be left with an out-of-date `app.configuration`.  Maybe this is OK?  The follower is designed to not need to know about the current configuration; it just applies changes.  I kept in the mutation so that the follower would have an up-to-date `app.configuration` though.

src/core/app.lua
```diff
@@ -119,66 +122,138 @@ end
...
-      if app_table[name].stop then
-         app_table[name]:stop()
-      end
-      if app_table[name].shm then
-         shm.delete_frame(app_table[name].shm)
+   -- As an efficiency hack, some apps rely on the fact that we add
+   -- links both by name and by index to the "input" and "output"
+   -- objects.  Probably they should be changed to just collect their
+   -- inputs and outputs in their :link() functions.  Until then, call
+   -- this function when removing links from app input/output objects.
```
eugeneia: :+1:

src/core/app.lua
```diff
@@ -119,66 +122,138 @@ end
...
+      app.output[linkname] = link
+      table.insert(app.output, link)
+      if app.link then app:link() end
+   end
+   function ops.link_input (appname, linkname, linkspec)
+      local app = app_table[appname]
+      local link = assert(link_table[linkspec])
+      app.input[linkname] = link
+      table.insert(app.input, link)
+      if app.link then app:link() end
```
eugeneia: Technically speaking, we don’t have to call `link` that often, from the [docs](http://snabbco.github.io/#app):

> Optional. Called any time the app’s links may have been changed (including on start-up). Guaranteed to be called before pull and push are called with new links.

E.g., its OK to just call each apps `link` methods once after reconfiguration, and its not necessary that the links have actually changed (apps may not rely on that).

wingo: I get what you're saying.  OTOH consider a configuration change that results in a big action list, and the follower is going to process that action list from another process.  You start streaming the actions across to the follower but for whatever reason (perhaps it is reading the actions as you write them, perhaps it limits the rate at which it takes actions, perhaps you fill the statically allocated ring buffer) it doesn't apply them all in the same breath.  In that case you could run the next push() after changing the app's links but before calling link().

I would prefer some static flag on apps indicating that their links changed, which they could check themselves in their pull functions.  I guess that could be a followup; wdyt?

src/core/app.lua
wingo: AFAIU also we should always unlink the sides -- if we're throwing away the link, then it was indeed linked before, and surely we should also unlink both sides.

src/core/app.lua
eugeneia: No, checking for link changes in a pull function is a big caveat imho, it would make a the respective pull callbacks ugly.

Its kind of mind bending to me that `Follower` and `Leader` are supposed to be apps, instead of programs. To me it seems simpler to stream configuration changes in between calls to `enginine.main`, and only apply them once they are complete.

src/core/app.lua
wingo: The code in this PR as-is introduces no race conditions AFAIU; do you agree?

As far as whether the leader/follower setup from https://github.com/Igalia/snabb/blob/yang/src/apps/config/README.md introduces race conditions: the follower does not read from `app.configuration` at all.  The leader has an idea of the app graph of the followers because it knows the current high-level configuration of the instance as a whole, and it can get the app graph (the thing you get when you call `core.config.new()`; there are two different things here with the name "configuration" and I am calling `core.config` the app graph to differentiate) via `local app_graph = setup_fn(conf)`.  When a new configuration comes, the leader computes a new app graph, calls `compute_config_actions` to get the config actions, then ships those actions to all followers.

The followers can lag the leader of course.  If there is an operation that for some reason requires that the follower state matches the leader state, the leader can just wait until the followers have consumed all their updates.

Hope that helps.  Do you see a race condition there?  I don't but I would be delighted if you had a counterexample.

src/core/link.h
```diff
@@ -15,12 +15,5 @@ struct link {
   //   read:  the next element to be read
   //   write: the next element to be written
   int read, write;
-  // Index (into the Lua app.active_apps array) of the app that
-  // receives from this link.
-  int receiving_app;
-  // True when there are new packets to process.
-  // Set when a new packet is added to the ring and cleared after
-  // 'receiving_app' runs.
-  bool has_new_data;
```
eugeneia: Then good riddance! :-)

src/core/app.lua
wingo: I think it's reasonable to check for link changes in a pull function, given that very few apps have link functions.  It makes things very obvious and matches the not-so-recent shift that we had from apps spawning timers to apps running timeouts manually in pull functions.  Your mileage varies evidently :-)  I am fine with whatever as my apps don't call link functions, and the quadratic cost only comes in if the link function exists and is linear in link count.

wingo:
Appears to be a significant regression: https://hydra.snabb.co/build/860978/download/2/report.html  I wonder how???

eugeneia:
@wingo See also `./snabb snabbmark nfvconfig` for a benchmarking configuration transitions:

```
  snabbmark nfvconfig  <config-file-x> <config-file-y> <n>
    Benchmark loading <config-file-x> and transitioning from <config-file-x>
    to <config-file-y> <n> times.

    Uses SNABB_PCI0.
```

wingo:
@eugeneia is that the regression?  Certainly snabbbot appears to think there is a 25% regression on basic1, which I guess is due to reconfiguration (and reconfiguration only afaiu).  My aim here is to allow us to separate compute_configure_actions and apply_configure_actions between processes; certainly we should not regress on that basic benchmark tho.

I don't understand how this could affect data plane performance tho.  What is the benchmark in the hydra results whose perf has gone down so much?  Is it reconfiguration as well?

wingo:
(incidentally snabb-bot did a great job in catching this perf problem and NACKing the PR :)

wingo:
optimize-murren on b551d27: https://hydra.snabb.co/build/866422/download/2/report.html

optimize-murren on c708b62: https://hydra.snabb.co/build/867116/download/2/report.html

benchmarks-murren-small on b551d27: https://hydra.snabb.co/build/864965/download/2/report.html

benchmarks-murren-small on c708b62: https://hydra.snabb.co/build/868183/download/2/report.html

wingo:
I don't understand the snabbnfv-iperf-1500-tunnel+crypto regression from snabbbot.  How would I test that?

eugeneia:
@wingo the reference I posted is not related to the regression, just an existing micro-benchmark for configuration transitions that I thought you might find useful for this PR.

You can run the snabbnfv-iperf-1500-tunnel+crypto benchmark on lugano-1/3 using:

```
SNABB_IPERF_BENCH_CONF=program/snabbnfv/test_fixtures/nfvconfig/test_functions/crypto-tunnel.ports \
sudo -E lock program/snabbnfv/selftest.sh bench
```

wingo:
tx for the tip!  I need to get better at being about being able to run NFV tests.

PR #1070, opened 2016-11-14
For next: Andy's topological app graph sort

kbara:


src/core/app.lua
```diff
@@ -229,27 +216,79 @@ function apply_config_actions (actions, conf)
...
+   for name, app in pairs(app_table) do
       if app.link then app:link() end
    end
+   compute_breathe_order ()
+end
+
+-- Sort the NODES topologically according to SUCCESSORS via
+-- reverse-post-order numbering.  The sort starts with ENTRIES.  This
+-- implementation is recursive; we should change it to be iterative
+-- instead.
```
lukego: This comment is a little cryptic for me:

> This implementation is recursive; we should change it to be iterative instead.

I don't immediately see a problem with recursion. Could we please extend this comment to explain what the problem with recursion is, or remove the comment if recursion is actually okay, or actually rewrite this iteratively in case it is really important?

src/core/app.lua
```diff
@@ -229,27 +216,79 @@ function apply_config_actions (actions, conf)
...
+   while #post_order > 0 do
+      table.insert(ret, table.remove(post_order))
+   end
+   return ret
+end
+
+breathe_pull_order = {}
+breathe_push_order = {}
+
+function compute_breathe_order ()
```
lukego: Question: is `compute_breath_order()` robust to cycles in an app network? (Could a cyclic link between apps cause an infinite recursion in the topological sort routine?)

wingo: Yes, it is robust.  Apps are added to the visited set before they are visited, and not revisited if they are in the visited set.  On a higher level, this is an implementation of a reverse post-order numbering via depth-first-walk, an algorithm that can be applied to graphs with cycles.

src/core/app.lua
wingo: I think a rewrite is not "really important"; any failure will not be silent, and it is unlikely to fail with normal (up to 100 apps) graphs.  I haven't tested it though.  I would be fine keeping both the implementation and the comment fwiw.

lukego: I am slightly concerned that a 100-app limit could make the NFV application explode with larger configurations. Our OpenStack test suite does have a "many VMs" test cases but that exercises less than the maximum number of apps. I suppose that ideally I would either beef that test up (we have throttled it due to resource constraints in the nested-virtualized test env) or write a new test case.

Generally question on my mind is: How much stack space is okay with LuaJIT? and how much is this algorithm going to need? You seem to be plugging in values from gut feeling here? I feel like some languages blow up with around 100 stack frames while others handle millions gracefully.

I did a quick test with a recursive factorial function and it seems okay up to around 30,000 stack frames:

```
$ ./snabb snsh -e 'function factorial (n) return (n==0) and 1 or n*factorial(n-1) end  print(factorial(30000))'
inf
```

and then exhausts the stack with 35,000:

```
$ ./snabb snsh -e 'function factorial (n) return (n==0) and 1 or n*factorial(n-1) end  print(factorial(35000))'
[string "function factorial (n) return (n==0) and 1 or..."]:1: stack overflow
```
If this topological sort is using O(n) stack space then we should be fine to well beyond 100 apps, right?

lukego:
Thanks for the explanation @wingo. I will merge onto `next`. If somebody contributes a test case showing problems with the recursive implementation before the next release then we may have to consider a revert but this seems unlikely based on quick happy testing.

src/core/app.lua
wingo: The stack space usage is O(n), yes.  We should be fine for more than 100 apps; I was just throwing out a number :)

PR #1071, opened 2016-11-16
Add Snabb logo to README

lukego:
Show the Snabb logo on the Github project page, which is also the http://snabb.co/ website.

PR #1072, opened 2016-11-16
Add Snabb logo to README

kbara:
Upstreaming of #1071


lukego:
Thanks Kat for humoring me and my efforts to keep the workflow decentralized even when it is admittedly a bit comical :).

kbara:
Gladly; if we don't do it for the trivial stuff, we'll start slipping on doing it where it matters. :)

PR #1073, opened 2016-11-17
Configure callback for “Macro apps”

eugeneia:
This new `configure` callback (as used in #1068) allows the creation of transparent “macro apps”, which are not instantiated, but instead modify the configuration they are configured on:

— Method **myapp:configure** *configuration*, *name*, *arg*

*Optional*. If this method is defined the `new` method is ignored and not
required, and when *myapp* is configured using `config.app` this method is
called instead of instantiating *myapp* using `new`. The `configure` method is
called with the *configuration*, *name*, and *arg* given to `config.app`, and
is intended to be used to add arbitrary apps and links to *configuration* using
`config.app` and `config.link`.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -85,10 +85,20 @@ unimplemented are marked as "optional".
 
 — Method **myapp:new** *arg*
 
-*Required*. Create an instance of the app with a given argument *arg*.
-`Myapp:new` must return an instance of the app. The handling of *arg* is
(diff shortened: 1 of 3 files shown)


wingo:
Excellent idea!  The docs should probably note that this is a class method, not a method on an instance, and that it's meant to add to *configuration* and not have other side effects.

PR #1075, opened 2016-11-18
Lwaftr v3.0.0

wingo:
This PR updates the `lwaftr` head in upstream Snabb for the latest downstream release, version 3.0.0.  Change log here: https://github.com/Igalia/snabb/blob/lwaftr/src/program/lwaftr/doc/CHANGELOG.md#300---2016-11-18


wingo:
TBH I am not sure what the upstream `lwaftr` head is for but as long as it's there, it seems like using it for released versions of the lwAFTR seems like a fine thing.  Feedback is welcome of course @kbara @lukego @eugeneia :)

PR #1076, opened 2016-11-24
Replace lib.timer() with lib.timeout() and lib.throttle()

lukego:
This branch contains several commits in a series.

The net effect is to replace the library function `lib.timer(duration, mode, timefun)` with a pair of simpler counterparts:

- `lib.timeout(seconds)` returns a function that starts returning true after *seconds* have elapsed.
- `lib.throttle(seconds)` returns a function that returns true but never twice in a *seconds* interval.

The idea is that two simple functions are better than one complex one. *Seconds* is always floating-point so you can write `0.25` for 250ms or `1e-9` for 1ns.

This started with commit a233c98 preventing `lib.timer()` from "saving up" timeouts like a token bucket but then it seemed worth taking a step further.

The case against keeping `lib.timer()`:
- Complex interface that requires many words to describe.
- Punts a hard problem to the user: what is the most appropriate time source?
- Duration units depend on the time source chosen (we are using both seconds and nanoseconds).
- Only called twice in the whole Snabb master code tree anyway.

The new timers are always using `engine.now()` as the time source but this function has been updated to return a fresh time value when called from outside the engine loop (instead of a potentially stale one.) One consequence is that the timers will not update during a breath but only between breaths or outside breaths.

Separately it could be worth seeing what Hydra says about dropping this whole time-caching business and always calling out for a fresh time value instead...

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -827,6 +827,15 @@ result of calling *timefun*, and is saved in the resulting closure. A
 *duration* has elapsed when its deadline is less than or equal the value
 obtained using *timefun* when calling the closure.
 
+— Function **lib.throttle** *seconds*
+
+Return a closure that returns `true` at most once during any *seconds*
+(a floating point value) time interval, otherwise false.
+
+— Function **lib.timeout** *seconds*
+
+Returns a closure that returns `true` if *seconds* (a floating point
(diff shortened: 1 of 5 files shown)


eugeneia:
LGTM!

PR #1077, opened 2016-11-26
core.main: Make supervisor child process instead of parent

lukego:
Reverse the "fork() into worker and supervisor" logic so that the parent process is the worker and the child is the supervisor.

Certain things become easier when the initially spawned process is the one that actually runs Snabb:

- The core.worker module can setup /var/run/snabb/$pid before exec()ing Snabb. It is important that this sets up the Snabb process rather than its supervisor.

- Running `strace snabb ...` will trace the process running Snabb rather than its supervisor.

- Simpler cleanup code because we don't have to handle the complicated case of interpreting SIGCHLD.

I will merge this straight into the multiprocess branch (#1021) because it fixes a major problem where the shared shm `group` folder is accidentally setup for the supervisor rather than the Snabb process. cc @kbara

### DIFF
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -179,6 +179,7 @@ if assert(S.fork()) ~= 0 then
 else
    -- child process: supervise parent & perform cleanup
    -- Subscribe to SIGHUP on parent death
+   S.prctl("set_name", "[snabb sup]")
    S.prctl("set_pdeathsig", "hup")
    -- Trap relevant signals to a file descriptor
    local exit_signals = "hup, int, quit, term"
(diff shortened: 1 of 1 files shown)


eugeneia:
I hope its OK I label this “merged” since its already merged in the respective upstream path.

wingo:
Good evening!  I don't think this will make the next release via the lwaftr, as that merge will probably take a little while (January release, but most Igalia folks are on holiday, so I could see that slipping too perhaps).  Anyway if you want it in master faster, we can work in parallel.

eugeneia:
Oops, I think I had this mixed up in my mind with the fd double close fix. I see no reason not to merge this into max-next, so that’s what I will do.

PR #1079, opened 2016-11-27
ipsec: Fix AUTH_SIZE

vanfstd:
We want AUTH_SIZE to be 12 since this is the size of what ends up as the integrity check value.  We can henceforth talk to Linux's IPsec stack (testcases follows in a separate PR)


eugeneia:
Good catch!

PR #1080, opened 2016-11-28
Another attempt at lwaftr tip + multiprocess orchestration.

kbara:
This cleanly merges in upstream's 1077 (supervisor as child) and updated upstream 1021 (multiprocess orchestration, with the double close fix). It also contains the channel lib.mkdir patch and @xray7224 's core.app fix. It is base on the current lwaftr, so that it can be reasonably tested by snabb bot.


kbara:
Sorry, wrong root tree; this was meant to target Igalia/lwaftr.

PR #1087, opened 2016-12-09
Snabb config testing with lwaftr

xray7224:
This introduces a bunch of front-end tests which run the `snabb config` commands we have at the moment (get, get-state, set, add and remove) and do its best to verify the data. The tests a a bunch of shell scripts with a helper file which contains code to do assets, start the lwaftr and ensure it closes too.

This runs as part of the `make test` testsuite in snabb. This should hopefully help us catch more bugs.

PTAL @wingo @kbara


xray7224:
Oops, posted to the wrong fork, this is what one gets for doing PRs when tired.

PR #1089, opened 2016-12-16
next: Changes queue for the v2016.12 release

lukego:
Here we go!

Coming in to 2017 we will have a few different models of device driver:

- `intel10g` that supports switching (but not rss or multiprocess operation.)
- `intel_rss` that supports RSS and fully independent multiprocess operation (but not switching.)
- `connectx4` (not submitted to upstream yet) that supports RSS and switching and multiprocess operation (but requires tighter coupling between the processes i.e. common shared memory objects via `core.worker` and a state machine protocol to mediate access to NIC resources.)

It would be lovely to reconcile these in a user-friendly way. but first: Christmas :snowman_with_snow: :-)

Hydra tests pending...


lukego:
Sorry all for being so slow on this and holding up the release train. @eugeneia has pointed out a performance regression in the Hydra tests and I am taking the time to develop new "timeline" tooling to make these easier to diagnose. On the one hand it is great to be able to bisect performance problems using Hydra benchmark campaigns, on the other hand I would like to extract much more information from each test run.

Please bear with me this month while I push forward on the new tooling. I think it will be much more effective when we can see exactly _where_ a performance difference comes from: which part of the engine, or which app, or which difference in workload, etc. Here is an early screenshot to show the kind of information I have in mind for being able to compare between multiple Snabb versions:

timeline-profile-cycles-per-packet

kbara:
That sounds amazing.

lukego:
@eugeneia This should be good to go now! @wingo fixed the performance regression with #1102.

lukego:
So! The timeline code didn't land in this release. Sorry about that @kbara. It's actually working well now but it's also evolving quite quickly - particularly the content of the messages that the engine logs - now that I am really using it. I'd like to stabilise my initial usage a bit before pushing it upstream. Have to make it easily accessible too: make the R tools easy to install and/or make Hydra run them automatically and publish on the web. (I'm also going to build a GUI for this but that's a medium-term thing.)

I also wrote a new sampling profiler for LuaJIT ("vmprofile") on the weekend and I'm finding it indispensable already. The idea is that it very efficiently profiles the Snabb process at all times and stores the samples in the `shm` folder where they can be accessed both while running and after termination (if preserved.) So profiling would be "always on." I have to make a PR out of this and send it up... meanwhile the basic mechanism is described at https://github.com/LuaJIT/LuaJIT/pull/290.

eugeneia:
@lukego SnabbBot reports a test failure: https://gist.github.com/SnabbBot/243a52b310a2545d2f65d1616df36c49#file-log-L33

lukego:
@eugeneia Interesting. The selftest for `core.app` is failing for the new breath-sorting code. It works for me when testing manually with `snabb snsh -t core.app`. The backtrace says that it is this assertion at [core/app.lua:497](https://github.com/snabbco/snabb/blob/next/src/core/app.lua#L497) failing for SnabbBot:

```lua
   assert(#breathe_push_order == 2)
```

I don't immediately understand why this assertion would work for me on the `next` branch but not for SnabbBot. I see that SnabbBot also saw the same problem with #1102. @wingo any idea?

lukego:
@eugeneia I was wrong about the problem only occurring on davos. Good. I made a fix in #1108 and I will merge that here if it passes the sniff test from @wingo.

lukego:
@eugeneia I merged the selftest fix. Hopefully good to go!

PR #1094, opened 2017-01-22
Add a CMake building system Skeleton

nzinfo:
This PR does not intend merge into master, but a demo of add CMake build system to snabb.
Although the current Makefile based build system works fine. But It might be have trouble when porting snabb to other platform. 
CMake might be better when porting snabb as a framework to other OS, eg. Windows.

@lukego @eugeneia


lukego:
Cool! I created a new label called "portability" to tag changes like this that will benefit future porting efforts.

PR #1097, opened 2017-02-08
compile_binding_table fails to compile

amanand:
compile_binding_table.lua:30: variable 'mtime_sec' is not declared
Issue is happening because recently in the line 24 input_stream was added in place of stream but this was missed out from line 30. Made this change and tested on my setup.

### DIFF
--- a/src/program/lwaftr/compile_binding_table/compile_binding_table.lua
+++ b/src/program/lwaftr/compile_binding_table/compile_binding_table.lua
@@ -27,6 +27,6 @@ function run(args)
       io.stderr:write(tostring(bt_or_err)..'\n')
       main.exit(1)
    end
-   bt_or_err:save(out_file, stream.mtime_sec, stream.mtime_nsec)
+   bt_or_err:save(out_file, input_stream.mtime_sec, input_stream.mtime_nsec)
    main.exit(0)
 end


dpino:
Thanks for the patch. Indeed, this was a regression introduced in fe286538e2f272a36aefe6ca1d7cc76af5d2d656. That said, I don't know whether is worth to merge this patch as the `compile_binding_table` command has been removed in the lwaftr development branch. Please, check out https://github.com/igalia/snabb for an up-to-date version of the lwaftr program.

amanand:
Thanks Diego. Can you please let me know the format of the new binding file and cfg/conf files. I tried to take a look at the README.md but the lwaftr link (https://github.com/Igalia/snabb/blob/lwaftr/src/program/lwaftr) takes me to the old format.

dpino:
@amanand The new lwAFTR configuration format is based in a YANG schema. The lwAFTR has a custom YANG schema called snabb-softwire-v1.yang (https://github.com/Igalia/snabb/blob/lwaftr/src/lib/yang/snabb-softwire-v1.yang). It also supports the latest version of the ietf-softwire schema (https://github.com/Igalia/snabb/blob/lwaftr/src/lib/yang/ietf-softwire.yang) (https://datatracker.ietf.org/doc/draft-sun-softwire-yang). Support of the ietf-sofwire schema is done through translation to snabb-softwire.yang (internally the lwAFTR works with this schema).

There's a tool to migrate the old configuration file plus binding table to the new format. The command is called "migrate-binding-table".

```
$ sudo ./snabb lwaftr migrate-configuration
Usage: migrate-configuration LWAFTR.CONF

Options:
  -h, --help                 Print usage information.
  -f, --from=VERSION         Specify version from which to migrate.

Migrate an old-style configuration and binding table to the new YANG
configuration.  LWAFTR.CONF should be the name of an old lwAFTR
configuration.  Available VERSION values are:

  legacy
    Configuration from pre-v3.0.0 lwAFTR.
  3.0.1
    lwAFTR versions where "container" nodes in schemas are missing
    corresponding nodes in the data unless "presence true" is
    specified.
  3.0.1.1
    lwAFTR development snapshot where "br" fields of softwires were
    0-based instead of 1-based.

The default version is "legacy".

The resulting up-to-date configuration will be printed on standard
output, ready to be saved to a new file.
```

Since you're coming from v2, your VERSION is legacy. The migrate configuration tool will translation your configuration to the most recent version. By default, output is printed to stdout, so use a pipe to redirect it to a file. Example

```bash
$ sudo ./snabb migrate-configuration lwaftr.conf > lwaftr-migrated.conf
```

Let me know how it goes.

wingo:
Closing this as the new lwaftr configuration format has landed in `master` with the last release.  Still, an embarassing bug, our apologies @amanand !

PR #1102, opened 2017-02-20
Sort links instead of apps for breathe order.

wingo:
This results in the invariant that each app's push function will be run after packets were pushed on its incoming links.  It might run before as well.


wingo:
benchmarks-murren-small: https://hydra.snabb.co/eval/4974 / https://hydra.snabb.co/build/987961

optimize-murren: https://hydra.snabb.co/eval/4973 / https://hydra.snabb.co/build/986242/download/2/report.html

PR #1090, opened 2017-01-04
Update pflua

kbara:
Upstreaming of #1088 , Asumu's recent improvements to pflua.

PR #1109, opened 2017-03-02
[sketch] upstreaming Snabbwall

takikawa:
This PR is an initial sketch for upstreaming Snabbwall, a layer 7 firewall app suite for Snabb that we've been developing at Igalia (@aperezdc, @dpino, me) funded by NLNet.

This follows up on some discussion on Slack about how we should go about upstreaming this.

As you can see from the diff, most of the changes are restricted to `src/apps/wall/` and `src/program/wall`. The main exceptions are the first commit, which makes a relatively small change to pflua to let `pfmatch` functions take extra arguments (CC @wingo, @kbara in case you have thoughts on this change---also if it should be separately committed to Igalia/pflua) and the addition of `lib/ljndpi`.

The latter `ljndpi` change depends on dynamically loading the nDPI library. The lack of nDPI at runtime shouldn't cause an error normally except when `snabb wall <subcommand>` is executed, but maybe it's a concern for running the tests.

I'd appreciate any thoughts on how appropriate it would be to merge, or if any big changes are needed to merge it (e.g., how to deal with the nDPI dependency).

Note: it would be best to merge this after https://github.com/snabbco/snabb/pull/1090 (pflua update) since some of the pflua changes are duplicated.


lukego:
Great stuff :). Thoughts - mostly using this PR as a vehicle for revisit the way we deal with dependencies in Snabb in case we can improve and/or clarify our strategy.

Curious: Did you consider the trade-offs between making nDPI a runtime dependency (shared library installed by the user) vs a build-time dependency (static library linked into `snabb` binary)? If so then how did you see the pros/cons and what lead you to choose the former? (Just curious because this will be a recurring theme for other people building Snabb applications that have external dependencies who may be able to benefit from your experience.)

I would prefer to upstream the real development branch rather than a squashed version. If every branch has the same view of the commit history then Git will be a powerful tool for keeping them synchronized even in exotic scenarios e.g. if we want to make a sweeping refactoring based on `master` that touches snabbwall and other applications. Having a squashed history would make the master branch a second-class citizen: if we touched any of the code that we don't have the full history for then we would end up with hard-to-resolve conflicts.

Relatedly: we have been squashing history of projects that we embed with `git subtree` e.g. LuaJIT and ljsyscall. This may be limiting in practice. I would like to be able to pull from multiple branches of the subtrees and end up with the union of all the commits. For example on LuaJIT we may pull `v2.1` for general bug fixes but also pull several other branches for specific features. Ordinarily `git merge` is smart about combining changes from multiple branches but with `git subtree pull --squash` it seems to me like the branches end up overwriting each other i.e. pulling from a new branch will revert code that was previously pulled from other branches. Could be that preserving the history is the solution. (This is bothering me at the moment when I want to pull an experimental change to LuaJIT for testing on Hydra but `git subtree` expects me to rebase it onto _exactly_ the same LuaJIT parent commit that Snabb most recently merged.)

So: Snabbwall looks good to me and I'd just like to take this opportunity to revisit how we handle dependencies and thrash out the best workflow. Sorry about that :).

lukego:
P.S. Cool that you already pushed nDPI support into nixpkgs :-). This will take care of the dependency from a CI perspective.

lukego:
One more topic: How should the snabbwall branch flow upstream? Should you PR it directly to `next` or via `wingo-next`, `kbara-next`, or `max-next`? These paths all lead fairly swiftly to `master` so it's mostly a social matter of who will do integration work like resolving conflicts from other branches and checking CI performance results.

takikawa:
Thanks for the comments! :)

With regard to dependencies, there are two reasons why we went with a runtime dependency. The first is that the [license](https://github.com/ntop/nDPI/blob/dev/COPYING) (LGPL) of nDPI is incompatible with statically including in Snabb as far as I know, but should be ok for dynamic linking.

The second is that Snabbwall's scanner backend is intended to be interchangeable (the backend just needs to implement the generic `Scanner` class), and so it may not make sense to include nDPI if the user wanted only a different backend. (in practice the nDPI backend is the only one available currently)

Re: squashing vs. whole history, I'll revise this PR to include the whole history instead of just the squashed commit. I didn't have a strong preference either way, so I'm happy to make the change.

For the upstream branch, maybe the coordination would be easier with another Igalian, but I'm happy to fit to whatever the usual workflow is. (@wingo or @kbara do you have any thoughts on that?)

aperezdc:
If it is not a big lot of work (and hopefully it shouldn't!), I also think it would indeed nicer to have the full history.

Related to @lukego's comment on `git subtree`: AFAIU it should be possible to keep the whole history of a subtree by using `git read-tree` when importing for the first time, then using `git pull -s subtree` to pick new commits ([more details](https://help.github.com/articles/about-git-subtree-merges/)). I was thinking of doing that when [https://github.com/aperezdc/ljndpi](aperezdc/ljndpi) was imported into the Snabb repository, but then I didn't because the other two modules under `lib/` have their history squashed, so I did the squashing for consistency.

lukego:
Groovy.

@aperezdc Happily I think we will be able to use `git subtree pull` without `--squash` to start including the history of our subtrees going forward, even though the older commits have been squashed. I am meaning to pull the latest LuaJIT and ljsyscall so I will try doing it this way and see how it looks.

takikawa:
Hi @lukego. Thanks, that plan sounds good to me! For the long-lived branch, my current plan is to host the branch on `Igalia/snabb`. I'll put together the pull request for `src/doc/branches.md` soon, and it looks like the next-hop won't be a problem either.

So I'll go ahead and close this sketch PR and proceed in separate PRs. Cheers.

PR #1110, opened 2017-03-03
Initial support for aarch64

jialiu02:
Hi,
It's the initial support for aarch64 platform.
Please note that the latest 2.1.0 luajit is needed to run on ARM.

Thanks!

### DIFF
--- a/src/doc/branches.md
+++ b/src/doc/branches.md
@@ -146,3 +146,10 @@ The current state of each branch with respect to master is visible here:
 
     Maintainer: Domen Kožar <domen@dev.si>
 
+#### aarch64
+
+    BRANCH: aarch64 git://github.com/jialiu02/snabb
+    Development branch for ARM aarch64 platform.
+
+    Maintainer: Jianbo Liu <jianbo.liu@linaro.org>
+


lukego:
Cool!

From past discussions the path upstream for new ports is first to maintain them on a branch and then to bring them upstream when they clearly represent a "net win" for the community. So if you are interested in driving an ARM port forward the first step is to create a branch that is "open for business" i.e. where you are accepting pull requests and merging new releases from upstream etc. If you (or anybody else) would like to do that the basic method is to update `doc/branches.md` to list your branch there & send a PR of that update to advertise it.

I think that ports have wonderful potential to bring in more developers, let us deploy applications on more platforms, encourage us to take a more disciplined approach to hardware dependencies, and so on.

I am curious about a bunch of aspects of an ARM port... do we need new NIC drivers to make this practical? (Is ARM64 usually an SoC? what ethernet controllers do they use?) How do we test the port on CI? (Do we run the tests in QEMU? Do we acquire ARM servers? Do we use an ARM public cloud?) and so on.

I also see costs for ports and the benefits need to outweigh these before we can merge a port onto the master branch. New architectures can introduce complexity like having to write code compatible with relaxed memory models, optimizing for divergent microarchitectures (e.g. both in-order and out-of-order execution), can be a disincentive to using special features of x86 like the TSC clock and performance counters, can be a disincentive to writing assembler code which some of us quite enjoy :-), and can be a problem for users if in practice it is hard to predict which features work on which platforms, and so on. Quite a basket! My feeling is that a well-maintained port for a practical platform can provide more than enough benefit to outweigh these costs, but it's definitely a high bar to clear before IMHO it makes sense to merge onto the master branch for all Snabb users.

End braindump! Great hacking :-) I know that a bunch of other people are interested in ARM64 too.

lukego:
(Would also be interesting to know why you are interested in having an ARM64 port of Snabb :-))

jialiu02:
I'd like to maintain a branch for aarch64 before it's merged into master.
Now you are pulling LuaJIT v2.1,  I will send a request to create aarch64 branch after that.

I know your concern about this porting. But we need to start the work, and let people join us.
The differences between x86 and arm are not hard to solve. CI and testing hardware are also not an issue if more and more people are interested in snabb switch in ARM.

Sorry for late reply because I'm busy with other things.

lukego:
@jialiu02 Sounds great!

Please pick where you want to maintain this branch - you can host it on this repository, or on your own fork, or on a new Github organization, etc, as you prefer - and send an update to `branches.md` describing the branch and letting everybody know that it is open for business.

(I'd suggest hosting the branch on this repo because then the community will have visibility of the PRs. However you may prefer another location if you want to build a sub-community, want to have direct control of the Github hooks, etc. It's up to you and you can change your mind later. See also related branch creation advice at https://github.com/snabbco/snabb/pull/1109#issuecomment-286693846.)

Aside: I am personally curious about what is driving the interest in ARM64. Is it users who want to deploy on ARM64? Or hardware vendors who want to bootstrap demand for ARM64 products? Or large customers who want to leverage ARM64 as a BATNA in price negotiations with Intel? You don't have to answer this, but further down the line I think the master branch will put the most weight on the needs of the first group i.e. people who are looking for the most practical deployment platform for Snabb applications. There are plenty of other projects like DPDK for vendors to use as battle grounds :).

Onward!

kbara:
I'd love to see an aarch64 port of Snabb, @jialiu02  . This PR looks like it's just getting started. Feel free to grab code/ideas from https://github.com/Igalia/snabb/tree/lwaftr_aarch64 , which is exploratory abandonware but got a bit farther along. I was doing the port on a Pine64 machine (which wouldn't be a useful production target, but is more than sufficient for getting the layers above the network cards working, in my opinion).

I share @lukego 's concerns about memory models - as http://preshing.com/20120930/weak-vs-strong-memory-models/ says, arm has a weaker memory model than x86, which I expect will lead to some interesting fun in the driver ports. The amount of dynasm in Snabb is also a minor hurdle for porting.

I reckon the same network cards would make sense - ie, there are linux kernel drivers for the Intel 82599 on aarch64 (and strictness of ordering is an issue - see https://www.spinics.net/lists/arm-kernel/msg568505.html for instance). Qemu's a painfully slow way to do aarch64 work. If this were to become a first-class port, I think the way to go would be a dedicated aarch64 machine with physical network cards in it, whether it's with other Snabb servers or in someone's cloud.

jialiu02:
@lukego I agree to host aarch64 branch here. And it's better to create a new branch after you finish merging luajit v2.1 because we need some new fixes in it for aarch64.
I understand your concern and we will be careful and not break the applications running on x86.

Thanks @kbara. It's glad to see ARM player here. I'll take a look at your repos, and you can also pull your changes to the new aarch64 after it's created. Yes, memory ordering is difficult in writing drivers,  the other thing is luajit as it supports amd64 not for a long time and may still have bugs. 
And don't worry about the devices, I have several different kinds of arm servers which can be tested on. :)

lukego:
> I'd love to see an aarch64 port of Snabb?

@kbara Just curious: why? I can see the costs of supporting multiple architectures but what are the benefits?

lukego:
> And it's better to create a new branch after you finish merging luajit v2.1 because we need some new fixes in it for aarch64.

Note: On the one hand you can wait for the luajit 2.1 update to land on `master` (couple of weeks or so) but on the other hand you could `git merge lukego/luajit-upgrade-history` to pull those changes in right now without having to wait.

Overall the intention is for `master` to be the "trailing edge" that keeps all other branches eventually consistent. It's not supposed to be on the critical path for new feature development - at least for features that are urgent.

kbara:
@lukego an unreasonable fondness of the platform, and an optimism about its future. No concrete short-term commercial reasons.

I'd be in favour of merging a solid, complete port, but not WIP ones; the costs you mention are very real.

lukego:
@kbara Cool. Yes, I would love to have a solid arm64 port living on a branch, and to merge that when there are some ARM64 platforms that kick Intel's butt for some interesting applications. Meanwhile x86-64 is improving all the time and so the bar is rising too :-)

(I have a weakness for RISC-V and lowRISC but that's too far down my priority stack to play with...)

PR #1115, opened 2017-03-17
Mailing list is abandoned

akostrikov:
Hello! The mailing list is abandoned.
We should remove it, maybe with mention of slack channel


lukego:
Good catch! I merged this onto next and added a note about the Slack chat.

PR #1120, opened 2017-04-04
next: Changes queued for the 2017.04 release

lukego:
Currently queued changes for 2017.04. Let's see what Hydra says.

Note: Many commits both due to importing LuaJIT history and also merging the Snabbwall branch.


eugeneia:
Exciting release! :-)

lukego:
@eugeneia I reckon this release is good to go although there are some interesting tidbits in the test results to acknowledge:

- Hydra shows [performance is the same as before](https://hydra.snabb.co/build/1217843/download/2/report.html#overall-summary) (good.)
- Hydra shows [snabbnfv breakage with QEMU 2.7.1 and 2.8.0](https://hydra.snabb.co/build/1217843/download/2/report.html#success-and-failure). This is a genuine problem that appears now because we did not have test coverage before (not due to changes in this Snabb release as we see from same errors with both master and next). I will create a separate issue to track this.
- SnabbBot is failing the build due to the Snabbwall `selftest.sh` not passing. This seems to be a software dependency issue and the solution I see is that we need to come up with a convention for adding nix expressions that specify the required dependencies for test cases. (Could also consider running those tests under Hydra instead if that makes life easier. Currently our Hydra jobsets are a bit spread out but we could unify them.)

So overall it seems like this branch is :+1: with new features and no genuine regressions. However, it seems like we should take some kind of short-term action to make peace between SnabbBot and the Snabbwall tests, otherwise SnabbBot will be failing every PR that comes in the next release cycle. Guessing it is okay to ship this branch and then correct that in SnabbBot either by providing the necessary dependencies or disabling that specific test case. What do you think @eugeneia?

eugeneia:
Ideally I would add the missing dependency to snabb_bot.nix before doing the release, but I don’t immediately see whats missing, so I might postpone the fix until after the release to forgo further stalling from my side?

lukego:
@eugeneia That sounds good to me. If we can make that fix without needing to hack the master branch, i.e. by editing out-of-tree code in snabb_box.nix, then it seems safe to release now to me.

eugeneia:
Heh, actually I can’t push the release onto master because SnabbBot did not sign off:

```
% git push snabbco master
Counting objects: 1, done.
Writing objects: 100% (1/1), 883 bytes | 0 bytes/s, done.
Total 1 (delta 0), reused 0 (delta 0)
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: Required status check "davos-eugeneia/snabb-nfv-test-vanilla" is failing.        
To git@github.com:snabbco/snabb.git
 ! [remote rejected] master -> master (protected branch hook declined)
error: failed to push some refs to 'git@github.com:snabbco/snabb.git'
```

lukego:
@eugeneia The machines have finally risen up and asserted their dominance :-).

I temporarily disabled that requirement on the master branch. You should be able to push now.

PR #1122, opened 2017-04-07
ljsyscall refresh: Import full history + latest changes

lukego:
This branch updates `lib/ljsyscall` with the same approach as the LuaJIT update from #1111 i.e. import all the latest code and also the complete git history.

Specific steps:

- First import the git history for the version of ljsyscall that we are already using (i.e. re-merge the same version but without squashing.) Resolve all conflicts in favour of our existing code so that only the history is updated and no the code.
- Second pull all new commits from ljsyscall upstream that we had not already merged.

The motivation for merging the complete history is so that git can properly support merges in the future. It also means that `git bisect` will be able to descend into ljsyscall changes when looking for obscure bugs.

One consequence is that this adds 5,187 commits to the Snabb git history. However, I feel that this is consistent with embracing git in the same way that Linux does, i.e. as a graph database for reconciling the _content_ of many changes by different people on different branches at different times. Having all the commits means that Git is able to merge branches intelligently with reference to the complete graph database of history.


lukego:
One change in our ljsyscall tree which is not upstream is f890ca0d1e183da51637066d910fb998829a042d which fixes a file descriptor double-close bug. However, I believe that an equivalent fix went upstream in justincormack/ljsyscall#188 by @aperezdc  and that this whole class of double-close bug was resolved with justincormack/ljsyscall#205. Just to be safe I cc @justincormack.

lukego:
(Aside: If there is a way for `git subtree` to squash the early history of a project and then continue with full history from a certain point onwards then that would be very interesting. I have experimented and only managed to import the complete history or a series of squashes.)

lukego:
@eugeneia Could the issue here be that SnabbBot is trying to fetch submodules, and we have referenced some via packages that we have imported? (Can you remind me where the SnabbBot code lives to check such details?)

lukego:
I pulled `master` partly to make SnabbBot repeat the test and see if the obscure Git error is the same.

eugeneia:
Regarding https://github.com/snabbco/snabb/pull/1122#issuecomment-293210223 (sorry, saw this just now), the SnabbBot module on davos is [this one](https://github.com/eugeneia/snabblab-nixos/blob/snabb_bot-fixes/modules/snabb_bot.nix#L14).

Looks like it could be indeed SnabbBot trying to run [git submodule update](https://github.com/eugeneia/snabb/blob/b6bed3018fe0b076867a9f644f393fdc3ae251a3/src/scripts/snabb_bot.sh#L69) (from back when we were using submodules instead of subtrees?). I think I remember @dpino reporting a similar problem, but can not find the issue right now.

dpino:
@eugeneia I remember I reported on some issues about snabb-bot when I first started dealing with the code, but I cannot remember if it was related with submodules :/

eugeneia:
I updated snabb bot on davos, it no longer attempts to fetch submodules and that appears to fix things.

The new test failure seems legit though,

```
src/testlog/program.snabbnfv.nfvconfig:
../lib/pflua/src/pf/types.lua:44: attempt to redefine 'bpf_insn'
```

Wonder if that is a clash with the new ljsyscall?

PR #1125, opened 2017-04-12
VMProfile (new LuaJIT sampling profiler) engine support

lukego:
I wrote an experimental sampling profiler for LuaJIT called VMProfile (LuaJIT/LuaJIT#290). It is very effective (although upstream didn't like it.)

This branch pulls that feature into Snabb and enables it by default. This means that each Snabb process will automatically create a small shared memory object that contains profiler samples saying which traces are hot. The samples are taken at 1ms intervals using a POSIX interval timer.

The profiler is low-level in the sense that it tells you the trace numbers that are hot. You then have to manually inspect the traces (e.g. with `-jdump`) to understand why. Sorry about that. (In the future I want to create a more efficient replacement for `-jdump` that can always be enabled and available for cross-referencing, but for now you have to do the JIT dumping yourself.)

This implementation actually creates one set of counters for the engine itself and a separate set of counters for each app class. This way you can choose whether to take a global view or whether to zoom in on particular apps.

The user interface is an external tool called [vmprofiler.R](https://github.com/studio/studio/blob/master/tools/vmprofiler/vmprofiler.R). This requires a little polish but is simple (~50 lines of code at the moment.)

- [ ] Check performance impact of the sampling (mandatory for enabled-by-default.)
- [ ] Provide simple instructions for running the `vmprofiler` frontend.

### Screenshots

Here is what the profiler can say about the `snabbmark basic1` benchmark:

Overall summary of where time is spent:

```
 profile percent
     Tee    34.7
    Sink    31.9
  Source    22.6
  engine    10.8
```

What kind of code is running? (The `head` of mcode traces, the `loop` of mcode traces, `foreign` C code, the garbage collector, the JIT assembler/recorder/etc):

```
    what percent
    loop    70.2
 foreign    16.0
    head    11.1
      gc     2.6
     asm     0.0
       c     0.0
  record     0.0
```

Where more specifically is this time being spent? (which trace numbers and which apps?)

```
 profile      where percent
    Sink    loop.20    26.3
    Sink      gc.22     2.6
    Sink    head.20     1.9
  Source    loop.13    16.4
  Source foreign.13     4.3
  Source    head.13     0.8
  Source    head.23     0.5
     Tee    loop.21    27.5
     Tee foreign.21     5.3
     Tee    head.21     0.9
  engine foreign.42     5.5
  engine    head.42     3.2
  engine    head.23     0.7
  engine    head.24     0.7
```

How about the 2.6% time spent garbage collecting?

```
 profile where percent
    Sink gc.22    99.9
```

(This makes finding & eliminating unsunk allocations as easy as shooting fish in a barrel: we can see that 99.9% of GC work is due to an allocation in trace number 22 which is being called from the `Sink` app.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ CSRC   = $(wildcard src/c/*.c)
 COBJ   = $(CSRC:.c=.o)
 PREFIX = /usr/local
 
-LUAJIT_CFLAGS := -include $(CURDIR)/gcc-preinclude.h
+LUAJIT_CFLAGS := -include $(CURDIR)/gcc-preinclude.h -DLUAJIT_VMPROFILE
 
 all: $(LUAJIT) $(SYSCALL) $(PFLUA)
 #       LuaJIT
--- a/lib/luajit/src/Makefile
+++ b/lib/luajit/src/Makefile
@@ -479,7 +479,7 @@ LJLIB_C= $(LJLIB_O:.o=.c)
 LJCORE_O= lj_gc.o lj_err.o lj_char.o lj_bc.o lj_obj.o lj_buf.o \
 	  lj_str.o lj_tab.o lj_func.o lj_udata.o lj_meta.o lj_debug.o \
 	  lj_state.o lj_dispatch.o lj_vmevent.o lj_vmmath.o lj_strscan.o \
-	  lj_strfmt.o lj_strfmt_num.o lj_api.o lj_profile.o \
+	  lj_strfmt.o lj_strfmt_num.o lj_api.o lj_profile.o lj_vmprofile.o \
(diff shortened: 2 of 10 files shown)


lukego:
@wingo are you up for being upstream on this one?

lukego:
I pushed the `vmprofiler` tool to studio/studio#5. Just now it has to be invoked via `nix-build` which is a bit verbose but I hope it's usable as a start! Going forward the idea is both to put a slicker front-end on (not manually typing nix expressions) and also take more advantage of nix (nice visualizations etc.)

eugeneia:
Closing because this has landed as part of RaptorJIT IIRC.

PR #1133, opened 2017-04-24
Merge Igalia/lwaftr upstream

wingo:
Hello kind Snabb hackers!

This is a PR that is manifestly too large, however it's what I got.  It merges hacking on the lwaftr to master, which hasn't been done for a while now (6 months or more).  Notably it merges in the YANG support.  Specific changes:

 * Add support for YANG schemas, configuration, and state data, allowing users to configure Snabb programs with pleasant textual schema-constrained configuration formats and have Snabb automatically translate that configuration into efficient native Lua data that is usable directly by a data plane.  See the [`lib.yang` documentation](https://github.com/Igalia/snabb/blob/lwaftr-2017-04-24/src/lib/yang/README.md) for full details.

 * New lwAFTR configuration format based on YANG.  To migrate old configurations, run `snabb lwaftr migrate-configuration old.conf` on the old configuration.  See the [`snabb-softwire-v1.yang` schema](https://github.com/Igalia/snabb/blob/lwaftr-2017-04-24/src/lib/yang/snabb-softwire-v1.yang) or [configuration.md](https://github.com/Igalia/snabb/blob/lwaftr-2017-04-24/src/program/lwaftr/doc/configuration.md) for full details on the new configuration format.

 * Add ability to reconfigure an entire lwAFTR process while it is running, including changes that can add or remove ingresss or egress filters, change NIC settings, or the like. Pass the `--reconfigurable` argument to `snabb lwaftr run`, then interact with the lwAFTR instance via `snabb config`.  This `--reconfigurable` facility is also available for other Snabb programs to use.  See [the `apps.config` documentation](https://github.com/Igalia/snabb/blob/lwaftr-2017-04-24/src/apps/config/README.md)

 * Add `snabb config` set of commands, to replace `snabb lwaftr control`.  These applications can query state or configuration of a running Snabb instance as well as update that configuration in an efficient way.  See [the `snabb config` documentation](https://github.com/Igalia/snabb/blob/lwaftr-2017-04-24/src/program/config/README.md) for full details.

 * Add support for named Snabb instances (and specifically named lwAFTR instances).  Pass `--name foo` to the `snabb lwaftr run` command to have it claim a name on a machine.   `snabb config` can identify the remote Snabb instance by name, which is often much more convenient than using the instance's PID.  The `snabb lwaftr query`, `snabb lwaftr monitor`, `snabbvmx query`, `snabbvmx top`, and `snabb top` tools can locate Snabb instances by name, and `snabb ps` can show all Snabb processes with their names.

 * Add support to the lwAFTR for fast binding-table updates to individual binding-table entries without causing the whole table to reload, via generic `snabb config` commands like `snabb config add /softwire-config/binding-table/softwire`.  See `snabb config` documentation for more on how to use `snabb config add` and `snabb config remove`.  This facility is also generic and can be used by other Snabb data planes.

 * Add support to lwAFTR for `ietf-softwire` alternate YANG schema.  Pass the `-s ietf-softwire` to `snabb config` invocations to use this schema.  Data planes expose a primary YANG model but there is a generic facility to allow the data plane to define transformations from other models to the primary model.

 * Adapt `snabbvmx` to configuration format changes.

 * Incorporate all PodHashMap functionality into `lib.ctable` and switch the lwaftr over to use `lib.ctable`, removing PodHashMap.

 * Incorporate multi-process work from @lukego in https://github.com/snabbco/snabb/pull/1021.

Review is very welcome.  My personal opinion is that all of this is "safe" and good to merge with the exception of the changes to `src/core/app.lua`: https://github.com/snabbco/snabb/compare/master...Igalia:lwaftr-2017-04-24#diff-e2bf611b7cf61fb539b42bf778b8fc9c.  These changes were designed and implemented here: https://github.com/snabbco/snabb/pull/1067.  I think that's the hard review.  I would suggest that we do perf testing on this merge and see if the regression found on that PR is still present in this one, and just see what happens.


wingo:
If I read https://hydra.snabb.co/build/1263403/download/2/report.html right, this branch has no significant perf impact.  Is that how you read it, @lukego?

lukego:
The CI report looks ambiguous to me. Have to take a closer look. Can be that the test run is still too small.

Great stuff on this branch :). I'm having a look at the changes to the core now. Somehow I am not finding the `magit` workflow that worked well for browsing large merges in the past. However now it seems like `git merge --no-commit --no-ff <branch>` does the job nicely i.e. it will stage all of the changes and let me explore the diff as if I were reviewing my own working copy.

lukego:
I'm running a really big benchmark of master vs lwaftr here: https://hydra.snabb.co/eval/5833.

The code looks good to me. I have only looked in detail at the changes to core code and my understanding of the YANG+CONFIG parts is superficial for now (looking forward to taking it for a spin in the future.)

It kind of looks to me like the changes to `app.lua` have been sorted out between @wingo and @eugeneia in previous discussions and that these changes will not cause any compatibility problems with other branches that haven't landed upstream yet. Is that the case?

wingo:
I guess the multi_copy test failed because it uses dynasm+AVX2 and the host that it runs on doesn't do AVX2.  I suppose I can skip the test in that case, or fall back to a different implementation.

wingo:
I believe that the core/app.lua changes have been hashed out, yes -- I understand that @eugeneia would still prefer that the app graph not cause manipulations to itself and I understand that :)  I think at some point I would like to refactor to make it so that worker processes that are participating in a leader/follower config relationship will poll for config changes outside of the app graph (e.g. as part of the main loop, possibly via a different main loop) instead of from inside it.  That should be a small change.

Additionally there was a performance question in https://github.com/snabbco/snabb/pull/1067#issuecomment-261184224.  I didn't understand it and didn't look into it farther, so we should make sure to check that particular case in our benchmarking runs.

lukego:
Is there a benchmark for the multi copy? I'm wondering if plain AVX would be the same speed if it's memory bound. There seems to be surprisingly little difference between avx/avx2 on memcpy routines in practice (glibc.)

I think it's a problem actually to error on pre-AVX2. Guessing most people want that (Ivy Bridge / Xeon v2) as a deployment option for another couple of years.

wingo:
Morning :) I did do benchmarking of a number of different options and this solution was fastest, though I suspect normal AVX would not be too terribly slower.  I didn't keep the other versions around though.

The multi_copy code is used in the streaming lookup interface to the ctable.  It's not used in the "normal" ctable lookup path and indeed couldn't really be used there -- it's really for streaming in multiple keys at a time, unlike the binary-search dasl that could be used with single lookups (but isn't currently).  So current ctable users aren't affected by this addition as their interfaces still work.  The only use of this code currently is the lwaftr, and we only deploy the lwaftr on machines with AVX2 so no problem there either afaiu.

dpino:
I want to comment on the failing SnabbVMX test reported by Snabb-bot.

Some weeks ago, we were setting up a new Snabb-bot CI in our snabb2 server. At some point we had two Snabb-bot instances running on every PR (and I think that's still the case today). Both Snabb-bot instances checked the same PRs, however on snabb1 tests passed but on snabb2 tests failed because of an SnabbVMX test.

After digging into this issue I reached to the conclusion that the cause for the failing test must be an environment problem. I compared the settings between snabb1 and snabb2 and a major difference was the kernel version used. snabb1 used 3.18.26 whereas snabb2 used 4.4.31.

I switched snabb2 to the same kernel version as snabb1 and all the tests passed.

In addition to that, snabbvmx was not the only failing test when using 4.4.31, there was also lib.pmu_x86. If I run this test in my laptop (4.8.0-49-generic) I got the following error:

```bash
$ sudo ./snabb snsh -t lib.pmu_x86
selftest: pmu_x86
nfixed  3       ngeneral        4
Segmentation fault (core dumped)
```

This error goes undetected in current master because the Makefile doesn't run .dasl tests (I think the lwAFTR merge should fix that).

What I suggest is to switch the Snabb-bot host to a 3.X kernel and check if tests pass as a first step. Eventually the failing snabbvmx test and lib.pmu_x86 should get fixed under 4.X.

wingo:
Regarding `lib.pmu_x86`, the bot does seem to run the test successfully: https://gist.github.com/SnabbBot/9b1632f41a542da562cefdb16927e523#file-log-L887

I am inclined to think that what must be a problem in snabbvmx or in its test suite should not cause the bot as a whole to have to use a particular kernel version -- given that the lib.pmu_x86 test passes, we should be able to figure out what's going wrong with this test on 4.x and fix it, or disable the test on master for the time being (instead of rolling back the kernel).  WDYT @dpino?  In the meantime we can file a bug to re-enable that test and prioritize that work.

dpino:
@wingo: OK, sounds good to me. I overlooked the results of lib_pmu tests, although I still got the error I pasted above if I run the test in my laptop (4.X kernel, AVX2 support). Since CI passes that's all we should care about.

I'd look deeper into the failing SnabbVMX test (program.snabbvmx.tests.selftest.sh). It could be the case that a Linux tool is missing in the VM. That would be an easy fix. But if that's not the cause, fixing the test could take longer. I'm OK with disabling it temporarily.

To disable it, I believe the easiest thing to do is to rename the test from "selftest.sh" to "selftest.disabled.sh", for example. This way the test won't be run on "make test".

wingo:
@eugeneia the snabb-bot run has been going overnight, since around 18h yesterday.  Is that normal?

eugeneia:
@wingo No it died :-/ Still working on sorting out davos.

lukego:
Hydra [seems happy](https://hydra.snabb.co/build/1281429/download/2/report.html)! Looks like this branch has improved some configurations/setups of the iperf benchmark for snabbnfv. Could be relatively obscure cases (e.g. running a patched QEMU with a larger vring) but always nice to move forward anyway.

lukego:
@wingo Do you want to do more work on this branch before merge?

wingo:
@lukego Sorry for the delay; I was on a short holiday from when you wrote.  I would like to fix snabb-bot before merging but I have no other functional changes queued.  I think the snabb-bot failures are related to configuration rather than to functionality, fwiw.

wingo:
Actually lookin at things again, I think the `intel_mp` test failure is spurious.  We certainly didn't change that code.  @eugeneia would you mind running the bot again on this PR?

wingo:
Just pushed a trivial commit to force the bot to run again.

eugeneia:
My guess is that the intel_mp selftest is somehow affected by the changes to core, given that it seems reproducible. Sorry for the SnabbBot delays, seems like we hit some new docker related bugs since I redeployed davos. Quite the pandora’s box I opened there…

wingo:
Well!  That was horrible.  I spent days ignorantly trying to understand things from a constructive perspective and then gave up and did a bisect.  It's wasn't quite a real bisect as the intel_mp driver only came in with v2017.04, but at each step between master and `lwaftr` I merged in #1086 and ran a test that just did two `testup.snabb` invocations on different queues on the same device.  If both complete with `true`, then great, otherwise one would hang and that would be the result.  Apparently the culprit is 9fe91ac51483232372abfc042e5120c5ad7d4be1 from the multi-process worker-and-supervisor code, and indeed it makes sense; the `intel_mp` driver expects bus mastering to not be tied to any particular process, I think.  Gotta knock off for today but will poke tomorrow.  /cc @lukego and @petebristow who will find this interesting I think.

src/core/worker.lua
```diff
@@ -0,0 +1,134 @@
...
+   local name = assert(lib.getenv("SNABB_WORKER_NAME"))
+   local parent = assert(lib.getenv("SNABB_WORKER_PARENT"))
+   print(("Starting worker %s for parent %d"):format(name, parent))
+
+   -- Create "group" alias to the shared group folder in the parent process
+   shm.alias("group", "/"..parent.."/group")
+
+   -- Run the engine with continuous configuration updates
+   local current_config
+   local child_path = "group/config/..name"
```
petebristow: Is the " in the wrong place?

wingo: Good question, it certainly looks like it.  Perhaps given that Luke is proposing #1021 separately, we comment there?

wingo:
The bot ran, yay!  From what I can tell, the failure noted in the bot run was a perf regression:

```
ERROR snabbnfv-iperf-1500-crypto -> 0.805118 of 1.016 (SD: 0.102489 )
```

However a previous run on this branch before this unrelated bus-mastering patch succeeded with 1.0, so I am inclined to think, no problem.

wingo:
This one's all ready to go, I think!  WDYT the next step is @lukego?  Do you want to do #1021 first and separately?  Should I just merge this to `wingo-next` ?

lukego:
Great hacking @wingo! This is merged onto #1148 with next hop `next`.

wingo:
Thank you for the review and the merge, yay!!!!

Issue #1141, opened 2017-05-15
Race condition in intel_mp 1g linkup test


wingo:
The `intel_mp` driver's self tests include a test that spawns 8 parallel invocations of `testup.snabb`, spread over 2 devices and 4 queues.  It waits for the last-spawned process to complete, then waits 2 seconds, then tests that all of the `testup.snabb` invocations have actually completed and produced output.  Usually parallel `testup.snabb` invocations on different queues of the same device will terminate together.

Unfortunately there is a race condition.  If the driver pokes the card and sees that it doesn't know the link status yet, it will wait 2 seconds.  Apparently an average time to know link status is right around 4 seconds, and the actual wait time is bimodal between 4 and 6 seconds (because of the sleep).  So many times, if you wait for the last-invoked process on device A and then sleep a couple seconds, it could be that you miss device B because it's already waiting 2 seconds longer.

The solution is to increase the timeout in the test, or to decrease the link status poll period.  I think I will try the latter.

wingo:
Example test run on master, FWIW:

```
[wingo@davos:~/snabb/src/apps/intel_mp]$ sudo lock sh -c 'for dev in 01:00.0 01:00.1; do for q in 0 1 2 3; do for i in 0 1 2 3 4 5 6 7 8 9; do time ./testup.snabb $dev $q; echo "./testup.snabb $dev $q result: $?"; done; done; done'
locking /var/lock/lab.. flock: getting lock took 0.000005 seconds
true

real	0m6.067s
user	0m0.030s
sys	0m0.022s
./testup.snabb 01:00.0 0 result: 0
true

real	0m4.073s
user	0m0.036s
sys	0m0.018s
./testup.snabb 01:00.0 0 result: 0
true

real	0m4.074s
user	0m0.038s
sys	0m0.019s
./testup.snabb 01:00.0 0 result: 0
true

real	0m4.071s
user	0m0.032s
sys	0m0.021s
./testup.snabb 01:00.0 0 result: 0
true

real	0m6.063s
user	0m0.027s
sys	0m0.017s
./testup.snabb 01:00.0 0 result: 0
true

real	0m4.074s
user	0m0.036s
sys	0m0.021s
./testup.snabb 01:00.0 0 result: 0
true

real	0m4.070s
user	0m0.032s
sys	0m0.020s
./testup.snabb 01:00.0 0 result: 0
true

real	0m4.073s
user	0m0.035s
sys	0m0.021s
[...]

PR #1142, opened 2017-05-16
intel_mp: Poll more frequently for linkup

wingo:
This patch lowers the link-up poll period from 2 seconds to 0.1
second, for faster reactions on link state change.  It also makes the
82599 linkup wait loop just wait on link state change, like the old
Intel82599 driver, instead of also doing reset logic in the wait loop.


wingo:
I wonder if the linkup poll period should be configurable though -- seems like a useless configuration option, one that we should just set appropriately.  I could make it a file-local variable instead.  WDYT @petebristow ?

petebristow:
The reset logic is in the loop because the card can wedge trying to come up. Particularly if the port is already seeing significant traffic during the reset. If you have it hooked up to a switch span port for instance. I wasn't able to come up with a test that would reliably fail if this code wasn't in place but if I ran selftest 1000 times I would get a few failures that went away with repeated resetting of the card. It's possible this is more an indication on the cheap cards / SFP+ DACs I'm using. 2 seconds appeared to be the sweet spot if it didn't come up after that period of time then it wasn't going to.

wingo:
This PR fixes #1141 fwiw.  I would say that 2s polling is not a good default and is specifically a regression relative to Intel82599 where it wasn't necessary.  Something needs to be done though, wdyt is the right way forward @petebristow ?

petebristow:
I don't follow, https://github.com/snabbco/snabb/blob/master/src/apps/intel/intel10g.lua#L126-L154 shows the Intel82599 doing a very similar thing of waiting on a link up and bouncing the card if it's not seen within a period of time. The control flow is a little more convoluted but the gist is the same. Are you objecting to the "something went wrong bounce the card until it works" methodology or that 2 seconds is too long to wait for the majority of cases where the card comes up ok first time?

wingo:
Hi, I see what you are saying, I didn't notice that bit before.  I have no objection to the loop actually, but I want a more reactive response to link-up.  You can see here that the older driver polls more frequently between resets: https://github.com/snabbco/snabb/blob/master/src/apps/intel/intel10g.lua#L394.

wingo:
I'll update the patch appropriately.

wingo:
Updated to re-introduce reset loop for 82599 cards and factor out the linkup wait loop.

wingo:
wdyt @petebristow ?

wingo:
The previous build failure was actually a bot failure I think.  Merged master to this branch and we'll see what the results are now.  @petebristow review still welcome, though @takikawa perhaps you would be available to review as well?

teknico:
LGTM. 0.1 seconds seems a good middle ground between the old driver behavior and 2 seconds.

PR #1146, opened 2017-05-17
Add core.version module and --version command-line argument

wingo:
This PR solves a problem I have had, which is verifying what version a particular Snabb binary is.  With this patch I get:

```
$ ./snabb --version
snabb 2017.04-1-g701bbc10
Copyright (C) 2012-2017 Snabb authors; see revision control logs for details.
License: <https://www.apache.org/licenses/LICENSE-2.0>

Snabb is open source software.  For more information on Snabb, see
https://github.com/snabbco/snabb.
```

All I really need is the version; I can trim off the rest if that's the right thing.

The only problem with this patch is that it always causes the dependencies of core/version.lua to be rebuilt -- the core_version_lua.o and the snabb binary.  I tried for a while to fix that but couldn't find the right way.  Oh well.  WDYT @lukego / @eugeneia ?


lukego:
I would also really like a precise `snabb --version` command!

I see a few issues related to running `git describe` in the Makefile:
- Is not available when building from a tarball (no `.git` database to query.)
- Is not available when building from nix (excludes `.git` because contents not deterministic.)
- Is potentially exposed to weird Git-isms (e.g. which _tags_ you have pulled into your repo.)

Is there a way to make this more deterministic so that it works with tarballs and nix too?

wingo:
Regarding tarballs: this patch also updates the "dist" target to copy a core/version.lua into the tarball.  Of course you might use some other way to create a tarball.

From Nix -- good question.  I guess we can use an environment variable (SNABB_VERSION or so), and default its value to `git describe` if it's unset and there is a .git.  FWIW Guile has a similar thing, and we ship a `.tarball-version` file to fall back on.

Regarding tags -- yes, can be weird.  Right now it restricts to tags that start with "v".  For the lwaftr this works for our purposes as we tag releases with `v3.1.9` etc... dunno though.  I think it will be OK though, as this is a function of your git checkout/repo, and you can control this, and when it really matters, you are setting a tag already because you're about to release.

wingo:
I should mention a couple of virtues of git-describe.  One, it includes the most recent tag; good for general orientation.  Two, for revisions that don't correspond to tags, it shows the number of commits since that tag -- so you can order two version numbers.  Third, for revisions that don't correspond to tags, it shows the abbreviated hash; useful.  Gives you a path back to the source from the version.  Finally it appends `-dirty` if you have local changes.  All of these are useful things to know.  When I log into a deployed machine and see a Snabb running, I would like to know what source it corresponds to, and if it's "new enough".  A simple checksum of sources is an oracle to verify that it's the same or different than a particular source checkout or a particular binary, but embedding a bit more git-isms lets me find the source more easily than a simple checksum, in addition to letting me compare versions.

lukego:
This looks like an excellent solution for the scenario where a developer is building and distributing a binary directly from their Git working tree, and may have commited or uncommitted changes since the most recent tag. This is an important case for sure, but it seems like we should cover other ones too:

- Build from a tarball like https://github.com/snabbco/snabb/archive/v2017.04.tar.gz.
- Build with tools like nix that don't provide `.git` but may provide another useful ID.
- Build with an explicit version to exclude all the magic e.g. for a release with own version scheme.

Do you agree? If so then what would be the simplest way?

One idea - best I have at the moment :) - would be to add a couple of optional version numbers to the Makefile arguments: `make VERSION=lukego-5.0 EXTRAVERSION=patch2` to build a binary that identifies as `lukego-5.0 (patch2)`. The default value of `VERSION` could be included in the source tree (e.g. master branch version.)  The default value of `EXTRAVERSION` could be picked up with magic like `git describe`, or the Nix derivation ID from the environment, etc.

lukego:
Let me try to be a proper upstream and make less rambling requests via the Review interface...

src/Makefile
```diff
@@ -72,6 +76,18 @@ $(EXE): snabb bin
 	@echo -n "BINARY    "
 	@ls -sh $@
 
+core/version.lua:
```
lukego: Request: Add an `else` to provide some meaningful version ID when `../.git` is not available.

(Ideally leave open the possibility for adding more ways to automatically detect the version ID in the future e.g. detecting a nix derivation name.)

src/Makefile
```diff
@@ -10,7 +10,11 @@ CC ?= gcc
 SRCDIR = $(shell find . -type d -not -regex './obj.*' -printf '%P ')
 OBJDIR = $(patsubst %,obj/%,$(SRCDIR))
 
-LUASRC = $(shell find . -regex '[^\#]*\.lua' -printf '%P ')
+GENLUASRC := core/version.lua
+LUASRC = $(shell for f in $(GENLUASRC); do \
+	           if test ! -f $$f; then echo -n "$$f "; fi; \
```
lukego: Request: Eliminate the special logic for building even when the `core.version` module is missing. Simpler to make sure this module always exists. Otherwise code that uses `core.version` will always need to check if it is there or not.

wingo: The intention was for the module to always exist; did I make an error?  The intention was for core/version.lua to be a "built source", like the lua modules that the dasl files compile into.

I didn't want to add version.lua to git because if the build process always overwrites it, then sadness would ensue.  WDYT?

wingo:
I think I addressed the review comments.  See changelog for https://github.com/snabbco/snabb/pull/1146/commits/fdf662d615b325eb6ae264d841aeffa970d22468.  An example:

```
wingo@rusty:~/src/snabb/src$ make
GEN       obj/core/version.lua.gen
LUA       obj/core/version_lua.o
LINK      snabb
BINARY    2.8M snabb
wingo@rusty:~/src/snabb/src$ ./snabb --version
snabb 2017.04 (2-gfdf662d6)
Copyright (C) 2012-2017 Snabb authors; see revision control logs for details.
License: <https://www.apache.org/licenses/LICENSE-2.0>

Snabb is open source software.  For more information on Snabb, see
https://github.com/snabbco/snabb.
wingo@rusty:~/src/snabb/src$ make VERSION=2017.05
GEN       obj/core/version.lua.gen
LUA       obj/core/version_lua.o
LINK      snabb
BINARY    2.8M snabb
wingo@rusty:~/src/snabb/src$ ./snabb --version
snabb 2017.05 (fdf662d6)
Copyright (C) 2012-2017 Snabb authors; see revision control logs for details.
License: <https://www.apache.org/licenses/LICENSE-2.0>

Snabb is open source software.  For more information on Snabb, see
https://github.com/snabbco/snabb.
```

lukego:
Great! The diff looks good to me, thanks for those changes.

Tarball build fails for me though:

```
$ wget https://github.com/Igalia/snabb/archive/snabb-version.tar.gz
$ tar zxf snabb-version.tar.gz
$ cd snabb*
$ make
...
GEN       obj/core/version.lua.gen
/bin/sh: obj/core/version.lua.gen: No such file or directory
/bin/sh: line 1: obj/core/version.lua.gen: No such file or directory
/bin/sh: line 11: obj/core/version.lua.gen: No such file or directory
/bin/sh: line 12: obj/core/version.lua.gen: No such file or directory
Makefile:78: recipe for target 'obj/core/version.lua.gen' failed
make[1]: *** [obj/core/version.lua.gen] Error 1
make[1]: Leaving directory '/tmp/snabb-snabb-version/src'
Makefile:11: recipe for target 'all' failed
make: *** [all] Error 2
```

wingo:
@lukego ah good catch.  fixed, i think.

lukego:
Thanks, this works for me now both with git and tarball builds.

> The only problem with this patch is that it always causes the dependencies of core/version.lua to be rebuilt -- the core_version_lua.o and the snabb binary. I tried for a while to fix that but couldn't find the right way. Oh well. WDYT

I have also spent some time on this now and not found a solution. I am a bit sad-panda about this. On the one hand it is nice to have version information included in the build by default, on the other hand I will miss the reassuring `make: 'snabb' is up to date.` message and the property of the `snabb` binary only being rebuilt when the code and/or version has actually changed. I will just suck that up and merge this week unless somebody else knows a way to get the best of both worlds?

wingo:
There must be a way to fix the rebuilding issue, even if it makes the makefile more complicated.  I will take a poke.

wingo:
PTAL; I think I fixed the rebuild issue.

lukego:
Sorry about the slow turnaround. This branch is not building for me, neither from tarball (command above) or from my working checkout.

```
/tmp/bcd/snabb-snabb-version]$ make -j
GEN       obj/version.lua.gen
/bin/sh: ../generate-version-lua.sh: No such file or directory
Makefile:56: recipe for target 'snabb' failed
make[1]: *** [snabb] Error 127
make[1]: Leaving directory '/tmp/bcd/snabb-snabb-version/src'
Makefile:11: recipe for target 'all' failed
make: *** [all] Error 2
```

lukego:
Clever! I did not think of including the version-generation as part of the step to build `snabb` so that it only happens when that would be done anyway.

Simple and works like a charm:

```
$ make
make: 'snabb' is up to date.
```

PR #1147, opened 2017-05-21
Illustrating performace issues with socket app

raj2569:
Socket app, just by its presence slows down the performance drastically. The snabbtest added with this
commit illustrates this performance drop.

```
sudo ./snabb snsh -jdump=+r,dump1.txt -jtprof -p snabbmark socket_test 100e6 8

Processed 100.0 million packets in 53.44 seconds (rate: 1.9 Mpps).
traceprof report (recorded 13158/13158 samples):
```
```
sudo ./snabb snsh -jdump=+r,dump1.txt -jtprof -p snabbmark socket_test 100e6 0

Processed 100.1 million packets in 3.01 seconds (rate: 33.2 Mpps).
traceprof report (recorded 756/756 samples):
```

### DIFF
--- a/src/program/snabbmark/snabbmark.lua
+++ b/src/program/snabbmark/snabbmark.lua
@@ -4,17 +4,20 @@ module(..., package.seeall)
 
 local usage = require("program.snabbmark.README_inc")
 
-local basic_apps = require("apps.basic.basic_apps")
-local pci           = require("lib.hardware.pci")
-local ethernet      = require("lib.protocol.ethernet")
-local lib = require("core.lib")
-local ffi = require("ffi")
-local C = ffi.C
+local basic_apps  = require("apps.basic.basic_apps")
+local socket      = require("apps.socket.unix")
+local pci         = require("lib.hardware.pci")
+local ethernet    = require("lib.protocol.ethernet")
+local lib         = require("core.lib")
+local ffi         = require("ffi")
+local C           = ffi.C
 
 function run (args)
    local command = table.remove(args, 1)
    if command == 'basic1' and #args == 1 then
       basic1(unpack(args))
+   elseif command == 'socket_test' and #args == 2 then
+      socket_test(unpack(args))
    elseif command == 'nfvconfig' and #args == 3 then
       nfvconfig(unpack(args))
    elseif command == 'solarflare' and #args >= 2 and #args <= 3 then
@@ -57,6 +60,37 @@ function basic1 (npackets)
    local finish = C.get_monotonic_time()
    local runtime = finish - start
    local packets = link.stats(engine.app_table.Source.output.tx).txpackets
+   engine.report()
+   print()
+   print(("Processed %.1f million packets in %.2f seconds (rate: %.1f Mpps)."):format(packets / 1e6, runtime, packets / runtime / 1e6))
+end
+
+
+function socket_test (npackets, nsockets)
+   npackets = tonumber(npackets) or error("Invalid number of packets: " .. npackets)
+   nsockets = tonumber(nsockets) or error("Invalid number of sockets: " .. nsockets)
+
+   local c = config.new()
+   config.app(c, "source",    basic_apps.Source)
+   config.app(c, "join",      basic_apps.Tee)
+   config.app(c, "sink",      basic_apps.Sink)
+
+   for i=1,nsockets do
+      config.app(c, "socket0"..i,  socket.UnixSocket, {filename="/tmp/socket0"..i, listen=true, mode="packet"})
+      config.link(c, ("socket0%d.tx  -> join.in%d"):format(i,i))
+   end
+   config.link(c, "source.tx    -> join.in")
+   config.link(c, "join.out     -> sink.in1")
+   engine.configure(c)
+
+   local start = C.get_monotonic_time()
+   timer.activate(timer.new("null", function () end, 1e6, 'repeating'))
+   while link.stats(engine.app_table.source.output.tx).txpackets < npackets do
+      engine.main({duration = 0.01, no_report = true})
+   end
+   local finish = C.get_monotonic_time()
+   local runtime = finish - start
+   local packets = link.stats(engine.app_table.source.output.tx).txpackets
    engine.report()
    print()
    print(("Processed %.1f million packets in %.2f seconds (rate: %.1f Mpps)."):format(packets / 1e6, runtime, packets / runtime / 1e6))


lukego:
Great reproducible benchmark!

Let me use this as an opportunity to test the latest profiling tool, VMProfile (#1125). This is a sampling profiler that collects ~1000 samples per second to show how much time is being spent in each app and in the engine itself, and also to break that time down into categories like JIT code vs FFI calls vs LuaJIT C code vs Lua interpreter vs Garbage collection.

Here is the high-level overview of the profiler report:

```
$ cat overview.txt 
     profile percent
 apps.socket      71
      engine      23
      Source       2
        Sink       1
         Tee       1
```

This is quite striking: hardly any time being spent in the apps that actually process packets (`Source`, `Sink`, `Tee`) and most time being spent in the `engine` (23%) and the `apps.socket` (71%). This is the opposite of what we want and expect i.e. for most time to be spent processing packets.

Here is another high-level global view of what kind of code is executing:

```
$ cat what.txt 
    what percent
  interp    39.9
       c    25.7
      gc    17.9
 foreign     7.6
    head     4.8
    loop     4.1
    exit     0.0
  record     0.0
```

Here we see that most of the time is spent in the code we don't want to run, like the Lua bytecode interpreter (`interp` 39.9%) and the LuaJIT C runtime system (`c` 25.7%) and the garbage collector (`gc` 17.9%). Little time is spent in more legitimate activities like JIT code (`head` 4.8% and `loop` 4.1%) and FFI calls (`foreign` 7.6%).

So both of these first reports confirm that something fishy is going on and we are not spending our time running nice fast machine code, which is what we want.

We can also zoom in on the specific activities of each app:

```
$ cat where.txt
     profile      where percent
        Sink    loop.35     1.2
      Source    loop.29     1.8
      Source foreign.29     0.5
         Tee    loop.33     1.0
 apps.socket     interp    31.8
 apps.socket          c    20.0
 apps.socket         gc    10.4
 apps.socket      gc.64     6.1
 apps.socket      gc.42     1.0
 apps.socket foreign.42     0.7
      engine     interp     8.0
      engine          c     5.7
      engine foreign.60     2.4
      engine foreign.66     1.1
      engine foreign.77     1.1
      engine    head.38     0.8
      engine foreign.38     0.7
```

Here we see specifically that the `Sink`, `Source`, and `Tee` apps are all running sensible code (machine code and FFI) while `apps.socket` is running mostly the interpreter and (probably as a side-effect) runtime system C and garbage collection routines. Suspiciously the engine is also running a significant amount of interpreted code which we also need to account for.

So: this tells us that we should be looking for the reason(s) that `apps.socket` and `engine` are running interpreted, which will most likely show up as trace abort/blacklist messages in the JIT trace/dump file. So - the next step is to read that.

Peeking ahead: Likely we will solve a couple of specific problems to make the code JIT instead of interpreting, and this will probably automatically resolve most/all of the runtime system calls (C/GC) by optimizing them away, and then if we are still not satisfied with the performance we could look at individual machine code traces to see the code that runs (for example `loop.35` in the profiler report is referring to the code we can find by opening the JIT dump and searching for `TRACE 35` and then scrolling down to the `--LOOP--` section of the machine code.)

Phew!

Ongoing related work here is to make the JIT tracing functionality be always enabled (see https://github.com/raptorjit/raptorjit/pull/63) and to make all the profiling/analysis tools really easy to run (see https://github.com/studio/studio/pull/6).

lukego:
So! The next step is to understand why some code is running in interpreted mode. We can look in our [dump1.txt](https://gist.github.com/c782d25532577f118f8c1ce5e00e2931
) file to work this out.

There are two main things we can expect to see in the dump: `abort` message where the JIT made an unsuccessful attempt to compile some code, and `blacklist` messages where the JIT gave up on compiling a piece of code because of too many aborts.

This is a bit of a needle-in-a-haystack situation because the dump is global and also contains e.g. a bunch of messages for things happening during startup and so on. So for the moment some manual legwork is needed, and in the future we need better tools for interpreting these dumps (some related work here: https://github.com/raptorjit/raptorjit/pull/63).


Just scanning down the dump file for `abort` and `blacklist` messages that seem relevant. Here is one:

```
---- TRACE 36 start syscalls.lua:362
0001  UGET     2   0      ; t
0002  TGETS    2   2   0  ; "fdset"
0003  CALL     2   2   1
0000  . FUNCC               ; ffi.meta.__call
---- TRACE 36 abort syscalls.lua:363 -- NYI: unsupported C type conversion
```

Here is what we can see:

- The JIT started tracing at [`syscalls:362`](https://github.com/snabbco/snabb/blob/master/lib/ljsyscall/syscall/syscalls.lua#L362) which is the function `mkfdset`.
- The trace aborted when it reached the [bytecode `FUNCC`](http://wiki.luajit.org/Bytecode-2.0) which is a call to a C function.
- The reason for the abort is a C type conversion that is "NYI" (i.e. non-JIT'able.)

Speculating a little bit I suspect this is due to this line at [socket/raw.lua:62](https://github.com/snabbco/snabb/blob/master/src/apps/socket/raw.lua#L62):

```
local t, err = S.select({readfds = {self.sock}}, 0)
```

Here we are using ljsyscall (module `S`) to call the C function `select()` and we are passing the parameter, a file descriptor set, as a Lua object. Probably the JIT is not able ("NYI") to generate machine code that converts a Lua object into a C fdset and so it forces the code to be interpreted.

So based on this theory the simplest fix that I can recommend would be to modify `socket/raw.lua` to avoid calling `select()` and instead to just call `read()` and accept that it will fail if no data is available. The `select()` is not really needed and it _seems like_ the performance problem is due to conversion from Lua to C of its file descriptor set argument.

lukego:
Just a few further interesting bits from the dump...

Here is the `select()` function being blacklisted by the JIT:

```
---- TRACE 79 start c.lua:664
0001  UGET     5   0      ; syscall
0002  UGET     6   1      ; sysselect
0003  UGET     7   2      ; int
0004  MOV      8   0
0005  CALL     7   2   2
0000  . FUNCC               ; ffi.meta.__call
0006  UGET     8   3      ; void
0007  MOV      9   1
0008  CALL     8   2   2
0000  . IFUNCF   4          ; c.lua:27
---- TRACE 79 abort c.lua:27 -- blacklisted
```

The last bytecode where it bails out is `IFUNCF` which is a call to an interpreted (blacklisted) function. So the blacklisting propagates because if one function is blacklisted then its callers will also be blacklisted, and their callers, and so on. This likely explains why the vmprofile is also showing the engine running interpreted code.

Overall it is a process for us to learn which specific Lua/LuaJIT language constructs cannot be JIT'ed and avoid using them in our apps. Looks like if we want to call `select()` we would need to explicitly create the fdset object instead of relying on automatic conversion from a Lua table. There are other [optimization tips](https://stackoverflow.com/questions/7167566/luajit-2-optimization-guide) around too.

I have a medium-term goal to make much better tools for understanding what the JIT is doing and why...

raj2569:
Updated the branch with `select()` replaced with `read()` in `socket/unix.lua`. While running I am getting an error:

```
syscall/helpers.lua:66: attempt to call method 'getfd' (a nil value)
stack traceback:
        core/main.lua:138: in function 'getfd'
        syscall/helpers.lua:66: in function 'getfd'
        syscall/syscalls.lua:90: in function 'read'
        apps/socket/unix.lua:101: in function 'can_receive'
        apps/socket/unix.lua:145: in function 'method'
        core/app.lua:127: in function 'with_restart'
        core/app.lua:429: in function 'thunk'
        core/histogram.lua:98: in function 'breathe'
        core/app.lua:376: in function 'main'
        program/snabbmark/snabbmark.lua:89: in function 'socket_test'
        program/snabbmark/snabbmark.lua:20: in function 'run'
        core/main.lua:56: in function <core/main.lua:43>
        [C]: in function 'xpcall'
        core/main.lua:179: in main chunk
        [C]: at 0x0044f270
        [C]: in function 'pcall'
        core/startup.lua:3: in main chunk
        [C]: in function 'require'
```
Also noticed that the `socket/unix.lua` file has a structure different from that of other apps. Only top level function is `UnixSocket:new(arg)` with every other function defined inside it, I don't know if that will make a difference though.

lukego:
I pushed a fix for this problem on #1153. Just now I am seeing this performance:
```
$ sudo ./snabb snsh -jdump=+rT,dump1.txt -p snabbmark socket_test 100e6 8
Processed 100.1 million packets in 10.65 seconds (rate: 9.4 Mpps).
```
That branch is based on `master` rather than this branch. You will get a conflict if you merge it. Here is one idea to resolve that:

```
git remote add lukego https://github.com/lukego/snabb
git fetch lukego socket-unix-push-pull
git merge lukego socket-unix-push-pull
git checkout --theirs src/apps/socket.lua
```

raj2569:
```
raj@1G-ethernet:~/snabb/src$ sudo ./snabb snsh -jdump=+rT,dump1.txt -p snabbmark socket_test 100e6 8

Processed 100.1 million packets in 5.45 seconds (rate: 18.4 Mpps).
raj@1G-ethernet:~/snabb/src$ sudo ./snabb snsh -jdump=+rT,dump1.txt -p snabbmark socket_test 100e6 16

Processed 100.1 million packets in 7.28 seconds (rate: 13.7 Mpps).
```

Much better results!

raj2569:
Closing as the issue has been fixed.

PR #1148, opened 2017-05-22
Merge multiprocess and lwaftr branches

lukego:
This integration branch is based on `next` and first merges multiprocess support (#1021) and then the latest lwAFTR (#1133).

The lwAFTR branch has already merged the multiprocess branch several times during development and so many commits are common to both branches. The reason that I merge both onto this integration branch is to make sure no commits are left behind and to detect any conflicts before landing on `next`. Git did not detect any conflicts so if the CI tests pass we are golden.

For anybody who wants to see how the new multiprocess support works: hop over to #1021 to see the [short README](https://github.com/lukego/snabb/blob/6d7e516fb306d46579e272e2ef4c3241938e6a14/src/README.md#multiprocess-operation-coreworker) for the API and [shortish diff](https://github.com/snabbco/snabb/pull/1021/files) for the code. Don't be distracted by all of the discussion threads on the #1021 PR because that is mostly preliminary chatter while we thrashed out a working design. The API and implementation are both quite small and straightforward.


wingo:
I don't quite understand the snabbdoc build error; do you, @lukego?  I would think we can merge this to `next` at this point though.

lukego:
Sorry about the wait. Hopefully fixed in 56a5acf.

lukego:
Yep that did it. Merged onto `next` now.

PR #1151, opened 2017-05-29
core.shm: Make shm objects world-readable

lukego:
Controversial idea for discussion: Make the /var/run/snabb/* shm directories world-readable.

This is more convenient for inspection during development and testing, but less convenient if you are running Snabb on a machine with unprivileged users (e.g. they could map the DMA memory to create backdoors.)

What do we think?

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -469,8 +469,16 @@ of this process:
 - Fully qualified: `/1234/foo/bar` ⇒ `/var/run/snabb/1234/foo/bar`
 
 Behind the scenes the objects are backed by files on ram disk
-(`/var/run/snabb/<pid>`) and accessed with the equivalent of POSIX shared
-memory (`shm_overview(7)`).
+(`/var/run/snabb/<pid>`) and accessed with the equivalent of POSIX
+shared memory (`shm_overview(7)`). The files are automatically removed
+on shutdown unless the environment `SNABB_SHM_KEEP` is set. The
+location `/var/run/snabb` can be overridden by the environment
+variable `SNABB_SHM_ROOT`.
+
+Shared memory objects are created world-readable for convenient access
+by diagnostic tools. You can lock this down by setting
+`SNABB_SHM_ROOT` to a path under a directory with appropriate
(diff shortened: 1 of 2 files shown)


wingo:
How would making /var/run/snabb readable directly enable backdoors?  I would have thought that would require write capability, but I am missing something :)

lukego:
@wingo Yes, I suppose that you probably need write access to do some obvious nefarious thing. Could be that it's not so unreasonable a default to allow read access to all objects, perhaps with the exception of DMA memory that could still be root-only (to avoid accidentally capturing user traffic.)

I suppose the main thing is to add a description to the manual. Then users can decide what steps they need to take for their deployment model (e.g. do they allow unprivileged users to execute arbitrary commands) and their threat model (e.g. do they care about local privilege escalation.)

WDYT?

wingo:
Sounds good to me.  We should note that this is a information leakage -- allowing access to mapped DMA memory is obviously allowing processes to read Snabb's memory.  We may need to add a knob in the future, but this is probably good enough for now.

lukego:
@wingo Thanks for the review. Addressed in c6d3a07. I suggest there that if you want to control permissions you can set `SNABB_SHM_ROOT=/var/run/some-locked-down-directory/snabb`.

lukego:
I made `/var/run/snabb` default to `rwxr-xr-x` permissions in 961fc43. This is stricter than the `rwxrwxrwxt` that we had before but seems likely to work for all use cases. @eugeneia is that okay for you? Looks like the original permissions came in with your commit 0df7f4af78a2dc071a7d326a5fe52ae597fa2e56.

eugeneia:
The permissions for `/var/run/snabb` (sticky, world writable—like `/tmp`) were chosen with a multi-user scenario in mind. I.e. given `/var/run/snabb` exists a non-superuser could start Snabb (e.g. create a pid sub-directory in `/var/run/snabb`). Since we agreed on a root-only user story, I think the world-writable sticky magic is obsolete.

:+1:

PR #1153, opened 2017-06-02
apps.socket.unix: Rewrote pull/push to avoid select()

lukego:
Avoiding select() for two reasons.

First, the way that we were passing the fdset argument to select() - as a Lua table - is an NYI for the JIT and forces the code to run interpreted.

Second, we don't really need select(), and can instead deal with failure of read() and write().

This is to help resolve the performance issue on #1147.


lukego:
Merged. The SnabbBot failure appears to be a false-negative performance regression test result.

PR #1158, opened 2017-06-09
Support AVX-only architectures in multi_copy.dasl

### LINKED ISSUE #1143: Don't require AVX2 in lib.multi_copy corner case
The `lib/multi_copy.dasl` code emits a `vpmovmaskd` instruction if it needs to copy the tail of a sequence that is not a multiple of 32 bytes long.  This instruction is the only AVX2 instruction in this file; we should use just AVX if that's what's supported on the system.

dpino:
Fixes #1143.

Explanation: double-words and single-precision floats have the same size (4-bytes). When it comes to moves and bitwise operations the type of data doesn't matter. That makes possible to replace `vpmaskmovd` (works on packed double-words) for `vmaskmovps` (workd on packed single-precision floats).

I made the PR 2 separated commits to check the AVX-only code works in an AVX2 architecture.

I didn't check performance of both versions. If there's no difference maybe is worth to use the AVX version only.

### DIFF
--- a/src/lib/multi_copy.dasl
+++ b/src/lib/multi_copy.dasl
@@ -92,10 +92,10 @@ function gen(count, size)
 
          if to_copy > 0 then
             for i = 0, stride-1 do
-               | vpmaskmovd ymm(i), ymm15, [Rq(8+i)]
+               | vmaskmovps ymm(i), ymm15, [Rq(8+i)]
             end
             for i = 0, stride-1 do
-               | vpmaskmovd [rdi + i*size], ymm15, ymm(i)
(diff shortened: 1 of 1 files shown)


wingo:
I think just use maskmovps then.  It's the same thing, runs just as fast.

dpino:
wingo: OK, so then I will just simply revert the last commit.

wingo:
lgtm!

lukego:
Nice!

lukego:
@wingo Land this one via wingo-next?

PR #1161, opened 2017-06-29
snabbmark esp: prevent GC from being invoked by making local vars local

eugeneia:
This is a small fix to a bug in the snabbmark esp micro benchmark that had noticeable negative effect on the results. See the commit message for more details.

I found this while benchmarking the (raw) ESP routines on lugano-1, here are the results with this patch:

snabb-esp-raw

### DIFF
--- a/src/program/snabbmark/snabbmark.lua
+++ b/src/program/snabbmark/snabbmark.lua
@@ -363,9 +363,8 @@ function esp (npackets, packet_size, mode, profile)
    if mode == "encapsulate" then
       if profile then profiler.start(profile) end
       local start = C.get_monotonic_time()
-      local encapsulated
       for i = 1, npackets do
-         encapsulated = packet.clone(plain)
+         local encapsulated = packet.clone(plain)
          enc:encapsulate(encapsulated)
          packet.free(encapsulated)
       end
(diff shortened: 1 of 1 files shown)


lukego:
Nice one.

Aside: I have a half-baked idea to reduce the need for these kind of optimizations in the future.

Generally it bothers me that we need to worry about this kind of optimization i.e. narrowing the scope of local variables that reference pointer/int64 values to prevent them from escaping from traces (ensuring they can be "sunk.") The trouble is that the trace structure is not obvious from looking at source code and so it is a black art to know when this is needed (basically you need to run `-jdump` and look for unsunk allocations and ponder their root cause.) I don't think we want Snabb hackers to be forced to dabble in black arts.

My idea is that we extend LuaJIT/RaptorJIT so that if a boxed constant (ptr/int64) cannot be sunk then it is allocated on the _stack_ rather than as a garbage-collected object on the heap. This way the basic penalty for an unsunk allocation would be much less: you spill the value from a register onto the stack with a few `mov` instructions and no heap allocation or garbage collection.

You would still need to pay the full price if the value then escapes from the stack e.g. if you store it in a table or return it from a function outside of a trace. But these situations should hopefully be easier to identify when looking at the source code.

lukego:
(The challenge with that idea is that you can't currently store a boxed value on the stack. So we would need to introduce wider stack slots where the first word can be a tagged pointer to the second word. Then we would need to make sure these values don't outlive the lifetime of that stack slot: if that would happen then we need to transport the value onto the heap and rewrite the pointer. This may be closely related to the Lua concept of [open/closed upvalues](http://www.cs.tufts.edu/~nr/cs257/archive/roberto-ierusalimschy/closures-draft.pdf) where local variables are only committed to closures when they escape the stack.)

PR #1167, opened 2017-07-03
next: Changes queued for 2017.07 release

lukego:
Still accumulating...


lukego:
@eugeneia This should be good to go :+1:.

There actually seems to be a significant [performance increase on the iperf benchmarks](https://hydra.snabb.co/build/1604171/download/2/report.html#iperf), with results clustering more tightly around 15G and less often falling down to 10G. I am very curious to account for this but the release needn't wait on that.

PR #1168, opened 2017-07-03
Randomly seed hash function used by ctables

wingo:
This PR changes ctable to randomly seed its SipHash hash functions.  This has two benefits: we prevent some kinds of DOS attacks if the user is able to programmatically make hash collisions, and we fix an accidentally-quadratic behavior if we incrementally populate a ctable from a sequence of entries that were serialized in hash order (https://github.com/Igalia/snabb/issues/654).

### DIFF
--- a/src/core/lib.lua
+++ b/src/core/lib.lua
@@ -706,6 +706,13 @@ function getenv (name)
    else return nil end
 end
 
+-- Wrapper around execve.
+function execv(path, argv)
+   local env = {}
+   for k, v in pairs(syscall.environ()) do table.insert(env, k.."="..v) end
+   return syscall.execve(path, argv or {}, env)
+end
+
 -- Return an array of random bytes.
 function random_bytes_from_dev_urandom (count)
    local bytes = ffi.new(ffi.typeof('uint8_t[$]', count))
--- a/src/core/lib.lua
+++ b/src/core/lib.lua
@@ -224,15 +224,6 @@ function comma_value(n) -- credit http://richard.warburton.it
    return left..(num:reverse():gsub('(%d%d%d)','%1,'):reverse())..right
 end
 
-function random_data(length)
-   result = ""
-   math.randomseed(os.time())
-   for i=1,length do
-      result = result..string.char(math.random(0, 255))
-   end
-   return result
-end
-
 -- Return a table for bounds-checked array access.
 function bounds_checked (type, base, offset, size)
    type = ffi.typeof(type)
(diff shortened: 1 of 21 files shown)


wingo:
Failing test appears to be a property-based test, that two runs produce the same output; of course ctable entries can indeed be reordered if the hash seed changes.  grr

wingo:
I did some performance testing; no regressions for lwaftr as far as I can tell.

src/core/lib.lua
```diff
@@ -715,6 +706,43 @@ function getenv (name)
...
+   if seed then
+      local msg = 'Using deterministic random numbers, SNABB_RANDOM_SEED=%d.\n'
+      io.stderr:write(msg:format(seed))
+      math.randomseed(seed)
+      -- When setting a seed, use deterministic random bytes.
+      random_bytes = random_bytes_from_math_random
+   else
+      -- Otherwise use /dev/urandom.
+      seed = ffi.cast('uint32_t*', random_bytes_from_dev_urandom(4))[0]
+      random_bytes = random_bytes_from_dev_urandom
```
dpino: Apparently `random_bytes` only exists if `randomseed` is called. It seems this is something the user should know beforehand before using `random_bytes`. Wouldn't be better that `randombytes` always exists and takes a default (possibly `random_bytes_from_dev_urandom`)?

src/lib/hash/siphash.dasl
```diff
@@ -45,23 +45,14 @@ local function finish (name, prototype, Dst)
...
+   return lib.random_bytes(16)
 end
 
-function reference_sip_hash_key()
-   return ffi.new('uint8_t[16]',
-                  0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15)
+function sip_hash_key_from_seed(seed)
+   local res = ffi.new('uint8_t[16]')
+   ffi.cast('uint32_t*', res)[0] = assert(tonumber(seed))
+   return res
```
dpino: I don't understand this code. `res` represents a 128-bit number, but seed is expected to be a uint32_t number. Why `res` is not an uint32_t number?

src/core/lib.lua
wingo: Snabb's `core.main` now calls `lib.randomseed`, so `lib.random_bytes` should always exist.

src/lib/hash/siphash.dasl
wingo: SipHash keys are 16 bytes log.  However a hash "seed" is much less -- in practice, just 4 bytes.  So we set 4 bytes of the SipHash key from the "seed", and leave the rest as zeros.

PR #1170, opened 2017-07-06
lib.pmu: Enable RDPMC via /sys/devices/cpu/rdpmc

lukego:
We want to access the CPU PMU hardware directly instead of via Linux
system calls. This requires the kernel to allow access to the RDPMC
instruction from userspace processes (including Snabb which the kernel
does not know is accessing the PMU.)

This requires the /sys/devices/cpu/rdpmc value to be 2 on newer
kernels (4.0) or 1 on older kernels.

This change makes Snabb attempt to raise the /sys/devices/cpu/rdpmc
value up to the highest value that the kernel supports.

### DIFF
--- a/src/lib/pmu_x86.dasl
+++ b/src/lib/pmu_x86.dasl
@@ -14,6 +14,7 @@ module(..., package.seeall)
 
 local debug = false
 
+local lib = require("core.lib")
 local ffi = require("ffi")
 local C = ffi.C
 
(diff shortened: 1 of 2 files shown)


src/lib/pmu_x86.dasl
```diff
@@ -130,8 +131,25 @@ function gen_rdpmc_multi (Dst)
 end
 rdpmc_multi = assemble("rdpmc_multi", "void(*)(void*)", gen_rdpmc_multi)
 
+-- Enable the RDPMC instruction in userspace via /sys/devices/cpu/rdpmc.
+-- Older kernels want value 1, newer kernels want value 2.
+-- See man perf_event_open(2) for gory details.
+function enable_rdpmc ()
+   local path = "/sys/devices/cpu/rdpmc"
+   local old = tonumber(lib.firstline(path))
```
wingo: add an `assert` around the `tonumber`, perhaps?

PR #1176, opened 2017-07-10
Avoid the most repetitive output in tests

teknico:
The test suite output has become very long and repetitive when there are errors, so much so that Github truncates the report ([example](https://gist.github.com/lwaftr-igalia/52ef86d3ab93dd8859d0726a180f486e)). This removes the worst offender, generating hundreds of lines.

### DIFF
--- a/src/core/lib.lua
+++ b/src/core/lib.lua
@@ -730,16 +730,12 @@ function random_bytes_from_math_random (count)
    return bytes
 end
 
-local lastseed=0
 function randomseed (seed)
    seed = tonumber(seed)
    if seed then
+      local msg = 'Using deterministic random numbers, SNABB_RANDOM_SEED=%d.\n'
+      io.stderr:write(msg:format(seed))
       -- When setting a seed, use deterministic random bytes.
-      if seed ~= lastseed then
-        lastseed = seed
-        local msg = 'Using deterministic random numbers, SNABB_RANDOM_SEED=%d.\n'
-        io.stderr:write(msg:format(seed))
-      end
       random_bytes = random_bytes_from_math_random
    else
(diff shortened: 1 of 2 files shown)


src/core/lib.lua
```diff
@@ -733,8 +733,6 @@ end
 function randomseed (seed)
    seed = tonumber(seed)
    if seed then
-      local msg = 'Using deterministic random numbers, SNABB_RANDOM_SEED=%d.\n'
```
wingo: I think this is not a good change.  If a test spawns so many sub-processes with this environment variable set that it's too spammy, that test needs to filter sub-process output.

teknico: I don't want to lose all stderr output and it's too complex to selectively filter the output at the test level, so I'll reinstate the message, but only when using a different seed from the last one.

src/core/lib.lua
```diff
@@ -730,12 +730,16 @@ function random_bytes_from_math_random (count)
...
-      io.stderr:write(msg:format(seed))
       -- When setting a seed, use deterministic random bytes.
+      if seed ~= lastseed then
+        lastseed = seed
+        local msg = 'Using deterministic random numbers, SNABB_RANDOM_SEED=%d.\n'
+        io.stderr:write(msg:format(seed))
+      end
```
wingo: I don't think this is necessary.  `lib.randomseed` should be called only once in a process's lifetime.  I think the multiple printouts that you see in the logs correspond to many subprocesses.

teknico: Right, and indeed it does not help. I looked again at the code and cannot find from where all the calls to `lib.randomseed` come, so I give this part up.

PR #1178, opened 2017-07-18
Return nil instead of asserting when a header is too big to overlay.

ichung:
If a header cannot be overlaid onto a piece of memory because it is too large, return nil instead of asserting.  This fits with the expectation of e.g. `Datagram:parse_match` that protocol header classes will return nil if they fail to parse a header.


alexandergall:
The base `new_from_mem` method from the `header` class has always had this `assert` and would never return `nil`, but extensions of `new_from_mem` from derived classes do return `nil` in some cases, e.g. the `gre` class when the header contains unsupported features. Historically, this is why the check for `nil` is present in `parse_match`, but it is also clearly wrong to distinguish the cases when calling the base method or an extended method.

So, I think this change should go forward, but please also fix the code where the check for `nil` is missing (in particular, in extended methods like that from the `gre` class).

alexandergall:
This looks good (wrt `lib.protocol`)

ichung:
@alexandergall These remaining commits assert pretty much every use of new_from_mem in snabb.  Most of the changes in 22fa8c8 in particular are kinda "unnecessary" since an error would be caused immediately by calling a function on the returned object, but I included that commit for completeness.  I believe the state after 20a1f19 contains all of the "necessary" assertions and checks.

alexandergall:
22fa8c8: In some instances, the code might have been save already due to previous consistency checks but asserts are cheap and probably more robust in case such pre-conditions are no longer met in future changes.
I believe this PR is ready to be merged.

lukego:
Merged via #1203.

PR #1185, opened 2017-07-27
[sketch] Share links between processes, balance packets using global freelist

eugeneia:
This is a proof of concept implementation of the ideas described in #772, using the spin-lock implementation of said PR.

To summarize, plain old links are allocated in shared memory and shared between snabb processes to interchange packets. This works because `struct link` is an inherently MP-safe single-producer/single-consumer queue. Packets are fed back into starved processes via a global freelist protected by a spin-lock.

We expect to be able to lift the performance to a practical level by optimizing `struct link` for multi-threaded using techniques borrowed from the MCRingBuffer described in the paper linked in #772.

This PR can be tested like so (only lightly tested so far):

```
$ sudo ./snabb snsh linkshare_test.lua
link report:
          11,679,306 sent on group/test.link (loss rate: 0%)
worker source: pid=21814 alive=true status=nil
1.1679306 Mpps
```

Notable issues regarding the implementation include:

- this PoC adds silly `attach_input`/`attach_output` functions to connect apps to inter-process links that should be replaced by a nicer extension to the configuration language
- apps connected to inter-process links need some special attention in `compute_breathe_order`, this PoC reverts the recent topological pull order changes instead
- the link structure currently includes pointers to its counters, this needlessly complicates the implementation (this PoC embeds the counters directly) and needs to be cleaned up

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -385,73 +385,22 @@ function apply_config_actions (actions)
       assert(ops[name], name)(unpack(args))
    end
 
-   compute_breathe_order ()
-end
-
--- Sort the NODES topologically according to SUCCESSORS via
--- reverse-post-order numbering.  The sort starts with ENTRIES.  This
--- implementation is recursive; we should change it to be iterative
--- instead.
-function tsort (nodes, entries, successors)
-   local visited = {}
-   local post_order = {}
-   local maybe_visit
-   local function visit(node)
-      visited[node] = true
-      for succ,_ in pairs(successors[node]) do maybe_visit(succ) end
-      table.insert(post_order, node)
-   end
-   function maybe_visit(node)
-      if not visited[node] then visit(node) end
-   end
-   for node,_ in pairs(entries) do maybe_visit(node) end
-   for node,_ in pairs(nodes) do maybe_visit(node) end
-   local ret = {}
-   while #post_order > 0 do table.insert(ret, table.remove(post_order)) end
-   return ret
+   compute_breathe_order()
 end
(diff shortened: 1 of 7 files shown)


eugeneia:
The previous results were actually bogus, the engine pacing just got confused by the changes and the processes would mostly idle, hence `busywait` affects the results a lot:

```
$ sudo ./snabb snsh linkshare_test.lua
link report:
         127,394,265 sent on group/test.link (loss rate: 0%)
worker source: pid=10443 alive=true status=nil
12.7394313 Mpps
```

eugeneia:
Today I experimented with using the ring buffer design from #1186 for `core.link`. The good is that this change (21410c9) doesn’t affect snabbmark basic1 results much, on the other hand it doesn’t improve the inter-process performance as much as I hoped, for reasons yet unknown to me.

```
$ sudo ./snabb snsh linkshare_test.lua
link report:
         154,885,679 sent on group/test.link (loss rate: 0%)
worker source: pid=11401 alive=true status=nil
15.4885729 Mpps
```

eugeneia:
By the way, I like this design more and more (as opposed to the dedicated inter-process links from #1186). In theory it should be even faster because there are less links in total, and the packet balancing is done in bulk instead of for every transmitted packet individually.

So actually I expect(ed) performance >= #1186. Which means either I made some errors while copying the ring buffer design or there is something else going on, both can probably be spotted using PMU analsysis?

eugeneia:
Closing as superseded by #1288.

PR #1186, opened 2017-07-27
[sketch] Specialized inter-process links

eugeneia:
This is a proof of concept implementation of specialized inter-process links with dedicated transmitter/receiver apps. This design is mostly similar to @javierguerragiraldez’ from #601. The underlying queue is based on the MCRingBuffer design described in the paper linked in #772. To avoid processes starving of packets, for each packet transmitted via an inter-process link one packet is given back in exchange, and subsequently freed by the transmitting process.

We expect to be able to apply the ring buffer design from this PoC to the design described in #1185, and this way hope to match the performance of this PR.

This PR can be tested like so (only lightly tested so far):

```
$ sudo ./snabb snsh apps/interlink/test.lua
link report:
         354,941,034 sent on rx.output -> sink.input (loss rate: 0%)
worker source: pid=15459 alive=true status=0
35.4941034 Mpps
```

The most interesting bit here is the special link implementation in `lib/interlink.lua`. The major difference when compared to `struct link` is that it features producer/consumer local state variables to eliminate false sharing. To that end this implementation requires explicit calls to publish the updated state after reading/writing elements to the producer/consumer.


wingo:
Does #1288 completely supersede this PR?

eugeneia:
Yep, I guess this and #1185 can be closed now i favor of #1228.

PR #1188, opened 2017-07-27
core.app: add a post_link() app method

alexandergall:
The existing link() method is called for all add/remove actions per
link. In some cases, it is convenient or even necessary to let an app
perform actions after link configuration is completed but before the
app is run.  The method is only invoked when at least one link was
changed (added or removed).

### DIFF
--- a/src/apps/config/action_codec.lua
+++ b/src/apps/config/action_codec.lua
@@ -10,8 +10,8 @@ local binary = require("lib.yang.binary")
 local shm = require("core.shm")
 
 local action_names = { 'unlink_output', 'unlink_input', 'free_link',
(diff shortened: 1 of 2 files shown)


wingo:
Looks about right but also needs support in `apps/config`.

alexandergall:
Ok. Unfortunately, `apps/config` is terra incognita for me (still living in the stone age of app configuration).

eugeneia:
@wingo Could I possibly test/merge this as-is? Support in apps/config (which I believe is in flux) could be added later on. I would add a note to https://github.com/snabbco/snabb/blob/master/src/apps/config/README.md explaining that `post_link` is not yet supported?

Rationale: this seems already useful for downstream projects that don’t use apps/config

PR #1189, opened 2017-07-28
Tap: support MTU, augment stats

alexandergall:
Support setting of the MTU on the tap device.  This is a bit tricky,
because the kernel only allows manipulation of the L3 (IP) MTU, while
Snabb considers the tap device to be L2 and includes the L2 header in
the MTU.  Various configuration options are introduces to work around
this issue in a flexible manner.

The app can create an ephemeral tap device or attach to a persistent
device managed outside of the Snabb process.  This was already
possible but this commit makes it clear in the README and also makes
sure that an ephemeral device's status is set to "up".

The interface's operational status is monitored and exposed as a stats
counter.


lukego:
Nice improvement! Merged via #1203.

PR #1190, opened 2017-07-28
intel_mp: move shm initialization to constructor

alexandergall:
Initialisation of the shm frame needs to be done in the constructor
method to avoid being overwritten by the frame created in core.app.

Add a copy of the mtu to the device-specific stats table, add a
counter to record the speed of the interface.

### DIFF
--- a/src/apps/intel_mp/intel_mp.lua
+++ b/src/apps/intel_mp/intel_mp.lua
@@ -260,10 +260,6 @@ Intel = {
       master_stats = {default=true},
       run_stats = {default=false}
    },
-   shm = {
-      mtu    = {counter},
-      txdrop = {counter}
-   }
 }
 Intel1g = setmetatable({}, {__index = Intel })
 Intel82599 = setmetatable({}, {__index = Intel})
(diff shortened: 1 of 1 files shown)


eugeneia:
> Initialisation of the shm frame needs to be done in the constructor
method to avoid being overwritten by the frame created in core.app.

Could you elaborate on that? I feel like I should understand this, but I don’t. `core.app` calls `shm.create_frame` only once after calling the `new` method of an app, so how does it get overwritten?

```
-bash-4.3$ ./snabb snsh -i
Snabb> Foo = { shm = "foo" }
Snabb> function Foo:new () return setmetatable({}, {__index=Foo}) end
Snabb> f = Foo:new()
Snabb> =Foo.shm
foo
Snabb> =f.shm
foo
Snabb> f.shm = "bar"
Snabb> =Foo.shm
foo
Snabb> =f.shm
bar
```

alexandergall:
When the frame is created, the counters are set to their default values, hence the original
```
counter.set(self.shm.mtu, self.mtu)
```
in `new()` has no effect. The patch sets the default value instead.

alexandergall:
Can we please proceed with this PR?

alexandergall:
Thanks. So can someone please merge?

PR #1191, opened 2017-07-28
apps.vlan: improve performance, configurable encapsulation

alexandergall:
The performance of the VlanMux app (apps.vlan.vlan) is improved by

  - avoiding use of pairs()
  - avoiding string composition
  - serialisation of loops to avoid branches

The forwarding strategy during de-multiplexing has been changed to
forward all packets from the trunk links and accepting drops on vlan
links which are full (instead of only forwarding as many packets as
the fullest vlan link can accept).

Support for different encapsulation schemes has been added: dot1q,
dot1ad and an explicit ethertype to cover non-standard encapsulations.

### DIFF
--- a/src/apps/vlan/README.md
+++ b/src/apps/vlan/README.md
@@ -6,49 +6,83 @@ can multiplex and demultiplex packets to different output ports based on tag.
 
 ## Tagger (apps.vlan.vlan)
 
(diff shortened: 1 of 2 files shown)


lukego:
Merged via #1203. The SnabbBot test failure is almost certainly a false-negative, showing a packetblaster regression when this branch is changing unrelated code.

PR #1192, opened 2017-07-28
Iftable_mib update

alexandergall:
Bug fixes and enhancements of `lib.ipc.shmem.iftable_mib`. The `l2vpn` program is still the only user of this code and it should eventually be re-written to use the standard `core.shm` API.


lukego:
Merged into #1203 and en route to master.

PR #1193, opened 2017-07-28
nd_light update

alexandergall:
Add two new options

   * **remote_mac** for static MAC address of `next_hop`
   * **quiet** to reduce verbosity

Add a `reconfig` method, fix a bug in NS processing.


lukego:
Merged via #1203.

PR #1195, opened 2017-07-31
Integrate ARP into IPFIX exporter

wingo:
This branch merges in the new ARP app to the IPFIX exporter, so that you don't have to specify MAC addresses when running the VNF.  It also adds some minimal `ipfix probe` documentation.

### DIFF
--- a/src/program/ipfix/probe/probe.lua
+++ b/src/program/ipfix/probe/probe.lua
@@ -38,7 +38,6 @@ local long_opts = {
    duration = "D",
    port = "p",
    transport = 1,
-   stats = "s",
    ["host-ip"] = "a",
    ["input-type"] = "i",
    ["output-type"] = "o",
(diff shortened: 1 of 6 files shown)


src/program/ipfix/probe/probe.lua
```diff
@@ -95,10 +91,6 @@ function run (args)
       c = function (arg)
          collector_ip = arg
       end,
-      -- TODO: this should probably be superceded by using ARP
-      M = function (arg)
-         collector_mac = arg
-      end,
```
dpino: These options should be removed too from the list of accepted options. Line 115:

```lua
args = lib.dogetopt(args, opt, "hsD:i:o:p:m:a:c:M:", long_opts)
```

dpino: And also removing in the list of long options:

```lua
stats = "s",
```

PR #1196, opened 2017-08-01
Add VMDq support to intel_mp driver

teknico:
This PR adds support for VMDq mode to the intel_mp driver, to bring it closer to feature parity with the intel10g one.

Specifically, this PR code supports VMDq mode with RSS (64 VM pools with 2 RSS queues each) with MAC/VLAN filtering on the Intel 82599. Some of the code is ported from the intel10g driver. For now, the driver just rejects VMDq mode on i210 and i350 NICs.


teknico:
Superseded by #1197.

PR #1199, opened 2017-08-03
Switch all code to intel_mp

teknico:
Switch all the various Snabb programs to use the `intel_mp` driver instead of `intel10g`.

Based on #1197, supersedes https://github.com/Igalia/snabb/pull/841 .


wingo:
Nice!  Any perf notes?

src/program/lwaftr/generate_binding_table/generate_binding_table.lua
```diff
@@ -43,8 +43,9 @@ end
 
 local function softwire_entry(v4addr, psid_len, b4)
    if tonumber(v4addr) then v4addr = to_ipv4_string(v4addr) end
-   return ("  softwire { ipv4 %s; psid %d; b4-ipv6 %s; }"):format(
-      v4addr, psid_len, b4)
+   local softwire = "  softwire { ipv4 %s; psid %d; b4-ipv6 %s; br-address %s;"
+   softwire = softwire .. " port-set { psid-length %d; }}"
+   return softwire:format(v4addr, psid_len, b4, br_address, port_set.psid_len)
```
wingo: Did you mean to commit this?  This change applies only to lwaftr, not Snabb:wingo-next.

teknico: Ah, I solved a conflict in the wrong direction, sorry. I'll revert.

lukego:
I setup a Hydra job running on `lugano` servers (i.e. with 82599 NICs) comparing the performance of `master`, `next`, `wingo-next`, and this branch (labelled `intel-mp`). This should help to catch any regressions related to the driver e.g. some obscure DMA prefetch setting.

The first report will be [on this build](https://hydra.snabb.co/build/1696040) and new builds will start when a branch is updated [on the evaluations page](https://hydra.snabb.co/jobset/snabb-new-tests/benchmarks-lugano-small-20x#tabs-evaluations).

This is a first step at least, we could run a different benchmarking campaign if we think of one that makes more sense.

lukego:
This branch looks really wonderful to me :). The `intel_mp` driver has really come up well! The initial Hydra test results with hardware NICs look fine too, I'll run a few more to be sure.

I am thinking through the "user experience" when other Snabb hackers pull this branch into their own tree. They probably have their own "downstream" applications that are using `intel10g`. Is pulling this branch going to break anything? what action is required to update their programs? will this catch them by surprise somehow? I think we should have a solid understanding of this and I am thinking about it.

One specific Q: what is the reason for 7d1397714b6f70009b897d2afcce8f9c775e147b?

lukego:
So! I am thinking we want to reach the point where we have one unified Intel driver that supports their many and varied cards. This branch is a big step in that direction because the `intel_mp` driver is The Right Thing and because it fully replaces the existing uses of the "legacy" `intel10g` driver.

So what remains to reach the final goal? when is it safe to delete `intel10g.lua`? and what are the risks of causing frustration to users and slipping backwards on the agenda?

The challenge that I see is to smoothly support the "downstream" code in people's applications. If somebody has forked Snabb and developed an application then they are probably using `intel10g`. Pulling this branch won't change anything - their code will still use the legacy driver - and if we pull their code upstream in the future then maybe we end up maintaining the legacy driver forever.

So it seems like some migration strategy is needed. I am tempted to suggest that we write down exactly how to migrate from the old driver to the new one and then print a loud "deprecation" warning when an application uses the legacy driver. Then we could hopefully remove `intel10g` in some reasonable timeframe e.g. the end of the year.

More things to consider:
- Should we rename `intel_mp` to just `intel` to position it as the default driver?
- Should we make migration automatic by having the `intel_app` API automatically forward to the new driver?
- Is there some immediate risk of breakage for downstream projects from merging this PR? For example, if people are using the PCI API to automatically detect which driver to use based on hardware, and this branch will switch from `intel10g` to `intel_mp`, then will this be transparent to the user? (I'm thinking the interfaces to the apps may be too different e.g. links named rx/tx on one driver and input/output on another?)

Sorry to slow down this fantastic work with such long and rambling discussions. I am just trying to put myself into the shoes of application developers who expect to be able to pull the master branch and get maximum new good stuff with minimum breakage that they have to stop and fix.

wingo:
Honestly as regards external users -- I think Snabb should offer no stability guarantees of any kind for out-of-tree code.  Correct and thorough-enough change logs yes, stability no.  In that regard I think it's an anti-goal to keep API or ABI stability, as it's cost that actively prevents benefits!  API includes but is not limited to app names, module names, link names, FFI type names, app configurations, and module exports.

Even for in-tree code I think we should offer only a best-effort approach.  If e.g. `intel_mp` changes name to `intel`, then it's easy to just rename all uses in the tree, no problem.  But if there's a bigger change and no one is maintaining an in-tree app or VNF or whatever, I think we should have an expectation that we can remove that code after some period of non-maintenance.  Snabb will change and that's a good thing; the challenge is how to enable change.

If a downstream out-of-tree user doesn't have time to update, that's fine, they don't have to update!  They have branched from some release, and when they are ready to update, they can as part of their internal processes of merging a new Snabb release.  System working as expected.

For me the concerns in this PR are not about compatibility from an API POV, they are about performance, reliability, and feature parity.  Is this driver rock-solid?  We don't know yet.  For *that* reason I would keep `intel10g` around for another release or three, in case people need to compare.  But then assuming all goes well I would delete it, sure, as it is providing no more benefit.  People will find out that the file is gone when their `require('apps.intel.intel_app')` starts throwing exceptions :)

tobyriddell:
Just a thought, but is 'intel' the correct name? How about 'intel_nic' so that it doesn't name-clash with the as-yet-unwritten code that takes a packet and passes it through the CPU-attached FPGA on the upcoming Cascade Lake series of chips (having previously programmed said FPGA to do deep packet inspection)?

lukego:
@wingo Thanks for the discussion :). I do think it makes sense to pull this branch "as is" and let the two Intel drivers co-exist for a while, as you say. Then we can reasonably hope that applications will gradually migrate to `intel_mp` over the coming months and then if that goes smoothly then we can remove `intel10g`. If there are hiccups then the lifetimes of the drivers can overlap a bit more as needed.

I think that downstream application developers will really appreciate all the work that you guys have done to make this code compatible. It's fantastic that you all made the effort to update the in-tree code. I surely appreciated that :) so thank you.

Like you say, the `master` branch needs to be focused on moving ahead and continuously improving. It also needs to be predictable enough that people will actually pull from it regularly :-) but this change seems nice in that respect & probably any extra compatibility hacks would end up backfiring (people not really knowing which driver they were using, etc.)

I am negotiating with Hydra to run a somewhat larger benchmark [here](https://hydra.snabb.co/build/1871556). I had manually aborted a few evaluations earlier to manage the test backlog and I'm working out how to tidy up that mess now :).

takikawa:
Whoops, I must not have notifications turned on correctly since I didn't see the last few comments until now. I just pushed some additional commits to fix some bugs, add tests, and resolve merge conflicts.

Thanks for the comments @lukego! In terms of what users should expect, we made an effort to support the documented interface of `intel10g` for `intel_mp` for the most part. One difference is the input and output link names (which you've already noticed). That might be a hurdle in switching over though, so it might make sense to use `intel10g`'s names.

BTW, commit 7d13977 was there to avoid a specific bug that happened on app reconfiguration. I think we may have fixed the underlying cause though, so we may be able to revert it now. I'll try it out and see.
 
I think it would be helpful to make sure the API for setting VMDq pools and Rx/Tx queues is the right one. We made the `rxq` and `txq` config parameters default to `0` (for selecting RSS queue numbers) and VMDq pools are selected automatically like `intel10g`. That way, the default behavior resembles `intel10g` with no arguments needed.

In case someone needs just RSS with no VMDq, providing explicit `rxq` and `txq` lets you have serve multiple queues (setting a default doesn't let you explicitly disable Rx/Tx on a specific app instance though, maybe we should allow this via a `"disable"` config value). For both VMDq + RSS, you can set `rxq` and `txq` from 0 to 3 for 4-way RSS but you probably need to explicitly choose pool numbers to ensure all the RSS queues are assigned to the same pool (the automatic selection will put them all on different pools). That last one does seem a little clunky perhaps.

If that API seems ok, then I think we're good. But we can also change it now if there's anything that looks off.

Also I agree with @wingo on needing to figure out the reliability. We've of course been testing the driver and writing more tests, but real-world testing is bound to find more issues. :)

lukego:
There is a performance regression on the snabbnfv iperf benchmark when running on `lugano` servers i.e. with a hardware 82599 NIC. This is a bit tricky to see because the R code is [erroring](https://hydra.snabb.co/build/1934699) on the Hydra build, I think that I need to update it somehow (could be that it is not prepared for how _few_ config permutations this test covers.)

Meanwhile here is a quick CDF based on the [CSV data](https://hydra.snabb.co/build/1934426) and an ad-hoc R script:

```R
library(ggplot2)
data <- read.csv("https://hydra.snabb.co/build/1934426/download/1/bench.csv")
iperf <- filter(data, benchmark == "iperf")
ggplot(iperf, aes(color=snabb)) + 
  stat_ecdf(aes(x=score)) +
  scale_y_continuous(labels = scales::percent) +
  labs(title = "iperf benchmark", y = "percent of scores below threshold")
```

iperf-ecdf

Have to think a little about a reasonable strategy for reproducing this for development purposes and for determining whether the difference is due to CPU/NIC hardware issues or JIT issues...

lukego:
Here is the [standard report](http://rpubs.com/lukego/302231) that I generated locally in RStudio on my laptop. Have not looked closely yet but I do notice there are quite a few failures on `intel-mp` branch and we should be able to browse the logs for those on Hydra. (Could also be that they are failures due to timeout e.g. JIT issue caused the test to run too slow and it was killed.)

I gotsta get CI building these updated reports...

takikawa:
@lukego Looking through some of the log files, it looks like something with `vhost_user` might be failing on some of the iperf tests. The snabb log has `vhost_user: Connection went down: vhost_A.sock` and qemu log has entries like

```
qemu-system-x86_64: unable to start vhost net: 1: falling back on userspace virtio
kvm_mem_ioeventfd_add: error adding ioeventfd: File exists
```

The l2fwd failure cases are puzzling, since the logs don't seem very different except that in the snabb logs it just stops transmitting packets after a bit.

I haven't been able to reproduce any failures by running the `selftest.sh bench` (for iperf) and `dpdk_bench.sh` (for l2fwd) on our Igalia snabb machines though. If you have any tips on reproducing it, I'd be happy to try things.

takikawa:
I wonder if the qemu error is related to this issue? https://github.com/snabbco/snabb/issues/1103 The error message appears to be the same as the one reported there.

lukego:
@takikawa Today I want to try to make this test setup a bit simpler: remove "known failing" cases from the CI jobs and make the tests easier to reproduce locally on arbitrary Linux boxes. Can chat on Slack if you like.

Just now for example if you like at the "Success and failure" graph you can see that all branches are failing on tests with QEMU versions 2.7.1 and 2.8.0 (because Snabb is not compatible today.) You can also see that with other QEMU versions the master/next branches have 100% success while intel-mp is more spotty. But, it's no fun to try and differentiate between "expected" and "unexpected" failures in the Hydra UI and so life would be simpler if we remove the unsupported QEMUs from these Hydra tests.

lukego:
@takikawa Just as a baby-step I updated the CI jobs to skip every test that is expected to fail (i.e. incompatible versions of QEMU/DPDK.) So now any failures on Hydra should represent real problems (whether crash or timeout.) The first updated result should be [Evaluation 6918](https://hydra.snabb.co/eval/6918).

I have been trying to hack up a little Nix script that would quickly and easily run a customized benchmark on an arbitrary Linux box, e.g. Ubuntu with or without a hardware NIC, but I didn't manage to get that working yet. Have to think about how to tease out the basic test cases from the overall "run a matrix of tests on a Snabb Lab NixOS server" framework.

takikawa:
I just pushed a commit that will hopefully address some of the iperf failures. Not totally sure it will fix it, but I guess we will see when the Hydra results are up.

takikawa:
I tried merging `next` into this branch to see if maybe some of the performance differences in iperf was due to other changes between the branches. It looks like that may have been the case. Here's a graph using the script @lukego posted above:

iperf-2017-09-07

which I found somewhat surprising, since there seemed to still be a performance difference when I tried this comparison on the Igalia machine. There's also that gap near score 13 that could be a real issue though.

eugeneia:
Is this PR still relevant given #1237 was merged? I guess not so I am cautiously closing this PR.

PR #1201, opened 2017-08-04
Add an integration test for ipfix probe using an ipfix collector

takikawa:
This PR fixes a few minor issues with `ipfix probe` and also adds an integration test that uses `nfcapd` from the `nfdump` suite to test the app with a real IPFIX collector program. It checks that the collector sees the right number of flows & packets from a short run of the probe & collector on test traffic.

The modifications include:

  * change ARP app to not read any incoming packets until the next hop MAC is resolved (to avoid losing template packets that the probe sends)
  * allow `tap` devices in `ipfix probe`

BTW: it doesn't matter for this test, but I noticed that if the `ipfix` app gets support for exporting over TCP/SCTP then it may need another input link for incoming responses from the collector.

Also, should the test script be a `selftest.sh` so it's automatically run? Wasn't sure if the dependency on `nfdump` would be a problem there.

### DIFF
--- a/src/apps/ipv4/arp.lua
+++ b/src/apps/ipv4/arp.lua
@@ -196,17 +196,15 @@ function ARP:push()
       end
    end
 
-   for _ = 1, link.nreadable(inorth) do
-      local p = receive(inorth)
-      if not self.next_mac then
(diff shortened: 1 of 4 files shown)


wingo:
Changes look great, thanks!

PR #1203, opened 2017-08-09
staging branch for merges into next [wip]

lukego:
This short-lived integration branch contains changes that I have merged for testing before pushing to `next`. The idea is to buffer changes here for some hours/days for some preliminary testing.


lukego:
Hm. I pushed a recent commit from this branch (c3ca3ae) to Hydra as `lukego-optimize` and it is showing 100% failure rate introduces on the iperf-with-pcap-filter benchmark. Have to isolate which change is responsible for this. It's tricky because this should be covered by SnabbBot but does not seem to be running (Github is showing a green tick but that seems to be only for generating the manual and not running the tests.)

I'll look for the root cause.

PR #1206, opened 2017-08-11
virtual_ether_mux: Treat vlan=0 as vlan=nil

lukego:
Commit da4f98b09caf015e800fb23b8ea0ef4a27bea85a makes apps.vlan validate the vlan ID argument. This correctly rejects the value zero as invalid (not allowed on the wire in 802.1Q standard.) However, the snabbnfv application does permit vlan=0 in its configuration file and this is used in various test vector files. This causes test failures.

There are two obvious solutions. The first is to update all configurations to never use the invalid vlan=0. The second is to interpret vlan=0 as meaning "no VLAN." This commit implements the second one.

This change will be pulled directly onto #1203 as a fix to test failures.

### DIFF
--- a/src/lib/io/virtual_ether_mux.lua
+++ b/src/lib/io/virtual_ether_mux.lua
@@ -54,12 +54,15 @@ function configure (c, ports, io)
          local name = port_name(port)
          local Switch_link = Switch.."."..name
          local Port_tx, Port_rx = Switch_link, Switch_link
-         if port.vlan then
+         local vlan = port.vlan
+         -- Backwards compat: VLAN 0 is accepted by nfvconfig but is not valid.
+         if vlan == 0 then vlan = nil end
+         if vlan then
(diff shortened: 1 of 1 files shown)


mwiget:
vlan=0 is valid and used to priority tag the traffic. If you google it, you'll find more info's about it:

https://www.cisco.com/c/en/us/td/docs/switches/connectedgrid/cg-switch-sw-master/software/configuration/guide/vlan0/b_vlan_0.html#con_1055352

So a switch will still treat the traffic as untagged, but it processes the 802.1p bits.

lukego:
@mwiget Thanks for the tip. Maybe we should loosen the assertion in `apps.vlan` to allow vlan=0 and revert this change.

What do you think @alexandergall?

lukego:
I reverted this change and updated the `vlan` app to accept id=0 in ebd0aa9174d67a25d1996bb7c715d95772c715ea on my integration branch #1203. cc @alexandergall no action required.

PR #1216, opened 2017-08-28
Stabilize breathe order

wingo:
This PR stabilizes the run-to-run breathe order for graphs that have cycles.  It is also pushed to `wingo/optimize` so that Hydra will have a go at it.


lukego:
@wingo want me to pull this now?

wingo:
@lukego Sure.  We have been running this on `lwaftr` and it's the only patch to `src/core/app.lua` (i.e. no followups), so I am pretty confident.

PR #1217, opened 2017-08-28
Allow LPM self-tests to complete even if no CPU affinity

wingo:


wingo:
Aaah, I thought I had pushed this one somewhere!  IN the meantime, dug it out and merged to #1231 via #1232.  Closing this one.

PR #1218, opened 2017-08-28
Fix Unix socket EOF handling

benagricola:
Hi @lukego, It looks like during the unix socket rewrite to remove `select()` a couple of months ago, the check for `local bytes = S.read()` returning 0 (EOF) was removed, and changed to `if not bytes`, which only catches errors, as ljsyscall converts -1 to a nil return value, and 0 is true.

This means EOF is not caught, so when the socket is terminated from the remote side, the app goes into a loop returning empty allocated packets.

This PR keeps handling for errors (`not bytes`) which includes EAGAIN and re-adds the `bytes == 0` check for EOF - resetting the socket.


lukego:
Thanks Ben! Merged. cc @alexandergall who I know has been working on related apps recently and may need to cover this case too.

PR #1220, opened 2017-09-04
IPsec ESP interoperability test with Linux

eugeneia:
This adds `apps/ipsec/selftest.sh` which boots a Linux guest configured for ESP in transport mode, and verifies that the ESP apps can successfully exchange packets with the Linux IPsec stack.

This test depends on some packages being installed in the guest, which are added in https://github.com/snabblab/snabblab-nixos/pull/99


lukego:
@eugeneia Could you get SnabbBot to test this please?

eugeneia:
@lukego Right, the problem here is that SnabbBot doesn’t have the environment required for `lib/ipsec/selftest.sh`, and that’s why it fails. I can think of some solutions:

 - skip the test when `SNABB_IPSEC_DONT_TEST_E2E` is set
 - add the required dependencies to the VM images in the Docker containers used by SnabbBot
 - make SnabbBot use Nix instead of Docker to test snabb

I am somewhat leaning towards the first option... burning containers for SnabbBot is somewhat annoying and I don’t see much gain here (the test is trivial to run in the lab or in hydra).

eugeneia:
@lukego alright, I went with the easy path and disabled this test for snabb bot (trusty doesn’t support extended sequence numbers, it appears). Ready to be merged!

PR #1225, opened 2017-09-13
siphash: avoid ctype diversity for hash functions

alexandergall:
When a hash table is re-sized automatically, a new hash function is
generated and cast to a generic function pointer type. However, due to
the semantics of ffi.typeof(), a new ctype ID is allocated each time,
which causes sub-optimal transitions to side-traces in already
compiled traces.  This commit makes sure that all hash function
pointers have the same ctype ID to avoid this problem.


lukego:
Merged on #1251. Thanks!

PR #1228, opened 2017-09-17
Set mask.size equals to MAX_NUMNODES if lower

### LINKED ISSUE #1200: NUMA get_mempolicy failing on Ubuntu 4.10 kernel
Try this:

```
sudo ./snabb snsh -e 'assert(require("syscall").get_mempolicy())'
```

Works locally on a GuixSD system with Linux 4.10.2 and just one NUMA node, works on a NixOS Linux 3.18.26 machine with two NUMA nodes, doesn't work on Ubuntu 17.04 with Linux 4.10-xx on any number of NUMA nodes.  The error is "invalid argument".  We need to fix this!

dpino:
Fixes #1200.

`get_mempolicy` syscall returns an EINVAL if argument maxnode is lower than the value of MAX_NUMNODES defined in the system:

```
int get_mempolicy(int *mode, unsigned long *nodemask,
                         unsigned long maxnode, void *addr,
                         unsigned long flags);
```

In ljsyscall `S.get_mempolicy` the argument maxnode is mask.size. What the patch does is to check the value of mask.size is lower than MAX_NUMNODES and in that case set mask.size to MAX_NUMNODES.

That fixes the call to `get_mempolicy` in Ubuntu 17.04. I think to properly know if this fix works, the right thing to do is to run the lwAFTR on such system (I cannot since I lack a 10Gb NIC).

### DIFF
--- a/lib/ljsyscall/syscall/linux/syscalls.lua
+++ b/lib/ljsyscall/syscall/linux/syscalls.lua
@@ -457,29 +457,42 @@ function S.sched_setaffinity(pid, mask, len) -- note len last as rarely used
   return retbool(C.sched_setaffinity(pid or 0, len or s.cpu_set, mktype(t.cpu_set, mask)))
 end
 
-local function get_max_numnodes()
-  local name = "/proc/self/status"
-  local fd, err = S.open(name, c.O.RDONLY)
-  if not fd then return nil, err end
-  local content, err = S.read(fd, nil, 8192)
-  fd:close()
-  for line in content:gmatch("[^\n]+") do
-    if line:match("^Mems_allowed:") then
-      line = line:gsub("^Mems_allowed:%s+", "")
-      return math.floor(((#line+1)/9)*32)
-    end
-  end
+local function get_maxnumnodes()
+   local function filesize(fd)
+      local size = 0
+      local bufsz = 1024
+      local buf = ffi.new("uint8_t[?]", bufsz)
+      while true do
+         local len, err = S.read(fd, buf, bufsz)
+         size = size + len
+         if len ~= bufsz then break end
+      end
+      fd:seek(0, "set")
+      return size
+   end
+   local function read(filename)
+      local fd, errno = S.open(filename, 0)
+      if not fd then error(errno) end
+      local content, errno = S.read(fd, nil, filesize(fd))
+      if not content then error(errno) end
+      fd:close()
+      return content
+   end
+   local content = read("/proc/self/status")
+   for line in content:gmatch("[^\n]+") do
+     if line:match("^Mems_allowed:") then
+       line = line:gsub("^Mems_allowed:%s+", "")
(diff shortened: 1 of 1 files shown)


lib/ljsyscall/syscall/linux/syscalls.lua
```diff
@@ -457,9 +457,28 @@ function S.sched_setaffinity(pid, mask, len) -- note len last as rarely used
   return retbool(C.sched_setaffinity(pid or 0, len or s.cpu_set, mktype(t.cpu_set, mask)))
 end
 
+local function get_max_numnodes()
+  local name = "/proc/self/status"
+  local fd, err = S.open(name, c.O.RDONLY)
```
wingo: Here we would need to return a default value, but then below we are only reading 8192 bytes, whereas for 1024 nodes that would already take 1024*9=9216 bytes.  So this won't scale arbitrarily.  I think it could be OK to instead of calling get_max_numnodes, just always use 4096 or something large.  Avoids many potential problems.  WDYT?

dpino: OK with me if it will always work.

dpino:
Setting a fixed number of max numnodes didn't work. In my testing case, where maxnumnodes is 1024, if I set a number larger than that I got a segfault error. I rewrote the `get_maxnumodes` so it doesn't depend on a fixed file size.

lib/ljsyscall/syscall/linux/syscalls.lua
```diff
@@ -457,10 +457,42 @@ function S.sched_setaffinity(pid, mask, len) -- note len last as rarely used
...
+      local buf = ffi.new("uint8_t[?]", bufsz)
+      while true do
+         local len, err = S.read(fd, buf, bufsz)
+         size = size + len
+         if len ~= bufsz then break end
+      end
+      fd:seek(0, "set")
+      return size
+   end
+   local function read(filename)
```
wingo: Can we use `core.lib.readfile` instead?

lib/ljsyscall/syscall/linux/syscalls.lua
```diff
@@ -457,10 +457,42 @@ function S.sched_setaffinity(pid, mask, len) -- note len last as rarely used
...
+      if not fd then error(errno) end
+      local content, errno = S.read(fd, nil, filesize(fd))
+      if not content then error(errno) end
+      fd:close()
+      return content
+   end
+   local content = read("/proc/self/status")
+   for line in content:gmatch("[^\n]+") do
+     if line:match("^Mems_allowed:") then
+       line = line:gsub("^Mems_allowed:%s+", "")
```
wingo: This needs a comment about what's going on here; why the 9 (I know, because NNNNNNNN,), why the 32.

wingo:
"Failure" appears to be a simple benchmark statistical anomaly unrelated to this patch; merging.  Thanks @dpino!

PR #1229, opened 2017-09-18
Implement VMDq, VLAN insert/remove for intel_mp

wingo:
This branch is based on https://github.com/snabbco/snabb/pull/1197, and adds additional `intel_mp`-only commits from https://github.com/snabbco/snabb/pull/1199.  As it's fairly well-tested in an `intel_mp` context and doesn't affect the rest of Snabb, it should be pretty safe to merge.

There are no incompatible changes.  For user-visible feature additions, see https://github.com/snabbco/snabb/compare/master...Igalia:merge-intel-mp?expand=1#diff-6abfd37084b87cfa568f82bc5d72bd0e.

PR #1231, opened 2017-09-18
Changes queue from wingo-next for v2017.10/v2017.11 release

wingo:
Contents:
 * #1229: Implement VMDq, VLAN insert/remove for intel_mp app
 * #1230: `intel_mp` app relies on `lib.parse` to supply default values
 * #1234: Driver independence
 * #1236: Driver independent lwAFTR
 * #1235: Adjust ctable selftest to avoid assertion failure


wingo:
Yay the tests pass, after fixing the bugs that I introduced!  I think this branch is ready to go!

PR #1232, opened 2017-09-18
Allow LPM self-tests to complete even if no CPU affinity

wingo:


wingo:
This patch went into the lwaftr via https://github.com/Igalia/snabb/pull/927, and allows tests to complete when there is no CPU affinity.

wingo:
Applied to `wingo-next` in #1231.

PR #1234, opened 2017-09-19
Driver independence

wingo:
This PR is a reworking of #1199 which tries to first adapt network functions to be agnostic as far as whether they are using the old intel10g driver or the new intel_mp driver.  It does this by adding `rx` and `tx` fields to the PCI driver info structure, letting network functions know what link names to use.  It renames the `ndescriptors` property of `intel_mp` to `ring_buffer_size`, and adds a corresponding initarg to the old intel10g driver.  Finally it enables VMDq whereever network functions were setting a MAC address; see #749.

I have a few more unpushed commits to make corresponding changes to the lwAFTR and the changes that @teknico and @takikawa made to the packetblaster / firehose, and the commit that actually changes over the `pci.device_info()` function to prefer `intel_mp`, but every patch in this series should work and cause no regressions, so it should be a safe merge on all downstreams of `wingo-next` (including `lwaftr`).

### DIFF
--- a/src/apps/intel_mp/intel_mp.lua
+++ b/src/apps/intel_mp/intel_mp.lua
@@ -482,6 +482,7 @@ function Intel:init_rx_q ()
       self.r.RXCTRL:set(bits{ RXEN=0 })
       self.r.DCA_RXCTRL:clr(bits{RxCTRL=12})
       if self.vmdq then
+         -- enable packet reception for this pool/VF (4.6.10.1.4)
          self.r.PFVFRE[math.floor(self.poolnum/32)]:set(bits{VFRE=self.poolnum%32})
       end
    elseif self.driver == "Intel1g" then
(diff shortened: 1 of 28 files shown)


src/apps/intel_mp/intel_mp.lua
```diff
@@ -569,12 +570,14 @@ function Intel:push ()
    if not self.txq then return end
    local li = self.input["input"]
    if li == nil then return end
---   assert(li, "intel_mp:push: no input link")
 
    while not empty(li) and self:ringnext(self.tdt) ~= self.tdh do
       local p = receive(li)
-      -- NB: see comment in intel10g for why this is commented out,
-      --     the rest of the loop body goes in an else branch
+      -- NB: the comment below is taken from intel_mp.lua, which disables
```
teknico: Should this say `intel10g` instead of `intel_mp.lua`?

wingo:
I am having trouble with the upstream bots for some reason; will PR this against `Igalia/snabb` to see what those bots think.

src/apps/intel_mp/intel_mp.lua
wingo: Yes you are right.  Will fix in a followup.  Thanks for the review!

wingo:
There is a test failure due to an `intel_mp` change that's already upstream, for intel1g cards only.  Merging this code as it's unrelated and we'll fix the intel1g problem later.

Issue #1212, opened 2017-08-21
SipHash / ctable badness


wingo:
In https://github.com/Igalia/snabb/pull/909, the first patch had a flaky CI failure in `lib.ctable`:

```
src/testlog/lib.ctable:
selftest: ctable
core/main.lua:26: 43

Stack Traceback
===============
(1) Lua function 'handler' at file 'core/main.lua:168' (best guess)
	Local variables:
	 reason = string: "core/main.lua:26: 43"
	 (*temporary) = C function: print
(2) global C function 'error'
(3) Lua global 'assert' at file 'core/main.lua:26'
	Local variables:
	 v = boolean: false
(4) Lua field 'selftest' at file 'lib/ctable.lua:637'
	Local variables:
	 bnot = C function: builtin#65
	 occupancy = number: 2e+06
	 params = table: 0x41586758  {initial_size:5000000, key_type:ctype<unsigned int [1]>, max_occupancy_rate:0.4 (more...)}
	 ctab = table: 0x410842a8  {scale:0.001164153451371, occupancy_lo:0, entries:cdata<struct 1397 *>: 0x2b9f00c00000 (more...)}
	 (for index) = number: 2
	 (for limit) = number: 2
	 (for step) = number: 1
i = number: 2
...
```
Full log here: https://gist.github.com/lwaftr-igalia/a39bdeaa48712b33a40254feb97323b8

(A flaky failure is a failure that only happens sometimes.  Appearance of a flaky failure doesn't necessarily indicate a problem in the PR in which is appears.)

This failure corresponds to a check in which a ctable is populated with the first 2 million integers as keys, and then we check that the max displacement on the table is below a certain number.  Code here: 
https://github.com/snabbco/snabb/blob/master/src/lib/ctable.lua#L604.  When we used the Jenkins hash, this value was 8.  Then when we switched to randomly-seeded SipHash, the value was often 8 or 9 and I thought, well, fine.  Of course this test has a probabilistic aspect to it but I do want to ensure that max displacement doesn't go off the rails -- otherwise the branchless lookup streamer won't work well.  (It takes N keys, hashes them to produce N hashes, streams in the M contiguous max-displacement entries for each key, then does binary search over those M entries.  Total cost N*M for memory, log(M) for search.  M is the max displacement, so if even one entry has a huge displacement for the table, it incurs a cost for all lookups.)

[...]

PR #1235, opened 2017-09-19
Adjust ctable selftest to avoid assertion failure

takikawa:
This PR should fix issue #1212. I tested it by running `status=0; while [ $status -eq 0 ]; do sudo ./snabb snsh -t lib.ctable; status=$?; done` and waiting for a failure. Without the commit, it fails in about a minute or so of waiting. With the commit, I've left it running for 20+ minutes.

I also tested it by adding an explicit `collectgarbage()` call to trigger the bug immediately.

### DIFF
--- a/src/lib/ctable.lua
+++ b/src/lib/ctable.lua
@@ -682,8 +682,12 @@ function selftest()
       end
       do
          local file = io.open(tmp, 'rb')
+         -- keep references to avoid GCing too early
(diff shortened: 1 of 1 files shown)


wingo:
Excellent investigation, thank you!  Note that we still do need to do some investigation to see how max_displacement varies with different random seed instances, but that can come later.

PR #1237, opened 2017-09-20
Use intel_mp driver instead of intel10g driver

wingo:
This PR flips the switch in `lib.hardware.pci.device_info` to prefer the intel_mp driver instead of intel10g.  It also ports over the `firehose` and `packetblaster` programs to intel_mp, also porting the `loadgen` app from intel10g.  All credit to @teknico and @takikawa for the great patches!

### DIFF
--- a/src/apps/intel_mp/intel_mp.lua
+++ b/src/apps/intel_mp/intel_mp.lua
@@ -154,6 +154,7 @@ RXCSUM      0x05000 -               RW Receive Checksum Control
 RFCTL       0x05008 -               RW Receive Filter Control Register
 RXCTRL      0x03000 -               RW Receive Control
 RXDGPC      0x02F50 -               RC DMA Good Rx Packet Counter
+TXDGPC      0x087A0 -               RC DMA Good Tx Packet Counter
 RXDSTATCTRL 0x02F40 -               RW Rx DMA Statistic Counter Control
 RUC         0x040A4 -               RC Receive Undersize Count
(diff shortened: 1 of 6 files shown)


wingo:
Rebased to fix a bug in packetblaster.

wingo:
This build failure is only `ERROR snabbnfv-iperf-1500 -> 0.828682 of 5.16 (SD: 0.709084 )`.  If I interpret this correctly, that would indicate perf of 4.28, which is 0.884 off from the mean; a bit more than a standard deviation, again if I understand it correctly.  I looked to see and it seems we don't have any pending perf patches in `lwaftr`.  So, I will push this branch to `wingo/optimize` to see what Hydra thinks of its perf.

lukego:
Thanks @wingo.

@eugeneia Perhaps we should disable the performance tests on SnabbBot since they are mostly firing on false positives? Hydra will always catch regressions a step or two before they merge to master so we have a safety net.

wingo:
FWIW @lukego I still get some value from these tests.  In particular the snabbnfv-loadgen-dpdk test on #1231 let me know that the `bench` mode of `snabbnfv` was broken with that branch, so I landed a fix.  In this case the benchmark could be an early warning perf error, but it's more of an indicator that I need to look more into Hydra results, whereas sometimes I don't bother.  So I appreciate them being there.

Whether the snabb-bot benchmarks should gate merges or not is another question :)

lukego:
Okidok. I am fine with any alternative including leaving things as they are.

PR #1240, opened 2017-10-05
Include IPV4/IPV6 in GRE inner protocol list

krawthekrow:
The protocol field for GRE should be an ether type; an implementation receiving a packet containing a Protocol Type which is not an ether type should discard the packet (RFC 2784, 2.4). GRE's inner protocol list (_ulp) should thus be the same as Ethernet's, so that GRE:upper_layer() behaves correctly.

### DIFF
--- a/src/lib/protocol/gre.lua
+++ b/src/lib/protocol/gre.lua
@@ -28,7 +28,11 @@ ffi.cdef[[
 -- Class variables
 gre._name = "gre"
 gre._ulp = {
-   class_map = { [0x6558] = "lib.protocol.ethernet" },
+   class_map = {
+                  [0x6558] = "lib.protocol.ethernet",
+                  [0x0800] = "lib.protocol.ipv4",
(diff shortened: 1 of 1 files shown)


alexandergall:
`0x6558` is a perfectly fine Ethertype (coincidentally, it's listed explicitly in the original GRE specification RFC1701), just like `0x8100` and `0x86dd` and everything else on https://www.iana.org/assignments/ieee-802-numbers/ieee-802-numbers.xhtml i.e. the list of the latter two is not the "correct list". The `class_map` only encodes what the current implementation of the `protocol` framework can decode automatically. It doesn't imply that any other values are filtered in any way. So, adding `0x8100` and `0x86dd` is, of course, fine, but don't remove the existing mapping. The title of this PR should be changed as well :)

krawthekrow:
Oops, that's correct. Fixed.

lukego:
Merged via #1251. Thanks @krawthekrow and @alexandergall for the review! Sorry about the latency.

PR #1239, opened 2017-10-03
Aborts in trace due to loop unroll limit and leaving loop in root trace

raj2569:
Creating a PR for loop aborts I am seeing so that @lukego can take a look.

This PR is based on `wingo-next` and has merged in `lukego/luajit-reset-penalty-cache`.

The test app is at `program/jit_loop/` and a helper app is at `apps/jit_loop/`. The app is basically a bridge which copies packets from one interface to another after changing mac and sends to next hop. Before running the app edit the PCI info, and MAC address in `jit_loop.lua`.

I am running this as: `sudo ./snabb snsh -jdump=+rsxaA,dump.txt -jtprof -p jit_loop`

Getting lots of trace abort as shown below:

```
---- TRACE 28 abort intel_mp.lua:787 -- failed to allocate mcode memory
---- TRACE 17 abort app.lua:511 -- leaving loop in root trace
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 22 abort app.lua:511 -- leaving loop in root trace
---- TRACE 30 abort intel_mp.lua:787 -- failed to allocate mcode memory
---- TRACE 17 abort app.lua:511 -- leaving loop in root trace
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 22 abort app.lua:511 -- leaving loop in root trace
---- TRACE 28 abort intel_mp.lua:787 -- failed to allocate mcode memory
---- TRACE 17 abort app.lua:511 -- leaving loop in root trace
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 21 abort app.lua:511 -- leaving loop in root trace
---- TRACE 28 abort intel_mp.lua:787 -- failed to allocate mcode memory
---- TRACE 16 abort app.lua:511 -- leaving loop in root trace
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 19 abort app.lua:512 -- loop unroll limit reached
---- TRACE 22 abort app.lua:511 -- leaving loop in root trace
[...]

PR #1242, opened 2017-11-05
engine: Separate JIT traces for each app callback

lukego:
This branch makes the engine ensure that JIT traces never cross between the engine and app callbacks. This is intended both to make JIT trace dumps more readable (don't mix engine and apps in the same trace) and to limit the impact of issues like JIT blacklistings (prevent problems in one app from propagating into other apps or the engine.)

This is achieved using a new primitive function `jit.tracebarrier()` that JIT traces cannot cross. During recording, a call to `jit.tracebarrier()` will immediately finish the current trace and then start a new root trace. This appears as a "trace stitch" in the log.

This branch makes the engine execute such a trace barrier immediately before and after each call to an app.

Performance impact needs to be evaluated. See also raptorjit/raptorjit#115.

### DIFF
--- a/lib/luajit/src/lib_jit.c
+++ b/lib/luajit/src/lib_jit.c
@@ -141,6 +141,12 @@ LJLIB_CF(jit_attach)
   return 0;
 }
 
+/* Calling this forces a trace stitch. */
+LJLIB_CF(jit_tracebarrier)
+{
+  return 0;
+}
+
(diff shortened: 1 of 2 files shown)


lukego:
Early feedback seems to be that this change does not introduce problems, but it remains to be seen how effective it really is for simplifying performance diagnostics. I will see if it helps to resolve #1239 before really proposing it for merge.

PR #1244, opened 2017-11-08
lj_record.c: Record IFUNC/IFUNCV the same as FUNC/FUNCV

lukego:
Record calls to blacklisted functions the same way as calls to
uncompiled functions. This allows for the possibility that a
blacklisted function can be successfully recorded as part of a larger
trace.

This is specifically intended to fix the problem where a tiny utility
function somewhere, such as `ntohs`, is not able to be compiled in
isolation for some reason and that ends up preventing all of its
callers from compiling too.

This has tended to happen for obscure reasons like "the utility
function ends with a tail-call to a C library routine" and with this
patch we should not need to worry about such minutiae anymore. (We
should be able to retire the "wrap in parenthesis to prevent a
tailcall" trick that some people are familiar with.)

More background:

- How the trace recorder treats function calls: raptorjit/raptorjit#118.
- Why this can be a problem: raptorjit/raptorjit#119.
- More on this specific solution: raptorjit/raptorjit#120.

### DIFF
--- a/lib/luajit/src/lj_record.c
+++ b/lib/luajit/src/lj_record.c
@@ -2411,8 +2411,6 @@ void lj_record_ins(jit_State *J)
   case BC_IFORL:
   case BC_IITERL:
   case BC_ILOOP:
-  case BC_IFUNCF:
-  case BC_IFUNCV:
     lj_trace_err(J, LJ_TRERR_BLACKL);
     break;
 
@@ -2424,6 +2422,7 @@ void lj_record_ins(jit_State *J)
   /* -- Function headers -------------------------------------------------- */
 
   case BC_FUNCF:
+  case BC_IFUNCF:
     rec_func_lua(J);
     break;
   case BC_JFUNCF:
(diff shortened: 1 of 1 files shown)


lukego:
I am running a performance regression test on this now. I would be very interested to hear if somebody can confirm that this change renders "the parenthesis trick" obsolete i.e. the function with the tailcall will still be blacklisted but we won't care because the calls will all be successfully inlined anyway (cc @alexandergall @eugeneia @wingo.)

eugeneia:
I have started a Hydra jobset that compares my base branch (vita), a version that includes attempts to manually inhibit tail calls to C functions (vita-stop-tco), a version with this PR merged (vita-record-blacklisted), and finally a version with both of the latter (vita-stop-tco+record-blacklisted, or "st+rb" in the report). Results will be here:

https://hydra.snabb.co/eval/7590

eugeneia:
https://hydra.snabb.co/eval/7621#tabs-still-succeed

I increased the number of iterations a bit, Tukey’s test still doesn’t see any differences. This may be very well because the outliers don’t impact the average result much.

To me it looks like `record-blacklisted` does have fewer "teeth" (i.e. dramatically bad traces) in the linegraph: https://hydra.snabb.co/build/2473885/download/2/report.html#line-graph

It also "looks" like the optimizations cancel each other out... no definitive answer from me here!

PR #1248, opened 2017-11-14
Fix adaptation of "snabb snabbnfv traffic" to be driver-agnostic

wingo:
Fix borked code introduced in 3f5d3efbb84a9033b8b7f1021ef47570f4ecce9c.


wingo:
Merging as it seems to fix the benchmark issue.

PR #1249, opened 2017-11-16
Incoming changes from wingo-next, November 2017 edition

wingo:
This PR merges in the following branches:

 * #1237: Flip the switch to use `intel_mp` rather than the old `Intel82599` driver.  Also update packetblaster and firehose to use `intel_mp` rather than the old driver.


wingo:
It would seem that perf results are OK: https://hydra.snabb.co/eval/7681

PR #1250, opened 2017-11-17
next: Changes queued for 2017.11 release

lukego:
This release will follow quickly on the heels of the much-delayed v2017.09 release.


lukego:
Huzzah! The new `intel_mp` driver is now adopted as the default and the `intel10g` driver (which represents some of the first lines of code ever written in Snabb) becomes a legacy backup option.

Great hacking to all the many `intel_mp` hackers !!! :+1:

lukego:
Sorry about the wait. I wanted to recommend this for release now but I see a [performance regression](https://hydra.snabb.co/build/2725133/download/2/report.html#l2fwd) that needs to be resolved. I will first check if this is related to overuse of trace barriers in #1242.

lukego:
Confirmed that the performance regression is due to the JIT barrier. The [lukego-optimize](https://hydra.snabb.co/build/2737087/download/2/report.html#l2fwd) performance with that change reverted matched master again. I will test using fewer barriers (on entry or exit to an app, but not both) and see if that is better. Otherwise the JIT barrier might not be suitable (too expensive) for app/engine transitions.

lukego:
I have reverted the calls to `jit.tracebarrier()` around app callbacks. These seem to be a bit of a performance drag overall and their benefit is not really established.

So the `jit.tracebarrier()` primitive still exists but the engine doesn't currently use it.

Let's wait for the standard Hydra tests to complete now and then we should be :+1: for the release.

lukego:
@eugeneia There is a performance regression on the iperf benchmark but I propose that we ship this now anyway.

I have been combing through recent CI results with @wingo on Slack and it seems like the problem seems to be caused by voodoo. I have another branch with almost identical contents that does not show the issue. The only difference is whether the `jit.tracebarrier()` primitive exists in the C code. Having that code present seems to provoke the problem even though it is never called.

I am reluctant to make a "nonsense" change to "solve" this problem. I would prefer to accept it for now and focus on finding the root cause. I see the new RaptorJIT tooling as the way to do this and so I want to spend my time now on integrating that with Snabb. Hence my willingness to accept this symptom of the root problem (wider variance on the iperf benchmark) for the moment.

Hypothetically if the problem is something obscure, like whether two Lua loop bytecode addresses hash into the same JIT hot counter, then there are probably very many different ways that it could be provoked (e.g. choice of C compiler version) and so I am not really confident that nailing down one such issue in the test environment would translate into a real world benefit. This would need to be solved more thoroughly in the JIT after seeing exactly what is really going on.

WDYT?

eugeneia:
Totally agree. I think #1244 is not included in this release, or is it? Maybe, as a last resort I would test to see if it resolves this regression like it did mine. This doesn’t block the release from my point of view though. Ready when you are!

lukego:
@eugeneia Glad to hear that this change has helped you. I pushed it here and let's see what Hydra says tomorrow.

lukego:
@eugeneia This was worth a try but the new benchmark results show the same issue: https://hydra.snabb.co/build/2822290/download/2/report.html#iperf.

wingo:
In a way that's good to know that the results are the same.  I was unsure whether to blame a voodoo change to the software itself or some problem with the statistics!

PR #1257, opened 2017-12-11
Flesh out multi-process reconfigurable network functions

wingo:
The multi-process facility is no longer a pair of weird apps; instead now it's a library called `lib.ptree` (for "process tree").  There is better developer-facing documentation in `src/lib/ptree/README.md` and more importantly there is example code in `src/program/ptree`, along with README.md there.  @xray7224 and @dpino wdyt?


src/lib/ptree/worker.lua
```diff
@@ -0,0 +1,118 @@
...
+   end
+   return ret
+end
+
+function Worker:shutdown()
+   -- This will call stop() on all apps.
+   engine.configure(app_graph.new())
+
+   -- Now we can exit.
+   S.exit(0)
```
dpino: Why not `main.exit(0)`?

src/program/lwaftr/csv_stats.lua
```diff
@@ -84,41 +71,81 @@ function CSVStatsTimer:add_app(id, links, link_names)
...
 
 function CSVStatsTimer:set_period(period) self.period = period end
 
 -- Activate the timer with a period of PERIOD seconds.
-function CSVStatsTimer:activate(write_header)
-   if write_header then
-      self.file:write(self.header..'\n')
-      self.file:flush()
+function CSVStatsTimer:start()
+   local function tick() return self:tick() end
```
dpino: What's the goal of wrapping the `self:tick()` call instead of calling `self:tick()` directly?

wingo: It's because timer functions are called without any arguments, and here we need to call tick(self).

src/lib/ptree/worker.lua
wingo: I do not know.  This comes from `follower.lua`.  I guess you mean `os.exit(0)` btw.  Can fix I guess :)

src/lib/scheduling.lua
```diff
@@ -0,0 +1,105 @@
...
+module(..., package.seeall)
+
+local S = require("syscall")
+local lib = require("core.lib")
+local numa = require("lib.numa")
+local ingress_drop_monitor = require("lib.timers.ingress_drop_monitor")
+
+local function fatal (msg)
+   print(msg)
+   main.exit(1)
```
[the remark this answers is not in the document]
wingo: I will make a note of it!

src/lib/ptree/worker.lua
eugeneia: Actually, `main.exit()` does exist but currently simply calls out to `os.exit()`. I think the idea was to be prepared for potential cleanup code in the engine?

wingo:
OK as this is merged to `lwaftr` upstream and `wingo-next` has the rest of the `lwaftr` updates, I will pull the trigger here and merge to `wingo-next`, so that it all goes into v2017.12.

PR #1258, opened 2017-12-14
Refactor apps.config into lib.ptree

wingo:
This merges https://github.com/snabbco/snabb/pull/1257.


wingo:
whoops, wrong target branch

PR #1261, opened 2018-01-09
Merge l2vpn topic branch

alexandergall:
This PR is a request to merge the `l2vpn` topic branch into master. It adds the directory `src/program/l2vpn` and doesn't touch anything else.


wingo:
Awesome!!!!  Given that it just affects program.l2vpn, I think we should merge as-is.  I would do it already if wingo-next weren't already blocked due to a perf regression; as is if no one else gets to it, I will merge when wingo-next flushes to next.

eugeneia:
Since `max-next` is free I will be the next hop instead as discussed on Slack with @wingo.

PR #1263, opened 2018-01-11
lib.yang: support for pattern restrictions

eugeneia:
Adds support for pattern and to lib.yang. The XSD Regex implementation is of questionable design (at least it was fun to hack), and has some restrictions:

* no unicode support (the implementation is 8‑bit clean though)
* no block escapes

Generally speaking, its not advisable to use it for sanitizing non-ASCII input. It’s good enough to parse things like the IP address patterns we use though.

I could add full unicode support (I have a branch that interprets the strings as UTF-8), it’s just that I have decided not to, because I wasn’t sure if sticking a whole Unicode databse into the `snabb` binary was worth it. If anyone actually needs Unicode support in Yang patterns, I will happily add it back in!

Cc @wingo


eugeneia:
Right, this can’t work right now since IP addresses are converted to numbers before the validators see them… I didn’t notice this before because I have disabled that behavior on my downstream branch, my bad!

I guess I can guard the pattern validator on the invariant that the value to be validated is a string.

eugeneia:
That being fixed, the next error appears to be a lwaftr test assuming ranges are single intervals:

https://gist.github.com/SnabbBot/47076ad6d12879c3ce4d9ba4bf8ccbd7#file-log-L816

eugeneia:
All good to go!

wingo:
Given that wingo-next has a lot of changes to some of the same stuff and that it's ready to go modulo some unrelated perf regression, I propose holding off here; unless you want to change the base branch to wingo-next of course.  WDYT?

eugeneia:
I am in no particular rush to get this merged, so I guess feel free to pick it up whenever convenient.

eugeneia:
Merged the current master. Note that besides the pattern validator this PR also...

* adds contextual errors to validators by passing in a parser
* fixes a bug where explicit values were not recognized in ranges

eugeneia:
Ok should be better now! :-)

PR #1172, opened 2017-07-06
Snabb on RaptorJIT [wip]

lukego:
This is a first attempt to run Snabb on [RaptorJIT](https://github.com/raptorjit/raptorjit).

RaptorJIT is a fork of LuaJIT intended to work better for Snabb and other x86/64 server applications. LuaJIT is in feature freeze but there are lots of amazing features that should be developed. So we in the Snabb community have an opportunity to roll up our sleeves and set a bold new direction.

The idea is to make the code more maintainable (stripped down features/ports), the community more broad (new maintainers welcome), and the features more appropriate (ubiquitous high-level tracing and profiling support.)

Here is some work being done in RaptorJIT to make life better for Snabb hackers:

- JIT IR Code visualizer: https://github.com/raptorjit/raptorjit/pull/63#issuecomment-310138536.
- JIT dumps enriched with profiler data: https://github.com/raptorjit/raptorjit/pull/63#issuecomment-309408425.
- Standard benchmark report to evaluate optimizations: https://github.com/raptorjit/raptorjit/issues/46#issuecomment-288061478.
- Top-down Optimization Guide: raptorjit/raptorjit#68.

Here are some TODO items for this branch:

- [ ] Get Snabb community to buy into the RaptorJIT concept.
- [ ] Get all tests passing.
- [ ] Get all benchmarks looking good.
- [ ] Replace introspection API calls with external-inspection dumps (see below.)
- [ ] Remove the build dependency on nix (see https://github.com/raptorjit/raptorjit/issues/58#issuecomment-289789193).

Here are a few relevant differences from LuaJIT that may present challenges:

- RaptorJIT only supports Linux/x86-64. This is to make it quicker, easier, safer, more fun to add new features to the VM. In other words to lower the hurdles to becoming a confident maintainer.
- RaptorJIT replaces introspection with inspection. Tools like `jit.dump` are removed. Raw data is exported for separate analysis tools instead. See https://github.com/raptorjit/raptorjit/issues/22.
- RaptorJIT only supports 64-bit heap ("GC64"). This is a new feature that @corsix added to LuaJIT but it is considered experimental and LuaJIT does not enabled it by default. So we have inherited some hard bugs that we need to fix e.g. https://github.com/raptorjit/raptorjit/pull/47 which went upstream into LuaJIT at LuaJIT/LuaJIT#305. This is real work but it will help us build confidence and also benefit other LuaJIT branches including upstream.

Questions?????

### Status

Current proof-of-concept state is that build and at least one selftest works:

[...]

PR #1264, opened 2018-01-15
Migrate to RaptorJIT

lukego:
This pull request is intended to migrate Snabb from LuaJIT to [RaptorJIT](https://github.com/raptorjit/raptorjit). This supersedes #1172 which is based on the same branch but was opened early as a preview into the development.

Consequences:

- JIT tracing and dumping will be always enabled. The shm file `audit.log` contains an efficient binary representation of the information from `jit.v` and `jit.dump`. This must be decoded with separate tools (Studio.)
- Profiling will be always enabled. The shm directory `vmprofile/` contains profiler datasets for the engine and for each individual app. The profiler data records which JIT traces were hot and separately tracks machine code, FFI code, VM code, GC, etc. This data must also be decoded with separate tools (Studio and also tools by @eugeneia.)
- The `jit.dump`, `jit.v`, and `jit.p` tools are no longer available. These are superseded by the features above which are superior because they are "always on" and available in production. Sorry! We will need to adjust to the new tools and identify the use cases that they currently miss.
- RaptorJIT has some different optimizations and maybe perform better/worse on certain things. We will need to identify and fix any regressions.
- We will be our own masters who can hack on the JIT compiler for fun and profit! :-)

TODO:

- [ ] I need to provide instructions for examining trace and profiler data with Studio.

I am hoping to land this branch in the next release. Thoughts?


wingo:
Is it possible to leave out the trace and profile generation?  Or to have it controllable by a switch?  Part of me wants to preserve that "quiet strace" thing that current Snabb dataplanes have.  Or is this irrational?

lukego:
> Is it possible to leave out the trace and profile generation?

@wingo Let's dig in to this a bit. On the one hand it is easy to make this configurable and that may well make sense. On the other hand it is a design goal to make the runtime behavior so smooth that you want to keep this enabled during benchmarking and in production.

The auditlog will make some `write()` system calls but only when the JIT is active. This should not happen in the steady-state. I believe that we also already have system calls when the JIT is working e.g. memory protection flags being updated where machine code is generated.

The profiler will not make any system calls in the steady-state. The logging is passive via file-backed shared memory. There will however be a `SIGPROF` signal every millisecond. This event will execute one small C function that bumps one counter in the current shared memory region.

How does that sound to you on paper? Still like something you would want to turn off?

wingo:
Ah good.  Yeah syscalls, writes, all that during JIT code generation are no problem at all.  We could spawn an Xeyes window and it would be fine :)  I am only concerned about steady-state syscall load.  

So what about profiling?  That isn't continuous and always-on and always writing to disk somewhere?

eugeneia:
> I need to provide instructions for examining trace and profiler data with Studio.

Vita comes with a tool [snabb trace](https://github.com/inters/vita/tree/master/src/program/trace) that analyzes vmprofile data and parses JIT dumps (think of it as a super-light battery-included alternative to Studio on terminals), I will update this tool to work with this branch and maybe upstream it? It would give users roughly the same abilities as the old traceprofile plus some more (e.g. get the byte/machine code of an interesting trace) in cases where Studio isn’t available (e.g. live in production).

eugeneia:
> So what about profiling? That isn't continuous and always-on and always writing to disk somewhere?

@wingo vmprofile just increments counters in a shared memory object. I have already been using it in Vita and can attest to the non-existant impact.

wingo:
What's the current timeline for getting trace/profile data out of a Snabb process?

How is this branch doing as far as performance goes?  The snabb-bot data is generally good but showing snabbnfv iperf having up to 100% slowdown.

I guess the test suite needs fixing up as well (above error is a core.memory error).

I am available to work on this soon, just trying to get your feeling here.

wingo:
I am working on raptorjit+snabb currently.  The branch is https://github.com/Igalia/snabb/tree/raptorjit.  Current additions relative to this branch are https://github.com/Igalia/snabb/pull/1032, a merge from master (https://github.com/Igalia/snabb/pull/1033), and a fix for DynASM 64-bit immediates (#1302).

wingo:
From what I can see of the log there are two blockers here:

 (1) Not so good perf on basic1-100e6; that could be statistical though.  Better to use hydra here
 (2) A failure in the tap app.  Seems to be legit.  I didn't see it locally because I don't run with SNABB_TAPTEST.

eugeneia:
Closing this one in favour of #1316.

PR #1266, opened 2018-01-26
Merge ipsec into next (v2018.01)

eugeneia:
Includes: #1262 (tested with #1220 )


eugeneia:
I did this the wrong way, I will merge ipsec onto max-next instead and do a second max-next -> next PR.

PR #1268, opened 2018-01-29
Allocate get_mempolicy mask of right size

wingo:
Before, if the kernel exposed more than a ulong's worth of NUMA nodes, we were inviting the kernel to overwrite our memory :grimacing:

### DIFF
--- a/lib/ljsyscall/syscall/linux/syscalls.lua
+++ b/lib/ljsyscall/syscall/linux/syscalls.lua
@@ -487,10 +487,11 @@ end
 
 function S.get_mempolicy(mode, mask, addr, flags)
   mode = mode or t.int1()
-  mask = mktype(t.bitmask, mask)
-  -- Size should be at least equals to maxnumnodes.
-  local size = ffi.cast("uint64_t", math.max(tonumber(mask.size), get_maxnumnodes()))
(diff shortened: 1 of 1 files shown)


wingo:
Cc @dpino for the correctness; @eugeneia I wonder if we introduced this lib.numa problem with the patch for #1200, and I wonder if this might fix it.

eugeneia:
@wingo No, this appears to be unrelated to #1269

dpino:
As far as I recall, #1200 was about fixing a bug in the numa library on systems that only have one NUMA node. This issue seems to be different. If I understand correctly, the issue is running the lib.numa selftest on systems that are not NUMA.

eugeneia:
Is this still an issue? Because SnabbBot seems to rightfully complain, but this seems like an important fix.

takikawa:
I put a fixed (I think) version of this PR on the lwaftr repo downstream (https://github.com/Igalia/snabb/pull/1198) and propose to merge it there, and then upstream it in a batch with other lwaftr changes.

eugeneia:
I cherry picked https://github.com/Igalia/snabb/pull/1198 onto max-next, closing this PR.

PR #1267, opened 2018-01-26
Merge max-next into next (v2018.01-2)

eugeneia:
Includes #1262 (tested with #1220) plus some fixes to tests:

 * `lib.lpm.*` selftests are less excessive by default (intensive tests are put behind a environment variable) Cc @petebristow 
* ~~`lib.numa` selftest is skipped when run on a non-NUMA system~~

PR #1269, opened 2018-01-29
lib.numa: selftest fails on non-NUMA systems

eugeneia:
As per https://github.com/snabbco/snabb/pull/1267#issuecomment-361235782

I still believe this to be expected behavior and possibly the correct fix (I have included an alternative approach that attempts to be... progressive), unless `has_numa` is supposed to behave differently than I expected. Let me explain:

I observe a difference in behavior on a system that has only a single NUMA node, and a system which has no NUMA at all. E.g.

Single NUMA node:
```
$ numactl -H
available: 1 nodes (0)
node 0 cpus: 0 1 2 3 4 5 6 7
node 0 size: 15943 MB
node 0 free: 13209 MB
node distances:
node   0 
  0:  10 
$ numactl -m 0 echo hello
hello
```

No NUMA support:
```
$ numactl -H
No NUMA available on this system
$ numactl -m 0 echo hello
numactl: This system does not support NUMA policy
```

On both of those systems `has_numa` returns false (maybe that is a bug in itself?). Without knowing too much about the subject matter, I had intuitively assumed that calling `bind_to_numa_node` can’t possibly work on a system that has no concept of NUMA nodes. I.e., `assert(S.set_mempolicy('bind', node))` will fail. I think this is different for systems with only a single NUMA node, where you can very well bind to the only node (maybe `has_numa` should return true on those?).

On the non-NUMA system the selftest of `lib.numa` will fail like so:

```
→ numactl -H
No NUMA available on this system
→ sudo ./snabb snsh -t lib.numa
selftest: numa
core/main.lua:26: Function not implemented

Stack Traceback
===============
(1) Lua function 'handler' at file 'core/main.lua:168' (best guess)
	Local variables:
	 reason = string: "core/main.lua:26: Function not implemented"
	 (*temporary) = C function: print
(2) global C function 'error'
(3) Lua global 'assert' at file 'core/main.lua:26'
	Local variables:
	 v = nil
(4) Lua global 'bind_to_numa_node' at file 'lib/numa.lua:116'
	Local variables:
	 node = number: 0
(5) Lua global 'bind_to_cpu' at file 'lib/numa.lua:103'
	Local variables:
	 cpu = number: 0
	 cpu_and_node = table: 0x419eb8c8  {cpu:0, node:0}
(6) Lua global 'test_cpu' at file 'lib/numa.lua:134'
	Local variables:
	 cpu = number: 0
	 node = nil
(7) Lua field 'selftest' at file 'lib/numa.lua:152'
	Local variables:
	 (for index) = number: 0
	 (for limit) = number: 1023
	 (for step) = number: 1
	 cpuid = number: 0
(8) Lua function 'opt' at file 'program/snsh/snsh.lua:31' (best guess)
	Local variables:
	 arg = string: "lib.numa"
(9) Lua field 'dogetopt' at file 'core/lib.lua:425'
	Local variables:
	 args = table: 0x4062b8d8  {1:-t, 2:lib.numa}
	 actions = table: 0x40754520  {t:function: 0x40e89798, q:function: 0x40754568, P:function: 0x40626f78 (more...)}
	 opts = string: "hl:p:t:die:j:P:q:"
	 long_opts = table: 0x40628810  {jit:j, help:h, program:p, test:t, load:l, package-path:P, debug:d, sigquit:q (more...)}
	 opts = table: 0x419e9460  {1:t}
	 optind = number: 3
	 optarg = table: 0x406288d8  {1:lib.numa}
	 (for generator) = C function: builtin#6
	 (for state) = table: 0x419e9460  {1:t}
	 (for control) = number: 1
	 i = number: 1
	 v = string: "t"
(10) Lua field 'run' at file 'program/snsh/snsh.lua:65'
	Local variables:
	 parameters = table: 0x4062b8d8  {1:-t, 2:lib.numa}
	 profiling = boolean: false
	 traceprofiling = boolean: false
	 start_repl = boolean: false
	 noop = boolean: true
	 program = nil
	 opt = table: 0x40754520  {t:function: 0x40e89798, q:function: 0x40754568, P:function: 0x40626f78 (more...)}
(11) Lua function 'main' at file 'core/main.lua:67' (best guess)
	Local variables:
	 program = string: "snsh"
	 args = table: 0x4062b8d8  {1:-t, 2:lib.numa}
(12) global C function 'xpcall'
(13) main chunk of file 'core/main.lua' at line 230
(14)  C function 'require'
(15) global C function 'pcall'
(16) main chunk of file 'core/startup.lua' at line 3
(17) global C function 'require'
(18) main chunk of [string "require "core.startup""] at line 1
	nil
```

Cc @wingo

### DIFF
--- a/src/lib/numa.lua
+++ b/src/lib/numa.lua
@@ -109,6 +109,7 @@ function unbind_numa_node ()
 end
 
 function bind_to_numa_node (node)
+   if not has_numa() then return end
    if node == bound_numa_node then return end
    if not node then return unbind_numa_node() end
    assert(not bound_numa_node, "already bound")
@@ -128,9 +129,6 @@ function prevent_preemption(priority)
 end
 
 function selftest ()
-   if not has_numa() then
-      os.exit(engine.test_skipped_code)
-   end
 
    function test_cpu(cpu)
       local node = cpu_get_numa_node(cpu)
@@ -138,11 +136,11 @@ function selftest ()
       assert(bound_cpu == cpu)
       assert(bound_numa_node == node)
       assert(S.getcpu().cpu == cpu)
-      assert(S.getcpu().node == node)
+      assert(not has_numa() or S.getcpu().node == node)
       bind_to_cpu(nil)
       assert(bound_cpu == nil)
       assert(bound_numa_node == node)
-      assert(S.getcpu().node == node)
+      assert(not has_numa() or S.getcpu().node == node)
       bind_to_numa_node(nil)
       assert(bound_cpu == nil)
       assert(bound_numa_node == nil)
--- a/src/lib/numa.lua
+++ b/src/lib/numa.lua
@@ -128,6 +128,9 @@ function prevent_preemption(priority)
 end
 
 function selftest ()
+   if not has_numa() then
+      os.exit(engine.test_skipped_code)
+   end
 
    function test_cpu(cpu)
       local node = cpu_get_numa_node(cpu)


eugeneia:
Oh right, I should add some info about the non-NUMA system:

```
→ uname -a
Linux perth 4.4.111workbench-hugetlbfs #1 SMP Thu Jan 18 13:18:28 CET 2018 x86_64 Intel(R) Core(TM) i3-4010U CPU @ 1.70GHz GenuineIntel GNU/Linux
```

The kernel isn’t very special except that I enabled support for hugetlbfs. Admittedly,

```
# CONFIG_NUMA is not set
```

so that might in fact make the difference.

dpino:
I cannot reproduce the bug as I haven't managed to disable NUMA in my system. Apparently it's possible to turn it off by passing 'numa=off' to the kernel command line arguments. I did that but when I run `numactl -H` it reports one NUMA node.

In any case, I realized there's a system call that reports whether a system has NUMA available. Apparently ljsyscall doesn't expose that syscall, which is part of libnuma. Could you try running the following program?

```lua
module(..., package.seeall)

function selftest()
   print("seltest")
   local ffi = require("ffi")
   local C = ffi.C
   ffi.cdef[[
   int numa_available(void);
   ]]
   local numa = ffi.load("numa")
   print(numa.numa_available())
end
```

```lua
sudo ./snabb snsh -t numa
seltest
0
```

In my system it returns 0 (OK). In systems where NUMA is not available should return -1.

What I would do is to make the selftest return SKIP_CODE if NUMA is not available.

dpino:
Another approach. Instead of depending on the `numa_available` syscall, the test could simply run the `numactl -H` command and check its status code (I hope the call returns a number >0 when NUMA is not available).

eugeneia:
https://github.com/snabbco/snabb/pull/1269#issuecomment-361420456
^ this does indeed return -1 on my system.

```
→ sudo ./snabb snsh -i
Snabb> ffi = require("ffi")
Snabb> ffi.cdef("int numa_available(void);")
Snabb> numa = ffi.load("numa")
Snabb> =numa.numa_available()
-1
```

What about the progressive approach in this PR though? Wouldn’t it be better to just not try to bind to numa nodes on systems where `has_numa` returns false?

dpino:
@eugenia works for me if that fixes the issue (sorry, I missed your PR changes).

eugeneia:
@wingo thoughts?

I realized there are two possible interpretations of `has_numa`:

1. “has NUMA capabilities” (i.e., supports `S.set_mempolicy('bind', node)`)
2. “has Non-Uniform-Memory-Architecture” (this is what `has_numa` currently implements, I think)

I had mistakenly assumed 1 in 090d617. The current state of this PR isn’t really affected by this, but we could add a function `supports_numa` (with the code from @dpino) that checks for 1  and use that where this PR uses `has_numa`. This way the code paths would stay uniform for all systems like before, except for systems like mine which presumably don’t have NUMA support compiled into the kernel. It’s more code though…

eugeneia:
Any idea what this [test failure](https://gist.github.com/SnabbBot/b41cebe4f38fa194a00c18086d0f03e7#file-log-L2722) could be about ([scroll way down](https://gist.github.com/SnabbBot/b41cebe4f38fa194a00c18086d0f03e7#file-log-L4339) for the error message)?

wingo:
I think the test failure is due to a race condition in the property-based test, whereby for some reason the tester couldn't connect to the dataplane being tested.  In general #1294 should solve the irritation, and we should fix these on the Igalia side of things.

PR #1274, opened 2018-02-08
Add "snabb unhexdump" tool

wingo:
See README for usage.


eugeneia:
Why not? :-)

lukego:
JFYI a similar tool is [text2pcap](https://www.wireshark.org/docs/man-pages/text2pcap.html), but who likes external dependencies anyway :).

PR #1276, opened 2018-02-09
:stop() apps on user program termination

Reperator:
This PR calls `main.exit()` once the snabb user program terminates. `main.exit()` then clears the `engine` config, thereby calling `:stop()` on all apps that implement the method.

Before that all links are killed and `:link()` is called on all apps that implement the method. If we don't want this behavior we could instead iterate over `engine`'s `app_table` and `:stop()` apps this way.

Previously apps weren't stopped after the last call to `engine.main()` which may prevent some apps from finalizing unless the user explicitly calls `engine.config()` with some config that doesn't include  these apps (e.g. an empty config).


Reperator:
Looks like the test fails because of a copy-paste-error which gets fixed in #1280.

lukego:
One related mechanism that we have is `main.shutdown()` that executes both on normal termination and also when the process crashes. This is where we are doing sensitive things like ensuring that DMA is disabled before we return HugeTLB memory to the kernel for reuse.

You can see how this works at the bottom of `main.lua` i.e. on startup we spinoff a separate process that will perform cleanup after the main process dies.

I suppose the question is how much do you care about the `stop()` method running? The engine can run it for normal termination but if that's not enough then we have the shutdown mechanism.

Reperator:
If something really blows up I'd say `:stop()` may not be necessary; at least as a user I wouldn't expect my apps to be shut down properly if something crashes hard.

OTOH having something like a packet capture or a logging app write stuff to disk in case of a crash seems useful for debugging. I'm not sure how I'd go about implementing that though.

alexandergall:
Here is a case where calling `stop()` from `main.shutdown` would be useful. I have a bunch of basically identical processes running on separate RSS queues. Each process is run as a separate `systemd` service. When a service is started, RSS is set up properly and the process is getting its share of traffic. But when it is terminated with `systemctl stop`, the RSS queue remains active and I'm losing traffic. Calling `stop()` in that case would clean up RSS and re-direct all traffic to the remaining instances.

eugeneia:
Note that, currently, `shutdown` is run by the process that supervises snabb, which doesn’t have access to the app graph.

The new test failure is weird btw: https://gist.github.com/SnabbBot/ec0315294ddd326fd1931e00ce17d5f3#file-log-L21

eugeneia:
Given that it is possible to have the behavior implemented in this PR by explicitly doing

```
-- ordered shutdown
engine.configure(config.new())
```

at the end of a Snabb program, I am inclined to pass on this PR. My rationale is that a program might genuinely don’t care about stopping its apps (test scripts, etc), so why prescribe that behavior?

@alexandergall What you are describing would probably be best served by a cleanup hook in the `intel_mp` app similar to this one:
https://github.com/snabbco/snabb/blob/master/src/lib/hardware/pci.lua#L195
https://github.com/snabbco/snabb/blob/master/src/core/main.lua#L186

Reperator:
Counter-argument: Apps that expose a `:stop()` method need to be stopped to release resources, finish writing files, etc. As a user I'd expect my apps to be stopped automatically once Snabb terminates.

Off the top of my head I can't think of a case where it'd hurt to stop apps (even if a script does not care about stopping) while I came across the problem when writing a custom pcap-App. It took me a while to find out why my app wasn't stopped upon termination. That may be on me for not knowing Snabb's internals but I'd say stopping apps would be more user-friendly.

We could make the stopping optional by implementing a `-nostop` or `-dostop` command line flag.
(I'm obvioulsy in favor of opt-out.)

lukego:
One compromise might be to add an `engine.stop()` that runs `engine.configure(config.new())`? 

Personally I quite like the view that "orderly termination is only a special case of process crashing, and not a particularly interesting one." That is, if you want your application to be robust then you need to handle abnormal termination (crash) correctly, and once you have written that logic the simplest approach is often to treat normal shutdown the same way.

Just from a user perspective I really like being confident that hitting a process with `^C` or `kill -9` won't screw things up e.g. create an invalid state that prevents the process from starting the next time.

PR #1277, opened 2018-02-09
[wip] [sketch] Naive i350 fixes

gmichael225:
Just the minimal changes required to get the I350 driver working on a PowerEdge R720 w/ Dell's . hybrid 2x I350 and 2x 82599ES module.


lukego:
The issue with polling for `DONE0` seems to be that the I350 is a quad-port controller but on this machine only two of the ports are actually used and whether the driver sees them as port 0/1/2/3 depends on complicated EEPROM settings. So if we poll for PHY to initialize for port 0 there is a risk that this never happens because port 0 is mapped to pins on the ethernet controller that are not actually connected to anything on the PCB.

Seems to me from the I350 data sheet that we should be allowed to skip polling for the PHY to come up. This seems to be necessary before changing PHY settings but our driver is instead accepting the defaults from EEPROM. Right?

PR #1280, opened 2018-02-09
apps.ipv4.reassemble: fix a copy paste error at end of selftest

eugeneia:
Fixes a copy paste error at end of apps.ipv4.reassemble’s selftest.


wingo:
Belated LGTM :)

PR #1287, opened 2018-02-14
core.app: remove garbage when using pace_breathing

alexandergall:
The accumulators `lastfrees`, `lastfreebytes`, `lastfreebits` for the
counters `frees`, `freebytes` and `freebits` are initialized as Lua numbers
but implicitly converted to cdata objects in the assignments in
 `pace_breathing()`. This causes allocations that cannot be removed by
the sink optimizer.  Conversion to Lua numbers avoids this and reduces
GC noise.

### DIFF
--- a/src/core/app.lua
+++ b/src/core/app.lua
@@ -505,9 +505,9 @@ function pace_breathing ()
       else
          sleep = math.floor(sleep/2)
       end
-      lastfrees = counter.read(frees)
-      lastfreebytes = counter.read(freebytes)
-      lastfreebits = counter.read(freebits)
+      lastfrees = tonumber(counter.read(frees))
+      lastfreebytes = tonumber(counter.read(freebytes))
+      lastfreebits = tonumber(counter.read(freebits))
    end
 end
 


src/core/app.lua
```diff
@@ -505,9 +505,9 @@ function pace_breathing ()
       else
          sleep = math.floor(sleep/2)
       end
-      lastfrees = counter.read(frees)
-      lastfreebytes = counter.read(freebytes)
-      lastfreebits = counter.read(freebits)
+      lastfrees = tonumber(counter.read(frees))
+      lastfreebytes = tonumber(counter.read(freebytes))
+      lastfreebits = tonumber(counter.read(freebits))
```
wingo: Is this really what you want?  At some point counter.read(frees) will stop being equal to lastfrees because the tonumber() may lose precision.  Seems to me better to make lastfrees etc to be uint64_t values in some allocated FFI data.

alexandergall: I'm not sure this is a problem in practice (it will occur after something like 10 years of processing 14Mpps or something, I guess) but, sure, using a "box" for that would be safe. BTW, why exactly are we keeping `lastfreebytes` and `lastfreebits`? They're not used anywhere.

eugeneia: I think @lukego introduced them in 45b23d4029e717fdb693e1e18ec2826507fd16a7 and 8b1bc0e13b275d34f020bdb6ccec24ffd916024b but even then they were not used. I’d say its safe to drop them.

PR #1288, opened 2018-02-16
Inter-process links, approaching upstream

eugeneia:
This branch represents my latest efforts on getting interlinks ready for upstream. I have once more punted on integrating things into `core.{link,app,config}`. I tell myself that this branch shares most of the benefits of #1185, but without most of the work required to consolidate with core. :-) I feel like its the path of least resistance, so to speak.

**To briefly summarize the design:** the apps `apps.interlink.{receiver,transmitter}` can be used to set up links across Snabb processes within the same group. They do that via a particular kind of ring buffer defined in `lib.interlink` that is optimized for multiprocessing. Compared with the sketch in #1186, this branch…

* uses the group-global shared packet freelist from #1185 to reclaim packets exchanged across processes. (More on this later.)

* drops the configuration parameter `name`, instead the SHM object name is derived from the app name. This removes a bit of redundancy from the configuration; setting up an interlink is now as simple as

```
config.app(c, "interlink", Transmitter)
config.link(c, "someapp.output -> interlink.input")
```

…and the other side would do…

```
config.app(c, "interlink", Receiver)
config.link(c, "interlink.output -> someapp.input")
```

(This imho beats introducing a new special syntax for `config.link`.)

* drops the configuration parameter `create=true|false`. It no longer matters in which order Receiver and Transmitter are started: the apps will synchronize while setting up the link on their own. (More on that later.)

**Notes:**

* There is a new function `packet.account_free` that is called on packets being enqueued on an interlink. This is necessary because the engine estimates load and paces its breaths based on the metric “packets freed per breath”. `account_free` bumps `free` related statistics without actually freeing the packets, and convinces the engine that work is being done.

* interlinks no longer work across different process groups. Does that matter?

* `core.app` no passes the app name as the second argument to `:new()` (the field `appname` is only accessible in later callbacks).

* `lib.interlink` synchronizes concurrent calls to `new` and `free` using a state machine (transitions done via atomic compare-and-swap). Only `new` can ever block, and blocks only if the target SHM queue is already in use, until it becomes available. The interlink apps rely on the fact that `core.app` ensures that old apps are stopped before new apps are started. If `core.app` would start a new app that uses the same interlink object as an existing app *before* stopping the existing app `lib.interlink` would deadlock. (Users can still configure interlinks in ways that deadlock, but we rely on invariants upheld by `core.app` so it doesn’t happen with sensible configurations.)

* When two processes `free` an interlink, the process that detaches last will free any remaining packets on the queue and unlink the SHM object. This can only reliably be performed if the process can be sure that the process on the other end is no longer writing or reading from the queue, hence it is done by the process that detaches last. ~~The flip side here is that interlinks can not be deallocated if either process crashes before freeing the queue. **FIXME?** I think I might know a way to fix this involving a handler in the supervisor process.~~ There are now shutdown handlers run by the supervisor that detach processes from queues they failed to detach from by themselves.

* The selftest’s synthetic micro-benchmark scores way lower on this branch than on #1186, but I have no reason to believe what it measures to be relevant. In more realistic environments (i.e, Vita) I have found this branch to perform equal or better than #1186 in all regards observable to me. While I would be very interested in why the micro-benchmark results differ that much in such an unexpected direction, I have for now decided to not trust them too much, and thus don’t worry about it.

* This branch adds a new module `core.sync` that contains some low-level MP synchronization primitives used by `lib.interlink` (atomic compare-and-swap) and the shared freelist (@xrme’s spin-lock). I guess its questionable if this is a good name for the module, or if it should be a module at all? Suggestions welcome.

* I haven’t written any real documentation for this yet, but promise to do so in case we decide this design to be upstream-worthy.

* More excuses: this branch meets all my immediate requirements. Interlinks and core links are different and it will be awkward to pretend they are the same. I fear if I add anything to the link syntax it will cease to be endearing. This way, the complexity is contained at least. …

* (This branch is a direct descendant of #1186)

Cc @alexandergall @wingo @lukego …


alexandergall:
I just gave it a quick try. It seems like packet frees are double-counted on the sending side:
```
 Kfrees/s        freeGbytes/s    breaths/s      
 7618.53         0.98            86200          

 Links (rx/tx/txdrop in Mpps)    rx      tx      rxGb    txGb    txdrop 
 queue                           3.80    3.80    0.98    0.98    0.00   
 input1.output -> rss.vlan1      3.80    3.80    0.98    0.98    0.00   
 rss.default_1 -> default_1.inpu 3.80    3.80    0.98    0.98    0.00   
 default                         3.80    3.80    0.98    0.98    0.00   
```

alexandergall:
A preliminary benchmark with my test case yields significantly lower performance compared to #1186, maybe around 20-30% in terms of CPU cycles. At a first glance, a lot of time is spent in the loop of `packet.allocate()`
```
      while freelist_nfree(group_fl) > 0
      and freelist_nfree(packets_fl) < packets_allocated do
         free(freelist_remove(group_fl))
      end
```
It's a pretty compact root trace but top of the list of `LLC-load-misses` and second in `cycles` in terms of `perf` metrics. Looks very similar as what I've seen happening in the `Transmitter` app of #1186 where balance packets are exchanged in the ring buffer.

eugeneia:
@alexandergall Just pushed a fix to the bogus free accounting bug you spotted. This little change also seems to positively impact performance on rudimentary testing, so it might be worth another test run.

src/apps/interlink/selftest.snabb
```diff
@@ -0,0 +1,33 @@
+#!snabb snsh
```
wingo: Does it make sense to run via `#!../../snabb` ?  That is what the `intel_mp` tests do, FWIW.

src/core/app.lua
```diff
@@ -517,8 +517,11 @@ function breathe ()
       end
    end
    counter.add(breaths)
-   -- Commit counters at a reasonable frequency
-   if counter.read(breaths) % 100 == 0 then counter.commit() end
+   -- Commit counters and rebalance freelists at a reasonable frequency
```
wingo: Do you have hydra comparisons that this additional rebalance_freelists call is perf-neutral?  If it would be possible to put this call in the pull or push function of an interlink app, that would be better, of course -- is it possible?  (Maybe not.)

src/core/packet.lua
```diff
@@ -75,15 +84,49 @@ local function freelist_nfree(freelist)
...
+
+local function freelist_unlock(freelist)
+   sync.unlock(freelist.lock)
+end
+
 local packet_allocation_step = 1000
 local packets_allocated = 0
-local packets_fl = ffi.new("struct freelist", max_packets, 0, max_packets)
+local packets_fl = ffi.new("struct freelist")
+      packets_fl.max = max_packets
```
wingo: probably should indent this assignment flush left

src/core/packet.lua
```diff
@@ -75,15 +84,49 @@ local function freelist_nfree(freelist)
...
+local group_fl
+if not shm.exists("group/packets.freelist") then
+   group_fl = shm.create("group/packets.freelist", "struct freelist")
```
wingo: There is a TOCTTOU here; probably better to always `shm.create`.  `shm.create` does not add `O_EXCL` so it should silently succeed if the file already exists.

wingo: More broadly, does it always make sense to have a group freelist, even if the system only has one process?  I guess it is OK but it is a point to consider.

src/core/packet.lua
```diff
@@ -75,15 +84,49 @@ local function freelist_nfree(freelist)
...
+   group_fl.max = max_packets
+else
+   group_fl = shm.open("group/packets.freelist", "struct freelist")
+end
+
+function rebalance_freelists ()
+   if freelist_nfree(packets_fl) > packets_allocated then
+      freelist_lock(group_fl)
+      while freelist_nfree(packets_fl) > packets_allocated
```
wingo: It sure would be nice to be able to transfer batches of packets to the group freelist, instead of transferring them one by one.  In that case you could potentially even avoid the spinlock around the group freelist and use atomic ops instead.  Anyway, something to consider for the future, it's clear that this works for now :)

src/apps/interlink/selftest.snabb
eugeneia: I guess it depends on how you intend run this. In this case `make test` will execute `apps/interlink/selftest.snabb` from `src/`, and that’s also where I tend to run the script from.

@petebristow champions this convention if I understood him correctly and I lean towards following suit. Since snabb is rarely in `PATH`, the next best guess seems to be that its in `PWD`. Of course you can always do `snabb snsh <script>.snabb`.

src/core/packet.lua
eugeneia: The code actually works as intended because the group parent will have created `group/packets.freelist` before it starts any children. I.e. `shm.exists("group/packets.freelist")` tests if we are the group parent, which is not something that changes over time.

I agree with the other point about not always having a group freelist, I’ll look into that.

eugeneia:
I have pushed some further commits that

1. install a shutdown handler that detaches crashed Receivers/Transmitters. It should now be possible to recover from a crashed process by restaring it.
2. add the glue required to have interlinks show useful info in `snabb top -l`

src/lib/interlink.lua
```diff
@@ -0,0 +1,245 @@
...
+-- and detach in any order, and even for multiple processes to attempt to
+-- attach to the same interlink at the same time.
+--
+-- Interlinks can be in one of five states:
+
+local FREE = 0 -- Implicit initial state due to 0 value.
+local RXUP = 1 -- Receiver has attached.
+local TXUP = 2 -- Transmitter has attached.
+local DXUP = 3 -- Both ends have attached.
+local DOWN = 4 -- Either end has detached; must be re-allocated.
```
wingo: I don't quite get this description; the state transition diagram below suggests that DXUP -> TXUP is valid.  I would have assumed that DXUP can only transition to DOWN, based on the description.  What am I missing?

wingo: Should the comment be, "both ends have detached" ?

wingo:
Looking really great.  I like the design and the implementation looks good too :)

For comparison, VPP's "memif" is not dissimilar, and it additionally allows for third-party apps to participate in a flow: https://docs.fd.io/vpp/18.01/libmemif_doc.html / https://docs.fd.io/vpp/18.01/db/df1/src_2plugins_2memif_2memif_8h_source.html

We might envision something like interlinks allowing this in some future, perhaps like `snabb firehose`.  Anyway, that's thinking a bit farther on :)  Great work!!!!

src/core/packet.lua
wingo: I believe you might even be able to pass `{ max = max_packets }` as the initarg!

src/apps/interlink/selftest.snabb
wingo: Surely PWD is not in your $PATH though, right?

src/lib/interlink.lua
eugeneia: Yep that’s right the comments were off.  :-/  An interlink can go from DXUP to {T,R}XUP back to DXUP. It must deallocated once both ends have detached. I.e., more than two processes can attach to a single interlink during its life time. As long as either end is attached, both receiver and transmitter can be superseded by new processes. Should be fixed now.

src/apps/interlink/selftest.snabb
eugeneia: IIRC prior to this change the line was `#!/usr/bin/env snabb snsh` which consults `PATH` but that doesn’t help much usually. On the other hand `#!snabb snsh` won’t consult `PATH` but execute `./snabb`.

eugeneia:
@alexandergall 154d63c should fix the issue where supervisors created SHM directories under SNABB_SHM_ROOT.

wingo:
Just missing documentation and benchmarking to merge!

eugeneia:
@wingo Should be good to go. I don’t think there is a perf regression: https://hydra.snabb.co/build/3241437/download/2/report.html

src/core/app.lua
```diff
@@ -339,7 +339,7 @@ function apply_config_actions (actions)
       configuration.apps[name] = nil
    end
    function ops.start_app (name, class, arg)
-      local app = class:new(arg)
+      local app = class:new(arg, name)
```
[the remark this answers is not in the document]
eugeneia: This is the bit that enables configuring inter process links without the somewhat redundant configuration by using its own name. I liked that you can simply `config.app(c, "SharedLink", Receiver)`, but I agree at a second glance that this might be unnecessairly magic. Alteratively, the synatx would be: `config.app(c, "AppName", Receiver, "SharedLink")`.

src/core/main.lua
```diff
@@ -172,17 +172,19 @@ end
 
 -- Cleanup after Snabb process.
 function shutdown (pid)
+   -- simple pcall helper to print error and continue
+   local function safely (f)
+      local ok, err = pcall(f)
+      if not ok then print(err) end
+   end
+   -- Run cleanup hooks
+   safely(function () require("apps.interlink.receiver").shutdown(pid) end)
```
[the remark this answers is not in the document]
eugeneia: Honestly, I don’t see the difference between core calling into lib/ vs apps/, its both outside core. As it stands, lib/interlink and apps/interlink form a single whole, the reason the shutdown hooks are in apps/interlink is because that is where SHM paths are chosen.

src/core/packet.lua
```diff
@@ -45,18 +47,25 @@ end
...
-ffi.cdef[[
+local max_packets = 1e6
+
+ffi.cdef([[
 struct freelist {
+    int32_t lock[1];
     uint64_t nfree;
     uint64_t max;
-    struct packet *list[?];
+    struct packet *list[]]..max_packets..[[];
```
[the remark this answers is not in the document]
eugeneia: `[?]` doesn’t work because this branch maps it as a SHM object and `core.shm` would then fail to determine the `sizeof` this ctype. So I guess the remaing purpose of `max_packets` is to give this value a name in the source.

eugeneia:
@wingo the `name` argument to `:new()` is now reverted (I was able to achieve the same by using `:link()`.) Take me to wingo-next! :-)

eugeneia:
Test failure seems to be lwatr related... one more nondeterministic test to squash? :-)

src/apps/interlink/receiver.lua
```diff
@@ -0,0 +1,57 @@
+-- Use of this source code is governed by the Apache 2.0 license; see COPYING.
+
+module(...,package.seeall)
+
+local shm = require("core.shm")
+local interlink = require("lib.interlink")
+
+local Receiver = {name="apps.interlink.Receiver"}
+
+function Receiver:new ()
```
wingo: I think the interlink apps would do better to use the standard way to declare app parameters.  It could default to the app name for the link name, but the interlink name should be configurable.  WDYT?

src/core/main.lua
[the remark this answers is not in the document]
wingo: Alternately we could move lib.interlink into core, and make it choose the SHM paths.  WDYT?

src/core/packet.lua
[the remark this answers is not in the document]
wingo: What I mean to say is that you could have:

```c
enum { MAX_PACKETS=1000000 };
struct freelist {
  int32_t lock[1];
  uint32_t nfree;
  struct packet *list[MAX_PACKETS};
};
```

That way you could still access max_packets by ffi.C.MAX_PACKETS, but since we are really *requiring* that freelists have a fixed size, we do away with the notion that the freelist size is configurable.  Also that uses uint32_t for the occupancy for better struct packing.  WDYT?

Or, if preserving variable-sized freelists is important (probably it is), then we need to use some other variable-size-friendly core.shm mapper.

src/apps/interlink/receiver.lua
eugeneia: Agree, have a patch for that in the barrel already.

eugeneia: Agree, have a patch just for that already in the barrel.

About the other bits (to which I can for some reason not reply individually)...

- It’s not like `max_packets` was configurable at run time before, so I see it as a net-zero to change it to your formulation, which I might also prefer. Replacing these `uint64_t`s with 32 bit ints seems like a good idea, regardless. Both of these don’t appear to me to be related to this PR in particularily, so I tend to pass.

- Yeah I could shuffle around the code some... but I honestly don’t see the benefit right now.

Let’s get this merged and improve on it incrementally?

wingo:
The reason I provided feedback on the final two points (max_packets and core->apps layering violation) is that they represent an accumulation of technical debt and cruft.  I think it's reasonable to ask for new features, especially those that touch core, to not add fixmes.  It's fine to fix in a followup of course.

eugeneia:
FWIW I didn’t mean to sound ungrateful, rather I think these issues are not directly related to interlink, and I have mixed feelings about sneaking in fixes to these in this imho unrelated PR. E.g.:

- Should shutdown handlers maybe be installed with `<coremod>.add_shutdown_handler(myhandler)`? That would resolve the perceived layering violation as well but is out of scope of this PR imho. Or maybe only core modules should be able to install shutdown handlers, and lib/memory.lua should really be in core/, then that would require formulating a policy and adapting that would possibly also affect lib/interlink.lua. Not something I want to decide right here, so I went with the “change as little as possible” route.

- Similarily for the `struct freelist` definition. Is it perfect? Probably not. Do the proposed changes affect this PR in particular? Not really. If I do them in this PR I would claim responsibility, even though I don’t have a particular opinion or insight on the topics at hand.

Thanks for the review! :tada: :balloon:

PR #1291, opened 2018-02-28
README.md: Add link to Slack auto-invite page

lukego:
Slack supports a link that allows anybody to self-invite themselves to the channel. This seems to be a new feature? Anyway it is a nicer user experience than having to email me for access :).


wingo:
Merged onto `wingo-next`, thanks!

Issue #1286, opened 2018-02-13
Can't map PCI memory if kernel driver is active in Linux >= 4.5


alexandergall:
This test case uses the current `master` (`2ad3ed2`) with a trivial graph:
```
local c = config.new()
config.app(c, "in", require("apps.intel_mp.intel_mp").Intel,
           { pciaddr = "0000:69:00.1" })
config.app(c, "out", require("apps.basic.basic_apps").Sink)
config.link(c, "in.output -> out.input")
engine.configure(c)
engine.main({ duration = 5, report = { showlinks = true } })
```
When the device is unbound from the kernel, it works as expected
```
$ sudo ./snabb pci_bind -u 0000:69:00.1
Unbound 0000:69:00.1, ready for Snabb.
$ sudo ./snabb snsh test.lua
link report:
          18,604,865 sent on in.output -> out.input (loss rate: 0%)
```
But when the device is still bound to the kernel, the result is
```
$ sudo bash -c "echo 0000:69:00.1 >/sys/bus/pci/drivers/ixgbe/bind"
$ sudo ./snabb snsh test.lua
core/main.lua:26: Invalid argument

Stack Traceback
===============
(1) Lua function 'handler' at file 'core/main.lua:168' (best guess)
        Local variables:
         reason = string: "core/main.lua:26: Invalid argument"
         (*temporary) = C function: print
(2) global C function 'error'
(3) Lua global 'assert' at file 'core/main.lua:26'
        Local variables:
         v = nil
(4) Lua field 'map_pci_memory_unlocked' at file 'lib/hardware/pci.lua:159'
        Local variables:
         device = string: "0000:69:00.1"
         n = number: 0
         lock = boolean: false
         filepath = string: "/sys/bus/pci/devices/0000:69:00.1/resource0"
         err = nil
(5) Lua method 'new' at file 'apps/intel_mp/intel_mp.lua:358'
        Local variables:
         self = table: 0x41190df0  {rss_tab:function: 0x41192338, transmit:function: 0x41192520, rss_tab_build:function: 0x41192358 (more...)}
         conf = table: 0x41193da8  {mtu:9014, linkup_wait_recheck:0.1, rate_limit:0, linkup_wait:120, master_stats:true (more...)}
         self = table: 0x41196da0  {shm_root:/intel-mp/69:00.1/, pciaddress:0000:69:00.1, rate_limit:0, vmdq:false (more...)}
         vendor = string: "0x8086"
         device = string: "0x10fb"
[...]

PR #1296, opened 2018-03-02
Intel driver unbinds PCI device from Linux before mapping

wingo:
Since Linux 4.5, devices that are bound to a kernel driver can't have
their memory mapped by userspace under the default configuration of
CONFIG_IO_STRICT_DEVMEM=y.

Should fix https://github.com/snabbco/snabb/issues/1286.

### DIFF
--- a/src/apps/intel_mp/intel_mp.lua
+++ b/src/apps/intel_mp/intel_mp.lua
@@ -356,6 +356,7 @@ function Intel:new (conf)
    self.max_q = byid.max_q
 
    -- Setup device access
+   pci.unbind_device_from_linux(self.pciaddress)
    self.base, self.fd = pci.map_pci_memory_unlocked(self.pciaddress, 0)
    self.master = self.fd:flock("ex, nb")
 
@@ -1119,7 +1120,6 @@ function Intel1g:unlock_fw_sem()
 end
 function Intel1g:init ()
    if not self.master then return end
-   pci.unbind_device_from_linux(self.pciaddress)
    pci.set_bus_master(self.pciaddress, true)
    pci.disable_bus_master_cleanup(self.pciaddress)
 
@@ -1311,7 +1311,6 @@ vmdq_enabled_t = ffi.typeof("struct { uint8_t enabled; }")
 
 function Intel82599:init ()
    if not self.master then return end
-   pci.unbind_device_from_linux(self.pciaddress)
    pci.set_bus_master(self.pciaddress, true)
    pci.disable_bus_master_cleanup(self.pciaddress)
 


alexandergall:
I think @petebristow was concerned that if the unbinding is done without obtaining the lock first, it could create some kind of collision between concurrent Snabb processes. I don't know if that's an issue.

wingo:
Are intel_mp tests not running on CI???  I don't see them in the log.

I don't see the collision possibility, do you? Unbinding is idempotent and we keep our same locking discipline as before.

alexandergall:
I agree. The statement I had in mind was this  https://snabb.slack.com/archives/C0H3TPBV2/p1518174897000123 I guess he didn't suggest that there is an actual problem.

alexandergall:
This appears to lead to a race condition. If I start multiple Snabb instances concurrently, only one runs successfully, the others die with the symptom of #1286. Adding a `C.sleep(1)` after the call to `unbind_device_from_linux()`, this problem goes away.

I don't know how exactly the unbinding is handled by the kernel. I suppose it's synchronous for the first process (otherwise that one would see the same problem). The other processes check for the writeability of `/sys/bus/pci/devices/<address>/driver/unbind`. The `driver` subdirectory is removed when the device is unbound. But if that happens before the actual unbinding, the process will attempt to map the memory immediately and fail. 

I guess the question is if there is a reliable check whether a device is bound or not.

alexandergall:
I guess my suspicion was correct. I still don't know what the best method is, but for now, I'm using this hack:
```
function map_pci_memory (device, n, lock, wait)
   assert(lock == true or lock == false, "Explicit lock status required")
   root_check()
   local filepath = path(device).."/resource"..n
   local f,err  = S.open(filepath, "rdwr, sync")
   assert(f, "failed to open resource " .. filepath .. ": " .. tostring(err))
   if lock then
     assert(f:flock("ex, nb"), "failed to lock " .. filepath)
   end
   local st = assert(f:stat())
   local mem, err = f:mmap(nil, st.size, "read, write", "shared", 0)
   if wait then
      while err and err.errno == const.E.INVAL do
         mem, err = f:mmap(nil, st.size, "read, write", "shared", 0)
      end
   end
   assert(mem)
   return ffi.cast("uint32_t *", mem), f
end
```
The precondition for using the new `wait` flag is that the caller has done a `unbind_device_from_linux()`. One should probably use a timeout as well.

alexandergall:
Argh, and there is yet another race condition in `pci.device_info`. The `net` subdirectory vanishes when the device is unbound. So, in the code
```
      info.interface = lib.firstfile(p.."/net")
      if info.interface then
         info.status = lib.firstline(p.."/net/"..info.interface.."/operstate")
      end
```
if the unbinding happens after the `lib.firstfile` call but before the `lib.firstline` call, the latter will fail. This just happened to me.

petebristow:
Perhaps moving the lock out of `map_pci_memory` would make sense ? That way you can grab the lock unbind and then release and avoid this. I only used the device entry for the lock because at the time `/var/snabb/*` didn't have a particularly obvious place to put it.

alexandergall:
But if you release the lock you can't use it to determine who's the master, can you?

eugeneia:
What is the status on this?

alexandergall:
Still unsolved I believe. Haven't looked at it since then.

PR #1300, opened 2018-03-19
Make supervisor setup more reliable

wingo:
There was a race condition when setting up the supervisor such that in
some cases it was possible to miss a signal when the parent process
died.  We could reproduce this with by running a "snabb lwaftr bench",
but only on our test machine with two NUMA nodes and only when setting
--cpu on the lwaftr.  In that case the problem would appear when
running "snabb lwaftr monitor" on the lwaftr; the monitor process
would hang reading from the signalfd.  Because the monitor process
still had stdout open, then when piping its output to "grep", the grep
process would hang because the write side of its stdin pipe would
still be open as well.

This patch fixes this error by making cleanup reliable.  It does so by
taking a POSIX lock on an unnamed file in the parent, then taking
another lock from the supervisor child process.  In this way we avoid
some of the more arcane parts of Linux.

### DIFF
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -225,23 +225,23 @@ end
 -- Fork a child process that monitors us and performs cleanup actions
 -- when we terminate.
 local snabbpid = S.getpid()
+local lockfile = os.tmpname()
+local lock = S.open(lockfile, "wronly")
+S.unlink(lockfile)
+S.sigprocmask("block", "hup, int, quit, term")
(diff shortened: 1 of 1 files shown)


wingo:
See comment at https://github.com/Igalia/snabb/pull/1031#issuecomment-374173979 for more details about the failure.

eugeneia:
Interviewed @wingo about this on Slack. LGTM, merging onto `max-next`. :-)

Issue #1307, opened 2018-03-21
[raptorjit] Occasional segfault running lib.ctable selftest


wingo:
With the raptorjit branch from #1306, I get occasional segfaults running the lib.ctable selftest.  They are not deterministic.  Here is one of them that I managed to catch in GDB:

```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000452481 in lj_alloc_free (msp=0x7ffff7fcf010, ptr=<optimized out>) at lj_alloc.c:1266
1266		unlink_chunk(fm, next, nsize);
(gdb) bt
#0  0x0000000000452481 in lj_alloc_free (msp=0x7ffff7fcf010, ptr=<optimized out>) at lj_alloc.c:1266
#1  0x0000000000419bbd in gc_sweep (g=g@entry=0x7ffff7fcf3d8, p=0x7ffff7fe4b28, lim=35, lim@entry=40) at lj_gc.c:383
#2  0x000000000041a835 in gc_onestep (L=L@entry=0x7ffff7fcf378) at lj_gc.c:610
#3  0x000000000041af4d in lj_gc_step (L=L@entry=0x7ffff7fcf378) at lj_gc.c:659
#4  0x000000000045b4c0 in lj_cf_ffi_cast (L=0x7ffff7fcf378) at lib_ffi.c:542
#5  0x00000000004176c5 in lj_BC_FUNCC ()
#6  0x000000000045897a in lj_cf_package_require (L=0x7ffff7fcf378) at lib_package.c:322
#7  0x00000000004176c5 in lj_BC_FUNCC ()
#8  0x000000000045897a in lj_cf_package_require (L=0x7ffff7fcf378) at lib_package.c:322
#9  0x00000000004176c5 in lj_BC_FUNCC ()
#10 0x0000000000412dc8 in lua_pcall (L=<optimized out>, nargs=<optimized out>, nresults=<optimized out>, errfunc=<optimized out>) at lj_api.c:1074
#11 0x000000000040d413 in main ()
(gdb) bt full
#0  0x0000000000452481 in lj_alloc_free (msp=0x7ffff7fcf010, ptr=<optimized out>) at lj_alloc.c:1266
        F = 0x7ffff6ced1c8
        B = <optimized out>
        I = 0
        nsize = <optimized out>
        p = 0x7ffff6ced200
        fm = 0x7ffff7fcf010
        psize = 80
        next = 0x7ffff6ced250
[...]

PR #1308, opened 2018-03-21
Fix out-of-bounds write in ctable test suite

wingo:
This is another instance of the bug from commit 93ef6bdbbd92eab4b96790f825cefec3988dc65a.  We didn't see any issue on upstream Snabb's test suites, but with RaptorJIT's new LJ_GC64 usage did manifest itself as intermittent heap corruption.

Fixes https://github.com/snabbco/snabb/issues/1307.

See https://github.com/Igalia/snabb/pull/659 for a previous iteration of this problem in which I was also the culprit.  Aaaaaggghhhh!!!!!

### DIFF
--- a/src/lib/ctable.lua
+++ b/src/lib/ctable.lua
@@ -682,7 +682,8 @@ function selftest()
          -- keep references to avoid GCing too early
          local handle = {}
          local function read(size)
-            local buf = ffi.new('uint8_t[?]', size, file:read(size))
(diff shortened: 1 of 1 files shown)


wingo:
FYI as with the last time I did do a grep on our source tree for '[?]', to find VLA FFI types that might be initialized with this same antipattern.  I didn't find any more other than this one.

PR #1309, opened 2018-03-22
RSS App

alexandergall:
This app is intended to be used with the upcoming inter-process link mechanism from #1288. It provides a software-based RSS mechanism. The feature set was motivated by my work on the IPFIX app (not yet published), in particular the "flow director" and "packet replication" features (see README). The flow director is, I believe, supported by the Intel NICs but hasn't been implemented in the drivers.

The app itself can be scaled by using hardware RSS if needed. This is useful if a feature is used that isn't available in hardware (e.g. packet replication).
 
The "meta-data" mechanism stores data in the packet buffer, which is a somewhat questionable practice. Comments on this are welcome.

### DIFF
--- a/src/apps/rss/README.md
+++ b/src/apps/rss/README.md
@@ -68,20 +68,17 @@ processing of the list stops.
 The default class can be disabled by configuration. In that case,
 packets not assigned to any class are dropped.
 
-## Packet duplication
+## Packet replication
 
-As described above, the regular flow-director assigns a packet to at
-most one class.  Any class can also be marked with the attribute
-`continue` to allow matches to multiple classes.  When a packet is
-matched to such a class, it is distributed to the set of ouput links
-as usual but processing of the remaining filter expressions continues.
-For all subsequent matches, a copy of the packet is created and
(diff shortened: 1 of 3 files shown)


eugeneia:
@wingo seems to be another flaky lwaftr test failure? https://gist.github.com/SnabbBot/ec04788225ade5801630df6389d1506e#file-log-L819

eugeneia:
> The "meta-data" mechanism stores data in the packet buffer, which is a somewhat questionable practice. Comments on this are welcome.

VPP seems to have a generic [metadata field](https://docs.google.com/presentation/d/1qLiI0YckBqyIxDCQuihuL7NgnSfEShX1MhOZSP1LEQk/pub?start=false&loop=false&delayms=3000#slide=id.p26) for such purposes (source: https://www.asumu.xyz/blog/2018/03/22/how-to-develop-vpp-plugins/). Adding a similar multi-purpose field to `struct packet` would potentially enable optimizations in lib.ipsec. Food for thought.

eugeneia:
LGTM. Lovely documentation! I took the liberty and included it in `genbook.sh`

PR #1316, opened 2018-04-04
RaptorJIT branch for upstreaming

lukego:
This `snabbco/raptorjit` is now the master branch for integrating RaptorJIT into Snabb. (See #1315.)

Supersedes #1264.


lukego:
The latest blockers identified by @wingo at this point are:

> 1. Not so good perf on basic1-100e6; that could be statistical though. Better to use hydra here
> 2. A failure in the tap app's selftest. Seems to be legit. I didn't see it locally because I don't run with SNABB_TAPTEST.

eugeneia:
Merged into max-next 🎉

PR #1319, opened 2018-04-10
Improve performance when busywait=false on raptorjit branch

takikawa:
Edited: this PR improves performance on the raptorjit branch when busywait=false. This should make the benchmarks pass on CI.


takikawa:
I think this is made redundant by commit acb499aa7da3b845590f8d0b890ede6eb26e1049 which is already in v2018.04 so I'll close it.

PR #1323, opened 2018-04-17
Add a SNABB_RSS_HASH_KEY env variable for testing

takikawa:
Makes intel_mp tests more deterministic to avoid occasional test failures in CI.

### DIFF
--- a/src/apps/intel_mp/intel_mp.lua
+++ b/src/apps/intel_mp/intel_mp.lua
@@ -31,6 +31,10 @@ local S           = require("syscall")
 local CallbackAlarm = alarms.CallbackAlarm
 local transmit, receive, empty = link.transmit, link.receive, link.empty
 
+-- This environment variable is for use in tests, where a random hash
+-- key is often undesirable (e.g., to ensure all queues see traffic in a test)
+local rss_hash_key = tonumber(lib.getenv("SNABB_RSS_HASH_KEY"))
+
 -- It's not clear what address to use for EEMNGCTL_i210 DPDK PMD / linux e1000
 -- both use 1010 but the docs say 12030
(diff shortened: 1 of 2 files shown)


src/apps/intel_mp/intel_mp.lua
```diff
@@ -31,6 +31,10 @@ local S           = require("syscall")
 local CallbackAlarm = alarms.CallbackAlarm
 local transmit, receive, empty = link.transmit, link.receive, link.empty
 
+-- This environment variable is for use in tests, where a random hash
+-- key is often undesirable (e.g., to ensure all queues see traffic in a test)
+local rss_hash_key = tonumber(lib.getenv("SNABB_RSS_HASH_KEY"))
```
wingo: We already have an environment variable for making deterministic random numbers for tests: SNABB_RANDOM_SEED.

src/apps/intel_mp/intel_mp.lua
```diff
@@ -729,7 +733,11 @@ function Intel:rss_enable ()
 end
 function Intel:rss_key ()
    for i=0,9,1 do
-      self.r.RSSRK[i](math.random(2^32))
+      if rss_hash_key then
+         self.r.RSSRK[i](rss_hash_key)
+      else
+         self.r.RSSRK[i](math.random(2^32))
```
wingo: If we were to keep this, we should reduce this to one line (`rss_hash_key or math.random(2^32)`.  But I think instead we should make the RSS tests set SNABB_RANDOM_SEED.  WDYT?

takikawa: Yeah that sounds reasonable, I will try the change out.

takikawa:
#1326 is a simpler solution for this, so I'll close this PR.

Issue #1313, opened 2018-04-02
Sporadic intel_mp selftest failure (test_10g_2q_blast_vmdq_auto)


eugeneia:
SnabbBot is seeing some sporadic intel_mp selftest failures, like this one:

https://gist.github.com/SnabbBot/975b2337e97b7365c74bc9a922fd3ea1#file-log-L847

```
FAILED: ./test_10g_2q_blast_vmdq_auto.sh
```

Cc @asumu @petebristow

takikawa:
Thanks, I'll look into this. I've been trying to reproduce it more consistently to narrow it down but haven't had much luck so far. Will keep trying.

PR #1326, opened 2018-04-18
Set SNABB_RANDOM_SEED for intel_mp tests

wingo:
This should fix some flakiness in these tests, as they use a
random number as the RSS key.

See also https://github.com/snabbco/snabb/pull/1323 for an alternate take on the same problem.

### DIFF
--- a/src/apps/intel_mp/selftest.sh
+++ b/src/apps/intel_mp/selftest.sh
@@ -14,6 +14,7 @@ TESTS=$(echo "$TESTS1G" "$TESTS10G" | grep -e "$FILTER" | sort)
 ESTATUS=0
 export SNABB_RECV_DEBUG=true
 export SNABB_RECV_MASTER_STATS=true
+export SNABB_RANDOM_SEED=0xacabba9e
 for i in $TESTS; do
    pkill -P $$ -f snabb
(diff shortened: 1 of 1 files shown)


eugeneia:
Just for the record: the test failures were due to intel_app sometimes putting multiple test traffic streams into the same RSS bucket, am I right?

takikawa:
@eugeneia Yes, that's correct. That was one of the sources of intermittent test failures. It looks like there may be some intermittent failures with the non-RSS tests too (like [this one](https://gist.github.com/SnabbBot/1b02680410d2840b765c74d5825189fe#file-log-L846) where the 2 queue version is failing) that might not be solved by this, so I'll keep looking into that.

takikawa:
Addendum to previous comment: the 2 queue test failure was probably due to the random hash key too. It turns out the 2 queue vmdq tests also ends up exercising the RSS functionality (see https://github.com/snabbco/snabb/blob/master/src/apps/intel_mp/test_10g_2q_blast_vmdq.sh#L10 where both queues 0 and 1 are used for RX, meaning that RSS is set up to distribute between two queues). So I think issue #1313 can be closed now.

PR #1327, opened 2018-04-22
fix a bug in tcp fields

xnhp0320:
tcp.ack is the bit set in tcp flags
while tcp.ack_num should be the ack seq number.


eugeneia:
Hi @xnhp0320 :-)

88da49f and ae53f4a look good to me, but the other commits seem out of scope of this PR (especially 3c77fb2 is suspicious, is this for a kernel without hugetlbfs support?) I will cherry pick those two commits if that’s ok?

xnhp0320:
Oh, sorry!

The other patches are used to make snabb run on our own servers, which seems have a 
old version of libc that does not support HUGETLB flags. 

You can just use the two patches. 

Sorry for the late response.

eugeneia:
No problem at all. Merged, thank you!

eugeneia:
Closing because the requested changes landed in v2018.06

PR #1339, opened 2018-05-03
Add snabb config test functionality

benagricola:
@wingo as mentioned on Slack earlier 🙂

Reworks `snabb config` command line parsing slightly to implement `snabb config test`.

`snabb config test` loads a schema passed with `-s` either by name (if already known by snabb) or by file path, then validates a passed configuration FILE against the schema.

If no `-s` is given but `ID` is passed, then `config test` attempts to connect to the instance and grab the current schema.

Also modifies the `log()` function so that informational log messages like the ones below are only printed when verbose is passed to `yang.load_configuration`:

```
config/snabb.conf: loading compiled configuration from config/snabb.o
config/snabb.conf: compiled configuration is up to date.
```

_NOTE_: WIP as I haven't tested this against running snabb instances yet,  will complete this next week.

### DIFF
--- a/src/lib/yang/yang.lua
+++ b/src/lib/yang/yang.lua
@@ -53,8 +53,10 @@ function load_configuration(filename, opts)
    end
    local function err(msg, ...) error(err_msg(msg, ...)) end
    local function log(msg, ...)
-      io.stderr:write(err_msg(msg, ...)..'\n')
-      io.stderr:flush()
+      if opts.verbose then
+         io.stderr:write(err_msg(msg, ...)..'\n')
+         io.stderr:flush()
+      end
    end
    local function assert(exp, msg, ...)
       if exp then return exp else err(msg, ...) end
--- a/src/program/config/common.lua
+++ b/src/program/config/common.lua
@@ -73,6 +73,7 @@ function parse_command_line(args, opts)
    function handlers.s(arg) ret.schema_name = arg end
    function handlers.r(arg) ret.revision_date = arg end
    function handlers.c(arg) ret.socket = arg end
+   function handlers.v(arg) ret.verbose = true end
    function handlers.f(arg)
       assert(arg == "yang" or arg == "xpath", "Not valid output format")
       ret.format = arg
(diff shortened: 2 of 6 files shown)


src/lib/yang/yang.lua
```diff
@@ -53,8 +53,10 @@ function load_configuration(filename, opts)
    end
    local function err(msg, ...) error(err_msg(msg, ...)) end
    local function log(msg, ...)
-      io.stderr:write(err_msg(msg, ...)..'\n')
-      io.stderr:flush()
+      if opts.verbose then
```
wingo: I think we need to change the default above at line 34 to `verbose={default=true}`, to keep the current behavior for existing callers.

src/program/config/common.lua
```diff
@@ -73,37 +73,58 @@ function parse_command_line(args, opts)
...
                         ['print-default']=0, format="f"})
+
    if #args == 0 then err() end
-   ret.instance_id = table.remove(args, 1)
-   local descr = call_leader(ret.instance_id, 'describe', {})
-   if not ret.schema_name then
-      if opts.require_schema then err("missing --schema arg") end
-      ret.schema_name = descr.default_schema
+
+   if opts.command == 'test' then
```
wingo: I dunno, I think since there's these two ways to deal with the schema -- either as provided by the user or as received from the remote -- I think it makes sense to move all argument handling to `test.lua`, to avoid making this `common.lua` too specific on the commands being used.  WDYT?

wingo (commented):
Looking good :)  Some comments.  Also, do you think it ever makes sense to use a compiled file?  I have the feeling that we should always be "testing" source.  (Should this command be named "snabb config compile" instead, and make it compile a configuration instead?)

eugeneia:
FWIW there is an ad-hoc implementation of this functionality in vita so this will be very useful to me! :heart:

benagricola:
@wingo re: should we ever config test a compiled file - no, that's a good point 😃- I'll rework this as `config compile`.

lukego:
(JFYI I didn't merge this on my sweep through open PRs just now because based on the discussion it seems like changes are pending.)

benagricola:
Yep, pending a rewrite I'm hoping to get around to after more `$work` related things :D

PR #1353, opened 2018-06-15
intel_mp: fix per queue statistics for Intel1g and add VMDq support for i350

eugeneia:
- 2814b7b  fixes a real bug wrt per queue stats accumulation

- 9c4e3ee is my attempt at adding VMDq support for i350 NICs. I generally like the way that turned out and think it cleans things up, but there is a remaining issue I couldn’t figure out. Help appreciated. From the commit message:

> NB: the behavior of the test_1g_2q_blast_vmdq_auto.sh is fishy and there is
> probably a bug here. Output looks like this:
> 
> nic1:
> load: time: 5.00s  fps: 1,483,058 fpGbps: 0.819 fpb: 2   bpp: 60   sleep: 1   us
> GPRC 1488238
> RPTHC 1488238
> nic2:
> load: time: 5.00s  fps: 5,132     fpGbps: 0.003 fpb: 0   bpp: 60   sleep: 99  us
> 
> I.e., both pools receive packets (and the right ones too) but the second pool
> receives way too little. FIXME?

I have ported the other relevant VMDq related selftests and they all pass (I have made the “failing” test pass to by being a bit more lax on the condition).

Further question: for Vita, I ideally want non-VMDq based MAC and VLAN support for all NICs supported by the intel_mp driver. Would be nice to have basic MAC and VLAN filtering for all supported devices. WDYT?


lukego:
@eugeneia whose review(s) do we need on this one?

eugeneia:
@takikawa and/or @petebristow maybe? I suppose 2814b7b and 33c0912 should be straight forward and can be cherry picked. OTOH, 9c4e3ee is a bit of a bummer, its seemingly 99% there an I have no idea which needle in which haystack the remaining 1% is. Reproducing my observation on a different i350 might rule out a hardware issue on my side?

takikawa:
I started reviewing this a few days ago (sorry for the delay), but I don't actually have access to a suitable machine with useable i350 NICs to test any changes. Is there a machine somewhere I could use for that?

lukego:
I've been a very lazy lab tech. I need to reinstall the OS on davos. I will put that onto my TODO list for today.

lukego:
I didn't quite get through the OS reinstallation process on `davos` in the time available today. I'll come back to that tomorrow. I have the machine sitting at the installer prompt at the moment. cc @eugeneia

takikawa:
I looked into the issue with `test_1g_2q_blast_vmdq_auto.sh` and I am suspecting that part of the issue is on the tx side in the test script. In particular, `testsend.snabb` uses the `Repeater` app to repeatedly send the contents of the test pcap. The test pcap has half packets from one MAC and half with another to exercise VMDq.

Changing `testsend.snabb` by adding in a `Sample` app after the `Repeater` fixes the imbalance, with both NIC apps receiving roughly the same number of packets. Maybe `testsend.snabb` is currently dropping packets such that mostly only the first half of the pcap gets transmitted to the receive interfaces.

I'm having trouble connecting to davos right now, but I'll keep looking into this later.

src/apps/intel_mp/README.md
```diff
@@ -72,18 +72,19 @@ For a given NIC, all driver instances should have this parameter either
 enabled or disabled uniformly. If this is enabled, *macaddr* must be
 specified.
 
-— Key **vmdq_queuing_mode**
+— Key **vmdq_queueing_mode**
```
takikawa: This is pretty minor, but apparently "queuing" and "queueing" are both valid spellings so I'd prefer not to change it to avoid breaking any existing uses.

eugeneia: While the change is backwards compatible (accepts both spellings), a quick internet search indicates that “queuing” is actually the more common spelling nowadays. Seems I have trusted my spell-checker too much. I suppose I will revert this for simplicity.

eugeneia:
I have pushed 703eb9a to mask the flakyness I observe with `test_1g_vmdq_mirror.snabb`, not really satisfying but all I can think of right now. From my side this PR is ready to be merged.

eugeneia:
Eh, and I found (and fixed) another bug (d29763d), and added test cases to cover the faulty behaviour. :-)

eugeneia:
I think this is ready, and unless anyone objects I will act as my own upstream for this PR.

PR #1355, opened 2018-06-15
Add support for RDTSC, apply to token bucket

alexandergall:
The new module lib.tsc provides a generic Time Stamp Counter mechanism
for measuring time intervals based on different time sources.  One
such source makes use of the CPU's TSC register via the rdtsc
instruction for low-latency timing purposes.

The rdtsc method is made the default time source for the token bucket,
formerly provided by core.lib, now moved to a separate module
lib.token_bucket. The commit also includes documentation for the token
bucket, which was previously missing.


eugeneia:
- Maybe we should move lib.logger out from core/lib to lib/ to avoid core/lib repending on lib/?
- Could we re-use this token bucket implementation for `RateLimiter` and lwaftr’s `RateLimitedRepeater`?

alexandergall:
Yes to both, though I don't know what exactly is required for the rate limiter of lwaftr.

alexandergall:
I'd prefer to do the logger stuff in separate PRs, if that was your implication. That would mean that we have a temporary dependency of `core.lib` to `lib.token_bucket` until that's sorted out. Would that be a problem?

eugeneia:
No, I don’t think so. As long as I can merge both onto max-next to get the desired result. :-)

wingo:
FYI the lwAFTR's loadgen uses its own implementation, copied from the rate limiter: https://github.com/snabbco/snabb/blob/master/src/apps/lwaftr/loadgen.lua#L39.

Issue #1207, opened 2017-08-11
ctable performance issue related to pointer arithmetic


alexandergall:
I started to use `ctable` indirectly through the IPv6 fragmentation code from `apps.lwaftr`. I was surprised to see much less performance than expected on the reassembly side. The profiler showed substantial amounts of interpreted code and GC occuring in the fast path.

A trace dump revealed a lot of instances of trace aborts like this
```
...
0014  . . . . TGETS    4   0   3  ; "remove_ptr"
0015  . . . . MOV      6   3
0016  . . . . CALL     4   1   3
0000  . . . . . FUNCF    7          ; ctable.lua:392
0001  . . . . . TGETS    2   0   0  ; "scale"
0002  . . . . . TGETS    3   0   1  ; "entries"
0003  . . . . . SUBVV    3   1   3
0000  . . . . . . . FUNCC               ; ffi.meta.__sub
---- TRACE 125 abort ctable.lua:394 -- bad argument type
```
This led to blacklisting of crucial parts of the reassembly code by the compiler.

After a bit of googling, I found a report that described this symptom: (https://experilous.com/1/blog/post/lua-optimizations-first-pass):

_"This one was nasty, and the trace abort message looked nasty too: “bad argument type”. It turns out that LuaJIT really strongly dislikes taking the difference between two pointers that point to some type whose size is not a power of 2."_

The `remove_ptr` method of `ctable` uses precisely this kind of pointer arithmetic
```
function CTable:remove_ptr(entry)
   local scale = self.scale
   local index = entry - self.entries
...
```

Here, `entry` is an element of the hash table constructed by the `make_entry_type()` function in `lib.ctable`.

In case of the fragmenter, the size of such an element is not a power of 2. I then applied a simple patch that pads the element to the next power of two:
```
local function make_entry_type(key_type, value_type)
   local cache = entry_types[key_type]
   if cache then
      cache = cache[value_type]
      if cache then return cache end
   else
      entry_types[key_type] = {}
   end
   local raw_size = ffi.sizeof(key_type) + ffi.sizeof(value_type) + 4
  local padding = 2^ceil(math.log(raw_size)/math.log(2)) - raw_size
   local ret = ffi.typeof([[struct {
         uint32_t hash;
         $ key;
         $ value;
         uint8_t padding[$];
      } __attribute__((packed))]],
      key_type,
      value_type,
      padding)
   entry_types[key_type][value_type] = ret
   return ret
end
```
[...]

PR #1360, opened 2018-06-19
ctable: fix power-of-2 issue

alexandergall:
I have brought this up in issue https://github.com/snabbco/snabb/issues/1207, which is still open. I propose this fix at least as an interim solution until RaptorJIT lands (provided it will contain the fix proposed by @lukego).

Removal of two tail-call optimizations is also part of this PR. I found that this reduces trace aborts in my ipfix app some time ago. I'm not sure if we have converged on a common strategy to deal with this sort of problem, so this could be debatable.


alexandergall:
Making this optional is probably not a good idea.

PR #1349, opened 2018-06-01
Merge wingo-next for v2018.05

wingo:
Includes:

 * #1263 
 * #1338 
 * #1315 
 * #1288

PR #1361, opened 2018-06-19
next: Changes queue for v2018.06

lukego:
This pulls the changes aggregated on `max-next` (#1350) and `wingo-next` (#1349) and should be good to go!

I reckon we can pull RaptorJIT onto `next` immediately after this release. It's working well for a bunch of people now (as discussed on Slack) but I don't think we need to squeeze it into this release. Seems worth a little extra time to e.g. review whether the README/documentation needs to be updated etc.

(Sorry for the relatively few cycles spent on upstreamery lately... It's a busy summer for me...)


eugeneia:
@lukego This is good to go except my `lib.numa` fixes (#1269) are broken, so we should revert 
3120d7c.

lukego:
@eugeneia Revert pushed! Sorry about the wait.

eugeneia:
@lukego I am afraid I found another regression: I have bisected a SnabbNFV selftest failure down to https://github.com/snabbco/snabb/pull/1361/commits/d4c2eca9fbee860a438217832e6661f789aad524 (the merge commit, not thecommit to be merged.) So https://github.com/snabbco/snabb/pull/1361/commits/d4c2eca9fbee860a438217832e6661f789aad524 should be reverted from the release so I can figure out what is wrong with it.

eugeneia:
Actually, never mind my last comment. The test was non-deterministic during bisect so the result is likely bogus.

PR #1364, opened 2018-06-28
Micro-optimizations and a small fix

wingo:
This branch contains miscellaneous patches to quieten the `-jv` output: fewer nested loops, mostly.  Also there's a fix to alarm:raise() to handle the case where the user passes in args, and a commit to expand the limits for JIT mcode in workers, taken from l2vpn.


wingo:
Bah, opened against the wrong repo; sorry about that!

PR #1366, opened 2018-07-02
Minor fixes to docs

takikawa:
This PR just has two minor documentation fixes


eugeneia:
(There is no need to close merged PRs, they will auto-close once they end up on master via a release. The merged label is to indicate that its on one of the upstream branches.)

PR #1369, opened 2018-07-24
Make vhost-user more compatible with new qemu versions

takikawa:
This should address issue #1103 about using snabb with newer versions of qemu. I encountered the problem myself while testing an app, and got it working with qemu version 2.8.1 with these changes.

The main change is adding support for `VHOST_USER_F_PROTOCOL_FEATURES` to allow queries for protocol extensions. Right now the app just reports not supporting any extensions (supporting extensions is a TODO for the future) and ignores requests to set extensions. It also adds support for `VHOST_USER_SET_VRING_ENABLE` which seems to be needed once `VHOST_USER_F_PROTOCOL_FEATURES` is supported.

There's also a minor commit that makes it easier to tell when you've used the wrong link names.


lukego:
Awesome!

lukego:
I'm running a CI job now to regression test this patch with 13 different QEMU versions from 2.1 to 3.0rc. I'll pull it onto next once that's looking good.

lukego:
The [CI report](https://hydra.snabb.co/build/3585149/download/2/report.html) looks good. It did compatibility tests on a matrix of QEMU and DPDK (in guest) versions and the success coverage is much better with this branch than with master.

Here is the relevant graph followed by some explanatory notes.

download

 The top panel (`lukego-optimize`) is this code and the bottom panel (`master`) is the baseline. There seems to be a specific issue with QEMU 2.7.1 and a few DPDK versions but that's not a major concern for us. There are also a couple of cases where <100% success was observed but I suspect these were spurious timeouts rather than anything systematic.

Great stuff! I merged this onto my integration branch at #1372 now.

PR #1371, opened 2018-07-30
Add engine.stop() to stop all apps

lukego:
Add `engine.stop()` as a way to trigger the orderly shutdown of apps via `:stop()` callback. This is a compromise solution to #1276.

If you want your apps to all shutdown in an orderly way with their `stop()` methods called then you can achieve this by calling `engine.stop()`. If your Snabb process terminates abruptly (`kill -9`, segfault, etc) then naturally this call will not be made and the stop methods will not run. (If you *really* need the shutdown actions to run then you can arrange for our separate supervisor process to do that, see #1276.)

Resolves #1276.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -258,6 +258,10 @@ following keys are recognized:
    printed.
 
 
+— Function **engine.stop**
+
+Stop all apps in the engine by loading an empty configuration.
+
 — Function **engine.now**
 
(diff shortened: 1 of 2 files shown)


lukego:
I'm taking the liberty of upstreaming this one myself since many maintainers have their hands full this summer.

Issue #1210, opened 2017-08-18
mmap: Invalid argument


leolovenet:
@lukego  When i use the latest version Snabb on CentOS6(2.6.32-696.3.2.el6.x86_64), I get the [error.txt](https://github.com/snabbco/snabb/files/1233265/error.txt) message, then I traced why，
I found it because `local tmpptr = syscall.mmap(nil, size, "read, write", "shared, hugetlb", fd, 0)` in [memory.lua](https://github.com/snabbco/snabb/blob/master/src/core/memory.lua#L149)， return `nil`, Then I look at the [history](https://github.com/snabbco/snabb/commits/master/src/core/memory.lua) of this file, Found that you rewrite the function `allocate_huge_page ` using Lua for supporting multiprocess, But, this function return `Invalid argument` with flags `shared, hugetlb`,  buf it is OK with flags `shared, anonymous, hugetlb`, I tested. 

At first I suspect that is the kernel version of the problem, So I changed the  [code](http://elixir.free-electrons.com/linux/v2.6.32.71/source/Documentation/vm/map_hugetlb.c) to test it under ubuntu 16.04 (4.4.0-63-generic).  But, It return the same, `mmap: Invalid argument` when flags was `shared, hugetlb`.

So the problem is to use MAP_ANONYMOUS to conflict with your design, What should i do ?

leolovenet:
I figured it out,  when i readed the code https://github.com/torvalds/linux/blob/master/tools/testing/selftests/vm/hugepage-mmap.c, Found no `MAP_HUGETLB` flag required, so I remove the `hugetlb` flage too, and It work. 
 :)

memory.lua，Should be like the following:

[149](https://github.com/snabbco/snabb/blob/master/src/core/memory.lua#L149): `local tmpptr = syscall.mmap(nil, size, "read, write", "shared", fd, 0)`
[154](https://github.com/snabbco/snabb/blob/master/src/core/memory.lua#L154): `local ptr = syscall.mmap(virt, size, "read, write", "shared, fixed", fd, 0)`

BTW, I think this is a bug.

dpino:
The `hugetlb` flag requests to mmap a region of memory as a hugepage. For some reason requesting a mmap as a hugepage is not working in your system. It seems from the error report you attached that hugepages are enabled in your system. Snabb was at least able to allocate some hugepages and then failed:

```
[memory: Provisioned a huge page: sysctl vm.nr_hugepages 503 -> 504]
[...]

PR #1372, opened 2018-07-30
Collection of ripe PRs

lukego:
This is an integration branch where I have merged a few PRs that look ripe for upstreaming.

Does anybody have objections to landing all of this on `next` provided that CI passes this branch?

Included PRs:

- #1367 `eugeneia/numa-selftest-nonnuma-fixed`
- #1362 `alexandergall/ctable-fix-non-power-of-2-size`
- #1359 `alexandergall/ctable-resize-callback`
- #1354 `eugeneia/pmu-amd`
- #1351 `eugeneia/core-memory-unmap-tmpptr`
- #1344 `benagricola/fix-yang-length-restriction`

PR #1373, opened 2018-07-31
Fixes: #1210 hugetlb parameter causes problem on Centos and since

kullanici0606:
memory is already mapped on hugetlbfs and with mlock, there is no need to explicitly use hugetlb parameter
(See: https://github.com/snabbco/snabb/issues/1210#issuecomment-409123442)


lukego:
Thanks! I've merged this onto my integration branch #1372 and I'll land that on `next` once I've tested that a bit more.

kullanici0606:
You're welcome. Thank you for this great project.

Issue #1352, opened 2018-06-12
Archive /var/run/snabb directory in case of crash


wingo:
The /var/run/snabb/PID directory contains counter data that can be useful in post-mortem debugging.  With raptorjit, we'll also have the auditlog as well; and in the future we'll also have historical counter values (https://github.com/Igalia/snabb/pull/1088), if enabled.

If a process crashes, we should arrange for the supervisor to archive the /var/run/snabb/PID directory somewhere; perhaps /var/run/snabb/crashes/XXXXXX or something.

PR #1378, opened 2018-08-29
Archive /var/run/snabb/PID directory in case of crash

dpino:
Solves #1352.

### DIFF
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -163,10 +163,40 @@ function initialize ()
    _G.main   = getfenv()
 end
 
+local function archive (src)
+   local function isdir (path)
+      local stat = S.stat(path)
+      return stat.isdir
+   end
+   local function copy (src, dst)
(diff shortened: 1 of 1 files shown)


wingo (commented):
I think this approach won't really solve the use case because there won't be any interesting counters any more.  Firstly I should say that I'm mostly concerned about ptree workers; handling the general case is possible but we definitely have to handle the ptree worker case.

In that regard, this approach won't work; often exits are orderly, and in that case the handler isn't called.  Disorderly exits don't always neatly propagate through an exception; if we want to handle that case, we should archive from the supervisor.

However more to the point, ptree does an `engine.configure(config.new())` before stopping, so all of the apps get shut down in an orderly way.  This will remove the counters and RRD files that we wanted to archive!

Archiving should also be optional.  Right now we won't cause the disk to fill up.  Systematically archiving could cause us to enter this state.  Consider controlling via an environment variable.

Also probably we should only archive `.counter` and `.rrd` files.

PR #1292, opened 2018-02-28
[WIP] Fibers

wingo:
This PR currently includes a timer wheel, a scheduler, and a fibers implementation, layered in that order.  Fibers are a lightweight concurrency facility.  See https://github.com/wingo/fibers/wiki/Manual for more documentation on what fibers could be.  This is a port from Scheme.

A to-do list, in some order:

- [ ] Implement IO for TCP sockets.  Depends on having a TCP implementation (is elsewhere).
- [x] Implement concurrent ML-like operations for inter-fiber synchronization.
- [x] Implement channels for inter-fiber communication.
- [ ] Add documentation.
- [ ] Improve API to be easier to use.
- [ ] Add implementation of a subset of HTTP to implement a fiber-based web server with user-space I/O.
- [x] Implement IO for non-blocking file descriptors, maybe.

My interest is using fibers to implement TCP services in Snabb.  This is just a WIP that I don't know if I will ever finish :)  Comments are welcome though!

PR #1379, opened 2018-09-04
Merge lwaftr v2018.06.01 to upstream Snabb

wingo:
This PR incorporates the 2018.04.02 and 2018.06.01 lwAFTR releases into Snabb.

Besides the many lwAFTR-specific changes detailed in https://github.com/Igalia/snabb/blob/lwaftr/src/program/lwaftr/doc/CHANGELOG.md#20180402 and https://github.com/Igalia/snabb/blob/lwaftr/src/program/lwaftr/doc/CHANGELOG.md#20180601, notable Snabb-relevant changes include:

 * Rewrite of Lua's file streams in Lua: https://github.com/snabbco/snabb/pull/1341

   Before, it was impossible to provide a new implementation of file-like objects without completely reimplementing the entire Lua API.  This is now fixed: new stream implementations just need to implement "read" and "write" methods; the buffering is handled by the common stream implementation.  This allows for streams that compress or decompress on the fly, or streams to or from bytes or strings in memory.  It also allows us to extend the API of Lua's streams is such a way that takes advantage of the buffering facilities, and to cover new areas such as efficient binary I/O.

   All I/O in the yang, ctable, and ptree code has been rewritten to use the new interfaces that extend the core Lua stream API.  The result is less overall memory use, as we can perform streaming parse and serialize operations instead of building up large string objects.

   See XXX for documentation.

 * Addition of Fibers, a lightweight concurrency facility:  https://github.com/snabbco/snabb/pull/1292

   Fibers allows for lightweight coroutines that communicate using primitives inspired by Concurrent ML.  Fibers also integrate with the new streams-in-Lua, suspending a fiber when a file read or write operation would block, and resuming it when the stream becomes readable or writable again.

   See XXX for documentation.

 * Addition of round-robin database library: https://github.com/Igalia/snabb/pull/1088
    
   This library allows us to easily record historical counter change rates.  Using the RRD format also allows us to use rrdtool to generate graphs.

   For more information: https://github.com/Igalia/snabb/blob/lwaftr/src/lib/README.rrd.md

 * Ptree manager aggregates worker counters, records historical counter change rates

   Aggregating counters allows operators to easily see at a glance how many packets per second are flowing through a multi-process Snabb network function.  The manager also records historical counter change rates to RRD files, allowing operators to see when interesting events occurred in the recent past.

   See https://github.com/Igalia/snabb/blob/lwaftr/src/lib/ptree/README.md for more information.

 * Rewrite of "snabb top" on top of fibers and RRD.

   The new "snabb top" monitors all Snabb instances on a machine.  Because it uses fibers, it can react to user input, restrict display to subsets of all information, and react and to window size changes.  It also skim back in time to show historical counter change rates.

   See https://github.com/Igalia/snabb/blob/lwaftr/src/program/top/README for more information.

 * Implement YANG alarm notifications: https://github.com/Igalia/snabb/pull/1055

   Operators can now receive notifications when a data-plane raises a YANG alarm, via the new "snabb alarms listen" subprogram.  See https://github.com/Igalia/snabb/blob/lwaftr/src/program/alarms/README.md#Notifications for details.

 * Add "influxdb" output format for "snabb config get": https://github.com/Igalia/snabb/pull/1058

   Pass `--format influxdb` to your `snabb config get` or `snabb config get-state` invocations.

 * Relax default NUMA memory binding policy: https://github.com/Igalia/snabb/pull/1072

   To allow for workloads that exceed available NUMA-local memory, such as multi-gigabyte routing tables, relax the default memory binding policy from "bind" to "preferred".  This prevents page faults from silently killing a process with SIGBUS.

 * Fix TTL in ICMPv4 and ICMPv6 echo response packets: https://github.com/Igalia/snabb/pull/1073

 * Improvements to Snabb's YANG support

   Snabb's YANG libraries now support the "leafref" data type.  See https://github.com/Igalia/snabb/pull/827 for more details.


eugeneia:
Looking great!

> Snabb's YANG libraries now support uniqueness constraints for YANG data,

I tested this by adding a unit test

```diff
modified   src/lib/yang/data.lua
@@ -1914,6 +1914,45 @@ function selftest()
       length_test "++++++++++++++++++++++";
    ]])
 
+   -- Test uniqueness restrictions.
+   local unique_schema = schema.load_schema([[module unique-schema {
+      namespace "urn:ietf:params:xml:ns:yang:unique-schema";
+      prefix "test";
+
+      list unique_test {
+         key "testkey"; unique "testleaf";
+         leaf testkey { type string; mandatory true; }
+         leaf testleaf { type string; mandatory true; }
+      }
+   }]])
+
+   -- Test unique validation (should fail)
+   local success, err = pcall(load_config_for_schema, unique_schema,
+                              mem.open_input_string [[
+      unique_test {
+         testkey "foo";
+         testleaf "bar";
+      }
+      unique_test {
+         testkey "foo2";
+         testleaf "bar";
+      }
+   ]])
+   assert(success == false)
+
+   -- Test unique validation (should succeed)
+   load_config_for_schema(unique_schema,
+                          mem.open_input_string [[
+      unique_test {
+         testkey "foo";
+         testleaf "bar";
+      }
+      unique_test {
+         testkey "foo2";
+         testleaf "bar2";
+      }
+   ]])
+
    influxdb_printer_tests()
 
    print('selfcheck: ok')
```

where “Test unique validation (should fail)“ does not fail. I had expected it to catch that `testleaf` is not unique.

Another minor nitpick I have is that `snabb bot` no longer supports dumb terminals, although I suspect I am the only person to care about that. It would be neat if `snabb top` had a mode (`--dumb`?) where it dumps the available information without requiring “advanced” terminal capabilities (also so you can do stuff like `sudo ./snabb top > foo.txt` which currently crashes `snabb top`).

wingo:
Thanks for the test case!

FWIW the consistency checker is actually run from `lib.yang.yang.load_config_for_schema` -- i.e. the `lib.yang.yang` module, not the `lib.yang.data` module.  This is because of some circularities between the`lib.yang.data` and `lib.yang.path_data` module.  Choosing the right place to run the constency checker wasn't obvious to us -- do you think it should be run elsewhere?

Regarding `snabb top`, I entirely agree that it should support piping to a file. It should detect whether the output is a TTY or not and act accordingly.  You can see there's the beginning of that code there but it's not yet implemented; I think it should be, though!

eugeneia:
> FWIW the consistency checker is actually run from...

I suppose I don’t mind as long as it is invoked from the public API in `lib.yang.yang`, moving the test case there doesn’t make it pass though. I suspect the uniqueness check is only done for key leaves?

wingo:
@eugeneia It looks like I actually misunderstood one of the patches, looking back... The thing that I thought was uniqueness constraints was something else, relating to influxdb output syntax (https://github.com/Igalia/snabb/pull/1058#issue-185456848, see the definition of ULEAF).  Uniqueness constraints are not yet implemented (but the consistency checker is the right place for it i think).  I will revise the PR description.  Thank you for the test case and apologies for the noise!

eugeneia:
I have some reservations about the new `snabb top`:

- it lacks the often used  (by me) `-l` feature to list arbitrary SHM objects (I think this could be added back in easily though)
- when I tested it with vita it turned out to be somewhat awkward to use because I was constantly struggling to see statistics that fell beyond the truncation threshold

As it is, this program is not a replacement for `snabb top` to me. I propose renaming it to `snabb monitor` (for example) and treating it as an addition instead of a replacement for `snabb top`. WDYT?

I have some ideas for the old `snabb top` to make it more useful alongside this monitoring program.

wingo:
Following a Slack conversation, it seems that removing the old "snabb top" was an error -- @eugeneia still uses it in some ways.  However we think that the new "snabb top" is more like "top" -- and the parts of the old one that were useful were more like "ps".  This PR also removes "snabb ps" incidentally -- so we've freed up that name.  Therefore the proposal is to re-add the old "snabb top", but as "snabb ps".

eugeneia:
Superseded by https://github.com/snabbco/snabb/pull/1401

PR #1380, opened 2018-09-05
Ensure vhost-user socket paths are NUL-terminated

wingo:
Found via GCC 8.2's -Werror=stringop-truncation.  NUL termination is required by the AF_UNIX socket API.


eugeneia:
LGTM!

PR #1384, opened 2018-09-20
Add expiration of incomplete reassembly buffers to IPv6

alexandergall:


src/apps/ipv6/reassemble.lua
```diff
@@ -264,24 +284,27 @@ function Reassembler:handle_fragment(h)
...
-   local max_data_offset = ether_ipv6_header_len + frag_start + frag_size
-   if max_data_offset > ffi.sizeof(reassembly.packet.data) then
-      -- Snabb packets have a maximum size of 10240 bytes.
-      return self:reassembly_error(entry)
+   -- Limit the scope of max_data_offset
+   do
+      local max_data_offset = ether_ipv6_header_len + frag_start + frag_size
+      if max_data_offset > ffi.sizeof(reassembly.packet.data) then
+         -- Snabb packets have a maximum size of 10240 bytes.
+         return self:reassembly_error(entry)
```
dpino: Maximum size of a packet is defined as a constant in `core/packet.lua:max_payload`.

src/apps/ipv6/reassemble.lua
```diff
@@ -264,24 +284,27 @@ function Reassembler:handle_fragment(h)
       else
          reassembly.final_start = frag_start
       end
-   elseif frag_size % 8 ~= 0 then
+   elseif bit.band(frag_size, 0x7) ~= 0 then
```
dpino: This changed isn't justified. I benchmarked % and "mod with logical and" and they're the same speed. What's indeed much slower is `math.fmod`. I think this change should be reverted.

alexandergall: Ok, I assume that `%` optimizes for this case?

src/apps/ipv6/reassemble.lua
alexandergall: Note that this is from the original code (I only added a `do...end` to limit the scope of locals). This would be a cosmetic change, I suppose?

dpino: Sorry I didn't realize this wasn't a new change. Rather than cosmetic, the difference is that the size is being computed every time this function is called, when it can be retrieved from a constant. I leave it up to you if you'd like to change it.

src/apps/ipv6/reassemble.lua
dpino: The results I got from benchmarking % and & showed no difference. Here's a gist of the code I tried https://gist.github.com/dpino/925a79494d3da27e3681e8b3acf67b32. I don't know what LuaJIT compares % to, though.

src/apps/ipv6/reassemble.lua
alexandergall: I'd have expected that this would be taken care of by the optimizer, but if it's not, that change would certainly be a good idea.

src/apps/ipv6/reassemble.lua
alexandergall: I just checked and the compiler does indeed specialize to bitops if the module is a power of 2.

eugeneia:
@alexandergall If you think this is ready I can I merge this onto max-next.

alexandergall:
Pushed the changes suggested by @dpino. Ready to merge for me.

PR #1385, opened 2018-09-20
Move logging facility to a separate module

alexandergall:
The code is moved from `core.lib` to `lib.logger`. 

Somewhat unrelated, the PR also adds caching or the calibration of the `rdtsc` time source in `lib.tsc` to avoid delays in some selftests.


eugeneia:
LGTM! (See #1355)

PR #1388, opened 2018-10-11
jit/bcsave: increment expected jit.version_num to match RaptorJIT

eugeneia:
I had to increment these to make the build work. Not sure if this is the right strategy or if this should be a `>=` instead. I guess it depends on whether that version number changes frequently or not?

Cc @lukego

### DIFF
--- a/lib/luajit/src/jit/bc.lua
+++ b/lib/luajit/src/jit/bc.lua
@@ -41,7 +41,7 @@
 
 -- Cache some library functions and objects.
 local jit = require("jit")
-assert(jit.version_num == 10000, "LuaJIT core/library version mismatch")
+assert(jit.version_num == 100000, "LuaJIT core/library version mismatch")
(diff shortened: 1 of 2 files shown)


lukego:
Hm. This appeared in commit 420fa7571432ba787aefa6a6e746adf1069762af. That changed `bc.lua`, `bcsave.lua`, and `luajit.h`. Looks to me like the real mistake was in `luajit.h` i.e. one trailing zero too many. Would it work to remove to change that instead so that all files say `10000`? This would symbolically mean `01.00.00` whereas `100000` would be `10.00.00` i.e. RaptorJIT 10.0.

eugeneia:
Works for me!

lukego:
Is this resolved by the merge of #1389 that should include an equivalent fix? I also added test coverage for bytecode version in upstream RaptorJIT.

PR #1390, opened 2018-10-23
core.packet: release packets to group freelist on shutdown

eugeneia:
In multi-process Snabb DMA huge pages are only released to the operating system
once the group’s parent shuts down. This commit makes it so that packets
allocated by workers are released into the group freelist when a worker shuts
down. This addresses two issues:

 - workers can have packets allocated by other workers on their freelist at the
   time of their shut down
 - DMA huge pages used for packets owned by workers that have shut down would
   otherwise leak (i.e. not be available to the OS and other processes until
   the group parent shuts down)

4da22e6  intends to handle both issues by recycling all allocated packets
within the process group (to match the DMA huge page management model.)

58fa77f is a minor, barely related fix.

Cc @lukego

### DIFF
--- a/src/core/main.lua
+++ b/src/core/main.lua
@@ -178,6 +178,7 @@ function shutdown (pid)
       if not ok then print(err) end
    end
    -- Run cleanup hooks
+   safely(function () require("core.packet").shutdown(pid) end)
    safely(function () require("apps.interlink.receiver").shutdown(pid) end)
    safely(function () require("apps.interlink.transmitter").shutdown(pid) end)
    -- Parent process performs additional cleanup steps.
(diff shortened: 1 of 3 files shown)


eugeneia:
@wingo I think this impacts lib.ptree’s configuration model: if a worker is stopped due to a configuration update, I believe its packets would leak without this change?

lukego:
Have to think about this one a bit. What kind of states can packets be in, how do we identify the packets in each state when a process terminates/crashes, what is required for each state to be safe to reuse?

States would include:

1. Packets on the process freelist.
2. Packets on inter-process links.
3. Packets on intra-process links in the local app network.
4. Packets in private use by apps e.g. queued by device drivers for DMA.

This branch does seem like a step in the right direction anyway i.e. when a worker process terminates the packets on its freelist should be fair game for reuse.

eugeneia:
Yep!

PR #1394, opened 2018-12-14
intel_mp: fix ctype diversity for txdesc_t

alexandergall:
Move the ctype declaration of txdesc_t out of the init_tx_q() method
to make its ctype identical in all instances of the driver.  This
eliminates certain side traces in the transmit code path.

### DIFF
--- a/src/apps/intel_mp/intel_mp.lua
+++ b/src/apps/intel_mp/intel_mp.lua
@@ -545,6 +545,8 @@ function Intel:init_rx_q ()
    end
    self:unlock_sw_sem()
 end
+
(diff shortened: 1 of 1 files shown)


eugeneia:
Minor nitpick: maybe it should still be a `local` variable?

alexandergall:
I used the same style as for other such definitions, e.g. https://github.com/snabbco/snabb/blob/8f7678224586746885e798c778da9a5b8f505bdc/src/apps/intel_mp/intel_mp.lua#L469

I'm not sure if that was deliberate either (do we need to expose these to users of the module?)

PR #1397, opened 2018-12-14
A little generic poptrie implementation

eugeneia:
This branch contains the beginning of an implementation of [Poptrie](http://conferences.sigcomm.org/sigcomm/2015/pdf/papers/p57.pdf). So far it can...

 - build the poptrie lookup structure for prefixes of up to 64 bits
 - look up keys in the poptrie fairly quick using a DynASM based lookup routine
 - be instantiated using a pre-built poptrie structure

Performance on an Intel(R) Xeon(R) CPU E3-1246 v3 @ 3.50GHz (Haswell, Turbo off)

```
PMU analysis (numentries=10000, keysize=32)
build: 0.1290 seconds
lookup: 13217.09 cycles/lookup 28014.35 instructions/lookup
lookup64: 122.94 cycles/lookup 133.22 instructions/lookup
build(direct_pointing): 0.1056 seconds
lookup(direct_pointing): 5519.01 cycles/lookup 11412.01 instructions/lookup
lookup64(direct_pointing): 89.82 cycles/lookup 70.72 instructions/lookup
```

Not implemented yet:

- ~~prefixes/keys of up to 128 bits (i.e., supporting IPv6)~~

Cc @petebristow

### DIFF
--- a/src/lib/poptrie_lookup.dasl
+++ b/src/lib/poptrie_lookup.dasl
@@ -19,7 +19,6 @@ function generate (Poptrie, keysize)
    -- Assert assumptions about lib.poptrie
    assert(Poptrie.k == 6)
    if Poptrie.direct_pointing then
-      assert(Poptrie.s == 18)
       assert(Poptrie.leaf_tag == bit.lshift(1, 31))
    end
    assert(ffi.sizeof(Poptrie.leaf_t) == 2)
@@ -66,9 +65,10 @@ end
 -- lookup(leaf_t *leaves, node_t *nodes, key) -> leaf_t
 function lookup (Dst, Poptrie, keysize)
    if Poptrie.direct_pointing then
-      -- v = extract(key, 0, s=18)
-      -- v = band(key, lshift(1, s=18) - 1)
-      | mov v, 0x3FFFF
+      -- v = extract(key, 0, Poptrie.s)
+      local direct_mask = bit.lshift(1ULL, Poptrie.s) - 1
+      -- v = band(key, direct_mask)
+      | mov v, direct_mask
       | and v, key
       -- index = dmap[v]
       | mov index, dword [dmap+v*4]
--- a/src/lib/poptrie_lookup.dasl
+++ b/src/lib/poptrie_lookup.dasl
@@ -1,0 +1,127 @@
+-- Use of this source code is governed by the Apache 2.0 license; see COPYING.
+
+module(..., package.seeall)
+
+local debug = false
+
+local ffi = require("ffi")
+local dasm = require("dasm")
+
+|.arch x64
+|.actionlist actions
+|.globalnames globalnames
+
+-- Table keeping machine code alive to the GC.
+local anchor = {}
+
+-- Assemble a lookup routine
+function generate (Poptrie, keysize)
+   local offsets = 
+   -- Assert assumptions about lib.poptrie
+   assert(Poptrie.k == 6)
+   assert(ffi.sizeof(Poptrie.leaf_t) == 2)
+   assert(ffi.sizeof(Poptrie.vector_t) == 8)
+   assert(ffi.sizeof(Poptrie.base_t) == 4)
+   assert(ffi.offsetof(Poptrie.node_t, 'leafvec') == 0)
+   assert(ffi.offsetof(Poptrie.node_t, 'vector') == 8)
+   assert(ffi.offsetof(Poptrie.node_t, 'base0') == 16)
+   assert(ffi.offsetof(Poptrie.node_t, 'base1') == 20)
+
+   local name = "poptrie_lookup(k="..Poptrie.k..", keysize="..keysize..")"
+
+   local Dst = dasm.new(actions)
+   lookup(Dst, Poptrie, keysize)
+   local mcode, size = Dst:build()
+   table.insert(anchor, mcode)
+
+   if debug then
+      print("mcode dump: "..name)
+      dasm.dump(mcode, size)
+   end
+
+   local prototype
+   if keysize <= 64 then
+      prototype = ffi.typeof("$ (*) ($ *, $ *, uint64_t)",
+                             Poptrie.leaf_t, Poptrie.leaf_t, Poptrie.node_t)
+   else error("NYI") end
+
+   return ffi.cast(prototype, mcode)
+end
+
+|.define leaves,  rdi -- pointer to leaves array
+|.define nodes,   rsi -- pointer to nodes array
+|.define key,     rdx -- key to look up
+|.define index,   r8d -- index into node array
+|.define node,    r8  -- pointer into node array 
+|.define offset,  r9  -- offset into key
+|.define v,       r10 -- k bits extracted from key
+|.define vec,     r11 -- 64-bit vector or leafvec
+
+-- lookup(leaf_t *leaves, node_t *nodes, key) -> leaf_t
+function lookup (Dst, Poptrie, keysize)
+   -- index, node, offset = 0, nodes[index], 0
+   | mov index, 0
+   | lea node, [nodes+0] -- nodes[0]
+   | mov offset, 0
+   -- while band(vec, lshift(1ULL, v)) ~= 0
+   |1:
+   -- v = extract(key, offset, k=6)
+   if keysize <= 64 then
+      -- v = rshift(key, offset)
+      | mov v, key
+      | mov rcx, offset
+      | shr v, cl
+      -- v = band(v, lshift(1, k=6) - 1)
+      | and v, 63
+   else error("NYI") end
+   -- vec = nodes[index].vector
+   | mov vec, qword [node+8]
+   -- rax = lshift(1ULL, v)
+   | mov rax, 1
+   | mov rcx, v
+   | shl rax, cl
+   -- rax = band(vec, rax)
+   | and rax, vec
+   -- is bit v set in vec?
+   | cmp rax, 0
+   | je >2 -- reached leaf, exit loop
+   -- rax = lshift(2ULL, v) - 1
+   | mov rax, 2
+   | mov rcx, v
+   | shl rax, cl
+   | sub rax, 1
+   -- rax = popcnt(band(vec, rax))
+   | and rax, vec
+   | popcnt rax, rax
+   -- index = base + bc - 1
+   | mov index, dword [node+20] -- nodes[index].base1
+   | add index, eax
+   | sub index, 1
+   -- node = nodes[index]
+   | imul index, 24 -- multiply by node size
+   | lea node, [nodes+index]
+   -- offset = offset + k
+   | add offset, 6
+   | jmp <1 -- loop
+   -- end while
+   |2:
+   -- rax = lshift(2ULL, v) - 1
+   | mov rax, 2
+   | mov rcx, v
+   | shl rax, cl
+   | sub rax, 1
+   if Poptrie.leaf_compression then
+      -- vec = nodes[index].leafvec
+      | mov vec, qword [node+0]
+   else error("NYI") end
+   -- rax = popcnt(band(vec, rax)) - 1
+   | and rax, vec
+   | popcnt rax, rax
+   -- return leaves[base + bc - 1]
+   | mov index, dword [node+16] -- nodes[index].base0
+   | add index, eax
+   | sub index, 1
+   -- assumption: at most ax is set at this point because popcnt(qword) < byte
+   | mov ax, word [leaves+index*2] -- leaves[index]
+   | ret
+end
--- a/src/lib/poptrie_lookup.dasl
+++ b/src/lib/poptrie_lookup.dasl
@@ -74,7 +74,8 @@ function lookup (Dst, Poptrie, keysize)
       | mov index, dword [dmap+v*4]
       -- eax = band(index, leaf_tag - 1) (tag inverted)
       | mov eax, index
-      | btr eax, 31 -- is leaf_tag set?
+      -- is leaf_tag set? (unsets bit)
+      | btr eax, 31
       | jnc >1 -- leaf_tag not set, index is a node
       | ret
       -- node, offset = nodes[index], s
@@ -102,7 +103,8 @@ function lookup (Dst, Poptrie, keysize)
    else error("NYI") end
    -- vec = nodes[index].vector
    | mov vec, qword [node+8]
-   | bt vec, v -- is bit v set in vec?
+   -- is bit v set in vec?
+   | bt vec, v
    | jnc >3 -- reached leaf, exit loop
    -- rax = lshift(2ULL, v) - 1
    | mov rax, 2
--- a/src/lib/poptrie_lookup.dasl
+++ b/src/lib/poptrie_lookup.dasl
@@ -4,6 +4,7 @@ module(..., package.seeall)
 
 local debug = false
 
+local lib = require("core.lib")
 local ffi = require("ffi")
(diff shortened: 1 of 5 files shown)


src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+      | ret
+      -- node, offset = nodes[index], s
+      |1:
+      | imul index, 24 -- multiply by node size
+      | lea node, [nodes+index]
+      -- offset = s
+      | mov offset, 18
+   else
+      -- index, node, offset = 0, nodes[index], 0
+      | mov index, 0
```
corsix: Unless you're explicitly preserving flags (which I assume you're not), `xor R, R` is typically better than `mov R, 0` (e.g. because it is a byte or two shorter to encode).

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+      | mov offset, 0
+   end
+   -- while band(vec, lshift(1ULL, v)) ~= 0
+   |2:
+   -- v = extract(key, offset, k=6)
+   if keysize <= 64 then
+      -- v = rshift(key, offset)
+      | mov v, key
+      | mov rcx, offset
+      | shr v, cl
```
corsix: If you can assume BMI2 (i.e. Haswell or later IIRC), this `mov`/`mov`/`shr` can probably be expressed as a single `shrx`.

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+   -- vec = nodes[index].vector
+   | mov vec, qword [node+8]
+   -- rax = lshift(1ULL, v)
+   | mov rax, 1
+   | mov rcx, v
+   | shl rax, cl
+   -- rax = band(vec, rax)
+   | and rax, vec
+   -- is bit v set in vec?
+   | cmp rax, 0
```
corsix: This `cmp` is entirely redundant - the immediately preceding `and` will already have set `ZF=1` if the result was zero.

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+   | je >3 -- reached leaf, exit loop
+   -- rax = lshift(2ULL, v) - 1
```
corsix: If you can assume BMI2, then this construction of `rax` and subsequent `and` can be replaced by something like:
```
lea rcx, [v+1]
bzhi rax, vec, rcx
```

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+      -- vec = nodes[index].leafvec
+      | mov vec, qword [node+0]
+   else error("NYI") end
+   -- rax = popcnt(band(vec, rax)) - 1
+   | and rax, vec
+   | popcnt rax, rax
+   -- return leaves[base + bc - 1]
+   | mov index, dword [node+16] -- nodes[index].base0
+   | add index, eax
+   | sub index, 1
```
corsix: This `sub` can be fused into the subsequent memory operand (i.e. `word [leaves+index*2-2]`).

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+   -- assumption: at most ax is set at this point because popcnt(qword) < byte
+   | mov ax, word [leaves+index*2] -- leaves[index]
```
corsix: This would probably be better as `movzx eax, ...`. The assumption in the comment is true, but the CPU is not clever enough to know this, and so if you do a `mov` into `ax`, then when `eax` is next used, it'll insert a micro-up to combine `ax` and the previous high 16-bits of `eax` (whereas `movzx` here is the same price as `mov`, and doesn't have the extra micro-op penalty down the line).

src/lib/poptrie_lookup.dasl
corsix: Alternatively, the entire sequence `mov`/`mov`/`shl`/`and`/`cmp`/`je` might be better as `bt vec, v`/`jnc`.

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+   | mov rax, 2
+   | mov rcx, v
+   | shl rax, cl
+   | sub rax, 1
+   -- rax = popcnt(band(vec, rax))
+   | and rax, vec
+   | popcnt rax, rax
+   -- index = base + bc - 1
+   | mov index, dword [node+20] -- nodes[index].base1
+   | add index, eax
```
corsix: There might be value in transposing this `add` and this `sub` - the `add` cannot execute until both the `mov` and the `popcnt` are done, and the `sub` cannot execute until the `add` is done - whereas if transposed, the `sub` could execute in parallel with the `popcnt`.

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+   | sub index, 1
+   -- node = nodes[index]
+   | imul index, 24 -- multiply by node size
+   | lea node, [nodes+index]
+   -- offset = offset + k
+   | add offset, 6
+   | jmp <2 -- loop
```
corsix: You could consider rotating the loop to get rid of this unconditional jump on every iteration, i.e. turning:
```
head:
body part 1
jump to tail if condition
body part 2
jmp head
tail:
```
into:
```
jmp middle
head:
body part 2
middle:
body part 1
jump to head if not condition
```
or:
```
body part 1
jump to tail if condition
head:
body part 2
body part 1
jump to head if not condition
tail:
```

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+   if Poptrie.direct_pointing then
+      -- v = extract(key, 0, Poptrie.s)
+      local direct_mask = bit.lshift(1ULL, Poptrie.s) - 1
+      -- v = band(key, direct_mask)
+      | mov v, direct_mask
+      | and v, key
+      -- index = dmap[v]
+      | mov index, dword [dmap+v*4]
+      -- eax = band(index, leaf_tag)
+      | mov eax, 0x80000000
```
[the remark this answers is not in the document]
corsix: Or perhaps `mov eax, index` / `btr eax, 31` / `jnc`, with the bonus of then eliminating the subsequent `mov`/`and`.

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,158 @@
...
+|.define vec,     r11 -- 64-bit vector or leafvec
+
+-- lookup(leaf_t *leaves, node_t *nodes, key) -> leaf_t
+function lookup (Dst, Poptrie, keysize)
+   if Poptrie.direct_pointing then
+      -- v = extract(key, 0, Poptrie.s)
+      local direct_mask = bit.lshift(1ULL, Poptrie.s) - 1
+      -- v = band(key, direct_mask)
+      | mov v, direct_mask
+      | and v, key
```
corsix: I'd probably write this `mov`/`and` as `mov v, key` / `and v, direct_mask`, as that can allow for a more compact encoding (`mov r32, imm32` always encodes the immediate as 32 bits, whereas `and r32, imm32` _can_ encode the immediate as 8 bits if it is small enough).

corsix: The other rationale here that I forgot last night: on contemporary architectures, `mov reg, reg` can be zero latency and not require any execution unit, whereas `mov reg, imm` does have latency and does require an execution unit (`op reg, reg` and `op reg, imm` are same cost for most/all ops other than `mov`).

eugeneia:
I have adopted some of the changes suggested by @corsix, which show quite a substantial improvement in performance. I have omitted the loop rotations, since they didn’t seem to make a measurable difference when testing and make the code harder to read imho.

I also didn’t do the BMI2 dependent changes since the oldest lab server doesn’t support them (grindelwald has an ivy bridge CPU), also I would have to add the BMI2 instructions to DynASM first. Might be worth to do a CPU feature detection here?

Before:
```
PMU analysis (numentries=400000, keysize=64)
build: 6.7974 seconds
lookup64: 576.12 cycles/lookup 215.73 instructions/lookup
build(direct_pointing): 7.3970 seconds
lookup64(direct_pointing): 441.36 cycles/lookup 150.59 instructions/lookup
```
After:
```
PMU analysis (numentries=400000, keysize=64)
build: 6.9591 seconds
lookup64: 507.08 cycles/lookup 186.47 instructions/lookup
build(direct_pointing): 7.5430 seconds
lookup64(direct_pointing): 404.69 cycles/lookup 132.36 instructions/lookup
```

corsix:
FWIW, the latest upstream DynASM should have BMI2 in it (c.f. https://github.com/LuaJIT/LuaJIT/blob/26f1023819efb843e10014232cd88bb1d52ea4f5/dynasm/dasm_x86.lua#L1690, present there since February 2018). Not having Haswell everywhere is a valid reason not to use it though (I would however be interested in knowing what performance improvement - if any - it could give here, if that experiment is something you're willing to try).

src/lib/poptrie_lookup.dasl
```diff
@@ -0,0 +1,146 @@
...
+   | shl rax, cl
+   | sub rax, 1
+   if Poptrie.leaf_compression then
+      -- vec = nodes[index].leafvec
+      | mov vec, qword [node+0]
+   else error("NYI") end
+   -- rax = popcnt(band(vec, rax)) - 1
+   | and rax, vec
+   | popcnt rax, rax
+   -- return leaves[base + bc - 1]
```
corsix: The `mov` and the `add` here could be fused, as in:
```
   | add eax, dword [node+16] -- nodes[index].base0
   | movzx eax, word [leaves+eax*2-2] -- leaves[index]
```
(No improvement in unfused-domain uops, but does save one fused-domain uop per function invocation, if that is even measurable against noise)

corsix:
Very nice; I like seeing concrete performance numbers.

eugeneia:
@petebristow I am having trouble patching in the BMI2 instructions from https://github.com/LuaJIT/LuaJIT/commit/fe651bf6e2b4d02b624be3c289378c08bab2fa9b

I get errors like

```
lib/poptrie_lookup.dasl:98: error: bad char `w' in pattern `F20F38wF7rM' for `shrx':
      | shrx v, key, offset
lib/poptrie_lookup.dasl:109: error: bad char `w' in pattern `0F38wF5rM' for `bzhi':
   | bzhi rax, vec, rcx
lib/poptrie_lookup.dasl:130: error: bad char `w' in pattern `0F38wF5rM' for `bzhi':
   | bzhi rax, vec, rcx
```
 and can’t seem to find where the `w` modifier is actually implemented?

corsix:
Aha, the `w` got lost as [the PR](https://github.com/LuaJIT/LuaJIT/pull/399/files) became [the upstream commit](https://github.com/LuaJIT/LuaJIT/commit/fe651bf6e2b4d02b624be3c289378c08bab2fa9b). That is unfortunate.

eugeneia:
@corsix On another note, I am currently updating this code to work with keys up to 128 bits (IPv6) and am wondering if you know any neat tricks for extracting 6 bits from a `uint8_t[16]` at an arbitrary offset in x86 asm? (I suppose what is currently an `uint64_t` argument will have to become a pointer.)

corsix:
shrd from a pair of uint64_t registers comes to mind. Not a cheap
instruction (e.g. four uops on Broadwell), but should do what you need.

eugeneia:
Without BMI2 (this branch):
```
PMU analysis (numentries=400000, keysize=64)
build: 6.7322 seconds
lookup: 55733.43 cycles/lookup 87694.39 instructions/lookup
lookup64: 504.83 cycles/lookup 186.46 instructions/lookup
build(direct_pointing): 7.5217 seconds
lookup(direct_pointing): 11152.73 cycles/lookup 25908.96 instructions/lookup
lookup64(direct_pointing): 416.15 cycles/lookup 132.36 instructions/lookup
```
With BMI2 (https://github.com/eugeneia/snabb/commits/poptrie-bmi2):
```
PMU analysis (numentries=400000, keysize=64)
build: 7.3056 seconds
lookup: 39330.65 cycles/lookup 72852.83 instructions/lookup
lookup64: 431.63 cycles/lookup 151.47 instructions/lookup
build(direct_pointing): 7.8106 seconds
lookup(direct_pointing): 12443.39 cycles/lookup 26317.73 instructions/lookup
lookup64(direct_pointing): 376.19 cycles/lookup 112.10 instructions/lookup
```

eugeneia:
@corsix Thanks for the suggestion! I ended up with something like this for extracting from a 16 byte array, probably not optimal... wdyt?

```
mov rxc, offset
mov rax, [key]
mov rbx, [key+8]
cmp rcx, 63
jng >1
sub rcx, 64
mov rax, rbx
1:
shrd rax, rbx, cl
and rax, 0x3F
```

corsix:
```
mov ecx, offset   -- Prefer 32-bit operations where possible
mov rax, [key]
mov rbx, [key+8]
test cl, 64       -- Prefer test to cmp (archaic habbit)
cmovnz rax, rbx   -- This is your mov within the branch
                  -- Do not need to sub rcx, 64
                  -- as shrd takes cl mod 64, and subtracting 64 mod 64 is a no-op
shrd rax, rbx, cl
and eax, 0x3F     -- Prefer 32-bit operations where possible
```

eugeneia:
@corsix Neat! Thanks!

I have added support for 128 (and 32) bit keys (the interface now expects pointers to byte arrays for keys), a well as conditional use of BMI2 instructions (since they have a substantial impact.)

I have adjusted the micro benchmark so that it does not measure a cold cache. It is not exactly representative of a real workload either, but yields more accurate feedback for testing changes. I have also noticed that for 400,000 entries the benchmark will now OOM, but could confirm that this will be fixed by switching to RaptorJIT.

I think the results tell us that its pretty fast with warm caches, and cost increases roughly linearly with key size (prefix length).

```
PMU analysis (numentries=400000, keysize=32)
build: 6.0666 seconds
lookup: 5604.90 cycles/lookup 14653.72 instructions/lookup
lookup32: 81.79 cycles/lookup 118.72 instructions/lookup
lookup64: 81.77 cycles/lookup 118.72 instructions/lookup
lookup128: 99.13 cycles/lookup 138.95 instructions/lookup
build(direct_pointing): 6.0043 seconds
lookup(direct_pointing): 2605.19 cycles/lookup 4805.54 instructions/lookup
lookup32(direct_pointing): 46.62 cycles/lookup 79.66 instructions/lookup
lookup64(direct_pointing): 47.05 cycles/lookup 79.65 instructions/lookup
lookup128(direct_pointing): 51.88 cycles/lookup 84.86 instructions/lookup
```
```
PMU analysis (numentries=400000, keysize=64)
build: 17.5607 seconds
lookup: 48630.60 cycles/lookup 58459.67 instructions/lookup
lookup64: 131.99 cycles/lookup 153.46 instructions/lookup
lookup128: 144.89 cycles/lookup 185.23 instructions/lookup
build(direct_pointing): 17.5885 seconds
lookup(direct_pointing): 4672.99 cycles/lookup 12343.01 instructions/lookup
lookup64(direct_pointing): 92.95 cycles/lookup 114.37 instructions/lookup
lookup128(direct_pointing): 96.92 cycles/lookup 131.13 instructions/lookup
```

eugeneia:
I’ll be my own upstream for this and I think this is ready, merged into max-next.

PR #1403, opened 2019-01-02
next: Changes queued for v2019.01 release

lukego:
Happy new year!

This branch contains changes queued for the v2019.01 release. There has been a haitus between Snabb releases for some months now basically because I have been short of time and thinly spread between many projects (notably getting RaptorJIT 1.0 out the door and keeping up with my two small children.)

There is a large backlog of *extremely* cool code waiting in Pull Request limbo. RRD counter files, automatic counter aggregation, a new `snabb top`, a lightweight concurrency library, and much more besides. My priority now is to get all of this merged and released and get back to development as usual. If this causes problems for people, e.g. due to some compatibility issues with the new `snabb top` etc, then let's take that with follow-up PRs.


eugeneia:
Mhm, seems something slipped in that breaks the build: https://gist.github.com/SnabbBot/0bc074efa5966f84146172c84f71d5fb#file-log-L159

PR #1411, opened 2019-01-29
Initial import of intel_avf driver

petebristow:


eugeneia:
Oooh neat!!

lukego:
So this is ~700 LOC to support all Intel NICs present and future? That would be nice!

lukego:
(I am really looking forward to reading this code properly once I have the next Studio release out the door...)

petebristow:
There are some tests, fixes and cleanup pending.

eugeneia:
Minor markdown formatting nitpick: markdown needs newlines between `— **thing** *args*` and the following paragraph to render the item “properly”. Also there seems to be some markdown parsing issue in the npackets diagram.

See: https://github.com/snabbco/snabb/blob/e50404a2fd81e3d582e24209a733f120e2caab7f/src/apps/test/README.md#npackets-appstestnpackets

PR #1413, opened 2019-01-31
Mellanox ConnectX-4/5/6 10G-200G device driver

lukego:
This branch contains the Mellanox device driver for upstreaming.

This was initially developed on ConnectX-4 and has since been debugged on ConnectX-5 by @alexandergall. Has not been interop tested with ConnectX-6 yet AFAIK.

Git history has been somewhat collapsed while transporting this driver between branches (all the more reason to finally land it upstream.)

TODO:
- [ ] Readme.
- [ ] Register ConnectX-5 device id (as per @alexandergall branch.)
- [ ] Revert some changes that are coming along for the ride.


lukego:
This PR is a bit malformed. I'll try again...

PR #1417, opened 2019-02-10
AF_XDP app 

dpino:
This has been taking longer than I expected, so I'd like to share the work done so far. At this point I have an AF_XDP app working. Here are some benchmarks (Load: 10M packets; packet size: 550 byte):

• Intel10G snabbmark:

```
$ sudo SNABB_PCI0=0000:82:00.0 SNABB_PCI1=0000:02:00.0 ./snabb snabbmark intel10g 10e6 550
No PMU available: single core cpu affinity required
Sending through nic0 (82:00.0); Receiving through nic1 (02:00.0)

Processed 10.0 million packets in 4.78 seconds (rate: 2.1 Mpps, 8.58 Gbit/s, 0.04 % packet loss).
```

• XDPSockApp (AF_XDP) snabbmark:

```
$ sudo SNABB_PCI0=ens6f0 SNABB_PCI1=ens2f0 ./snabb snabbmark xdpsocket 10e6 550
No PMU available: single core cpu affinity required
Packets: 10000000; Packet Size: 550; Timeout: 1000
Sending through ens6f0 (90:e2:ba:ac:b9:b4); Receiving through ens2f0 (90:e2:ba:ac:b9:48)

Processed 10.0 million packets in 14.87 seconds (rate: 0.7 Mpps, 2.76 Gbit/s, 0.00 % packet loss).
```

• RawSocketApp (AF_PACKET) snabbmark:

```
$ sudo SNABB_PCI0=ens6f0 SNABB_PCI1=ens2f0 ./snabb snabbmark rawsocket 10e6 550
No PMU available: single core cpu affinity required
Packets: 10000000; Packet Size: 550; Timeout: 1000
Sending through ens6f0 (90:e2:ba:ac:b9:b4); Receiving through ens2f0 (90:e2:ba:ac:b9:48)

Processed 10.0 million packets in 117.35 seconds (rate: 0.1 Mpps, 0.35 Gbit/s, 6.79 % packet loss).
```

The AF_XDP app is way much faster than the traditional RawSocket app (8 times faster), but it's still slower than the native Intel10G driver (4 times slower). I think there's room for improving the performance. The code of the app is based in the Linux kernel AF_XDP example (https://github.com/torvalds/linux/blob/master/samples/bpf/xdpsock_user.c). This example reads and writes packets in batches with a BATCH_SIZE of 16. However in the app, to simplify the code and make it work, I used a BATCH_SIZE of 1. Also, the Linux kernel example uses 4 sockets whereas the Snabb AF_XDP app only opens one socket.

Being able to send and receive packets in batches plus using several sockets underneath should definitely help to improve the performance. Once that is done, likely rewriting as much code of the app as possible in Lua may help to improve performance too (since more code could be trace jitted by RaptorJIT).

I'm currently working in increasing the BATCH_SIZE.

Note: To use the AF_XDP app is necessary to use a Linux kernel >=4.18 (built with AF_XDP support). At this moment the app also depends on the Linux kernel source code. So to build Snabb with AF_XDP enabled do: `LINUX_SOURCE_DIR=/path/to/linux/source make -j`.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -13,7 +13,11 @@ OBJDIR = $(patsubst %,obj/%,$(SRCDIR))
 LUASRC = $(shell find . -regex '[^\#]*\.lua' -printf '%P ')
 PFLUASRC = $(shell cd ../lib/pflua/src && \
 	         find . -regex '[^\#]*\.lua' -printf '%P ')
-CSRC   = $(shell find . -regex '[^\#]*\.c' -printf '%P ')
+ifdef LINUX_SOURCE_DIR
+	CSRC   = $(shell find . -regex '[^\#]*\.c' -printf '%P ')
+else
+	CSRC   = $(shell find . -regex '[^\#]*\.c' | egrep -v 'xdpsock' | tr '\n' ' ')
+endif
 CHDR   = $(shell find . -regex '[^\#]*\.h' -printf '%P ')
 ASM    = $(shell find . -regex '[^\#]*\.dasl' -printf '%P ')
 PFLUAASM = $(shell cd ../lib/pflua/src && \
@@ -40,6 +44,11 @@ EXE    := bin/snabb $(patsubst %,bin/%,$(PROGRAM))
 
 LUAJIT_A := ../lib/luajit/src/raptorjit.a
 
+XDPSRCDIR = apps/socket/xdp
+XDPOBJDIR = obj/apps/socket/xdp
+CLANG = clang
+LLC = llc
+
 # TESTMODS expands to:
 #   core.memory core.lib ...
 # for each module that has a top-level selftest () function.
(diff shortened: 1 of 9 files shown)


eugeneia:
Neat! I feel like this looks reasonably straight forward, seems within grasp for us to copy the a C function signatures and structures and do this 100% using LuaJIT ffi and ljsyscall, no? :-)

dpino:
Right, eventually the code could be rewritten in Lua + ljsyscall. In the meantime, I prefer to keep the differences respect the original xdpsock example small. Once there's support for batch fetching and using multiple sockets, I think the next step is to rewrite the code in Lua.

I also think it should be possible to add the Linux source code dependencies as part of Snabb code base, instead of requiring to download the kernel sources.

dpino:
Implemented `BATCH_SIZE` support.

Apparently the results are slightly worse, however I think the previous code was not working totally ok and this performance regression comes at the cost of fixing bugs in the previous code (for instance a `packet.clone` was needed in `pull`). Here are the results for the same benchmark:

```
$ sudo SNABB_PCI0=ens6f0 SNABB_PCI1=ens2f0 ./snabb snabbmark xdpsocket 10e6 550
No PMU available: single core cpu affinity required
Packets: 10000000; Packet Size: 550; Timeout: 1000
Sending through ens6f0 (90:e2:ba:ac:b9:b4); Receiving through ens2f0 (90:e2:ba:ac:b9:48)

Processed 10.0 million packets in 18.37 seconds (rate: 0.5 Mpps, 2.23 Gbit/s, 0.65 % packet loss).
```

If I try `BATCH_SIZE=1`, like before, results are much worse than `BATCH_SIZE=16`.

```
$ sudo SNABB_PCI0=ens6f0 SNABB_PCI1=ens2f0 ./snabb snabbmark xdpsocket 10e6 550
No PMU available: single core cpu affinity required
Packets: 10000000; Packet Size: 550; Timeout: 1000
Sending through ens6f0 (90:e2:ba:ac:b9:b4); Receiving through ens2f0 (90:e2:ba:ac:b9:48)

Processed 10.0 million packets in 25.31 seconds (rate: 0.4 Mpps, 1.62 Gbit/s, 0.00 % packet loss).
```

It seems `BATCH_SIZE=16` is a hard limit. If try `BATCH_SIZE=32`, there are too many packet drops:

```bash
$ sudo SNABB_PCI0=ens6f0 SNABB_PCI1=ens2f0 ./snabb snabbmark xdpsocket 10e6 550
No PMU available: single core cpu affinity required
Packets: 10000000; Packet Size: 550; Timeout: 1000
Sending through ens6f0 (90:e2:ba:ac:b9:b4); Receiving through ens2f0 (90:e2:ba:ac:b9:48)

Processed 10.0 million packets in 13.48 seconds (rate: 0.7 Mpps, 3.04 Gbit/s, 37.21 % packet loss).
```

eugeneia:
I have looked a little bit into [AF_XDP](https://www.kernel.org/doc/html/latest/networking/af_xdp.html). I am just going to braindump some thoughts to get a discussion going, correct me if I am wrong.

I imagine this should work like our regular NIC drivers to some extent:

- There is a concept of `UMEM` to which the kernel (used to be: hardware NIC) can write packets and read packets from. UMEM is just a memory region, the only difference with regard to DMA memory seems to be that we have to register it to the kernel and bind it to a specific XDP interface (XSK).
- Each XSK has four kinds of descriptor rings: FILL, RX, TX, COMPLETION
- We put descriptors (pointers into UMEM) on the FILL ring, once the kernel is done with the descriptor it will appear filled with a packet in the RX ring and can be reused
- We put descriptors filled with packets to send on the TX ring, once the kernel is done they will appear on the COMPLETION ring and can be reused

So far so good, pretty much business as usual as compared with our polling NIC drivers? Now to get the most out of this interface could we maybe do something like:

 - register all our `struct packet` memory as UMEM for a given interface?
 - fill the rings with descriptors that point to packet `data`?

I.e., have the kernel write out packets directly instead of copying between UMEM and DMA memory?

> An AF_XDP is socket linked to a single UMEM, but one UMEM can have multiple AF_XDP sockets.

One roadblock I see here is that we apparently can have only one contiguous UMEM region per socket, so maybe that clashes with our HUGETLB chunk allocation policy. Dunno! :-)

Cc @lukego

PR #1435, opened 2019-07-17
lib.yang: unique and min/max restrictions, choice fixes

eugeneia:
This extends lib.yang to support unique and min/max restrictions as well as choice statements.

Contains: https://github.com/Igalia/snabb/pull/1188

### DIFF
--- a/src/lib/ptree/ptree.lua
+++ b/src/lib/ptree/ptree.lua
@@ -151,6 +151,7 @@ function Manager:state_change_event(event, ...)
 end
 
 function Manager:set_initial_configuration (configuration)
+   path_data.consistency_checker_from_schema_by_name(self.schema_name, true)(configuration)
    self.current_configuration = configuration
    self.current_in_place_dependencies = {}
 
--- a/src/lib/yang/data.lua
+++ b/src/lib/yang/data.lua
@@ -184,8 +184,12 @@ function data_grammar_from_schema(schema, is_config)
    function handlers.choice(node)
       local choices = {}
       for choice, n in pairs(node.body) do
-         local members = visit_body(n)
-         if not is_empty(members) then choices[choice] = members end
+         if n.kind == 'case' then
+            local members = visit_body(n)
+            if not is_empty(members) then choices[choice] = members end
+         else
+            choices[choice] = { [choice] = visit(n) }
+         end
       end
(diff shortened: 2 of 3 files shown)


eugeneia:
Cc @wingo @dpino @xray7224

eugeneia:
Cc @alexandergall

eugeneia:
This seems to break `program/lwaftr/tests/selftest.sh`. I think one bug here is that the consistency checkers don’t handle ctype objects such as the one in question:

```
{
  ...
  key_ctype = "struct { uint32_t ipv4; uint16_t psid; }",
  keys = { ... },
  type = "table",
  values = { ... },
  value_ctype = "struct { uint8_t b4_ipv6[16]; uint8_t br_address[16]; uint16_t padding; struct { uint8_t psid_length; uint8_t reserved_ports_bit_count; } port_set; }",
}
```

So let’s say I add

```diff
diff --git a/src/lib/yang/path_data.lua b/src/lib/yang/path_data.lua
index 7923aa82e..ac27c515c 100644
--- a/src/lib/yang/path_data.lua
+++ b/src/lib/yang/path_data.lua
@@ -562,6 +562,7 @@ function uniqueness_checker_from_grammar(grammar)
       elseif grammar.type == 'table' then
          -- visit values
          for name, value in pairs(grammar.values) do
+            local pairs = grammar.key_ctype and cltable.pairs or pairs
             for k, datum in pairs(data) do
                visit_unique_and_check(value, datum[name])
             end
@@ -616,6 +617,7 @@ function minmax_elements_checker_from_grammar(grammar)
       elseif grammar.type == 'table' then
          -- visit values
          for name, value in pairs(grammar.values) do
+            local pairs = grammar.key_ctype and cltable.pairs or pairs
             for k, datum in pairs(data) do
                visit_minmax_and_check(value, datum[name], name)
             end
```

However, the `data` here seems to be a ctable (not a cltable, so can’t use `cltable.pairs`) so I am not sure what to do here. How can the data for a grammar with `key_ctype` not be a `cltable`?

eugeneia:
Figured it out. If `key_ctype` and `value_ctype` are present in the grammar the data will be represented unless `string_key` is also present. I hadn’t accounted for that.

From my side this is good to go! Any objections if I upstream this via max-next?

PR #1436, opened 2019-07-21
Unbind PCI devices before mapping their memory (take two)

eugeneia:
Make sure PCI devices are unbound before attempting to map their memory in intel_mp. This a second take on #1296:

> Since Linux 4.5, devices that are bound to a kernel driver can't have
their memory mapped by userspace under the default configuration of
CONFIG_IO_STRICT_DEVMEM=y.

Why do I think we avoid races this time?  Basically what we have during device initialization is a state machine `FREE -> INIT -> SHARED`.  Master will acquire the exclusive lock, unbind the device, enable bus mastering, and then map device memory and do the init before allowing the lock to be shared (`INIT` state). Other processes will wait for the shared lock and map device memory only after master has finished unbinding and initialization (`SHARED` state).

During looking at this I also stumbled upon the fact that intel_mp bypassed the `lib.hardware.pci` shutdown safety routine, and ended up with https://github.com/snabbco/snabb/commit/af820842ee9618114d8983f42684e7e611a92eeb.

One remaining question that came upon me is if we have a strategy for when two unrelated Snabb process try to grab the same NIC?

Testing, feedback, review would be appreciated @alexandergall @wingo @petebristow 

Resolves: #1286 

Supersedes: #1296

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1016,6 +1016,7 @@ Groups of Snabb processes each have the following special properties:
   mastering (DMA) is disabled upon termination before any DMA memory
   is returned to the kernel. This prevents "dangling" DMA requests
   from corrupting memory that has been freed and reused.
+  See [lib.hardware.pci](#pci-lib.hardware.pci) for details.
 
 The `core.worker` API functions are available in the main process only:
 
--- a/src/apps/intel/intel10g.lua
+++ b/src/apps/intel/intel10g.lua
@@ -91,9 +91,10 @@ function new_sf (conf)
 end
 
 function M_sf:open ()
+   self.fd = pci.open_pci_resource_locked(self.pciaddress, 0)
    pci.unbind_device_from_linux(self.pciaddress)
    pci.set_bus_master(self.pciaddress, true)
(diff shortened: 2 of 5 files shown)


eugeneia:
> One remaining question that came upon me is if we have a strategy for when two unrelated Snabb process try to grab the same NIC?

Could do something along the lines the following patch to detect when the current master is not in our group:
```diff
diff --git a/src/apps/intel_mp/intel_mp.lua b/src/apps/intel_mp/intel_mp.lua
index c46f4d89b..767e82a5e 100644
--- a/src/apps/intel_mp/intel_mp.lua
+++ b/src/apps/intel_mp/intel_mp.lua
@@ -408,6 +408,8 @@ function Intel:new (conf)
       -- Other processes wait for the shared lock before memory mapping the and
       -- loading registers.
       self.fd:flock("sh")
+      assert(shm.exists("group/dma/pci/"..self.pciaddress),
+             "NIC is already in use: "..self.pciaddress)
       self.base = pci.map_pci_memory(self.fd)
       self:load_registers(byid.registers)
    end
```

irevoire:
This fix works for me.

PR #1446, opened 2019-10-26
Integrated intel_avf driver

eugeneia:
This PR includes @petebristow’s Intel AVF driver in #1411, as well as some minor polish and a bug fix or two. Tested on lugano-1 with both x710 and xl710 cards (rev 01) like so:

```
sudo SNABB_AVF_PF0=0000:03:00.0 apps/intel_avf/selftest.sh
```

(It’s possible to test this with a single unwired port/PF afterall.)

Cc @lukego 

Would be neat to get this into the pending release!

### DIFF
--- a/src/apps/intel_avf/README.md
+++ b/src/apps/intel_avf/README.md
@@ -1,0 +1,42 @@
+# Intel AVF (adaptive virtual function) app (apps.intel_avf.intel_avf)
+
+The `intel_avf.Intel_avf` app provides drivers for the Virtual Functions exported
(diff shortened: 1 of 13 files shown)


eugeneia:
Just a heads up that this is integrated in Vita here (https://github.com/inters/vita/pull/106) and works with `snabb loadtest` too. Next steps I think would be to implement the VF-side RSS configuration which is sadly not fully documented in the spec. Either way I think this is good to go for an initial merge.

PR #1452, opened 2019-12-12
Various YANG fixes related to choice statements and the consistency checker

eugeneia:
This fixes a few oversights regarding the handling and validation of choice bodies. 
e96ae1a in particular could use a set of eyes familiar with lib.yang. Was there a purpose for the `order` parameter that I am not seeing?

Cc @wingo @dpino @xray7224

### DIFF
--- a/src/lib/yang/data.lua
+++ b/src/lib/yang/data.lua
@@ -880,6 +880,7 @@ function xpath_printer_from_grammar(production, print_default, root)
          if translator ~= nil then
             local statements = translator(keyword, production)
             for k,v in pairs(statements) do translated[k] = v end
+            order = nil
          else
             translated[keyword] = production
          end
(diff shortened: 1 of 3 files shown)


dpino:
About the `order` argument in the body printer, I think the purpose was to print the keys of an output sorted.

You commented on 19072b4 that the _pre-ordering always matches the internal ordering logic
of body_printer_. If that's the case, I think removing the `order` argument is fine. But I would suggest checking the body printers on their different formats (YANG, XML and InfluxDB), to confirm the removal doesn't affect the output.

It seems the only printer tested is InfluxDB. For the other two, I would suggest launching a lwAFTR instance and run several config-get commands (there are several examples in https://github.com/Igalia/snabb/pull/843)

```bash
$ sudo ./snabb lwaftr run --name lwaftr --conf lwaftr.conf --on-a-stick 82:00.1
```
```
$ sudo ./snabb config get lwaftr /softwire-config/external-interface
...
$ sudo ./snabb config get --format=xpath lwaftr /softwire-config/binding-table  
...
```

Consider also querying the whole configuration ('/) or the binding table ('/softwire-config/binding-table'). If the results keep sorted, removing the `order` arguments seems right.

It would be great also (probably in an future PR) to add tests for the YANG and XML printers, like the one it currently exists for InfluxDB.

dpino:
FWIW, the other changes LGTM.

PR #1457, opened 2020-04-17
arch/checksum: fix bug in carry propagation

eugeneia:
This fixes a bug in our scalar x86 IP checksum implementation, where some
carries where missed. Really hard to find with random testing, added a
systematic test case for carry propagation.


dpino (approved):

dpino:
Sorry for the late review, I didn't see your PR.

PR #1461, opened 2021-11-12
Driver for Mellanox Connect-X 4/5/6 cards

eugeneia:
Finally a PR to include all the hard work by @capr @lukego @alexandergall and myself in the next release (TBA)!

This work was made possible by Mellanox releasing a public Programmer’s Reference Manual (PRM) for their Connect-X 4 cards!

This driver can operate Mellanox Connect-X 4 and 5 cards, and has been tested at SWITCH and in lwAFTR.

It supports RSS and MAC+VLAN switching (and combinations thereof). Currently not supported is VLAN insertion and stripping (use `apps.vlan`).

There is a branch https://github.com/eugeneia/snabb/compare/mellanox-2021...eugeneia:mellanox-2021-vlan-strip-insert that implements VLAN stripping and insertion in the driver, but I’m not sure if the extra code is worth its weight or if its cleaner to delegate to `apps.vlan` instead. Opinions welcome!


lukego:
Great going Max :)

eugeneia:
Seems that the driver also works for at least some Connect-X 6 cards, which is cool.

We should extend the tests a bit more before merging, and I think a blocker that remains is proper support for `app:stop()`. Right now the implementation is a stub and doesn’t really work. Needs to:
- transition queues from IDLE->DEAD and delete the objects
- free remaining packets on queues

eugeneia:
Added support for `stop()`, extended the tests a bit more, and added a README.

I noticed that local-loopback between queues is currently not implemented. I tend towards leaving that as a TBD. (Couldn’t quite figure out how to enable it from a quick survey of the PRM.)

PR #1462, opened 2021-11-15
Updated Intel AVF driver with VLAN+RSS support

eugeneia:
This extends the Intel AVF driver to support multiple RSS queues via an IO app to attach to each individual queue (just like our Connect-X driver).

This also adds support for VLAN tagging/untagging in the driver, fixes a perf issue, canonicalizes the SHM counter location of the driver, and includes a fix to make `snabb top` correctly handle interfaces that do not specify a link speed (as typical with AVF devices).

Depends on: https://github.com/snabbco/snabb/pull/1451

### DIFF
--- a/src/apps/intel_avf/README.md
+++ b/src/apps/intel_avf/README.md
@@ -18,20 +18,55 @@ The links are named `input` and `output`.
 
 *Required*. The PCI address of the NIC as a string.
 
+— Key **vlan**
+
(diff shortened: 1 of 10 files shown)


eugeneia:
Cc @petebristow

eugeneia:
Extended tests some more and found an obvious bug where somewhere in my tinkering I made it so that the driver would transmit 4 bytes too many.

Issue #1007, opened 2016-09-02
ConnectX: Review N*SQ 64B transmit performance mellanox (Rev 2)


lukego:
This [benchmark report](http://rpubs.com/lukego/205901) for single-core transmit ("packetblaster") over multiple Send Queues superceeds #1006. This is based on a new Snabb version that performs better due to different DMA memory allocation.

![download 2](https://cloud.githubusercontent.com/assets/13791/18199349/3320e9aa-7100-11e6-9674-fc0d14c08b38.png)

These results look much clearer and simpler.

Summary:
- ~16 Mpps per Send Queue.
- ~64 Mpps with SendQueues>=4 and QueueSize>=1024.
- Max observation ~67 Mpps.

lukego:
I am reading an absolutely fascinating book called [Statistical Modeling: A Fresh Approach](http://www.mosaic-web.org/go/StatisticalModeling/). It's about how to make models (functions) that succinctly account for the variations in data sets. The notion is that formal models can be used to capture details that would be too complex to present visually.

I am using this ConnectX-4 data set as a first modeling exercise. The idea is to make a model that accounts for how performance (`Mpps`) depends on other variables (focusing on `SendQueues` which has the biggest effect).

Here is what the raw data looks like. (The points appear in groups because several different queue lengths are tested.)

![raw](https://cloud.githubusercontent.com/assets/13791/18267921/bc4800c8-7421-11e6-9407-9ebec5b76ed7.png)

The exercise here is to make a series of models that try to account for the `Mpps` values.

### Constant

The first naive model supposes that `Mpps` is always the same. This can be expressed by taking a linear model (`lm`) for the formula `Mpps ~ 1` and "fitting" that with the value that best suits this data:

```{r}
> m1 <- lm(Mpps ~ 1, data = d)
> coefficients(m1)
(Intercept) 
   60.11094 
```

This model predicts that performance will always be 60.1 Mpps. We can visualize the model as a line overlayed on the data:

![constant](https://cloud.githubusercontent.com/assets/13791/18267928/c53190d2-7421-11e6-9dfb-58de8c31c00b.png)

On the one hand this is useful information but on the other hand it does not account for any of the variations in the measured values.

### Linear

The second model supposes that `Mpps` increases linearly i.e. following a straight line. The line is defined by the "intercept" (the notional Mpps with 0 send queues) and a slope (how much Mpps changes for each added send queue).

```{r}
[...]

PR #1469, opened 2022-02-02
apps.mellanox.connectx: benchmark matrix

eugeneia:
This adds some scripts and scaffolding for running benchmark matrices on our Connect-X driver. At its heart is a Snabb script that, given a system with two connected Connect-X ports, let’s you run a benchmark like so:

```
./snabb snsh apps/mellanox/benchmark.snabb \
  --pci-a 81:00.0 --pci-b 81:00.1 --cores-a 6-11 --cores-b 12-17 \
  --mode source|source-sink|source-fwd
  --nworkers <int> --nqueues <int> --npackets <int> --pktsize <int> --nmacs <int> --nvlans <int> \
  --opt <string> # Additional options for apps.mellanox.connectx
```

If you wrap this in a shell script like so you can generate matrix CSV data to then investigate with R...

```bash
qsizes="512
1024
2048
4096"

echo i,workers,queues,qsize,pktsize,rate,drop,error
for i in `seq 1 3`; do
    for w in `seq 1 1`; do
        for q in `seq 1 16`; do
            for qsize in $qsizes; do
                for s in "64"; do
                    out=$(./snabb snsh apps/mellanox/benchmark.snabb -a 81:00.0 -b 81:00.1 -A 6-11 \
                            -m source -w $w -q $q -s $s -n 100e6 -o "{sendq_size=$qsize}")
                    rate=$(echo "$out" | grep "Tx Rate is" | cut -d " " -f 4)
                    drop=$(echo "$out" | grep "Tx Drop Rate is" | cut -d " " -f 5)
                    error=$(echo "$out" | grep "Tx Error Rate is" | cut -d " " -f 5)
                    echo "$i,$w,$q,$qsize,$s,$rate,$drop,$error"
                    #echo $out
                done
            done
        done
    done
done
```

…and maybe end up with a plot like this:

mellanox-nqueues-qsize

We already used this in an attempt to reproduce https://github.com/snabbco/snabb/issues/1007 and debug a performance issue on EPYC systems. A report on the results can be found here: https://rpubs.com/eugeneia/860725


eugeneia:
I forgot to mention, but [7f967e4](https://github.com/snabbco/snabb/pull/1469/commits/7f967e4d6500bcbfd738b2dddec397aa6404e327) and [9adad52](https://github.com/snabbco/snabb/pull/1469/commits/9adad52634d226af9707cd7a829b44aaada145ba) are interesting commits in their own right. They (re)introduce a packetblaster mode for the Connect-X driver that should be on-par with the trusty 52899 [packetblaster](https://github.com/snabbco/snabb/tree/master/src/program/packetblaster) in transmitting packets with low CPU overhead. I’ve double checked it to use the equivalent transmit routine @lukego used in the 2016 expirements/measurements.

This specialized transmit routine should be able to tease out hardware limits as it has negligible CPU overhead. It doesn’t however necessarily reflect “real” application transmit performance (as its sending the same buffer worth of packets over and over).

PR #1486, opened 2022-10-04
Timeline 2022

eugeneia:
We want to finally merge @lukego’s timeline feature to help debug latency issues in Snabb applications.

- [x] This latest merge with master is not very good, have to manually triage changes before upstreaming
- [x] Expose flag to disable the timeline
- [x] Run some performance CI to understand what the timeline costs us today


eugeneia:
The timeline is now disabled by default on this branch. It can be enabled via `require("core.timeline").enabled=true` or `lib.scheduling`.

To reduce the perf impact with the timeline disabled we
- disabled the `packet_freed` event
- removed arguments from some breath and app events

Performance impact with these changes looks like this:

download__5_

PR #1489, opened 2022-10-20
Configurable interlink queue, freelist, and group freelist sizes

eugeneia:
This set of changes allows to adjust interlink queue, freelist, and group freelist sizes at runtime.

This is useful because some applications might choose larger interlink queue sizes (just like an application might select larger NIC driver receive queue buffers) to allow for longer breath latencies before packets are dropped.

### DIFF
--- a/src/lib/interlink.lua
+++ b/src/lib/interlink.lua
@@ -96,11 +96,12 @@ ffi.cdef([[
 -- Interlinks can be in one of six states:
 
 local INIT = 0 -- Implicit initial state due to 0 value.
-local FREE = 1 -- Queue is in free state, ready to attach.
-local RXUP = 2 -- Receiver has attached.
-local TXUP = 3 -- Transmitter has attached.
-local DXUP = 4 -- Both ends have attached.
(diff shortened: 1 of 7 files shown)


src/lib/interlink.lua
```diff
@@ -130,15 +135,20 @@ local DOWN = 4 -- Both ends have detached; must be re-allocated.
...
    waitfor(
       function ()
          -- Create/open the queue.
-         r = shm.create(name, "struct interlink")
-         -- Return if we succeed to initialize it.
-         if initialize(r) then return true end
+         r = shm.create(name, "struct interlink", size)
+         -- Initialize queue (only one process can set size).
+         if sync.cas(r.state, INIT, FREE) then
+            r.size = size
```
eugeneia, line 148: Looking at this again I think this change introduces a race condition, because another process can transition `FREE->*` and assert the interlink size before it is set. Need an additional state?

eugeneia: Should be fixed in [052d8c2](https://github.com/snabbco/snabb/pull/1489/commits/052d8c256d6b440031c3a05b902107bac0151d63)

PR #1494, opened 2023-02-09
core.group_freelist: fix a deadlock

eugeneia:
[core.group_freelist: fix a deadlock](https://github.com/snabbco/snabb/commit/45d3ac768bcb3d394b986062a3a474715f2eb817) 

chunk.sequence was not reloaded on further iterations of the loop.

Use sync.load() to explicitly reload lock-free states.

Panic if we fail to progress after 1000 iterations.


alexandergall (commented):
I agree with the changes but suggest to leave out the deadlock detection.

PR #1495, opened 2023-02-27
lib.poptrie: fix bug where address bits where interpreted in reverse order

eugeneia:
We used to shift in bits from right to left, thinking that since we are on little endian the addresses are strictly in reverse order. That is not the case! The bytes are swapped but the bits are not! :-)

In the x86 assembly routines we now do `bswap` and then shift down bits from right to left.


alexandergall (approved):
Independently verified with Prefix->ASN mappings from the GeoLite2 ASN Maxmind database.

PR #1501, opened 2023-04-28
apps.ipfix.maps: re-create map ctable if source file has changed

alexandergall:
If mk_map() is called via ipfix:reconfig() because the name of a map file has changed, the map ctable will still not be re-created due to its presence in the cache used by mk_map(). This commit uses the ctime of the file to detect whether the ctable needs to be re-created.

ipfix:reconfig() is also enhanced to only re-create the mapping tables if the rest of a template's configuration is unchanged.

### DIFF
--- a/src/apps/ipfix/ipfix.lua
+++ b/src/apps/ipfix/ipfix.lua
@@ -120,6 +120,17 @@ end
 
 FlowSet = {}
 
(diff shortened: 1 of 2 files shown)


eugeneia (approved):
lgtm

PR #1505, opened 2023-06-07
Ipfix tls template

alexandergall:
The `v4_HTTPS` and `v6_HTTPS` templates extract the TLS Server Name Indication (SNI) from the Client Hello handshake message. Due to lack of a standardised IPFIX element, the name is stored in a proprietary `string` field with PEN 39499 to be compatible with the Progress/Flowmon netflow collector. The field is using a fixed width of 64 bytes and the actual length of the name is stored in a separate element.


alexandergall:
I'm going to clean up the code and create a new PR.

PR #1525, opened 2024-07-11
Remove historic Snabb programs

eugeneia:
This removes:

 - l2vpn
 - lisper
 - snabbnfv
 - snabbvmx
 - wall
 
We currently do not actively maintain those, hence their removal. We can always revive them on demand. For now this lets us simplify maintenance and CI via snabblab-nixos.


eugeneia:
Cc @alexandergall

PR #1534, opened 2026-07-02
lib.yang.schema: fix refinement of grouping instances

alexandergall:
When a grouping is applied with the "uses" statement, create copies of the grouped elements rather than references to them. This fixes the bug that a "refine" would otherwise modify the grouping itself rather than the instance at the place of insertion.


eugeneia:
@alexandergall Hey I'm on holiday until July 26, but I'll take a look at the open PRs when I'm back. :)

eugeneia (approved):

PR #1542, opened 2026-09-23
Handle EOF in ptree JSON unicode escapes

j1sk1ss:
This fixes a crash in the ptree JSON parser when a string ends with a truncated Unicode escape such as `"\\u`.

Previously, `consume_pat()` called `ch:match(...)` before checking whether `read_char()` returned `nil`. At EOF this raised a nil dereference instead of reporting the malformed input. The EOF check now happens first, so truncated escapes fail with `unexpected EOF`.

A regression test was added to `lib.ptree.json.selftest()` for the truncated Unicode escape case.

Tested with:

```sh
SNABB_SHM_ROOT=/tmp/snabb-shm-json-test make -C src DEBUG= lib.ptree.json
```

Found this with [Svace](https://svace.pages.ispras.ru/svace-website/en/) static analyzer


eugeneia:
Hi! Can I ask what brings you to Snabb and this code path in particular? The regression test seems to be missing fwiw.

j1sk1ss:
At this moment we're writing a paper about Svace for Lua, and Snabb's source code has been included to a test corpus as a big code base.

